PR #15, opened 2019-01-01
Tutorial Suggestions

rdaruwala:
I made a few suggestions to the tutorial to address some of the questions / roadblocks I hit while implementing MathComponent


LeStarch:
@rdaruwala thanks for this feedback.  A cursory glance of the changes look good.  Since it affects a tutorial, I'll do a full "from scratch" run through this weekend.  If that looks good, I'll follow-up with where to go from there.

LeStarch:
@timcanham I recommend we merge this.  Note: the link order is wrong in the modules.mk definitions, and we'll need to update that.

PR #37, opened 2019-05-18
CMake + RPI Demo Updates

rdaruwala:
I've made the following modifications:

1. Added if/else statements into `Os/CMakeLists.txt` to support compiling on other platforms + cross-compiling with the Raspberry Pi
2. Added CMake compile capabilities to the components in `Drv`/`Svc/LinuxTimer` and the `RPI` demo application
3. Updated the RPI demo application documentation to include a small snippet about compiling & cross-compiling using CMake

This is one of my first times working with CMake so I'd love advice on how to optimize some of the `CMakeLists.txt` files, especially the one in `Os` which contains a lot of repetitive includes between the various platforms


LeStarch:
I'll try to test this tomorrow and respond.  However, if you'd jot down your notes on the usability of the cake build system, I'd be very grateful.

Os/CMakeLists.txt
```diff
@@ -1,32 +1,149 @@
-####
+  ####
```
LeStarch: Superfluous whitespace added?

LeStarch:
@rdaruwala did you try building for the Ref app on Linux and/or MacOS?  Did you run the UTs? Nice work, I just would like to know how much testing you have done in general

I still need to run the build, but the style looks fine.

LeStarch:
@rdaruwala for clarity, would you recommend leaving the Os CMake lists as are, or break them into separate includes?  Since Linux or Mac all look the same for many of these, should we collapse the if cases or leave them explicit?  I'd appreciate your opinions.

rdaruwala:
I thought that the CMake README was very thorough and did a great job walking me through how to compile. No complaints there.

I built Ref on both Linux and MacOS. Built the RPI demo on Linux and ran it on my Raspberry Pi (cross-compiled), though I only made sure that it ran and didn't check for functionality/correctness. I didn't run any UTs, if you'd like I can address them this weekend. 

I first attempted to separate out the common source files by having two "Sets": One before the conditional statements for all the common files, and then the platform-specific ones being included when needed. I believe I ran into some errors when trying to implement it so I reverted to having each platform's entire source file set being included in the conditional. 

My 2 cents would be that if there are source files that are common for all the different platforms that F' supports, they should be included together once before the platform-specific checks. I think it would help keep the file smaller as more platforms are added as well as make it easier to update if there are future changes to one of those common files.

LeStarch:
@rdaruwala I verified your build runs on RPI. Nice work! I am running it through our CI today to look through problems.  Otherwise, it is looking good.

rdaruwala:
Thank you! Sounds good. Let me know if anything comes up during the CI verification

LeStarch:
@rdaruwala I'll have to run the test manually.  Looks like there is an issue with the CI system.

LeStarch:
@rdaruwala looks like the available tests pass.

LeStarch:
@timcanham what do you think of the PR, it adds CMake support to the RPI package.  It passes builds, manual RPI check, UTs, and UTs build individually.

timcanham:
@LeStarch @rdaruwala Looks good! If you two can certify it's good, let's promote it.

LeStarch:
I already approved.  Merging now.

PR #53, opened 2019-06-17
mstarch: adds tcp server binary, better logging.

LeStarch:
@timcanham this adds a run_tcpserver script for running the server independently.  It also prints the log file should an error occur at the startup of the server.


LeStarch:
This doesn't work.  Use 'a' for appending.

LeStarch:
This is no longer useful.

PR #70, opened 2019-09-03
GDS Stripchart Fixes

Oxymoren:
I've been experimenting with the Ref application, and ran across a few issues in the Stripchart section of GDS (TK Implementation). 

Here's the issues and steps taken to fix them: 

- Stripchart was not getting any message data.
  - Add message passing in the channel listener. A StripChartListener comment said it should be called from there.

- The "Channel Telemetry Filter Selection" GUI was cutting off buttons. 
  - Expand the window vertically to 450px.

- Clicking the Settings button caused an unhanded exception.
  -  A config field "stripchart_update_rate" wasn't initialized. Added the field and removed a duplicate field.  

- Clicking the Save All button caused an unhanded exception. 
  - Fixed the loop that gathered the graphs. It was digging too deeply into the data structure.

- Canvas.show() is depreciated.
  - Followed TK's warning and switched to canvas.draw().

Also, I don't see any contribution guidelines for this repository. Please let me know if you need any more information, would like to split the PR, create issues, switch target branches, etc.

### DIFF
--- a/Gds/src/fprime_gds/tkgui/controllers/channel_listener.py
+++ b/Gds/src/fprime_gds/tkgui/controllers/channel_listener.py
@@ -142,6 +142,9 @@ class ChannelListener(consumer.Consumer):
       # Descriptor and length of message have already been decoded
       self.__status_bar_updater.put_data((8 + len(data), 0))
 
+      # Pass message to the stripchart panel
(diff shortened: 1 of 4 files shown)


LeStarch:
Looks ok to me.

PR #82, opened 2019-09-17
don't try to commit build directories

SterlingPeet:
This change adds the standard cmake ``build`` or ``build-and_some_other_stuff`` formatted folders to the ``.gitignore``, which makes it harder to accidentally commit artifacts back into the repository.

Most people who are even marginally familiar with the cmake build process, even if all they have done is just build and install a project, use this convention.  This is a convenience to avoid having to sort through a large number of files in your git status for files that actually need to be committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+build
+build*
+
 AutoXML/
 test_harness/src/test_harness-C/application.out
 /.project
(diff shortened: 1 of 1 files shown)


LeStarch:
We should add `cmake-build` and `cmake-build*` too.  Merging this, no code changes.

@SterlingPeet thanks!!!

PR #83, opened 2019-09-18
Devel

tlee753:
Working updates for local build raspberry pi tutorial with some conditions:
after `make rebuild_rpi`, you have to go to the `Gds/src/` directory to run (`../../RPI/scripts/run_rpi.sh`) because spawning the lxterminal for the app doesn't maintain environment variables (namely `PYTHONPATH`) from the previous terminal. If you spawn from `Gds/src/` then your new terminal will have the `appwrapper` module in its path.

If you run in gnome-terminal, environment variables are kept in a spawned terminal, but I think this is the exception just from messing around with my mac/lxterminal on ubuntu. To maintain the environment variables and keep everything isolated, you may want to `printenv > ~/env.temp`, then source that file from the spawned window, then delete the file.

Minor update to the `run_deployment.py` file as a `proc.returncode` of `0` is a success. A few other clean ups just to make things easier.


Gds/src/fprime_gds/executables/run_deployment.py
```diff
@@ -222,7 +222,7 @@ def kill_wait():
     print("[INFO] Waiting 5 seconds for {0} to start".format(name))
     time.sleep(5)
     proc.poll()
-    if proc.returncode is not None:
+    if proc.returncode:
```
LeStarch: What is the rationale behind not running this code if it returns 0?  A zero may be a successful completion of the program, but it is still not running after 5 seconds.

LeStarch: Thus, what was intended to be a long-running process has exited after only 5 seconds, which to me represents an error condition.

RPI/README.md
```diff
@@ -43,7 +45,7 @@ If you are cross-compiling, this would apply only to the compile host:
 ```
 ./mk/os-pkg/ubuntu-packages.sh
 sudo pip install -r  mk/python/pip_required_build.txt
-sudo pip install -r  mk/python/pip_required_gui.txt
+sudo pip install -r  mk/python/pip_required_gse.txt
```
LeStarch: These are fine for now, but eventyally will need to be changed to (or the sudo equivalent):
```
pip install -r Autocoders/Python/requirements.txt
pip install -r Gds/requirements.txt
```

RPI/README.md
```diff
@@ -31,6 +31,8 @@ If the UART port is not set up correctly, there will be a file open error.
 |IO23|IO25|GPIO|
 |IO24|IO17|GPIO|
 
+ ![`Raspberry Pi Pinout`](img/pi_pinout.png "Pi Pinout")
```
LeStarch: Nice.  Thank you!

RPI/scripts/run_rpi.sh
```diff
@@ -0,0 +1,17 @@
...
+DIRNAME="`dirname $0`"
+
+# Set BUILD_ROOT if unset or "" set the BUILD_ROOT to be the above dir
+if [ -z ${BUILD_ROOT} ]
+then
+    export BUILD_ROOT="`cd ${DIRNAME}/../..; pwd`"
+fi
+
+DEPLOY=`cd ${DIRNAME}/..; pwd;`
+${BUILD_ROOT}/Gds/bin/run_deployment.sh --deploy "${DEPLOY}" "$@"
```
LeStarch: Does this line need to add the `-n` flag, as the Ground Station is typically not run on the RPI, and thus should not attempt to run the FSW.

tlee753:
Thanks, you and Tim are doing awesome work! These changes are designed around running the whole demo, ground software included on the pi, just an update after going through the readme. But I agree with you, the ground station really shouldn't be run on the pi.

I'll try it in a minute on mac and xterm on ubuntu, but the environment variables problem seems like a decent concern. Are you running on gnome or another platform?

RPI/README.md
tlee753: I'll try this on a clean pi install, likely should just be updated to what you have :)

timcanham:
FYI, originally I had a setup to run the FSW + GUI on the Pi, but we had a number of cases where it just wouldn't work for mysterious reasons, so we gave up and told people to run it on another machine.

tlee753:
I saw your issue, I think the issues you were having were the pythonpath environment variable not existing in the new terminal as well as the app not starting because the process of the appwrapper around the app has a return code of 0 whereas run_deployment.py does a kill if its not "None." It works fine for me if a spawn the new terminal from the path with the appwrapper submodule (Gds/src/) but thats obviously a hacky solution...

ljreder:
Good point I think a great solution would use setup tools and install all our python into virtual environment.  No more paths to worry about.  In fact I @mstarch might even already have some of this working.   

Leonard J. Reder

RPI/scripts/run_rpi.sh
tlee753: I think I'm confused, doesn't -n not run the app and only run the ground station/tcpserver?

tlee753:
Agreed, that's a better solution then my "exporting environment variables to a file, and sourcing that file in the new terminal before deleting" proposal from above. I actually tried to implement my solution but the spawned terminal is really only meant to run command it seems. I knew there was a better solution out there to keep things isolated (not editing everyone's bashrc's) just couldn't think of it.

LeStarch:
Another possible solution is to pipe the logging output to a tab in the ground system.  Thus removing the dependency on a terminal, as these terminals are all implemented differently.  Then it is a raw-run of the application and the output is captured into the Gds.

If a user wants or needs a terminal, they can run the software in their own terminal, as historically done.

RPI/scripts/run_rpi.sh
LeStarch: It does, I assumed you were running on the ground, not the pi.

Gds/src/fprime_gds/executables/run_deployment.py
tlee753: Can you please explain None vs 0 to me? The lxterminal process gives a 0 which I know as the "successful" return code, I couldn't quite figure out how why different terminals where giving me None vs 0

LeStarch: It isn't the terminal.  `0` means the process ended in success.  A non-zero integer means the process ended in failure.  However `None` means the process is still running.  We want the process to continue running, so `None` is the only correct value, as that means the terminal and thus the FSW is still running 5 seconds later.  Any exit 0 or non-zero before that means the FSW stopped running, and thus an error occurred.

tlee753: Ah, makes more sense, appreciate the explanation. The app is still running and continues to run in lxterminal when it returns 0 (using above change so it doesn't exit), not sure why it considers this a successful exit if its still running.

ljreder:
Yeah but will not be able to pipe the output as I and others like to do.

Leonard J. Reder

Gds/src/fprime_gds/executables/run_deployment.py
tlee753: This should definitely stay as is, lxterminal is just different.

tlee753:
I'm closing this PR as its not worth the hassle on the pi which shouldn't be running the ground station anyway. I tried using virtualenv but the spawned terminal isn't in the virtualenv and thus the environment variables aren't passed along. Also, large scientific packages like scipy don't seem to work well in virtual environments, especially on a pi (my pi 3 froze entirely). Env variables seemed to be passed along just fine on mac/gnome-terminal/xterm so just a lxterminal bug to take note of.

PR #87, opened 2019-10-21
Add I2C component and unit test

aaronbuchwald:
Add LinuxI2CDriver component and a unit test for it

Unit test depends on having hardware attached. This unit test was developed and run while using the ArduCam multiplexer here: https://www.arducam.com/arducam-multi-camera-multiplexer-raspberry-pi-4-application/. Which is at address 0x70. In order to be run successfully, a device at address 0x70 must be attached or reads and writes will obviously fail.


LeStarch:
First of all, thanks for taking a swing at this.  I'll add comments using the review feature.

Drv/LinuxI2CDriver/Events.xml
```diff
@@ -0,0 +1,89 @@
+<events>
+    <!-- Open error -->
+    <event id="0" name="GP_OpenError" severity="WARNING_HI" format_string = "Error opening GPIO device %d: %d (%s)" >
```
LeStarch: All you events are using the GPIO naming convention.  You should rename these to be more clear for I2C

Drv/LinuxI2CDriver/Events.xml
```diff
@@ -0,0 +1,89 @@
...
+        <comment>
+        Open error
+        </comment>
+        <args>
+            <arg name="gpio" type="I32">
```
LeStarch: You'll need to go through this whole file, and ensure that it is for I2C not GPIO.

Drv/LinuxI2CDriver/Makefile
```diff
@@ -0,0 +1,22 @@
+# This Makefile goes in each module, and allows building of an individual module library.
```
LeStarch: We'll need to add a CMakeList.txt for this component too.

Drv/LinuxI2CDriver/old-test/mod.mk
```diff
@@ -0,0 +1 @@
+SUBDIRS = ut
```
LeStarch: Is `old-test` supposed to be checked in or has it been replaced by `test`?

Drv/LinuxI2CDriver/LinuxI2CDriverComponentAi.xml
```diff
@@ -0,0 +1,15 @@
...
+<?xml-model href="../../Autocoders/Python/schema/ISF/component_schema.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
+
+<component name="LinuxI2CDriver" kind="passive" namespace="Drv" modeler="true">
+
+    <import_port_type>Drv/LinuxI2CPorts/LinuxI2CWritePortAi.xml</import_port_type>
+    <import_port_type>Drv/LinuxI2CPorts/LinuxI2CReadPortAi.xml</import_port_type>
+    <ports>
+        <port name="write" data_type="Drv::I2CWritePort" kind="guarded_input">
+        </port>
+        <port name="read" data_type="Drv::I2CReadPort" kind="sync_input">
```
LeStarch: If you use a sync_input here, won't multiple reads collide?

Drv/LinuxI2CDriver/LinuxI2CDriverComponentImpl.cpp
```diff
@@ -0,0 +1,90 @@
...
+// ======================================================================
+
+
+#include <Drv/LinuxI2CDriver/LinuxI2CDriverComponentImpl.hpp>
+#include "Fw/Types/BasicTypes.hpp"
+
+#define I2C_FILE (char*)"/dev/i2c-1"
+
+int init_address(int addr) {
+  int fd = open(I2C_FILE, O_CLOEXEC | O_RDWR);
```
LeStarch: This open call appears to never be closed.  Also why `O_CLOEXEC`?

Drv/LinuxI2CDriver/LinuxI2CDriverComponentImpl.cpp
```diff
@@ -0,0 +1,90 @@
...
+  FW_ASSERT(fd >= 0);
+  int ioctlRes = ioctl(fd, I2C_SLAVE, addr);
```
LeStarch: You use raw `int` types here?  Perhaps these should be made into `NATIVE_INT_TYPES`?

Drv/LinuxI2CDriver/LinuxI2CDriverComponentImpl.cpp
LeStarch: Also, are you required to open the file anew for every call, or can it be opened once and shared with all other callers.

Drv/LinuxI2CDriver/LinuxI2CDriverComponentAi.xml
aaronbuchwald: Should this be switched to protected input then?

Drv/LinuxI2CDriver/LinuxI2CDriverComponentImpl.cpp
aaronbuchwald: I chose to use O_CLOEXEC because this means that it will be closed on the execution of a read or write operation, so it is intended to be used once. Is it preferred that an i2c address be opened, then used for transactions, and then closed? If so, I could switch to this flow and remove the O_CLOEXEC, but I thought that this simplified the API in a very nice way.

Drv/LinuxI2CDriver/LinuxI2CDriverComponentImpl.cpp
aaronbuchwald: Do you mean uint_8 or one of the NATIVE_INT_TYPES defined by F Prime? If the latter, what's the syntax to use it?

Drv/LinuxI2CDriver/Events.xml
aaronbuchwald: Ok will do

Drv/LinuxI2CDriver/old-test/mod.mk
aaronbuchwald: Meant to be removed, I will delete it and update the branch

Drv/LinuxI2CDriver/Makefile
aaronbuchwald: Can I easily utilize a CMakeList.txt file from another driver and only change a few things?

Drv/LinuxI2CDriver/Events.xml
aaronbuchwald: Where can I find documentation for creating an Events.xml file?

genemerewether:
@LeStarch - not sure if you want to pull it in / unify this, but here are I2C ports and driver (for Hexagon DSP on Snapdragon, but shell code is in place for Linux):

https://github.com/quest-fw/quest-fw/tree/quest-master/Drv/LinuxI2CDriver
https://github.com/quest-fw/quest-fw/tree/quest-master/Drv/I2CDriverPorts

Drv/LinuxI2CDriver/Events.xml
LeStarch: You are not emitting events, it is best to remove the file.

Drv/LinuxI2CDriver/LinuxI2CDriverComponentAi.xml
LeStarch: Use a `guarded_input` like the write port.

Drv/LinuxI2CDriver/Events.xml
LeStarch: Just delete this file, it is not used.  Nor is it included in your ComponentAi file.

Drv/LinuxI2CDriver/LinuxI2CDriverComponentImpl.cpp
LeStarch: https://github.com/nasa/fprime/blob/master/Fw/Types/BasicTypes.hpp defines the types, but here is a quick rundown.

```
U<bit width> defines unsigned int of that size. i.e. U8, U16, U32, and U64
I<bit width> defines signed int of that size. i.e. I8, I16, I32, and I64
F32, F64 are float and double.
```

Drv/LinuxI2CDriver/Makefile
LeStarch: Yes, you can!  Grab any CMakeList.txt from the drivers directory.  The key portions
```
set(SOURCE_FILES
...Ai.xml and cpp files needed
)
add_fprime_module()
```
We'll avoid the UTs for now, as they are run automatically in CI, and that CI doesn't have an I2C.  So we'll take that discussion offline for now.

Drv/LinuxI2CDriver/old-test/mod.mk
LeStarch: Thanks.

LeStarch:
@genemerewether thanks for the heads up.  I'll add an issue to do this.

LeStarch:
@aaronbuchwald did you fix this?

Drv/LinuxI2CDriver/LinuxI2CDriverComponentImpl.cpp
timcanham: This is an older PR, but we should make the interface consistent with the other drivers. The pattern there is:

```
open(filename)
read...
write...
destructor closes file
```
I don't think the file name should be hard coded in case a linux flavor uses a different file name.

Having the port guarded allows the `ioctl` call and then the `read`/`write` to be atomic.

Issue #93, opened 2019-11-01
Make Build Error


acxz:
I am trying to build fprime and I encountered the following issue:
```
[  0%] Generating ../../../../../../../../Autocoders/Python/src/fprime_ac/generators/templates/topology/includes1TopologyCpp.py, ../../../../../../../../Autocoders/Python/src/fprime_ac/generators/templates/topology/includes1TopologyH.py, ../../../../../../../../Autocoders/Python/src/fprime_ac/generators/templates/topology/publicInstanceTopologyCpp.py, ../../../../../../../../Autocoders/Python/src/fprime_ac/generators/templates/topology/publicInstanceTopologyH.py, ../../../../../../../../Autocoders/Python/src/fprime_ac/generators/templates/topology/publicTopologyCpp.py, ../../../../../../../../Autocoders/Python/src/fprime_ac/generators/templates/topology/publicTopologyH.py, ../../../../../../../../Autocoders/Python/src/fprime_ac/generators/templates/topology/publicTopologyID.py
/bin/sh: cheetah-compile: command not found
make[2]: *** [F-Prime/Autocoders/Python/src/fprime_ac/generators/templates/CMakeFiles/codegen.dir/build.make:276: ../Autocoders/Python/src/fprime_ac/generators/templates/topology/includes1TopologyCpp.py] Error 127
make[1]: *** [CMakeFiles/Makefile2:309: F-Prime/Autocoders/Python/src/fprime_ac/generators/templates/CMakeFiles/codegen.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

Steps to reproduce:

1. Clone fprime
2. cd fprime
3. mkdir build_dir
4. cd build_dir
5. cmake ..
6. make
7. See error above

System:
OS: ArchLinux
python-cheetah3: 3.2.4

I believe the issue is caused because `python-cheetah3` ships with `cheetah-compile3` instead of `cheetah-compile` as in `python2-cheetah` in ArchLinux. I checked the Debian package and it is the same thing, Debian `python3-cheetah` ships with `cheetah-compile3` instead of `cheetah-compile`. (https://packages.debian.org/de/sid/amd64/python3-cheetah/filelist) This makes me believe that `fprime` does not build under the latest python (v3) at all. I can change all the `cheetah-compile`s to `cheetah-compile3`s but that would make it incompatible for older pythons. Maybe we can add a check every time we need to call `cheetah-compile` to make sure the correct one is being used? I am not sure about a good solution for this issue.

[...]

PR #94, opened 2019-11-02
Cheetah compile py3 fix

acxz:
This allows template code generation to occur using a system installation of `python-cheetah3`. 
We check if we are using python 2 or if we not using a system install of python, i.e. if we are using a venv. If that is satisfied then we use `cheetah-compile`. Else if we are using a system install of python3 then we use `cheetah-compile3`. This fixes build issues on Arch and Debian based systems.

One caveat is that if you are building `fprime` with python 3 outside a venv/virtualenv then you cannot use `pip`'s version of `cheetah3`. This is fine though, since using pip at the system wide level is not recommend in the first place. It should only be used in virtualenvs or at the user install level.

Confirmed with both system wide python3 install and virtualenv python3 install. Have not tested with python2, but I am sure it works.

Note: I could not think of a better way to check if the python interpreter was not a system one besides checking if the path starts at `/usr`. There is a CMake variable `Python_FIND_VIRTUALENV`, but that only works if you are already inside a virtualenv, defeating the purpose of the check in our case.

Resolves: #93

### DIFF
--- a/cmake/support/AC_Utils.cmake
+++ b/cmake/support/AC_Utils.cmake
@@ -21,7 +21,7 @@ function(cheetah CHEETAH_TEMPLATES)
   # forces cheetah templates to be compiled and up-to-date before running the auto-coder.
   string(REPLACE ".tmpl" ".py" PYTHON_TEMPLATES "${CHEETAH_TEMPLATES}")
   # Setup the cheetah-compile command that runs the physical compile of the above statement. This
-  # controls the work to be done to create PYTHON_TEMPLATES from ${CHEETAH_TEMPLATES. 
+  # controls the work to be done to create PYTHON_TEMPLATES from ${CHEETAH_TEMPLATES.
   add_custom_command(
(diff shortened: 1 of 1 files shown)


LeStarch:
Why not use a FIND_PROGRAM call?

https://cmake.org/cmake/help/v3.0/command/find_program.html

```
find_program(CHEETAH_EXE NAMES "cheetah-compile" "cheetah-compile3")
```
Then you don't need all the checks.

acxz:
Ah that is much smarter, I'll look into it later this week and see if it resolves the issue for me.

acxz:
I messed up the commits on my branch, but the force push should clean up the commit history when it gets merged. @LeStarch, let me know if you want me to create a cleaner PR for this.

acxz:
@LeStarch Can we provide closure to this PR?

LeStarch:
I'll take a look at this.

PR #105, opened 2019-12-02
 Fix Dockerfile path issues

matzipan:


mk/docker/README.md
```diff
@@ -70,9 +70,9 @@ implications in mounting volumens into Docker.
 
 ### Build the F´Docker Image
 ```bash 
-fprime>docker build -t fprime -f docker/Dockerfile mk/
+fprime>docker build -t fprime -f mk/ocker/Dockerfile .
```
LeStarch: Should say `mk/docker/` yes? Otherwise I concur with the changes.

PR #108, opened 2019-12-02
Fix Dockerfile path issues

matzipan:


matzipan:
Opened as new merge request because it was quicker to do it in the github webinterface like this.

PR #116, opened 2019-12-06
Move CMake builds to python3

matzipan:
When calling the python command there is no guarantee that your script will run with Python 3.  Therefore, explicitly call python3 for cross-platform support.

This PR only touches the scripts called by the Ref deployment in the CMake flow.


timcanham:
Good idea!

PR #118, opened 2019-12-12
lestarch: installation and utility bug fixes

LeStarch:
This adds quick fixes for installation and running on fresh systems.  Namely it does the following:

1. `fprime-util` checks for `cmake` before trying to run.  Corrects a confusing error when cmake is not installed.

2. Fixes python 2.7 install to fix scipy errors.


LeStarch:
Merging.

PR #139, opened 2020-02-14
lestarch: this adds troublshooting to the INSTALL guide.

LeStarch:
@timcanham what do you think of these additions?  Issues we saw from JPLers and others trying to install are now covered in this new section.

Do you think this is ok to tack-on at the end of the document?


LeStarch:
@timcanham, fixed.

Issue #71, opened 2019-09-08
CMake System is not dependent on AcConstants.ini


pelmini:
After editing AcConstant.ini CMake did not detect the change in the file.

LeStarch:
Affects `cmake` system.  Work around is to regenerate cmake build from scratch.

LeStarch:
Current work around: delete CMake build directory and start over.

LeStarch:
Still broken, prioritizing fix.

LeStarch:
Fixed.

PR #151, opened 2020-03-03
lestarch: adding ini dependency to autocode calls, fixes nasa/fprime#71

LeStarch:


LeStarch:
@timcanham can you take a glace?  This will force a dependency on the Ac.ini file.  Projects may override by setting:

PROJECT_AC_CONSTANTS_FILE variable

LeStarch:
Approved by @timcanham .

Issue #103, opened 2019-12-02
Building reference application with CMake doesn't get past generating the application topology


matzipan:
```
andrei@PC-40:~/Workspace/fprime/build$ cmake ../Ref/
-- The C compiler identification is GNU 8.3.0
-- The CXX compiler identification is GNU 8.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- F prime core directory set to: /home/andrei/Workspace/fprime/cmake/..
-- Target build platform: Linux
-- Including /home/andrei/Workspace/fprime/cmake/platform/Linux.cmake
-- Requiring thread library
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- [VALIDATION] Running: python;/home/andrei/Workspace/fprime/cmake/support/validation/pipsetup.py
-- F prime BUILD_ROOT currently set to: /home/andrei/Workspace/fprime/cmake/..
-- Installation directory: /home/andrei/Workspace/fprime/Ref
-- Adding library: Fw_Buffer
-- 	Found serializable: Buffer from BufferSerializableAi.xml
-- 	Found port: BufferGet from BufferGetPortAi.xml
-- 	Found port: BufferSend from BufferSendPortAi.xml
-- Adding library: Fw_Com
-- 	Found port: Com from ComPortAi.xml
-- Adding library: Fw_Cmd
-- 	Found port: Cmd from CmdPortAi.xml
-- 	Found port: CmdReg from CmdRegPortAi.xml
-- 	Found port: CmdResponse from CmdResponsePortAi.xml
-- Adding library: Fw_Log
-- 	Found port: Log from LogPortAi.xml
-- 	Found port: LogText from LogTextPortAi.xml
[...]

PR #152, opened 2020-03-03
lestarch: formalizing Python 3.5+ support. Other versions are end-of-life, resolves nasa/fprime#103

LeStarch:
This moves our requirement to Python 3.5+.  This is consistent with Python's end-of-lifed packages.

See: https://devguide.python.org/#status-of-python-branches


LeStarch:
@timcanham has approved.

PR #154, opened 2020-03-04
Fix uninformative exception in TopoFactory.py

aaronbuchwald:
Fix blank exception thrown while reading Topology file


aaronbuchwald:
Also this exception is being thrown after changing all of the base ids, so that there is no longer a conflict. Would it be possible to change this, so that when all of the base ids in the current version of the topology file do not overlap there is no exception thrown at all?

LeStarch:
Great!  Let me take a look.

LeStarch:
Looks good to me.  I see a better message in the Exception when violated.  Thank you!

LeStarch:
I switched the input branch to `devel`

PR #159, opened 2020-03-08
Fix incorrect path to rpi cross compiler

aaronbuchwald:
Fixes an extra slash at the end of the path to RPi cross compiler in raspberrypi.cmake toolchain


aaronbuchwald:
Closing to instead make PR to devel

PR #160, opened 2020-03-08
Fix incorrect path to rpi cross compiler

aaronbuchwald:


LeStarch:
Not opposed to this, but why is it needed?

aaronbuchwald:
The way it was before it resulted in there being two slashes in the path to the rpi compiler

PR #163, opened 2020-03-24
Implement LinuxI2CDriver

aaronbuchwald:
Happy to add a unit test for this, could you point me in the direction of the new documentation for the unit test framework?


timcanham:
Since this is a more generic *inux driver, it should be put in the `Drv::` name space instead of `RPI::`.

aaronbuchwald:
It doesn't allow me to build when I put the driver in the Drv subdirectory. How do I get it to build when adding a component in the Drv directory?

LeStarch:
I think we have an I2C driver now.  I will close this PR, but feel free to reopen if more is needed.

PR #178, opened 2020-06-04
Fixes in preparation for release 1.4.1

LeStarch:


LeStarch:
@timcanham can we approve/merge this for devel?

Utils/TokenBucket.cpp
```diff
@@ -0,0 +1,134 @@
+// ====================================================================== 
+// \title  TokenBucket.cpp
+// \author vwong
+// \brief  cpp file for a rate limiter utility class
```
timcanham, line 4: Comment is incorrect.

LeStarch: I'll file a ticket with the submitters.

PR #179, opened 2020-06-09
lestarch: GitHub Actions Based CI

LeStarch:
This PR adds in CI through GitHub Actions.  This will allow all future PRs to be auto-checked.


LeStarch:
Also contains some new contributions to Os.

PR #180, opened 2020-06-09
lestarch: adding tutorial remove orphaned files

LeStarch:


docs/Tutorials/GpsTutorial/Tutorial.md
```diff
@@ -817,7 +817,9 @@ application's files and modify them to construct our topology.
 From the `GpsApp` directory run the following:
 ```shell
 cp -r ../Ref/Top ./Top
+rm ./Top/RefTopologyAppDictionary.xml ./Top/RefTopologyAppAi_IDTableLog.txt ./Top/RefTopologyAppID.csv
 ```
+**Note:** do not worry if the removed files do not exist, the remove is there to clean-up files generated during a build.
```
timcanham, line 822: Why do we have to remove them? I never do.

LeStarch: If you leave a generated dictionary file in that directory, there is a chance that dictionary file will be read instead of the newly generated one for this deployment.  Thus, if you will clone-and-own, you need to remove the dictionary file at least.

timcanham: Okay, I get it. I thought this was a general instruction for cleaning up. This is directions for cloning the directory as an example for making your own deployment.

Issue #173, opened 2020-05-21
Math Component F' Tutorial Section 4 GDS Error


csmith608:
In section 4 when the ground system is running and the MS_DO_MATH command is sent, the  error below is produced.  Other commands are sent and information shows up correctly.  
<img width="1333" alt="Screen Shot 2020-05-21 at 2 52 16 PM" src="https://user-images.githubusercontent.com/63731123/82603969-87d21500-9b81-11ea-9aad-731212002e95.png">

JHDeerin:
I'm getting the same error, but am also getting it for any command that takes a typed-in argument (e.g. `MR_SET_FACTOR_1`,  `factor2_PRM_SET`, `cmdDisp.CMD_TEST_CMD_1`, etc.). Math commands that don't take an argument, as well as commands with drop-down menus for selecting arguments, etc., seem to work fine and are recorded in the event log.

kbiesiadecki141:
Any update on this? I'm coming across this exact error only for commands that take arguments (mathSender.MS_DO_MATH, mathReceiver.MR_SET_FACTOR1, etc.) like JHDeerin above

LeStarch:
Yes.  It seems to be an issue with float and double arguments.   A fix is in progress.

LeStarch:
Once number #182 is merged, this problem should be fixed.  Nothing y'all did wrong.

LeStarch:
Closing this issue.

lewisjaggi:
Hello,
I did the tutorial yesterday and I encountered this problem.
Can you tell me what can i do to fix it ?

PR #182, opened 2020-06-12
Update/float argument handling

LeStarch:
Fix for issue: #173 

There is still a bit of an issue with validating input in the HTML browser, but it doesn't prevent sending commands....just highlights them wrong.  Will look into this next.


LeStarch:
@timcanham this is blocking some users.  Can we merge?

PR #185, opened 2020-06-16
Feature/fprime gds cli

JHDeerin:
## F´ Pull Request
|**Field**|**Value**|
|:---|:---|
|**_Submission Date_**| June 16th, 2020 |
|**_Submitter_**| Jake Deerin (JHDeerin) |
|**_Originating Project/Creator_**| F' |
|**_Base Branch_**| `devel` |
|**_Short Description_**| Added basic-functionality CLI tool for interacting with the GDS |
|**_Affected Component_**| GDS |
|**_Affected Architectures(s)_**| *N/A*  |
|**_Related Issue(s)_**|  *N/A* |
|**_Has Unit Tests (y/n)_**| y (but only ~30% coverage, if that) |

---
## Change Description

This adds a functional command-line interface tool that developers can use to interact with the GDS, viewing telemetry data, events, and available commands, and sending commands through the GDS. **This tool is functional but not complete, and all interfaces/output formats are subject to change.**

After cloning, the tool can be installed and used as follows:

1. `cd` into the `Gds` folder
2. Run `python setup.py install`
3. Run `fprime-cli -h` to run the tool and view the help menu for further instructions

## Rationale

A CLI for the GDS would be a useful supplement to the existing GUI, making testing easier over the network or in other situations where it can be tedious to set up a GUI, as well as providing developers who prefer the command line an alternate option. Interfacing with the existing UNIX tool suite allows for massive flexibility when handling GDS output.

## Testing Recommendations

- A cursory check of the tool's coding style/organization should be done to ensure the tool will remain straightforward to maintain as F' evolves.
- More unit tests should be added to `Gds/tests/fprime_gds/common/gds_cli` to cover the remaining commands and ensure functionality is working and stable.
- Furthermore, brief UX testing is recommended to ensure the tool meets developer needs, isn't missing important tools, and is optimized for ease-of-use with the most important functions.

## Future Work

- Tuning the output format to be usable by testers is important for the tool to actually be useful.
- Specifying the dictionary location each time is annoying; being able to keep this as an option while having a configuration file/etc. to set a default project location would alleviate this.
- Adding quality-of-life features like tab completion, manual warning limits, colored output, etc. would also be helpful if time permits.
- Due to having to connect to the API anew for each command, CLI performance is somewhat sluggish; finding a way around this would be nice.


LeStarch:
@JHDeerin can you merge with "hunterpaulson:black".

JHDeerin:
> @JHDeerin can you merge with "hunterpaulson:black".

@LeStarch I don't think I can edit this pull request to do that (it isn't showing up as an option), so I think I'd need to delete this one and open a new PR merging into Hunter's fork. Is that okay?

LeStarch:
@JHDeerin, merge his branch into yours.

```
git remote add hpaulson https://github.com/hunterpaulson/fprime.git
git fetch hpaulson
git merge black
```

JHDeerin:
Ah, understood. Should've just merged his branch into this one.

Gds/setup.py
```diff
@@ -33,6 +33,8 @@
 from setuptools import find_packages
 from setuptools import setup
 
+import src.fastentrypoints
```
LeStarch, line 36: Why is this imported as src.fastentrypoints?

LeStarch: Are you running pip install Gds/ or python setup.py install or both?  Does this file work for both?

Gds/setup.py
```diff
@@ -85,7 +87,8 @@
     # standard use of utilities that ship as part of F prime.
     ####
     entry_points={
-        "gui_scripts": ["fprime-gds = fprime_gds.executables.run_deployment:main"]
+        "gui_scripts": ["fprime-gds = fprime_gds.executables.run_deployment:main"],
+        "console_scripts": ["fprime-cli = fprime_gds.executables.fprime_cli:main"],
```
LeStarch, line 91: Awesome!

Gds/src/fprime_gds/common/gds_cli/channels.py
```diff
@@ -0,0 +1,60 @@
...
+            item_dictionary=project_dictionary.channel_id,
+            search_filter=filter_predicate,
+            template_to_data=ChData.get_empty_obj,
+        )
+        return channel_list
+
+    @classmethod
+    def _get_upcoming_item(
+        cls,
+        api: IntegrationTestAPI,
```
LeStarch, line 50: Type hinting....nice!

Gds/test/fprime_gds/common/gds_cli/TestDictionary.xml
```diff
@@ -0,0 +1 @@
+<!-- Blank as a blocked writer's book -->
```
LeStarch, line 1: What is this for?

docs/INSTALL.md
```diff
@@ -96,17 +96,21 @@ fprime-util install --jobs 32
 fprime-gds -g html -d <path to fprime checkout>/Ref
 ```
 **Note:** `Ref` should contain pre-built dictionaries and binaries for the user's system. This can
-be achieved by running the autocoder installation test (the user must have a working autocoder
+be achieved by running the Autocoder installation test (the user must have a working Autocoder
 installation).
 
+## (Optional) Installing Tab Completion
+
+Several of F´s command line utilities support tab completion. To enable these tools to use it, see the [instructions here](../Gds/docs/gdsCLI/AUTOCOMPLETE.md).
```
LeStarch, line 104: You should put this document in the normal `docs` folder.

`fprime/docs/gds/AUTOCOMPLETE.md`

LeStarch:
@JHDeerin we'll need to move that document into the top-level `docs` path so it is published correctly.

Also I had a few comments, and accolades.

Gds/test/fprime_gds/common/gds_cli/TestDictionary.xml
JHDeerin: Ah, this was just for some parser testing functionality (the parser would exit early if it couldn't find an XML or .py dictionary, so I added this as a dummy dictionary for that). I'll try to remove this (or put something more descriptive inside at least if I can't delete it without breaking tests)

docs/INSTALL.md
JHDeerin: Can do!

JHDeerin: Should we move the Test API documentation too, or is that different because it isn't needed for the installation process?

Gds/setup.py
JHDeerin: Hey Michael, there wasn't any particular reason I did this; that's just where I put this script and this was how I knew to import it. Both methods work for me if I run this command from within the Gds directory. If I run it from the root, `pip install Gds/.` works fine, but `python Gds/setup.py install` fails.

Weirdly, even if I completely remove this line, calling `python Gds/setup.py install` fails with the following message until I run it from within the Gds folder:

```bash
(fprime-venv) vagrant@vagrant:~/jpl/fprime$ python Gds/setup.py install
running install
running bdist_egg
error: error in 'egg_base' option: 'src' does not exist or is not a directory
```

JHDeerin:
@LeStarch Just pushed some changes to address those issues + Rob's openpyxl issues, and to speed up the code a bit.

While I was doing that, there were 2 weird optimization things I noticed that we might want to look into further:

1. When running the code, the biggest impact on import time comes from the Test API's `api.py`, at ~280ms - but when I did some profiling, the openpyxl imports in `test_logger.py` account for *250ms of that time*. Now that the code doesn't require openpyxl, it actually runs noticeably faster if openpyxl is uninstalled; finding some way to delay that import until logging is actually used might be worthwhile (since it's pretty unintuitive to punish users for having an extra pip package installed).

2. When actually using the Test API, you can see in the image below (from calling `command-send`) that a good chunk of time is spent trying to *disconnect,* and specifically waiting for a thread to join. If this is expected then that's alright, but if we can figure out why this takes so long on `client_socket` and make it faster, sending commands would be substantially faster, and polling would have less overhead.

command_send_timings

LeStarch:
Could you file these performance items as issues?  We'll work on them later.  You could look at an as-needed import for openpyxl for the test API.

PR #189, opened 2020-07-01
Fix/gds cli missing events

JHDeerin:
## F´ Pull Request
|**Field**|**Value**|
|:---|:---|
|**_Submission Date_**| 7/1/2020 |
|**_Submitter_**| JHDeerin |
|**_Originating Project/Creator_**| fprime|
|**_Base Branch_**| devel |
|**_Short Description_**| Fixed a bug in the GDS CLI where some data wasn't printed |
|**_Effected Component_**| GDS CLI |
|**_Effected Architectures(s)_**| N/A |

---
## Change Description

Fixed a bug where the GDS CLI would miss some events by adding a new predicate type (for comparing another predicate against a `SysData` object's `TimeType`) and wrapping the time-updating check in `base_commands.py` in this predicate, making the comparisons valid.

I also updated some broken unit tests for the GDS CLI while I was at it (a previous commit had added a flag that wasn't represented in the parsing tests), and created a few new ones to make sure this change worked.

## Rationale

This bug was originally reported by Aadil Rizvi, who noticed an issue when doing the following:

1. `cd` into the `Ref` folder and run it normally via `fprime-gds`
2. In another terminal, run the command `fprime-cli events`
3. Send a `cmdDisp.CMD_NO_OP` command to the GDS
4. Note that the GUI Events window shows 3 new events after sending this command, but only 1 of them is printed in the terminal

After debugging, it seems this error was caused in `base_commands.py` by creating a `predicates.greater_than` that expected to be passed in a `TimeType`, when the Test API actually passed in `SysData` objects to this predicate, meaning there was an invalid comparison occurring (which didn't throw an error for some reason, possibly because of the implementation of `__ge__` on `TimeType`). The new predicate wraps this comparison by only using `SysData.get_time()`, making the comparison valid.

Keeping information parity between the CLI tools and the GUI is clearly important to avoid getting confusing, contradictory outputs, and to make sure test results stay consistent regardless of which tool is used. Doubly-so when this difference in output was clearly a bug.

## Testing Recommendations

Besides running the unit tests, follow the above bug-reproducing steps and verify the console/GUI output now matches. Check the `channel` output for similar issues.

## Future Work

N/A (Assuming the bug is truly fixed)

### DIFF
--- a/Gds/src/fprime_gds/common/gds_cli/base_commands.py
+++ b/Gds/src/fprime_gds/common/gds_cli/base_commands.py
@@ -195,12 +195,16 @@ class QueryHistoryCommand(BaseCommand):
             item = cls._get_upcoming_item(api, search_filter, "NOW", timeout)
             cls._log(cls._get_item_string(item, json))
         else:
+
             def print_upcoming_item(min_start_time="NOW"):
                 item = cls._get_upcoming_item(api, search_filter, min_start_time)
                 cls._log(cls._get_item_string(item, json))
                 # Update time so we catch the next item since the last one
                 if item:
                     min_start_time = predicates.greater_than(item.get_time())
+                    min_start_time = filtering_utils.time_to_data_predicate(
+                        min_start_time
+                    )
                 return (min_start_time,)
 
             misc_utils.repeat_until_interrupt(print_upcoming_item, "NOW")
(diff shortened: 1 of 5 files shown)


JHDeerin:
Hey @LeStarch, just pushed a fix for the junk log files in the home directory when running the CLI. It looks like the StandardPipeline class defaulted to putting its log files in the home directory if not given a specific logging path; to fix this, I changed the default behavior to be to not log at all. I didn't see any way to disable logging entirely on StandardPipeline before this.

I did a quick grep and didn't see any other python files that depended on this default StandardPipeline logging, so I don't think this breaks anything, but let me know if this seems problematic.

JHDeerin:
Just checked the GUI and I can confirm it's still logging (but the CLI tool is not) after this change.

PR #191, opened 2020-07-14
Performs automatic semmle analysis on all pull requests

hunterpaulson:
## F´ Pull Request
|**Field**|**Value**|
|:---|:---|
|**_Submission Date_**| 2020-07-14 |
|**_Submitter_**| @hunterpaulson |
|**_Originating Project/Creator_**| FPrime |
|**_Base Branch_**| devel |
|**_Short Description_**| Adds lgtm to fprime, which automatically performs semmle on all PR's |
|**_Effected Component_**| all |
|**_Effected Architectures(s)_**| n/a |

---
## Change Description
LGTM.com (Looks good to me) is an external tool that can be configured to run semmle on our code. Currently it is configured to run on C/C++, Python, and JavaScript code. Before a pull request is merged it checks to make sure no new isses were introduced to the code as well as keeping track of older issues that have been fixed. All of the issues semmle finds can be found on [LGTM.com](https://lgtm.com/projects/g/nasa/fprime/alerts/?mode=list)
## Rationale
Now we can be aware of issues as soon as they are introduced as well as track our progress of fixing current issues.
## Testing Recommendations
Introduce a new issue
## Future Work
Fix the outstanding issues.


Gds/src/fprime_gds/common/adapters/ip.py
```diff
@@ -77,14 +76,13 @@ def open(self):
             self.thudp.start()
             # Start up a keep-alive ping if desired. This will hit the TCP uplink, and die if the connection is down
             if IpAdapter.KEEPALIVE_INTERVAL is not None:
-                self.keepalive = threading.Thread(
```
LeStarch, line 80: Why did you remove this keep-alive thread?

Gds/src/fprime_gds/common/history/chrono.py
```diff
@@ -40,17 +40,17 @@ def __init__(self, filter_pred=None):
 
         self.retrieved_cursor = 0
 
-    def data_callback(self, data_object, sender=None):
```
LeStarch, line 43: Why?

Gds/src/fprime_gds/common/utils/PortFinder.py
```diff
@@ -57,7 +57,7 @@ def IsPortUsed(port):
     except:
         # print 'error:', sys.exc_info()[0]
         portUsed = True
-        if isinstance(s, socket):
```
LeStarch, line 60: This doesn't seem right.

Gds/src/fprime_gds/executables/tcpserver.py
```diff
@@ -32,7 +32,7 @@
 GUI_ids = []
 
 
-def signal_handler(signal, frame):
```
LeStarch, line 35: This won't work (I don't think).  This will be passed a signal and a frame, you'll want to replace these with _, _

Gds/src/fprime_gds/common/history/chrono.py
hunterpaulson: Pylint was complaining that "Parameters differ from overridden 'default' method". (where the variable is data)
This is somewhat of a false positive as the difference between 'data' and 'data_object' is practically zero.

However Ashley in [this issue](https://github.com/PyCQA/pylint/issues/1631) makes a good point on the subject that 

"Why do you consider this a false positive? Renaming arguments means that as a user I could no longer do something like...
```
encoder = MyEncoder()
kwargs = {'o': to_encode}
encoder.default(**kwargs)
```

Gds/src/fprime_gds/common/adapters/ip.py
hunterpaulson: I deleted the variable because threading.Thread(...).start() doesn't have a return value and the vairavle was. Thus pylint was complaining that it was "Assigning result of a function call, where the function has no return". The key here is the .start() has no return, but 'threading.Thread()' does. Therefore the correct solution is as follows
```
self.keepalive = threading.Thread(
    target=self.th_alive, args=[float(self.KEEPALIVE_INTERVAL)]
)
self.keepalive.start()
```

Gds/src/fprime_gds/common/history/chrono.py
LeStarch: Nope, I am happy with the answer.  I agree with Ashley too.

Gds/src/fprime_gds/common/utils/PortFinder.py
LeStarch: `socket` in this context is a module.  You want the following, I think:

```
if isinstance(s, socket.socket):
    
```

LeStarch:
@hunterpaulson please make sure to run the GDS and make sure it is still working with these changes.  Then we can merge.

LeStarch:
@hunterpaulson did you submit this fix? (*_) or is that coming?

hunterpaulson:
@LeStarch (*_) is part of c74e03f

PR #194, opened 2020-07-18
lestarch: fixing bugs in Gds layer

LeStarch:


LeStarch:
Closing.  An all-encompassing fix is coming.

PR #202, opened 2020-08-06
Docker update

csmith608:
Allows docker to begin with the fprime-venv launched.  fprime-util displays the expected message and in opt/fprime/Ref the generate and build commands can be run.


csmith608:
I branched off master rather than devel so closing while I fix this

PR #203, opened 2020-08-06
Docker update venv

csmith608:
Allows docker to begin with the fprime-venv launched. fprime-util displays the expected message and in opt/fprime/Ref the generate and build commands can be run.

### DIFF
--- a/mk/docker/Dockerfile
+++ b/mk/docker/Dockerfile
@@ -35,7 +35,7 @@ ENTRYPOINT ["/bin/bash"]
 FROM fprime-base AS fprime-docker
 RUN     git clone --quiet https://github.com/nasa/fprime.git /opt/fprime && \
         python3 -m venv /opt/fprime-venv/ && . /opt/fprime-venv/bin/activate && \
-        pip install wheel && \
+        pip install -U wheel setuptools pip && \
         pip install /opt/fprime/Fw/Python/ && \
         pip install /opt/fprime/Gds/ && \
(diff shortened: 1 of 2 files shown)


mk/docker/Dockerfile
```diff
@@ -32,19 +32,17 @@ ENV LC_ALL C.UTF-8
...
-#        rm -r ~/.cache/pip && \
-#        chown -R fprime:fprime /opt/fprime-venv && \
-#        chmod -R 775 /opt/fprime-venv
-#ENV VIRTUAL_ENV "/opt/fprime-venv"
-#ENV PYTHONHOME "/opt/fprime-venv"
-#ENV PATH "/opt/fprime-venv/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+FROM fprime-base AS fprime-docker
+RUN     git clone --quiet https://github.com/nasa/fprime.git /opt/fprime && \
+        python3 -m venv /opt/fprime-venv/ && . /opt/fprime-venv/bin/activate && \
+        pip install wheel && \
```
LeStarch, line 38: we should make this `pip install -U wheel setuptools pip` as this is the latest F´ install recommendations.

LeStarch: Specifically line 38 here, replace `pip install wheel` with the above.

csmith608: Addressed in bbce27c

PR #210, opened 2020-09-16
fixed key checks to reference correct dictionary

AaronMcDaniel:
I was getting errors about command_elem being referenced before assigned. Looking at the check I think they were referencing the wrong dictionary. This fixed my errors, does this look like a good fix?

### DIFF
--- a/Autocoders/Python/bin/codegen.py
+++ b/Autocoders/Python/bin/codegen.py
@@ -534,7 +534,7 @@ def generate_topology(the_parsed_topology_xml, xml_filename, opt):
                         command_elem_set.attrib["component"] = comp_name
                         command_elem_set.attrib["mnemonic"] = parameter.get_name()+ "_PRM_SET"
                         command_elem_set.attrib["opcode"] = "%s"%(hex(int(parameter.get_set_opcodes()[0],base=0) + comp_id))
-                        if ("comment" in list(command_elem.attrib.keys())):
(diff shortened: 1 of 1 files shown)


Autocoders/Python/bin/codegen.py
```diff
@@ -581,7 +581,7 @@ def generate_topology(the_parsed_topology_xml, xml_filename, opt):
                         command_elem_save.attrib["component"] = comp_name
                         command_elem_save.attrib["mnemonic"] = parameter.get_name()+ "_PRM_SAVE"
                         command_elem_save.attrib["opcode"] = "%s"%(hex(int(parameter.get_save_opcodes()[0],base=0) + comp_id))
-                        if ("comment" in list(command_elem.attrib.keys())):
+                        if ("comment" in list(command_elem_set.attrib.keys())):
```
AaronMcDaniel, line 584: Should this be referencing the command_elem_save variable instead? I made this change because it was the smallest change to avoid errors, but it looks weird to reference command_elem_set here compared to the other block on line 537

LeStarch:
@timcanham can you review this change?

Autocoders/Python/bin/codegen.py
LeStarch: I believe so.  I am going to merge this PR and I'll take the action to correct this.

LeStarch (approved):

PR #214, opened 2020-10-21
Allow GroundInterface and QueuedComponentBase to work with FW_OBJECT_NAMES disabled

nathancheek:
There are currently a couple of issues which cause F Prime to not compile when FW_OBJECT_NAMES is disabled. I have patched 2 issues below.

GroundInterfaceComponent currently has a malformed constructor definition when FW_OBJECT_NAMES is disabled. The first commit patches this issue.

QueuedComponentBase tries to call `Os::Queue::getNumQueues()` without using an object, which is not allowed and leads to this error:

>error: cannot call member function 'NATIVE_INT_TYPE Os::Queue::getNumQueues()' without object

The second commit fixes this issue by calling `getNumQueues()` via the `m_queue` object.

### DIFF
--- a/Fw/Comp/QueuedComponentBase.cpp
+++ b/Fw/Comp/QueuedComponentBase.cpp
@@ -33,7 +33,7 @@ namespace Fw {
         queueName = this->m_objName;
 #else
         char queueNameChar[FW_QUEUE_NAME_MAX_SIZE];
-        (void)snprintf(queueNameChar,sizeof(queueNameChar),"CompQ_%d",Os::Queue::getNumQueues());
+        (void)snprintf(queueNameChar,sizeof(queueNameChar),"CompQ_%d",this->m_queue.getNumQueues());
         queueName = queueNameChar;
 #endif
     	return this->m_queue.create(queueName, depth, msgSize);


LeStarch:
Changing the base branch to `devel`.

LeStarch:
@nathancheek I can fix this conflict, but it would require me to commit to your master branch.   Thoughts?

My fix is:

- [ ] Take fprime `devel`'s version of `Svc/GroundInterface/GroundInterface.cpp`  it is updated

nathancheek:
I see `GroundInterface.cpp` was updated in `devel` so that fix is no longer applicable. I rebased on top of `nasa/devel` so now there's just the one fix for `QueuedComponentBase.cpp`.

Fw/Comp/QueuedComponentBase.cpp
```diff
@@ -33,7 +33,7 @@ namespace Fw {
         queueName = this->m_objName;
 #else
         char queueNameChar[FW_QUEUE_NAME_MAX_SIZE];
-        (void)snprintf(queueNameChar,sizeof(queueNameChar),"CompQ_%d",Os::Queue::getNumQueues());
+        (void)snprintf(queueNameChar,sizeof(queueNameChar),"CompQ_%d",this->m_queue.getNumQueues());
```
LeStarch, line 36: Does this still compile with 1.5.0?   getNumQueues is now a static class function is is therefore traditionally called with:
```
Os::Queue::getNumQueues()
```

LeStarch:
My instinct is to close this PR, as the same change was made twice. However, I want you to contribute PRs in the future, because I really value your contributions.

We are trying to bring more development publicly such that these collisions don't happen in the future.

Fw/Comp/QueuedComponentBase.cpp
nathancheek: Ah ok, I had not actually tested against `devel` so I didn't realize @Joshua-Anderson had turned `getNumQueues` into a static function. That seems like the better fix anyway!

nathancheek:
Yes, it looks like @Joshua-Anderson fixed both of these bugs, so there's no need for this PR.

PR #218, opened 2020-10-22
mstarch: fixing impl and testimpl usage of AC constants

LeStarch:
This fixes a critical bug in how the AcConstants.ini was being passed in for custom configurations.  In short, building works, but to use this in template generation would break.  This corrects the issue.  Fixes #218


cmake/target/impl.cmake
```diff
@@ -37,7 +37,8 @@ function(add_module_target MODULE_NAME TARGET_NAME AC_INPUTS SOURCE_FILES AC_OUT
     add_custom_target("${TARGET_NAME}" 
       COMMAND ${CMAKE_COMMAND} -E chdir ${CMAKE_CURRENT_SOURCE_DIR}
       ${CMAKE_COMMAND} -E env PYTHONPATH=${PYTHON_AUTOCODER_DIR}/src:${PYTHON_AUTOCODER_DIR}/utils BUILD_ROOT="${FPRIME_BUILD_LOCATIONS_SEP}"
-      PYTHON_AUTOCODER_DIR=${PYTHON_AUTOCODER_DIR}
+      FPRIME_AC_CONSTANTS_FILE="${FPRIME_AC_CONSTANTS_FILE}"
+      PYTHON_AUTOCODER_DIR=${PYTHON_AUTOCODER_DIR} DICTIONARY_DIR=${DICTIONARY_DIR}
```
Joshua-Anderson, line 41: `DICTIONARY_DIR` is unset and can be removed

LeStarch: Good point.

LeStarch: Given this is parallel to the codegen.py call, I'll take an action to clean this up after we are out of hotfix territory.

Joshua-Anderson (approved):

PR #223, opened 2020-10-25
Regex correction

pelmini:
Changed regex pattern in fprime build utility file


LeStarch:
Note: this needs to be a hotfix 1.5.2 release.

LeStarch (approved):

PR #224, opened 2020-10-26
Hotfix 1.5.2 Fixes CMake -D argument pass through for paths

LeStarch:


Fw/Python/src/fprime/util/build_helper.py
```diff
@@ -32,7 +32,7 @@
 )
 
 
-CMAKE_REG = re.compile(r"-D([a-zA-Z0-9_]+)=([a-zA-Z0-9_]+)")
+CMAKE_REG = re.compile(r"-D([a-zA-Z0-9_]+)=(.*)")
```
Joshua-Anderson, line 35: This allows 0+ length values. Do we want to allow setting a variable to nothing? I.E. `-DSOMETHING=`

LeStarch: I believe so, as it would allow you to set a -D to a ""

Joshua-Anderson (approved):

PR #232, opened 2020-11-05
Strict aliasing issue

kevin-f-ortega:
Reverting these lines of code to the NASA-v1.3.1 version. The NASA-v1.5.2 version generates strict-aliasing issues when compiling.


LeStarch (approved):
Reversion to remove undefined behavior.

PR #242, opened 2020-11-10
Packaged build artifacts into artifacts directory

### LINKED ISSUE #236: CMake Install and GDS Package
Currently the CMake install and usage with the GDS causes the following issues:

1. dictionary is placed in Top as opposed to a folder like the other pieces (e.g. Ref/dict)
2. GDS requires the user to run install or it loses updates. This is counter intuitive
3. Install triggers build-all, not just the deployment itself. This is confusing.

Joshua-Anderson:
Currently, `fprime-util install` installs files into a variety of different
locations, which makes it difficult to archive all the build artifacts.

Now, `fprime-util install` has been removed and `fprime-util build`
automatically ensures all build artifacts are installed into the location
provided by the `install_dest` settings option, which defaults to `build-artifacts`.

It has the following structure:

- <PLATFORM>/bin: deployment binaries
- <PLATFORM>/lib/static: deployment static libraries
- <PLATFORM>/dict: command dictionary, autogenerated html command,
  channel, and event documentation

The GDS has also been updated to automatically use the settings.ini file to look
into the artifacts directory to find the deployment application and dictionary.

Resolves #236


LeStarch:
@Joshua-Anderson you probably need to pull from `devel` now

Ref/README.md
```diff
@@ -26,22 +26,19 @@ fprime-util generate
...
+F´ ships with a browser-based test ground system. This system is designed to help developers of F´
+projects quickly test and work with F´ code without much overhead. This ground system can be run
+with the following commands. Please note: the Ref application's binary will also be run
+automatically. This allows for quick testing on Linux and Mac OSX. Before running the GDS, make sure
+that you have built the Ref example.
 
 ```
 cd fprime/Ref
-fprime-gds -d .
+fprime-gds
```
LeStarch, line 41: Why wouldn't you improve the documentation to run from the packaged output?

Ref/README.md
```diff
@@ -50,13 +47,13 @@ application should an error arise.  See: Logs -> Ref.log to see standard output
 To run the ground system without starting the Ref app:
 ```
 cd fprime/Ref
```
LeStarch, line 49: Same here

cmake/API.cmake
```diff
@@ -301,7 +301,15 @@ function(register_fprime_executable)
     endif()
     get_nearest_build_root(${CMAKE_CURRENT_LIST_DIR})
     # Explicit call to module register
-    generate_executable("${EX_NAME}" "${SC_IFS}" "${MD_IFS}")
+    generate_executable("${EX_NAME}_exe" "${SC_IFS}" "${MD_IFS}")
```
LeStarch, line 304: You should document this change in the above comment.

cmake/API.cmake
```diff
@@ -301,7 +301,15 @@ function(register_fprime_executable)
...
+    set_target_properties("${EX_NAME}_exe" PROPERTIES OUTPUT_NAME "${EX_NAME}")
+    add_custom_target(${EX_NAME} ALL)
+    add_dependencies("${EX_NAME}" "${EX_NAME}_exe")
+
+    # Only install into artifacts directory in release builds when SKIP_INSTALL is not set.
+    if (NOT DEFINED SKIP_INSTALL AND CMAKE_BUILD_TYPE STREQUAL "RELEASE")
```
LeStarch, line 310: This implies that for the "Testing" build, you will not make the installed directory?  Thus how do I run the gds against a testing build?

cmake/support/AC_Utils.cmake
```diff
@@ -98,7 +98,7 @@ function(acwrap AC_TYPE AC_FINAL_SOURCE AC_FINAL_HEADER AI_XML XML_FILE_DEPS MOD
       COMMAND ${CMAKE_COMMAND} -E chdir ${CMAKE_CURRENT_SOURCE_DIR}
       ${CMAKE_COMMAND} -E env PYTHONPATH=${PYTHON_AUTOCODER_DIR}/src:${PYTHON_AUTOCODER_DIR}/utils BUILD_ROOT="${FPRIME_BUILD_LOCATIONS_SEP}"
       FPRIME_AC_CONSTANTS_FILE="${FPRIME_AC_CONSTANTS_FILE}"
-      PYTHON_AUTOCODER_DIR=${PYTHON_AUTOCODER_DIR} DICTIONARY_DIR=${DICTIONARY_DIR}
+      PYTHON_AUTOCODER_DIR=${PYTHON_AUTOCODER_DIR}
```
LeStarch, line 101: Why did you remove DICTIONARY_DIR here?

cmake/target/dict.cmake
```diff
@@ -17,19 +17,33 @@ function(dictgen MODULE_NAME AI_XML MOD_DEPS)
   string(REGEX REPLACE "Ai.xml" "Dictionary.xml" DICT_XML "${AI_XML}")
   string(REGEX REPLACE "Ai.xml" "Ac" AC_BASE "${AI_XML}")
   string(REPLACE ";" ":" FPRIME_BUILD_LOCATIONS_SEP "${FPRIME_BUILD_LOCATIONS}")
-  # Building here with a fake dictionary output file, so this command will always rebuild.
+  get_filename_component(DICT_XML_NAME ${DICT_XML} NAME)
```
LeStarch, line 20: Does this imply that you are copying the dictionary straight to the package_gen?

docs/Tutorials/GpsTutorial/Tutorial.md
```diff
@@ -971,7 +971,7 @@ separate terminal run the following:
 
 ```shell
 cd fprime/GpsApp
-scp bin/arm-linux-gnueabihf/GpsApp pi@<raspberry pi IP>
+scp build-artifacts/arm-linux-gnueabihf/bin/GpsApp pi@<raspberry pi IP>
```
LeStarch, line 974: Is this build directory correct?

LeStarch (commented):
Just a bunch of comments.

Ref/README.md
Joshua-Anderson: I figured `fprime-gds` was cleaner than `fprime-gds -r build-artifacts`. Since we added artifact root autodetection we might as well take advantage of it, right?

LeStarch:
I ran this, works as expected.

cmake/API.cmake
Joshua-Anderson: Oh, interesting. I figured the only use of testing builds was unit tests, but I guess there's also the use case of debug binaries. I want to be certain that people don't mix release and debug builds. We could do somethign like `build-artifacts/<platform>/<RELEASE|TESTING>/...` for organization. It adds a new directory level to the build artifacts but does separate debug from release builds.

Is adding that additional complexity better than making users dig into the cmake build directory to find the debug binary? I'd say that the usage debug builds are rare enough that it's probably not worth adding another directory level. It could also cause user confusion because the RELEASE and TESTING directories are built separately, so they can get out of sync.

cmake/support/AC_Utils.cmake
Joshua-Anderson: DICTIONARY_DIR is never set, so it makes the dictionary target logic less confusing to have it removed. I think I removed it everywhere but DICTIONARY_DIR got removed in other changes, so after rebasing this was the only spot left, making it seem a little random.

cmake/target/dict.cmake
Joshua-Anderson: Yes, it's created, copied into build artifacts, then deleted from the source tree.

docs/Tutorials/GpsTutorial/Tutorial.md
Joshua-Anderson: Yep, I previously tested to confirm. We use the platform from cmake, which isn't necessarily the same name as the toolchain name passed to fprime.

Ref/README.md
LeStarch: Won't this fail because there is no settings.ini to tell it which artifact directory to use?

Joshua-Anderson: No, because even if the settings.ini is missing it'll provide the default location `build-artifacts`.

Detection order: `--root` argument, else `settings.ini`, else `build-artifacts`.

LeStarch (approved):

PR #247, opened 2020-11-12
mstarch: updating Fw::Buffer API

LeStarch:
This PR is to change Fw::Buffer's API in several ways:

1. U64 m_data is not U8* m_data to more cleanly represent the type
2. Users may now ask for Fw::SerializeBufferBase representations to serialize/deserialize *into* this buffer removing the need to manually wrap the data
3. Naming conventions where followed e.g. setdata -> setData
4. Reworked setData/setSize pair to setData(data, size) to reduce risk of leaking memory by forgetting the size
5. managerId and bufferId have been merged into a single agnostic context member

**Note:** in order to accomplish these changes, the implementation was hand written not auto-generated.

`Fw::ComFile` was removed in this PR, as it is completely unused and confuses the understanding of how file donwlink works.


LeStarch:
@timcanham I am really done now, just cleaning up the UTs and RPI stuff!

Svc/BufferManager/BufferManager.cpp
```diff
@@ -270,10 +270,31 @@ namespace Svc {
...
+      FW_ASSERT(bufferId <= 0xFFFF, bufferId);
+
+      U32 context = (managerId << 16) | bufferId;
+      FW_ASSERT(context != Fw::Buffer::NO_CONTEXT); // Ensure we didn't collide with not context
+      return context;
+  }
+
+  U32 BufferManager :: getManagerIdFromBufferContext(const U32 context) {
+      FW_ASSERT(context != Fw::Buffer::NO_CONTEXT); // Ensure we didn't receive with not context
+      return (context >> 16) & 0x7FFF;
```
Joshua-Anderson, line 285: Why `0x7FFF`? Aren't you storing manager as a 16 bit number

Fw/Buffer/Buffer.hpp
```diff
@@ -0,0 +1,172 @@
...
+    //! Wraps the given data pointer with given size in a buffer. The context by default is set to NO_CONTEXT but can
+    //! be set to specify a specific context.
+    //! \param data: data pointer to wrap
+    //! \param size: size of data located at data pointer
+    //! \param context: user-specified context to track creation. Default: no context
+    Buffer(U8* data, U32 size, U32 context=NO_CONTEXT);
+
+    //! Assignment operator to set given buffer from another
+    //!
+    Buffer& operator=(const Buffer& src);
```
Joshua-Anderson, line 72: You may want to clarify in the comment that this is ***not*** a copy.

Fw/Buffer/Buffer.hpp
```diff
@@ -0,0 +1,172 @@
...
+        NO_CONTEXT = 0xFFFFFFFF //!< Value representing no context
+    };
+
+    //! Construct a buffer with no context nor data
+    //!
+    //! Constructs a buffer setting the context to the default no-context value of 0xffffffff. In addition, the size
+    //! and data pointers are zeroed-out.
+    Buffer();
+
+    //! Construct a buffer by copying data from a reference to anothe buffer
```
Joshua-Anderson, line 58: This description implies that this constructor is supposed to copy the contents of the buffer to another buffer, but this does not seem to be the case from the implementation.

s/anothe/another/

Joshua-Anderson (commented):
Codes changes are looking pretty good to me!
All I had were some questions regarding the comments.

Do you think you need a unit test for the new Buffer class?

Fw/Buffer/Buffer.hpp
LeStarch: Agreed.

Fw/Buffer/Buffer.hpp
LeStarch: Agreed.

Svc/BufferManager/BufferManager.cpp
LeStarch: 0xFFFFFFFF is an invalid context. Thus this code forces the top bit to be 0.

LeStarch:
I'll need to write a UT and adjust the 2 comments.

Svc/BufferManager/BufferManager.cpp
Joshua-Anderson: That makes sense, you may want to assert above in `getBufferContext` that manager ID is less than `0x7FFF`, rather than `0xFFFF` to avoid confusion, since I saw the mismatch as a red flag because the bounds checks were different.

Drv/LinuxSerialDriver/LinuxSerialDriverComponentImpl.cpp
```diff
@@ -352,8 +352,8 @@ namespace Drv {
               }
 
               stat = ::read(comp->m_fd,
-                            reinterpret_cast<unsigned char*>(buff.getdata()),
-                            buff.getsize());
+                            reinterpret_cast<unsigned char*>(buff.getData()),
```
timcanham, line 355: same cast comment.

Joshua-Anderson:
Do you think it's worth adding a method to copy to contents of the buffer?

Drv/LinuxSerialDriver/LinuxSerialDriverComponentImplCommon.cpp
```diff
@@ -59,7 +59,7 @@ namespace Drv {
             }
         }
         this->m_readBuffMutex.unLock();
-        FW_ASSERT(found,Buffer.getbufferID(),Buffer.getmanagerID());
+        FW_ASSERT(found,Buffer.getContext());
```
timcanham, line 62: Why are we doing this check again?

Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
```diff
@@ -41,14 +41,14 @@ namespace Drv {
...
-        DEBUG_PRINT("Writing %d bytes to SPI\n",writeBuffer.getsize());
+        DEBUG_PRINT("Writing %d bytes to SPI\n",writeBuffer.getSize());
 
         spi_ioc_transfer tr;
         // Zero for unused fields:
         memset(&tr, 0, sizeof(tr));
-        tr.tx_buf = writeBuffer.getdata();
-        tr.rx_buf = readBuffer.getdata();
-        tr.len = writeBuffer.getsize();
+        tr.tx_buf = (U64)writeBuffer.getData();
```
timcanham, line 49: Should we update the SPI driver to U8*?

Fw/Buffer/Buffer.cpp
```diff
@@ -0,0 +1,175 @@
...
+    this->m_serialize_repr.setExtBuffer(m_data, m_size);
+    // Force a reset of usage
+    this->m_serialize_repr.resetSer();
+}
+
+void Buffer::setContext(const U32 context) {
+    this->m_context = context;
+}
+
+Fw::SerializeBufferBase& Buffer::getSerializeRepr() {
```
timcanham, line 88: Not sure I like this accessor having a side effect. Why not have one accessor, then the user can call `resetSer()` or `resetDeser()` on their own?

Svc/BufferManager/BufferManager.cpp
```diff
@@ -270,10 +270,31 @@ namespace Svc {
 
   }
 
-  // ----------------------------------------------------------------------
+  U32 BufferManager :: getBufferContext(const U32 managerId, const U32 bufferId) {
+      // Context is a concatenation of IDs, so check for overflows
+      FW_ASSERT(managerId <= 0x7FFF, managerId);
```
timcanham, line 275: Why < 0x7FFF? You have 16-bits, don't you?

timcanham (changes requested):
In general:

1) Get rid of old unneeded casts
2) Consider one accessor for underlying SerializeBufferBase.

Drv/LinuxSerialDriver/LinuxSerialDriverComponentImplCommon.cpp
LeStarch: This check is to see if the buffer is found in the above for loop.  The buffer.getContext() is just the description provided as an arg, in the case it is not found, this is printed.

Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
LeStarch: This type is defined as U64 in the Linux SPI header.  It is not in our control?

LeStarch:
@timcanham "Consider one accessor for underlying SerializeBufferBase."  The code must set the `serLoc` member to length if deserializing and 0 if serializaing.

So I can make a single accessor with a bool for that switch, or I can keep it as 2 accessors.

Thoughts?

Drv/LinuxSerialDriver/LinuxSerialDriverComponentImplCommon.cpp
LeStarch: I don't believe this is an issue, just an indication of the buffer identifier (context) that was not found.

Fw/Buffer/Buffer.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+    //! Construct a buffer to wrap the given data pointer of given size
+    //!
+    //! Wraps the given data pointer with given size in a buffer. The context by default is set to NO_CONTEXT but can
+    //! be set to specify a specific context.
+    //! \param data: data pointer to wrap
+    //! \param size: size of data located at data pointer
+    //! \param context: user-specified context to track creation. Default: no context
+    Buffer(U8* data, U32 size, U32 context=NO_CONTEXT);
+
```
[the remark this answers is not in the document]
LeStarch: Do we need the set?  I removed it as it was an extra method.  No one was using the set.

Fw/Buffer/Buffer.cpp
LeStarch: This make sense I will make this change.

Svc/BufferManager/BufferManager.cpp
LeStarch: Yes, but the ID 0xFFFFFFFF is reserved and that one bit is taken out of the manager ID's space.

Drv/LinuxI2cDriver/LinuxI2cDriverComponentImpl.cpp
```diff
@@ -80,8 +80,8 @@ namespace Drv {
 
 #if DEBUG_PRINT
       printf("I2c addr: 0x%02X\n",addr);
-      for (U32 byte = 0; byte < serBuffer.getsize(); byte++) {
-    	  printf("0x%02X ",((U8*)serBuffer.getdata())[byte]);
+      for (U32 byte = 0; byte < serBuffer.getSize(); byte++) {
+    	  printf("0x%02X ",((U8*)serBuffer.getData())[byte]);
```
timcanham: I think this cast is wrong as well. U8 should go in `0x%02X`, not (U8*), which is a pointer. I don't think you need the cast.

Drv/LinuxSerialDriver/LinuxSerialDriverComponentImplCommon.cpp
timcanham: We would have to guarantee that 0 is never a valid value for the context.

Drv/SocketIpDriver/SocketIpDriverComponentImpl.cpp
```diff
@@ -71,7 +71,7 @@ namespace Drv {
 
           // If the network connection is open, read from it
           if (self->m_helper.isOpened()) {
-              BYTE* data = reinterpret_cast<U8*>(self->m_buffer.getdata());
+              BYTE* data = reinterpret_cast<U8*>(self->m_buffer.getData());
```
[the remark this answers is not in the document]
timcanham: This should still be fixed; either change the cast to BYTE* or get rid of the cast. I don't think it's needed.

Fw/Buffer/Buffer.hpp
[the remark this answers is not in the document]
timcanham: I've used it on my projects when I have all the values available. It saves multiple calls to set everything.

Fw/FilePacket/CancelPacket.cpp
```diff
@@ -31,8 +31,8 @@ namespace Fw {
     toBuffer(Buffer& buffer) const
   {
     SerialBuffer serialBuffer(
-        reinterpret_cast<U8*>(buffer.getdata()),
-        buffer.getsize()
+        reinterpret_cast<U8*>(buffer.getData()),
```
timcanham: Remove cast...

Fw/FilePacket/DataPacket.cpp
```diff
@@ -43,8 +43,8 @@ namespace Fw {
     toBuffer(Buffer& buffer) const
   {
     SerialBuffer serialBuffer(
-        reinterpret_cast<U8*>(buffer.getdata()),
-        buffer.getsize()
+        reinterpret_cast<U8*>(buffer.getData()),
```
timcanham: remove cast

Fw/FilePacket/EndPacket.cpp
```diff
@@ -37,8 +37,8 @@ namespace Fw {
     toBuffer(Buffer& buffer) const
   {
     SerialBuffer serialBuffer(
-        reinterpret_cast<U8*>(buffer.getdata()),
-        buffer.getsize()
+        reinterpret_cast<U8*>(buffer.getData()),
```
timcanham: remove cast

Fw/FilePacket/FilePacket.cpp
```diff
@@ -23,8 +23,8 @@ namespace Fw {
     fromBuffer(const Buffer& buffer)
   {
     SerialBuffer serialBuffer(
-        reinterpret_cast<U8*>(const_cast<Buffer&>(buffer).getdata()),
-        const_cast<Buffer&>(buffer).getsize()
+        reinterpret_cast<U8*>(const_cast<Buffer&>(buffer).getData()),
```
timcanham: remove cast

Fw/FilePacket/StartPacket.cpp
```diff
@@ -42,8 +42,8 @@ namespace Fw {
     toBuffer(Buffer& buffer) const
   {
     SerialBuffer serialBuffer(
-        reinterpret_cast<U8*>(buffer.getdata()),
-        buffer.getsize()
+        reinterpret_cast<U8*>(buffer.getData()),
```
timcanham: remove cast

Svc/BufferLogger/BufferLogger.cpp
```diff
@@ -63,8 +63,8 @@ namespace Svc {
     )
   {
     if (m_state == LOGGING_ON) {
-      const U8 *const addr = reinterpret_cast<U8*>(fwBuffer.getdata());
-      const U32 size = fwBuffer.getsize();
+      const U8 *const addr = reinterpret_cast<U8*>(fwBuffer.getData());
```
timcanham: remove cast

Svc/BufferManager/BufferManager.cpp
timcanham: Ah - that's right.

Svc/GroundInterface/GroundInterface.cpp
```diff
@@ -97,8 +97,7 @@ namespace Svc {
   void GroundInterfaceComponentImpl::frame_send(U8 *data, TOKEN_TYPE size, TOKEN_TYPE packet_type) {
       // TODO: replace with a call to a buffer manager
       Fw::Buffer buffer = m_ext_buffer;
-      Fw::ExternalSerializeBuffer buffer_wrapper(reinterpret_cast<U8*>(m_ext_buffer.getdata()),
-                                                 m_ext_buffer.getsize());
+      Fw::SerializeBufferBase& buffer_wrapper = buffer.getSerializeRepr();
       // True size is supplied size plus sizeof(TOKEN_TYPE) if a packet_type other than "UNKNOWN" was supplied.
       // This is because if not UNKOWN, the packet_type is serialized too.  Otherwise it is assumed the PACKET_TYPE is
```
timcanham, line 103: Typo, UNKNOWN

Svc/GroundInterface/test/ut/GroundInterfaceRules.cpp
```diff
@@ -65,13 +65,13 @@ namespace Svc {
     }
     // Pick a rule and downlink
     void FileDownlinkRule :: action(Svc::Tester &state) {
-        char data_holder[2048];
-        Fw::Buffer buffer(0, 0, reinterpret_cast<U64>(data_holder), sizeof(data_holder));
-        Fw::ExternalSerializeBuffer buffer_wrapper(reinterpret_cast<U8*>(buffer.getdata()), buffer.getsize());
+        unsigned char data_holder[2048];
```
timcanham, line 68: You could make this U8 to get rid of Coverity warnings.

Svc/GroundInterface/test/ut/GroundInterfaceRules.cpp
```diff
@@ -65,13 +65,13 @@ namespace Svc {
...
+        Fw::Buffer buffer(data_holder, sizeof(data_holder));
+        Fw::ExternalSerializeBuffer buffer_wrapper(reinterpret_cast<U8*>(buffer.getData()), buffer.getSize());
```
timcanham: Remove cast

timcanham (changes requested):
More comments - missed some casts

Drv/LinuxSerialDriver/LinuxSerialDriverComponentImplCommon.cpp
LeStarch: Why?  The getContext() is not part of the assert, it is for the message args.

Fw/Buffer/Buffer.hpp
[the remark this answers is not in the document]
LeStarch: Alright, all add it.

Drv/LinuxI2cDriver/LinuxI2cDriverComponentImpl.cpp
LeStarch: I think this is fixed, we index into a U8* to get a U8.

Drv/LinuxSerialDriver/LinuxSerialDriverComponentImplCommon.cpp
timcanham: Duh! Brain short. Thought `getContext()` was the thing being asserted.

Fw/Buffer/Buffer.cpp
```diff
@@ -0,0 +1,177 @@
...
+void Buffer::set(U8* const data, const U32 size, const U32 context) {
+    this->m_data = data;
+    this->m_size = size;
+    if (m_data != NULL) {
+        this->m_serialize_repr.setExtBuffer(m_data, m_size);
+    }
+    this->m_context = context;
+}
+
+Fw::SerializeBufferBase& Buffer::getSerializeRepr() {
```
timcanham, line 95: This is the one I think we should change. I would not have the `resetSer()` call and just return the buffer. The user can reset it if that's what they want to do. They could have logic that appends to what is there, and this would break it.

timcanham (changes requested):

Fw/Buffer/Buffer.cpp
LeStarch: Good catch.  That was my intention, it should be fixed now.

LeStarch:
@timcanham @Joshua-Anderson I think I resolved everything on the list.  Anything remaining that I need to address?

timcanham (approved):

LeStarch:
Merge it! Merge it!

PR #253, opened 2020-11-13
Create module source and dependency graph during generation

Joshua-Anderson:
During cmake generation, every module will generate two file - a source file list and a dependency list.

These files can then be used by the metrics tool to generate source lines of code (SLOC) reports.

The SLOC tool will look for executable modules (denoted by ending in `_exe`). It will then start at the executable and generate a report by walking all the executables dependencies.

The tool will generate a per-module and overall SLOC report with the following data.

- Total lines of code
- Lines of xml code
- Lines of user-written C++ code (comments excluded)
- Lines of F' generated C++ code (comments excluded) (as matched by `*Ac.cpp|hpp`)
- Lines of code per file.

This needs a cmake unit test, but I wanted to get the design reviewed before writing tests.


LeStarch:
Decided to go another direction on sloc.

PR #255, opened 2020-11-14
mstarch: cleaning up issues detected by LGTM

LeStarch:
Cleans up a number of changes required by LGTM.  Also removes wxgui GDS.  This can be found in release 1.5.3.


Autocoders/Python/src/fprime_ac/generators/formatters.py
```diff
@@ -1413,7 +1413,7 @@ def argFilter(self, module, args, name="foo"):
                 new_args.append((id2, "U16", comment, "", "", [], []))
 
             elif type == "SCLK":
-                type = module[0].upper() + module[1:] + "ArgSclk"
+                _ = module[0].upper() + module[1:] + "ArgSclk"
```
Joshua-Anderson, line 1416: Can't we just delete this line?

Autocoders/Python/src/fprime_ac/generators/visitors/ChannelVisitor.py
```diff
@@ -94,7 +94,7 @@ def DictStartVisit(self, obj):
         output_dir = os.environ["DICT_DIR"] + "/channels"
         if not (os.path.isdir(output_dir)):
             os.makedirs(output_dir)
-        pyfile = output_dir + "/" + obj.get_name() + ".py"
+        _ = output_dir + "/" + obj.get_name() + ".py"
```
Joshua-Anderson, line 97: same here

Joshua-Anderson (changes requested):
There's quite a few places where you're assigning to _, which is a no-op (aside from the mutation from the assignment itself).

Nearly all of these lines can just totally be removed, and a new of them could probably be refactored to avoid the assignment.

Autocoders/Python/src/fprime_ac/generators/formatters.py
LeStarch: I made a lot of switches from variable -> _ but left the line in.  Given the overall quality of the code base, I cannot be sure that there aren't side-effects to function calls and what not.

Thus clean ups more substantial will require a full refactor.

Autocoders/Python/src/fprime_ac/generators/visitors/ChannelVisitor.py
LeStarch: I made a lot of switches from variable -> _ but left the line in.  Given the overall quality of the code base, I cannot be sure that there aren't side-effects to function calls and what not.

Thus clean ups more substantial will require a full refactor.

Joshua-Anderson: Yeah, there may be some unexpected affects, but if that's the case can't you just directly call the functions you're concerned have side-effects?

Also, just scrolling through the changes, most of them seem this one, where removing the entire line seems pretty safe (unless get_name is doing something insane).

I'd argue this cleanup is pointless without actually removing the offending lines (other than appeasing the linter), but if it doesn't make sense to invest the effort in removing these lines now, I guess at least clearly marks the lines for cleanup later.

LeStarch:
Agreed to fix the obvious assignment only statements.

LeStarch:
@Joshua-Anderson  should be ready for re-review

Joshua-Anderson (approved):

PR #256, opened 2020-11-15
replace 32 with total cpu thread(s)

Uniminin:
`$(nproc || printf '%s\n' 1)"` will return total number of cpu thread(s) if it returns `0` else `1`.


Uniminin (commented):
All fine!

PR #258, opened 2020-11-15
Replace 32 with total cpu thread(s)

Uniminin:
`"$(nproc || printf '%s\n' 1)"` will return total number of cpu thread(s) if it returns `0` else `1`.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -87,7 +87,7 @@ layer components. Should the user not have installed the given component, the co
 ```
 cd Ref
 fprime-util generate
-fprime-util build --jobs 32
+fprime-util build --jobs "$(nproc || printf '%s\n' 1)"
 ```
 
 **Testing F´ GDS Installation Via Running HTML GUI**


Uniminin (commented):
All fine!

LeStarch:
What do you think of setting it to a small number, say `4` instead?  Since this is the Install guide we want it very clear what this code does.

Alternatively, we could pull this into a note on how to build with all machine cores.

Thoughts?

LeStarch (commented):
What do you think of setting it to a small number, say 4 instead? Since this is the Install guide we want it very clear what this code does.

Alternatively, we could pull this into a note on how to build with all machine cores.

Thoughts?

Uniminin:
> What do you think of setting it to a small number, say 4 instead? Since this is the Install guide we want it very clear what this code does.
> 
> Alternatively, we could pull this into a note on how to build with all machine cores.
> 
> Thoughts?

Setting it 4 isn't a good idea. What if anyone is building on a single or dual core processor. 4 jobs won't really help or may make it worse.

If not best but better way will be if it detects the total number of cpu threads on a system and apply the jobs according to it. Which is what i did on my commit on this PR.

Uniminin:
My commit is simple. `nproc` returns total number of cpu threads (check `man nproc`.
Brief Explanation: https://github.com/nasa/fprime/pull/258#issue-521143662

Uniminin:
> Alternatively, we could pull this into a note on how to build with all machine cores.
Not necessary. people can understand. we could add a small note if we want but again that's not very necessary.

LeStarch (approved):

LeStarch:
Decided I like this fix.  It is better then recommending 32 processes for even small machines.  Thanks!!!

PR #265, opened 2020-11-20
mstarch: reworked FileDownlink

LeStarch:
This file downlink doesn't have the thundering herd problem.

### DIFF
--- a/Svc/FileDownlink/FileDownlink.cpp
+++ b/Svc/FileDownlink/FileDownlink.cpp
@@ -116,7 +116,6 @@ namespace Svc {
           this->mode.set(Mode::DOWNLINK);
       }
 
-      //If back-to-back transfer mode attempt the next downlink
       this->downlinkPacket();
   }
 
--- a/Svc/FileDownlink/FileDownlink.cpp
+++ b/Svc/FileDownlink/FileDownlink.cpp
@@ -1,18 +1,18 @@
 // ====================================================================== 
 // \title  FileDownlink.hpp
-// \author bocchino
+// \author bocchino, mstarch
 // \brief  hpp file for FileDownlink component implementation class
 //
 // \copyright
 // Copyright 2009-2015, by the California Institute of Technology.
 // ALL RIGHTS RESERVED.  United States Government Sponsorship
 // acknowledged.
-// 
 // ====================================================================== 
 
 #include <Svc/FileDownlink/FileDownlink.hpp>
 #include <Fw/Types/Assert.hpp>
 #include <Fw/Types/BasicTypes.hpp>
+
 
 namespace Svc {
 
@@ -23,16 +23,25 @@ namespace Svc {
   FileDownlink ::
     FileDownlink(
         const char *const name,
-        const U16 downlinkPacketSize
+        const U32 timeout,
+        const U32 cycleTime
     ) :
       FileDownlinkComponentBase(name),
-      downlinkPacketSize(downlinkPacketSize),
       filesSent(this),
       packetsSent(this),
       warnings(this),
-      sequenceIndex(0)
+      sequenceIndex(0),
+      timeout(timeout),
+      curTimer(0),
+      cycleTime(cycleTime),
+      curOpCode(0),
+      curCmdSeq(0),
+      bufferSize(0),
+      byteOffset(0),
+      endOffset(0),
+      lastCompletedType(Fw::FilePacket::T_NONE),
+      lastBufferId(0)
   {
-
   }
 
   void FileDownlink ::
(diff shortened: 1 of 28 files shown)


Svc/FileDownlink/FileDownlink.cpp
```diff
@@ -49,160 +58,269 @@ namespace Svc {
...
+          this->lastCompletedType == Fw::FilePacket::T_CANCEL) {
+          finishHelper(this->lastCompletedType == Fw::FilePacket::T_CANCEL);
+          return;
+      }
+      //If waiting and a buffer is in-bound, then switch to downlink mode
+      else if (this->mode.get() == Mode::WAIT) {
+          this->mode.set(Mode::DOWNLINK);
+      }
+
+      //If back-to-back transfer mode attempt the next downlink
```
LeStarch, line 119: Confusing comment as reported by @timcanham

LeStarch:
@Joshua-Anderson ready for review.

Svc/FileUplink/File.cpp
```diff
@@ -40,13 +40,16 @@ namespace Svc {
...
+    if (status != Os::File::OP_OK) {
+        return status;
+    }
 
     NATIVE_INT_TYPE intLength = length;
-    status = this->osFile.write(data, intLength);
-    if (status != Os::File::OP_OK)
-      return status;
+    //Note: not waiting for the file write to finish
+    status = this->osFile.write(data, intLength, false);
```
Joshua-Anderson, line 49: Could changing write behavior to not block cause any side affects?

Svc/FileDownlink/FileDownlink.hpp
```diff
@@ -191,14 +190,14 @@ namespace Svc {
           void fileOpenError(void);
 
           //! Issue a File Read Error warning
-          void fileRead(void);
+          void fileRead(const Os::File::Status sttatus);
```
Joshua-Anderson, line 193: sttatus->status

Svc/FileDownlink/FileDownlink.cpp
```diff
@@ -49,160 +58,268 @@ namespace Svc {
...
+    this->sequenceIndex = 1;
+    this->curTimer = 0;
+    this->byteOffset = 0;
+    this->endOffset = this->file.size;
+    this->lastCompletedType = Fw::FilePacket::T_START;
+    this->log_ACTIVITY_HI_SendStarted(this->file.size, this->file.sourceName, this->file.destName);
+  }
+
+  void FileDownlink ::
+    SendPartial_cmdHandler(
```
Joshua-Anderson, line 177: Would it make sense to refactor SendPartial and SendFile into a single common function? It seems like there's a some duplicated logic.

Svc/FileDownlink/FileDownlink.cpp
```diff
@@ -49,160 +58,268 @@ namespace Svc {
...
   {
-
-    const U32 fileSize = this->file.size;
-    FW_ASSERT(byteOffset < fileSize, byteOffset);
-    const U16 maxDataSize = static_cast<U16>(this->downlinkPacketSize);
-    const U16 dataSize = static_cast<U16>((byteOffset + maxDataSize > fileSize) ?
-      fileSize - byteOffset : maxDataSize);
+    FW_ASSERT(byteOffset < this->endOffset);
+    const U32 maxDataSize = FILEDOWNLINK_INTERNAL_BUFFER_SIZE - Fw::FilePacket::DataPacket::HEADERSIZE;
+    const U32 dataSize = (byteOffset + maxDataSize > this->endOffset) ?
```
Joshua-Anderson, line 270: Nitpick, but a ternary operator wrapped across multiple lines like this is difficult to read. Maybe break into an if statement?

Svc/FileDownlink/FileDownlink.hpp
```diff
@@ -210,6 +209,13 @@ namespace Svc {
           FileDownlink *const fileDownlink;
 
       };
+      //! Enumeration for packet types
+      //! Each type has a buffer to store it.
+      enum PacketType {
+          FILE_PACKET,
+          CANCEL_PACKET,
+          COUNT_PACKET_TYPE
```
Joshua-Anderson, line 217: Is this the best way to get the number of elements in an enum? I'm guessing you did this because there isn't a better way, but I just have to ask 😂 . It took me a second to figure out how this worked.

Joshua-Anderson (changes requested):
Looking good and seems like a nice improvement! Just had a few questions/comments.

Svc/FileDownlink/FileDownlink.cpp
LeStarch: I agree, but I don't want that as this PR.  The existing code is as-is from the contributing project.

Svc/FileDownlink/FileDownlink.hpp
LeStarch: We have a pattern of adding an extra type like that on enums of this style for assertion purposes.  This is done all over F´.  I don't know of a better way given the older versions of C++ we support.

Svc/FileUplink/File.cpp
LeStarch: I don't believe so, as this wasn't a true non-block operation before.  This just doesn't force a "sync" call through the OS before moving on. `sync` calls are fairly heavy weight and rarely needed these days.  They do however fight with the OSes ability to buffer files.

Joshua-Anderson (approved):

PR #266, opened 2020-11-24
Fix dictionary target dependencies

Joshua-Anderson:
The dictonary target wasn't depending on the topology xml file,
so changes to the topology weren't causing the dictionary to rebuild.

The autocoding step also generates dictionaries as a side affect of
generating the topology autocode (might be worth considering if we can
somehow combine these steps so we don't have to run the topology autocoder
twice). This could sometimes run after the dictionary generation step,
leading to leftover dictionary files in the source tree. Adding a dependency
on the modules autocoded files (including the autocoded topology files)
ensures that this step always runs after the topology autocode step
and can properly clean up.

### DIFF
--- a/cmake/target/dict.cmake
+++ b/cmake/target/dict.cmake
@@ -13,33 +13,37 @@
 #
 # Generate a dictionary from any *AppAi.xml file that we see
 ####
-function(dictgen MODULE_NAME AI_XML MOD_DEPS)
+function(dictgen MODULE_NAME AI_XML DEPS)
   string(REGEX REPLACE "Ai.xml" "Dictionary.xml" DICT_XML "${AI_XML}")
+  string(REGEX REPLACE "Ai.xml" "ID.csv" ID_CSV_XML "${AI_XML}")
+  string(REGEX REPLACE "Ai.xml" "Ai_IDTableLog.txt" ID_LOG_XML "${AI_XML}")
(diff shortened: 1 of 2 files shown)


cmake/target/dict.cmake
```diff
@@ -81,7 +85,7 @@ function(add_module_target MODULE_NAME TARGET_NAME GLOBAL_TARGET_NAME AC_INPUTS
         # Only generate dictionaries on serializables or topologies
         if (AC_IN MATCHES ".*Topology.*\.xml$")
             fprime_ai_info("${AC_IN}" "${MODULE_NAME}")
-            dictgen("${MODULE_NAME}" "${AC_IN}" "${MODULE_DEPENDENCIES};${MOD_DEPS};${FILE_DEPENDENCIES}")
+            dictgen("${MODULE_NAME}" "${AC_IN}" "${AC_OUTPUTS}")
```
LeStarch, line 88: Don't you also need to depend on on the module dependencies?  It seems just depending on AC_OUTPUTS will not detect changes in constituent components.

Joshua-Anderson: You're right, I thought since the included xml files weren't compiled I didn't need to depend on them, but I forgot about triggering rebuilds. I'll fix that.

Joshua-Anderson:
@LeStarch Ready for a re-review

LeStarch (approved):
This is what I expected.  Nice work.

PR #268, opened 2020-11-24
lestarch: adding custom addons to GDS

LeStarch:
This adds in a very simple addon system to the HTML GDS allowing users to design custom Vue.js components.

### DIFF
--- a/Gds/src/fprime_gds/flask/static/addons/enabled.js
+++ b/Gds/src/fprime_gds/flask/static/addons/enabled.js
@@ -1,0 +1,2 @@
+// Add addon imports here, try used to prevent errors from crashing GDS
+import "./image-display/addon.js"
--- a/Gds/src/fprime_gds/flask/static/addons/image-display/addon.js
+++ b/Gds/src/fprime_gds/flask/static/addons/image-display/addon.js
@@ -1,0 +1,43 @@
+/**
+ * ImageDisplay: an example component addon.
+ *
+ * This file defines the "image-display" component used to display the last downlinked image in the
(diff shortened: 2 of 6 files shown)


Joshua-Anderson:
I think the code looks good, but before merging this we should probably discuss that the long term plan for fprime-gds dashboards/visualization is. There's a lot of visualization and graphing tools out there (openmct, grafana, etc...), so it may be worth considering if any of those would meet our requirements rather than rolling our own.

LeStarch:
@Joshua-Anderson I looked into this and came to the conclusion (a second time) that we should continue maintaining the GDS layer. In short, the other available solutions would require substantial effort to get commanding/file support/events working and the GDS supports those already.

The F´  GUI is a command and basic telemetry interface.   Most other solutions can visualize, but few other solutions handle file uplink/downlink, nor commanding.  In addition, events tend to break other specific telemetry interfaces as most don't have this concept.  Although, solutions could be built these missing features would essentially require reinvention.  The goal of the GDS is to provide a working, out-of-the-box interfaces for developers to test and should a project need a better or different ground tool, they are free to use said tool and integrate with it.

Thus, at least as far as I stand, the future plan is to run maintenance on the GDS rather than invest a large effort (and maintenance) adapting tools that lack key features.

LeStarch:
@Joshua-Anderson can you approve this.  We discussed that the recommendation is that this will leave our GDS in a state where users can expand as they see fit without F´ needing to support XYZ dashboard development.

Joshua-Anderson (approved):
LGTM!

PR #270, opened 2020-11-24
Fix clang compiler warnings

Joshua-Anderson:
Fixes a variety of compiler warnings and causes running
`fprime build --all` and `fprime check --all` on macOs Catalina
to compile with the `-Wall` compiler flag added.

### DIFF
--- a/Drv/LinuxI2cDriver/LinuxI2cDriverComponentImpl.hpp
+++ b/Drv/LinuxI2cDriver/LinuxI2cDriverComponentImpl.hpp
@@ -64,9 +64,10 @@ namespace Drv {
           Fw::Buffer &serBuffer 
       );
 
+      // Prevent unused field error when using stub
+      #ifndef STUBBED_LINUX_I2C_DRIVER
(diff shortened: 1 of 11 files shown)


Drv/LinuxI2cDriver/LinuxI2cDriverComponentImpl.hpp
```diff
@@ -64,9 +64,10 @@ namespace Drv {
           Fw::Buffer &serBuffer 
       );
 
+      // Prevent unused field error when using stub
```
LeStarch, line 67: Should we just not compile this driver for MAC?  It doesn't work.

LeStarch (approved):

LeStarch:
A stub driver is nice to have.  It allows a consistent compile on Mac.

Joshua-Anderson:
Yep, that's exactly what I thinking with keeping the stub functional. You can test-compile linux fsw on macOS without any modifications (though obviously it wouldn't actually run properly). I figured this was worth keeping stub around for now, and we can revisit in the future.

PR #276, opened 2020-11-30
fprime-util: fix --path argument

Joshua-Anderson:
Because `settings.py` was explicitly using the `cwd`, setting the path manually on the command line didn't change where fprime-util looked for the framework path.


LeStarch (approved):
Looks good.

PR #277, opened 2020-11-30
Update/active logger min

timcanham:
This PR updates `Svc::ActiveLogger` to remove circular buffers and the command to dump them to a file. These were never used and are better handled by `Svc::ComLogger`. The circular buffers consumed memory and CPU cycles, so this will help smaller embedded targets reclaim some memory and CPU resources.


Svc/ActiveLogger/test/ut/ActiveLoggerTester.cpp
```diff
@@ -150,43 +150,43 @@ TEST(ActiveLoggerTest,FatalTesting) {
 
 }
 
-TEST(ActiveLoggerTest,CircularBufferDump) {
+// TEST(ActiveLoggerTest,CircularBufferDump) {
```
LeStarch, line 153: @timcanham is there some reason we don't just delete these disabled tests, since the functionality is gone?

Svc/ActiveLogger/test/ut/TestMain.cpp
```diff
@@ -0,0 +1,15 @@
+// ----------------------------------------------------------------------
```
LeStarch, line 1: It looks like this file is still in its stub form.  We should probably remove it.

LeStarch (changes requested):

LeStarch:
@timcanham this looks really good.  Just a few minor changes basically removing deadcode.

Svc/ActiveLogger/test/ut/ActiveLoggerTester.cpp
timcanham: Removed commented out code.

Svc/ActiveLogger/test/ut/TestMain.cpp
timcanham: Removed unused file.

LeStarch (approved):

LeStarch:
@timcanham we update SDD.

LeStarch:
@timcanham collapse the double filters into a single filter for the current time.

LeStarch:
FYI ci is failing on invalid names used in the integration test:

https://github.com/nasa/fprime/blob/566ab7247417f37396cdbb51ce569e62ae479fab/Ref/test/int/ref_integration_test.py#L113-L117

timcanham:
Got it – command names and arguments changed when I removed the post filter.

Tim Canham
Mars Helicopter Operations Lead

timcanham:
You can probably figure it out by looking at the XML.

Tim Canham
Mars Helicopter Operations Lead

LeStarch (approved):
Looks good to me.

Svc/ActiveLogger/docs/sdd.md
```diff
@@ -1,23 +1,20 @@
...
-AL-002 | The `Svc::ActiveLogger` component shall store a configurable circular buffer of events by severity. | Unit Test
-AL-003 | The `Svc::ActiveLogger` component shall write the circular buffers to a file upon command | Unit Test
-AL-004 | The `Svc::ActiveLogger` component shall have commands to filter events based on event severity. | Unit Test
-AL-005 | The `Svc::ActiveLogger` component shall filter events on receipt of the event on the caller's thread but before being placed in the event buffer. | Unit Test
-AL-006 | The `Svc::ActiveLogger` component shall filter events on the thread of the component before sending them as packets, but after they have been stored in the circular buffers | Unit Test
-AL-007 | The `Svc::ActiveLogger` component shall filter events based on the event ID. | Unit Test 
+AL-001 | The `Svc::ActiveLogger` component shall queue events and compose them into downlink packets. | Inspection; Unit Test
+AL-002 | The `Svc::ActiveLogger` component shall have commands to filter events based on event severity. | Unit Test
+AL-003 | The `Svc::ActiveLogger` component shall have commands to filter events based on the event ID. | Unit Test
```
Joshua-Anderson, line 17: Is there no requirement for sending out fatal events on the FatalAnnounce port?
I guess it wasn't there before, so this isn't a blocker to me.

Joshua-Anderson (commented):
Unit Test and SDD look good. I had one non-blocking comment about a requirement that seems to have been missing previously.

Joshua-Anderson (approved):
:shipit:

PR #283, opened 2020-12-01
Update/buffer manager

timcanham:
This pull request updates `Svc::BufferManager` to allow allocation and deallocation in any order, and allows the user to specify a set of differently sized pools to optimize memory usage.


Svc/BufferManager/BufferManagerComponentImpl.cpp
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  void BufferManagerComponentImpl ::
+    schedIn_handler(
+        const NATIVE_INT_TYPE portNum,
+        NATIVE_UINT_TYPE context
+    )
+  {
+    // write telemetry values
+    this->tlmWrite_HiBuffs(this->m_highWater);
```
cruise2coast, line 212: It may be a good idea to toggle these telemetry reports on/off via commands. By default, these telemetry channels could be turned off and not reported to avoid constantly generating telemetry packets. When needed, these channels can be turned on for reporting via a command. This would help control telemetry packet generation/output especially in radio RF bandwidth constrained environments.

Svc/BufferManager/BufferManagerComponentImpl.cpp
```diff
@@ -0,0 +1,222 @@
...
+        if (this->m_bufferBins.bins[bin].numBuffers) {
+            memorySize += 
+                this->m_bufferBins.bins[bin].bufferSize * this->m_bufferBins.bins[bin].numBuffers // allocate each set of buffer memory
+                + sizeof(AllocatedBuffer) * this->m_bufferBins.bins[bin].numBuffers; // allocate the structs to track the buffers
+            this->m_numStructs += this->m_bufferBins.bins[bin].numBuffers;
+        }
+    }
+
+    NATIVE_UINT_TYPE allocatedSize = memorySize;
+    bool recoverable; //!< don't care if it is recoverable since they are a pool of user buffers
```
bocchino, line 170: This variable should be initialized to a specific value (true or false), even if the value doesn't matter. Pick one. Also, we should run static analysis and make sure there are no more uninitialized variables.

Svc/BufferManager/BufferManagerComponentImpl.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+    // walk through entries and initialize them
+    NATIVE_UINT_TYPE currStruct = 0;
+    for (NATIVE_UINT_TYPE bin = 0; bin < BUFFERMGR_MAX_NUM_BINS; bin++) {
+        if (this->m_bufferBins.bins[bin].numBuffers) {
+            for (NATIVE_UINT_TYPE binEntry = 0; binEntry < this->m_bufferBins.bins[bin].numBuffers; binEntry++) {
+                // placement new for Fw::Buffer instance. We don't need the new() return value, 
+                // because we know where the Fw::Buffer instance is
+                U32 context = (this->m_mgrId << 16) | currStruct;
+                (void) new(&this->m_buffers[currStruct].buff) Fw::Buffer(bufferMem,this->m_bufferBins.bins[bin].bufferSize,context);
```
bocchino, line 191: This code is using more memory than it needs to. The `bufferMem` pointer is stored twice. The other information in the `Fw::Buffer` can be computed when the buffer is allocated. It doesn't need to be stored in the reserved memory area.

Svc/BufferManager/BufferManagerComponentImpl.cpp
bocchino: A related issue is that the previous version had a mechanism for suppressing out-of-memory warnings, so repeated requests wouldn't cause spam. I don't think that mechanism has been incorporated here. Maybe add event throttling to that event?

bocchino (commented):
Looks good. Just a few comments in the code. General comments:

I suggest using rules-based testing for the unit tests. I have a memory allocator that I wrote for the Autonomy research project that uses rules-based testing.

Consider using a more efficient algorithm for allocation. In the approach suggested below, if the bins are sized well so that a buffer of minimal size is always available, then the allocation time is constant. If a minimal buffer is not available (which also wastes memory or denies a request), then the allocation time is worst-case linear in the number of bins.

1. Use a separate data structure for the bin metadata and for the allocated memory. For the bin metadata, you can have an array of bins. Each bin can store (1) an array of pointers to memory blocks and (2) a size, which is the number of buffers reserved from that bin (initially zero for each bin). The allocated memory can be an array of arrays of memory blocks. On setup, it can look something like this, assuming 2 bins with 6 pointers and 4 pointers:

buffer-manager-1

2. To allocate, (1) start at the smallest bin that will fit the size and (2) reserve a buffer from the first bin that has a buffer available (based on the number of reserved buffers). To reserve a buffer from a bin _B_, (1) use the buffer whose pointer is at the end of the array of pointers in _B_ and (2) increment the number of reserved buffers for _B_. Here is an example after three allocations from the lower bin. Pointers marked with an X point to reserved buffers. The lower bin has three reserved buffers, so the next allocation from that bin will come from the buffer whose pointer is at index 3 in the bin. The allocation time is constant once a bin with available buffers is found.

buffer-manager-2

3. To deallocate a buffer, if the buffer to be deallocated has its pointer at the end of its bin, just decrement the number of reserved buffers for that bin, like this:

buffer-manager-3

4. If the buffer to be deallocated does not have its pointer at the end of the bin, then swap its pointer with the pointer at the end before decrementing the number of reserved buffers, like this:

buffer-manager-4

Another standard approach is to use two linked lists for each bin, an allocated list and a free list. This approach has the same algorithmic complexity as what I sketched above, but I find the code for this approach to be less elegant.

Svc/BufferManager/BufferManagerComponentImpl.cpp
bocchino: Also, it's a good idea to assert that `currStruct` is in bounds before using it to access `m_buffers`.

bocchino:
Another variant is to use a bounded FIFO queue of indices for each bin. The max size of each queue is the number of buffers in the bin. The queue stores the array indices of buffers that are available. Initially, each queue is full (it contains one copy of each buffer index). Then allocation by dequeuing and index and deallocation is by enqueuing an index. You can implement the bounded FIFO queue with an array and a size. A bin has buffers available if its size is greater than zero. I think this way is slightly easier than what I described above.

LeStarch:
Note: The heritage BufferManager is available here: https://github.com/fprime-community/fprime-contributions/tree/master/BufferManager

Svc/BufferManager/test/ut/Tester.cpp
```diff
@@ -46,23 +115,395 @@ namespace Svc {
...
+
+      for (NATIVE_UINT_TYPE b=0; b<BIN1_NUM_BUFFERS; b++) {
+          NATIVE_UINT_TYPE entry;
+          while (1) {
+              entry = rand() % BIN1_NUM_BUFFERS;
+              if (not returned[entry]) {
+                  returned[entry] = true;
+                  break;
+              }
+          }
```
saba-ja, line 256: This part causes an infinite loop which halts execution of Unit Test.

saba-ja: This issue is resolved in pull request https://github.com/nasa/fprime/pull/338

Svc/BufferManager/BufferManagerComponentImpl.cpp
saba-ja: Will be followed under issue https://github.com/nasa/fprime/issues/342

Svc/BufferManager/BufferManagerComponentImpl.cpp
saba-ja: This issue is resolved in pull request #338

Svc/BufferManager/BufferManagerComponentImpl.cpp
saba-ja: Will be followed under issue https://github.com/nasa/fprime/issues/340

LeStarch:
Work migrated to #366.

PR #290, opened 2020-12-01
mstarch: repairing seqgen

LeStarch:
Does the following:

1. Adds in `fprime-seqgen` tool to python environment
2. Corrects example files
3. Adds documentation
4. Adds an integration test to test sequences

### DIFF
--- a/Ref/test/int/ref_integration_test.py
+++ b/Ref/test/int/ref_integration_test.py
@@ -2,6 +2,7 @@ import os
 import sys
 import time
 import platform
+import subprocess
 from enum import Enum
 
 filename = os.path.dirname(__file__)
@@ -34,6 +35,7 @@ class TestRefAppClass(object):
         logpath = os.path.join(filename, "./logs")
         cls.api = IntegrationTestAPI(cls.pipeline, logpath)
         cls.case_list = []  # TODO find a better way to do this.
+        cls.dictionary = path
 
     @classmethod
     def teardown_class(cls):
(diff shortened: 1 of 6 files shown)


Ref/test/int/ref_integration_test.py
```diff
@@ -264,3 +266,9 @@ def test_active_logger_filter(self):
             self.api.assert_event_count(pred, actHI_events)
         finally:
             self.set_default_filters()
+
+    def test_seqgen(self):
+        """ Tests the seqgen code """
+        sequence = os.path.join(os.path.dirname(__file__), "test_seq.seq")
+        assert subprocess.run(["fprime-seqgen", "-d", self.dictionary, sequence, "/tmp/ref_test_int.bin"]).returncode == 0, "Failed to run fprime-seqgen"
```
Joshua-Anderson, line 273: You may want to make this a temporary file/directory with a random name https://docs.python.org/3/library/tempfile.html

Makes the system slightly more resilient against unexpected state, like where the file already exists (like where the integration test previously crashed).

docs/UsersGuide/gds/seqgen.md
```diff
@@ -0,0 +1,58 @@
...
+command of any command sequencer instance defined in the flight system.
+
+## Writing an F´ Sequence File
+
+F´ sequence files consist of lists of commands. These commands start with a time argument, followed by the command
+mnemonic, and lastly any arguments to command.  See the example below, which are all pulled from the example sequence
+file found here: [simple_sequence.seq](https://github.com/nasa/fprime/blob/devel/Gds/examples/simple_sequence.seq).
+
+```
+A2015-075T22:32:40.123 cmdDisp.CMD_NO_OP
```
Joshua-Anderson, line 17: Could you possibly clarify the time format of the relative and absolute date strings?

Joshua-Anderson (commented):
LGTM, just left some minor suggestions you can take or leave.

Ref/test/int/ref_integration_test.py
LeStarch: tmpfile can create longer file paths and there is an issue with cmdSeq and long paths.  So I'll leave this for now.

docs/UsersGuide/gds/seqgen.md
LeStarch: Fixed.

Ref/test/int/ref_integration_test.py
Joshua-Anderson: How does the length of the filepath on the host system affect seqgen? I can get that the remote filepath length will be limited, but the host path shouldn't be used anywhere should it?

Joshua-Anderson (approved):

LeStarch:
Canceled stalled workflow. No code changes since last success.

PR #293, opened 2020-12-02
Fix GCC Compiler Warnings

Joshua-Anderson:
I also enable '-Wall -Werror' on the root fprime project, which is only used
by CI and fprime developers to ensure that compiler warnings aren't accidentally
introduced in the future

### DIFF
--- a/Drv/LinuxGpioDriver/LinuxGpioDriverComponentImpl.cpp
+++ b/Drv/LinuxGpioDriver/LinuxGpioDriverComponentImpl.cpp
@@ -50,7 +50,11 @@ namespace Drv {
 
         // TODO check value of len
         len = snprintf(buf, sizeof(buf), "%d", gpio);
-        (void) write(fd, buf, len); // TODO check return value
+        if(write(fd, buf, len) != len) {
+            (void) close(fd);
+            DEBUG_PRINT("gpio/export error!\n");
+            return -1;
+        }
         (void) close(fd);
 
 	/* NOTE(mereweth) - this is to allow systemd udev to make
@@ -77,7 +81,11 @@ namespace Drv {
 
         // TODO check value of len
         len = snprintf(buf, sizeof(buf), "%d", gpio);
-        (void) write(fd, buf, len); // TODO check return value
+        if(write(fd, buf, len) != len) {
+            (void) close(fd);
+            DEBUG_PRINT("gpio/unexport error!\n");
+            return -1;
+        }
         (void) close(fd);
 
 	/* NOTE(mereweth) - this is to allow systemd udev to make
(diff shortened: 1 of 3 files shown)


Joshua-Anderson:
@LeStarch I had to touch the linux gpio driver to resolve some warnings. I highly doubt I broke anything, but would it be possible for you to quickly test gpio functionality when testing the 2.0 release on the raspberry pi?

Drv/LinuxGpioDriver/LinuxGpioDriverComponentImpl.cpp
```diff
@@ -105,12 +113,13 @@ namespace Drv {
...
         }
 
-        // TODO check return value
-        if (out_flag) {
-            (void) write(fd, "out", 4);
-        }
-        else {
-            (void) write(fd, "in", 3);
+        const char *dir = out_flag ? "out" : "in";
+        len = strlen(dir);
```
LeStarch, line 117: Do we need strnlen here?  Strlen can malfunction if the string is not terminated.

Drv/LinuxGpioDriver/LinuxGpioDriverComponentImpl.cpp
```diff
@@ -311,64 +324,63 @@ namespace Drv {
   void LinuxGpioDriverComponentImpl ::
     intTaskEntry(void * ptr) {
 
-      FW_ASSERT(ptr);
-      LinuxGpioDriverComponentImpl* compPtr = (LinuxGpioDriverComponentImpl*) ptr;
+    FW_ASSERT(ptr);
+    LinuxGpioDriverComponentImpl* compPtr = (LinuxGpioDriverComponentImpl*) ptr;
```
LeStarch, line 328: Can you use a static_cast here.  (LinuxGpioDriverComponentImpl*) can be dangerous.

LeStarch: Is all of this just formatting?  If so, don't change it.

LeStarch (approved):

LeStarch:
@Joshua-Anderson yes we can test it.

Drv/LinuxGpioDriver/LinuxGpioDriverComponentImpl.cpp
Joshua-Anderson: In this case it's looking to two constant strings "out" and "in". I though making it dynamic was better than hard coding the string length and having somebody accidentally break it into the future.

Drv/LinuxGpioDriver/LinuxGpioDriverComponentImpl.cpp
Joshua-Anderson: This is just formatting cause it was mixed tabs, 2 and 4 space indentation. I can revert this I guess but it was pretty bad.

PR #304, opened 2020-12-08
Separate fprime-util generate for different build types

### LINKED ISSUE #295: Separate Release and Unit Test Generation
Often when working with fprime, the release build targets one platform and unit tests target another platform.
Right now we always generate a release and testing build cache for a platform, regardless of whether they're used.

By only generating build caches that are actually used, we can cut down on build time.
We've also seen issues with deployments like fprime-sphinx where generating the testing build cache fails for their target platform. Even though that platform doesn't need the testing build cache, the current fprime-util doesn't offer a way to skip generating the cache and avoid the generation errors.

Steps:

- Remove `--build-type` argument in `fprime-util`: The `--ut` flag implies a build type of testing.
- Only generate the supplied build type build cache in `fprime-utl generate`.
- Update error message when building without a build-cache to specify which arguments to `fprime-util generate` are necessary to create the specific build cache.
- Add `default_ut_toolchain` option to `settings.ini`: This sets the default toolchain for the testing build type. If this argument is not set, fall back on a native build *not* `default_toolchain`, the default toolchain for release builds. It's rare for unit tests to target the embedded platform, so it's better to default to building unit tests for the host rather than default target.

Joshua-Anderson:
As part of this, I got rid of the `--build-type` flag and instead had the build type default to RELEASE unless the `--ut` flag is passed, in which case the build type is set to TESTING. This should allow us to add a `--debug` flag in the future and support a DEBUG build type.

This plays a little strangely with the fprime-util cmake targets. I thought about having the cmake targets use the global build type, but ran into some obstacles:

- Most targets only make sense for a single build type. I.E. `check` will only run on TESTING build types.
- Even for targets that make sense to run in multiple build types, I.E. build, by using the global build type we lose the ability to show all the potential target configurations in `fprime-util info`. So if you listed targets you'd only see `build` and not `build` + `build --ut`.

Instead, I figured the simplest solution was to never use the global build type when building a cmake target. Instead, every cmake target has a specific build type it can run against (as opposed to a list as before). The downside of this is that we have to specifically to list every potential target and build type configuration, but due to the limited number of build types and the fact that most targets only run against a single build type I don't see this as a major obstacle.

Resolves #295

### DIFF
--- a/Fw/Python/src/fprime/fbuild/builder.py
+++ b/Fw/Python/src/fprime/fbuild/builder.py
@@ -66,7 +66,7 @@ class Target(ABC):
         self,
         mnemonic: str,
         desc: str,
-        build_types: List[BuildType] = None,
+        build_type: BuildType = None,
         flags: set = None,
         cmake: str = None,
     ):
@@ -81,10 +81,8 @@ class Target(ABC):
         """
         self.mnemonic = mnemonic
         self.desc = desc
-        self.build_types = (
-            build_types
-            if build_types is not None
-            else [BuildType.BUILD_NORMAL, BuildType.BUILD_TESTING]
+        self.build_type = (
+            build_type if build_type is not None else BuildType.BUILD_NORMAL
         )
         self.flags = flags if flags is not None else set()
         self.cmake_target = cmake if cmake is not None else mnemonic
@@ -141,7 +139,6 @@ class Target(ABC):
         Returns:
             single matching target
         """
-        # matching = [target for target in cls.get_all_targets() if target.mnemonic == mnemonic and flags == target.flags]
         matching = []
         for target in cls.get_all_targets():
             if target.mnemonic == mnemonic and flags == target.flags:
@@ -257,8 +254,11 @@ class Build:
             not self.build_dir.exists()
             or not (self.build_dir / "CMakeCache.txt").exists()
         ):
+            gen_args = " --ut" if self.build_type == BuildType.BUILD_TESTING else ""
             raise InvalidBuildCacheException(
-                "{} invalid build cache. Please (re)generate.".format(build_dir)
+                "'{}' is not a valid build cache. Generate this build cache with 'fprime-util generate{}'".format(
+                    self.build_dir, gen_args
+                )
             )
 
     def get_settings(
@@ -341,10 +341,7 @@ class Build:
             and isinstance(target, LocalTarget)
         ]
         global_targets = [
-            target
-            for target in BUILD_TARGETS
-            if isinstance(target, GlobalTarget)
-            if self.build_type in target.build_types
+            target for target in BUILD_TARGETS if isinstance(target, GlobalTarget)
         ]
 
         relative_path = self.cmake.get_project_relative_path(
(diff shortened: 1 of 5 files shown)


Fw/Python/src/fprime/fbuild/builder.py
```diff
@@ -344,7 +340,6 @@ def get_build_info(self, context: Path) -> dict:
             target
             for target in BUILD_TARGETS
             if isinstance(target, GlobalTarget)
-            if self.build_type in target.build_types
```
LeStarch, line 347: I don't understand this removal.  Should this happen anyway?

Fw/Python/src/fprime/fbuild/builder.py
```diff
@@ -580,52 +575,47 @@ class AmbiguousToolchainException(FprimeException):
     LocalTarget(
         "build",
         "Build unit tests",
-        build_types=[BuildType.BUILD_TESTING],
```
LeStarch, line 583: Are these targets no longer supported on multiple build types?

Fw/Python/src/fprime/util/build_helper.py
```diff
@@ -71,21 +67,9 @@ def get_build(
         build meeting the specifications on build type
     """
     build_types = target.build_types if target is not None else BuildType
```
LeStarch, line 69: Wouldn't this need to be build_type not build_types.  You reduced the targets to a single item.

Fw/Python/src/fprime/fbuild/builder.py
Joshua-Anderson: If you saw my explanation above, this was the compromise I made. Targets supporting multiple build types caused issues cause they wouldn't list properly in `fprime-util info` and implicitly setting the build type to release required users to manually pass in `--ut` in order to run targets like `fprime-util check`.

LeStarch (commented):
Some comments.  I will look again when CI passes.

Fw/Python/src/fprime/fbuild/builder.py
LeStarch: So what happens (eventually) when we run build --debug?  Would this not allow us to build something using the `debug`?

Fw/Python/src/fprime/fbuild/builder.py
Joshua-Anderson: This would filter the output of `fprime-util info` based on the build type. Since I don't see a major use case for filtering based on build type, I figured it was easier to remove this functionality than to update it to work with the other code changes

Joshua-Anderson:
@LeStarch CI passing, ready for re-review

LeStarch:
@Joshua-Anderson we probably should update the install guide to point this out too.

Joshua-Anderson:
@LeStarch The fprime installation guide doesn't actually run the unit tests (though it still references `fprime-util install` 😬 ) , so it doesn't technically need to run `fprime-util generate --ut`. If a new users does try to run the unit tests, they'll get an error message to run `fprime-util generate --ut`. Do you think adding `fprime-util generate --ut` is necessary?

LeStarch:
@Joshua-Anderson I see your point.  No need to include it.

LeStarch (approved):

PR #306, opened 2020-12-09
lestarch: add in issue templates

LeStarch:
Adds in PR/bug/feature templates for creating issues and PRs!!

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,0 +1,30 @@
+---
+name: Bug Report
+about: Report a problem using F´
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+|**Field**|**Value**|
+|:---|:---|
+|**_Report Date_**| [DATE] |
+|**_Reported By_**| [USER] |
+|**_Branch/Hash_**| |
+|**_Short Description_**| |
+|**_Affected Component_**|   |
+---
+## Problem Description
+
+A description of the problem with sufficient detail to understand the issue.
+
+## How to Reproduce
+
+1. 
+2.
+3.
+
+## Expected Behavior
+
+A description of the expected behavior.
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -6,13 +6,9 @@ labels: bug
 assignees: ''
 
 ---
-
-|**Field**|**Value**|
+| | |
 |:---|:---|
-|**_Report Date_**| [DATE] |
-|**_Reported By_**| [USER] |
-|**_Branch/Hash_**| |
-|**_Short Description_**| |
+|**_F´ Version_**| |
 |**_Affected Component_**|   |
 ---
 ## Problem Description
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,23 @@
+|**Field**|**Value**|
+|:---|:---|
+|**_Submission Date_**| [DATE] |
+|**_Submitter_**| [USER] |
+|**_Originating Project/Creator_**| |
+|**_Base Branch_**| devel |
+|**_Short Description_**| |
+|**_Affected Component_**|  |
+|**_Affected Architectures(s)_**|  |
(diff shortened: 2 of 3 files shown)


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,30 @@
...
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+|**Field**|**Value**|
+|:---|:---|
+|**_Report Date_**| [DATE] |
+|**_Reported By_**| [USER] |
```
Joshua-Anderson, line 13: I'm wondering if we can remove Report Date / Reported By fields.  90% of the time, this will be the user who opened the issue and the date the issue was opened. In the rare cases where somebody is opening a ticket on behalf of somebody else they can directly note it in the description.

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,30 @@
...
+|**_Branch/Hash_**| |
```
Joshua-Anderson, line 14: Since most users will be using a fprime release we may want to call this fprime version or release rather than branch.

.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
...
+|**_Submission Date_**| [DATE] |
+|**_Submitter_**| [USER] |
+|**_Originating Project/Creator_**| |
+|**_Base Branch_**| devel |
+|**_Short Description_**| |
+|**_Affected Component_**|  |
+|**_Affected Architectures(s)_**|  |
+|**_Related Issue(s)_**|  |
+|**_Has Unit Tests (y/n)_**|  |
+|**_Build Checked (y/n)_**|  |
```
Joshua-Anderson, line 12: Build Checked sounds a little awkward to me. Would 'Changes Compile' possibly sound a bit better?

.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
+|**Field**|**Value**|
+|:---|:---|
+|**_Submission Date_**| [DATE] |
+|**_Submitter_**| [USER] |
+|**_Originating Project/Creator_**| |
+|**_Base Branch_**| devel |
```
Joshua-Anderson, line 6: Do we need this if the base branch is a part of the pull request itself?

.github/pull_request_template.md
```diff
@@ -0,0 +1,23 @@
...
+|**_Short Description_**| |
+|**_Affected Component_**|  |
+|**_Affected Architectures(s)_**|  |
```
Joshua-Anderson, line 9: Is this relevant for most changes? In the rare case where a particular architecture is affected I'd expect it to come up as a part of the change description.

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,30 @@
...
+|**_Short Description_**| |
```
Joshua-Anderson, line 15: What's the difference between the title and the short description? Do we need both? I feel like when I'm filing this out I'd just copy and paste in the title.

Joshua-Anderson (changes requested):
I think adding a template is a good idea, but we're going to be filling this out a lot so I'd like to advocate for sliming the templates down as much as possible.

.github/pull_request_template.md
LeStarch: This is here because the team wanted a quick way of tracking changes in C++ code that affect certain architectures.

Joshua-Anderson: Is a field in the template the best place to track this? Architecture specific changes are rare and because it's a field in the header it's not easy to search for. What about tagging to pull requests that affect specific architectures?

LeStarch: This field header is really to make reviewing and testing quick.  The reason this field is there is so the following process can be done:

1. Look at architecture
2. Spin-up proper architecture
3. Follow testing recommendations

It isn't so much about searching, but specifically calling out if a change needs to be tested with specific hardware.

Joshua-Anderson (approved):

Joshua-Anderson (approved):

LeStarch:
Doesn't affect build or LGTM.  Merging.

PR #317, opened 2020-12-16
FileDownlink: Add file downlink queue and SendFile port

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**| FileDownlink  |
|**_Related Issue(s)_**|  |
|**_Builds Without Errors (y/n)_**| y  |

---
## Change Description

SendFile now adds files to a file downlink queue. FileDownlink automatically pulls and downlinks files from the queue,
allowing users to queue up a list of files to downlink.

A SendFile port was added, allowing components to add files to the file downlink queue. When a file downlink initated by a port completes, a notification is broadcast on the FileComplete port, allowing components to react when files they requested were downlinked.

In order to prevent a queue of files from saturating the communication link, a cooldown state was added, allowing users to specify a delay between when a file downlink completes and the next one begins.

## Rationale

It can be useful to allow components to initiate a file downlink. Previously file downlink only supported downlinking a single file at a time. If multiple components tried to downlink files, one would get a busy error message. In order to support multiple components well, we either needed to add error and retry handling to every component enqueuing files, or allow queuing filles to filedownlink.

- There's no less that a half dozen string types file names go through, I'm a little concerned that the size restrictions on a particular type may cause problems.
- I only added a single new test, a test where a file is downlinked via a port, verifying that the file was transfered and a FileComplete message was sent. I feel this is sufficient, but if anybody thinks of test cases that would be valuable to cover let me know.

## Future Work

None


Svc/FileDownlink/FileDownlink.hpp
```diff
@@ -209,6 +210,22 @@ namespace Svc {
           FileDownlink *const fileDownlink;
 
       };
+
+      //! Sources of send file requests
+      enum CallerSource { COMMAND, PORT };
+      
+      //! Used to track a single file downlink request
+      struct FileEntry {
+        char srcFilename[101]; // Name of requested file
```
LeStarch, line 219: Why are these lengths hard-coded?

Svc/FileDownlink/FileDownlink.cpp
```diff
@@ -127,106 +228,158 @@ namespace Svc {
...
+          sendResponse(SendFileStatus::OK);
       }
       return;
     }
 
     // If the amount to downlink is greater than the file size, emit a Warning and then allow
     // the file to be downlinked anyway
     if (startOffset >= this->file.size) {
-        this->enterIdle();
+        this->enterCooldown();
```
LeStarch, line 372: I think we can skip cool-down if this file failed to start...as we likely just had one.

Svc/FileDownlink/FileDownlink.cpp
```diff
@@ -127,106 +228,158 @@ namespace Svc {
...
+    case SendFileStatus::INVALID:
+      return Fw::COMMAND_VALIDATION_ERROR;
+    case SendFileStatus::BUSY:
+        return Fw::COMMAND_BUSY;
+    default:
+        // Trigger assertion if given unknown status
+        FW_ASSERT(false);
+    }
+
+    // It's impossible to reach this, but added to suppress gcc missing return warning
```
LeStarch, line 325: If FW_ASSERT is turned of then this is possible.

LeStarch (changes requested):
Some static buffer sizes and a few other comments. Otherwise looks pretty good.

Svc/FileDownlink/FileDownlink.hpp
LeStarch: It matches the port sizes.

LeStarch (approved):

PR #324, opened 2021-01-07
Update fprime-gds option and path being used in INSTALL.md

Ruben1729:
## Change Description

It's just an update to the INSTALL.md file. `fprime-gds -g html -d <path to fprime checkout>/Ref` does not work. The option -d doesn't exist, I'm currently using -r instead. Additionally, the path I've used to successfully run the gds has been <path to prime checkout>/Ref/build-artifacts.


LeStarch:
Awesome!!!  I have requested a review from the developer who mad this original change.

Joshua-Anderson (approved):
Good catch!

PR #330, opened 2021-01-11
fix: Updated failing Framer UT

saba-ja:
## Change Description

Framer component ports were renamed for consistency which caused the UT
to break. "bufferReturn" is renamed "bufferDeallocate", "allocate" is renamed to “framedAllocate”.
Framer UT is updated now and tests pass.


LeStarch (approved):
Thanks!!

PR #336, opened 2021-01-19
update: fixes #335

### LINKED ISSUE #335: Receives [ERROR] Unknown arguments: -b when running ./docs/doxygen/generate_docs.bash
| | |
|:---|:---|
|**_F´ Version_**| 1.5>=|
|**_Affected Component_**|all   |
---
## Problem Description

Receives [ERROR] Unknown arguments: -b when running:
```./docs/doxygen/generate_docs.bash```

## How to Reproduce
run the following from fprime directory:
```./docs/doxygen/generate_docs.bash```

## Expected Behavior
Generate doxygen and sphinx documents

### PR DESCRIPTION
**Pull request by saba-ja**
## Change Description

Failing issue of generate_docs.bash when generating doxygen and sphinx is resolved. 

1. The script will use a temporary folder name with a prefix that is used by fprime-util build. 
2. Sphinx dependency packages are checked before attempting to run sphinx.

### DIFF
--- a/docs/doxygen/generate_docs.bash
+++ b/docs/doxygen/generate_docs.bash
@@ -50,18 +50,18 @@ function clobber
     cd "${FPRIME}"
     clobber "${DOXY_OUTPUT}"
     (
-        mkdir -p "${FPRIME}/docs-build-for-docs"
-        cd "${FPRIME}/docs-build-for-docs"
+        mkdir -p "${FPRIME}/build-fprime-automatic-docs"
+        cd "${FPRIME}/build-fprime-automatic-docs"
         cmake "${FPRIME}" -DCMAKE_BUILD_TYPE=RELEASE
     )
-    fprime-util build -b "${FPRIME}/docs-build-for-docs" -j32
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: docs/doxygen/generate_docs.bash [hunk 1]
```diff
@@ -75,6 +75,7 @@ function clobber
 # Python
 (
     clobber "${PY_OUTPUT}"
+    pip install sphinx sphinx-rtd-theme sphinx-autoapi sphinx-autoapi recommonmark sphinxcontrib.mermaid
```
LeStarch, line 78: Why is this line needed?  It seems like it should already be part of the VENV.

saba-ja: I agree, they could be moved to 
https://github.com/nasa/fprime/blob/261ae546d93feb779536bee393caeb8c5140167d/Fw/Python/setup.py#L87

extras_require={"dev": ["black", "pylama", "pylint", "pre-commit"]},

saba-ja:
Will re-submit the pull request with the modification specified above.

PR #346, opened 2021-02-01
fix: missing GDS command name in event tb

saba-ja:
| | |
|:---|:---|
|**_Originating Project/Creator_**| Fprime|
|**_Affected Component_**|  GDS|
|**_Affected Architectures(s)_**|  events.js|
|**_Related Issue(s)_**|  GDS command dispatch is too vague|
|**_Has Unit Tests (y/n)_**|  N|
|**_Builds Without Errors (y/n)_**|  Y|
|**_Unit Tests Pass (y/n)_**|  NA|

---
## Change Description
Reported by @bcmetz:
image

A lot of instruments have similar functions, the above picture is an example, lots of instruments may want to adjust various voltage or light "amplitudes". A user cannot tell where this command came from without looking up the opcode and does not know where it is going without looking up the port number. We are going to have many sequences dispatching commands in parallel so the command dispatch EVRs should be as clear as possible.

Request

Include the component name and the command name in the event description, example for above bp_C4D.AMPLTIUDE
Add the connection name from the XML next to the port number, this will allow us to understand which instance of the component it is going to
Example Command Description:
```bp_C4D.AMPLITUDE dispatched to port bp_C4DCmd (36)```

Now I now exactly what command went out and to which instance it went to. We have components in the system that are going to be used multiple times (the CE_FPM component comes to mind).

## Rationale
image

Command name is now shown in the table with the following format:
```Command.mnemonic```

## Future Work

Port name needs to be added.
Per @LeStarch the second request (labeling the destination port) is considerably more complicated to the point of possibly needing to re-architect all of F´, or rewriting large portions of the F´ GDS to support this.

### CODE REVIEW

File: Gds/src/fprime_gds/flask/static/js/vue-support/event.js [hunk 1]
```diff
@@ -75,15 +75,18 @@ Vue.component("event-list", {
             // Remap command EVRs to expand opcode for visualization pruposes
             let groups = null
             if (item.template.severity.value == "EventSeverity.COMMAND" && (groups = display_text.match(OPREG)) != null) {
+                let component = "UNKNOWN";
                 let mnemonic = "UNKNOWN";
                 let id = parseInt(groups[1]);
                 for (let command in this.commands) {
                     command = this.commands[command];
                     if (command.id == id) {
+                        component = command.comp_name;
```
LeStarch, line 84: Try using "command.full_name", it should already be comp.command and is what we used to execute commands.

LeStarch (commented):
We should make this change against devel.  I will update your base-branch.

LeStarch:
@saba-ja any chance you can cherry-pick the changes into devel.  I would prefer to code-frees this ground/prototype branch at this point since we are ready to break it into PRs.  That means less stuff we have to remember to transition over.

Gds/src/fprime_gds/flask/static/js/vue-support/event.js
saba-ja: Fixed this in [38f4f7e](https://github.com/nasa/fprime/pull/346/commits/38f4f7e42f647624faa18b3525cea17887c4c0b9)

Issue #348, opened 2021-02-04
Type Enums are printed as NOSPEC to stdout


saba-ja:
| | |
|:---|:---|
|**_F´ Version_**|1.5> |
|**_Affected Component_**|   all |
---
## Problem Description

Enum types are printed as NOSPEC to stdout
The following example is from SignalGen

![image](https://user-images.githubusercontent.com/35859004/106944614-a9f1d000-66db-11eb-9dbc-c8019530fd66.png)

The Signal types are defined here:
https://github.com/nasa/fprime/blob/ca71e1278cf61fba4b6faf67c9f049a3a5f423e5/Ref/SignalGen/SignalTypeEnumAi.xml#L5-L8

## How to Reproduce

1. `fprime-gds -n`
2. `./fprime/Ref/build-artifacts/Darwin/bin/Ref -a 127.0.0.1 -p 50000`
3. In the GUI commanding page
![image](https://user-images.githubusercontent.com/35859004/106945338-854a2800-66dc-11eb-8fd5-1916eaf8f625.png)

Observe the printed EVR in stdout as following 
![image](https://user-images.githubusercontent.com/35859004/106944614-a9f1d000-66db-11eb-9dbc-c8019530fd66.png)

## Expected Behavior
The EVR should correctly print out the ENUM name instead of NOSPEC

saba-ja:
Per @LeStarch :
The fix would be to improve the Autocoder template for enums to define a “toString” method that produces the correct string. It requires a change to the Autocoder.

saba-ja:
## Root cause:
Enum toString is not implemented in:
Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_cpp.tmpl

This causes the parent Serializable class Fw/Types/Serializable.cpp be used which only returns NOSPEC.

## Fix:
The fix can be a patch similar to the following:

In file `Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_hpp.tmpl` add:
```
\#include "Fw/Types/EightyCharString.hpp"
```
...
```
\#if FW_SERIALIZABLE_TO_STRING || BUILD_UT
    //! ENUM toString
    void toString(Fw::StringBase& text) const;  //!< generate text from serializable
\#endif
```

In file `Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_cpp.tmpl` add:

```
  \#if FW_SERIALIZABLE_TO_STRING || BUILD_UT
[...]

PR #350, opened 2021-02-07
Update/ground prototype

saba-ja:
| | |
|:---|:---|
|**_Originating Project/Creator_**| FPrime|
|**_Affected Component_**|  All|
|**_Affected Architectures(s)_**| FSW |
|**_Related Issue(s)_**|  [#348](https://github.com/nasa/fprime/issues/348)

---
## Change Description

Enum toString is not implemented in:
Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_cpp.tmpl
This causes the parent Serializable class Fw/Types/Serializable.cpp be used which only returns NOSPEC.

## Rationale

The fix can be a patch similar to the following:

In file `Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_hpp.tmpl` add:

```
\#include "Fw/Types/EightyCharString.hpp"
```
...

```
\#if FW_SERIALIZABLE_TO_STRING || BUILD_UT
    //! ENUM toString
    void toString(Fw::StringBase& text) const;  //!< generate text from serializable
\#endif
```

In file `Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_cpp.tmpl` add:

```
  \#if FW_SERIALIZABLE_TO_STRING || BUILD_UT
    void ${name} :: toString(Fw::StringBase& text) const {
    const ${name}::t e = this->e;
    switch (e) {
#for $item_name,$item_value,$item_comment in $items_list:
        case ${name}::${item_name}:
          text = "${item_name}";
          break;
#end for
        default:
          text = "[invalid]";
          break;
    }
  }
  \#endif
```

The following is the stdout after the fix:
image

## Future Work

Now the remaining question before adding the ENUM toString is the following:
Is `FW_SERIALIZABLE_TO_STRING` sufficient as condition for implementing the toString for ENUM or a new condition needs to be added in `config/FpConfig.hpp`?


LeStarch:
@saba-ja can you replay your changes on devel?  They look good, but should be merged with devel not ground prototype.

PR #358, opened 2021-02-11
lestarch: adding in fprime-util new

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**| fprime-util |
|**_Related Issue(s)_**|  |
|**_Builds Without Errors (y/n)_**| n  |
|**_Unit Tests Pass (y/n)_**| n  |
|**_Documentation Included (y/n)_**| y -- CLI |

---
## Change Description

Adds in a `fprime-util new` command to create new components.  Thanks to Georgia Tech's cookiecutter.

## Rationale

Simplifies the process of making components.

## Future Work

1. Add in Port and Topology cookiecutters
2. Allow specifying project-supplied cookiecutter package
3. Clean-up versioning of cookiecutters

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -320,6 +320,7 @@ confparse
 Connectedoutput
 Consolas
 consts
+cookiecutter
 cooldown
 coor
 coravy
--- a/Fw/Python/setup.py
+++ b/Fw/Python/setup.py
@@ -29,8 +29,8 @@ setup(
     # information should match the F prime description information.
     ####
     name="fprime-tools",
-    use_scm_version = {
-        "root": os.path.join("..",".."),
+    use_scm_version={
+        "root": os.path.join("..", ".."),
         "relative_to": __file__,
     },
     license="Apache 2.0 License",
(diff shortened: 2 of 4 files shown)


LeStarch:
For safety, when possible I like to avoid deleting files. Is it possible to send a PR to remove the makefiles from the cookiecutter template instead?

LeStarch:
Just deleting the first 11 lines of the file feels really brittle and dangerous, what if the cookiecutter template is updated? Is there a smarter way we can merge the files?

LeStarch:
I don't think we want to have fprime using the latest version of the cookiecutter template. What if the template is updated for the latest version of F', but the user is using an older version of fprime?

What I would do is version the cookiecutter template with fprime, and have fprime-util pull the template version corresponding to the users fprime version.

I'd also prefer to ship the cookiecutter template as part of the fprime-util package (to install it when you pip install fprime-util), since it's nice not to pull in internet dependencies when possible, but this might conflict with however the cookiecutter template is versioned and will require more thought.

LeStarch:
Comments pulled from #357 context available there.

LeStarch:
Decisions: add a "prototype" warning, and ship as a prototype feature.  We can add the ability to change cookiecutter templates later.

Joshua-Anderson (approved):
Since CI no longer unit tests fprime-util, approval is contingent on fprime-util unit tests passing locally.

LeStarch:
Confirming local UT run passes.

PR #399, opened 2021-02-20
lestarch: fixing missing autocodes with time role port

LeStarch:
| | |
|:---|:---|
|**_Originating Project/Creator_**| infrastructure |
|**_Affected Component_**| Autocoders |
|**_Related Issue(s)_**|  |

---
## Change Description

Adds in autocoding when a time role port is specified by-hand (not only calculated by existence of telemetry or events).

## Rationale

`this->timeGet();` should be available for use when without requiring events or channels such that users may easily request time for other purposes.

### DIFF
--- a/Autocoders/Python/src/fprime_ac/generators/visitors/ComponentVisitorBase.py
+++ b/Autocoders/Python/src/fprime_ac/generators/visitors/ComponentVisitorBase.py
@@ -442,6 +442,9 @@ class ComponentVisitorBase(AbstractVisitor.AbstractVisitor):
         c.has_output_ports = len(c.output_ports) > 0
         c.has_typed_output_ports = len(c.typed_output_ports) > 0
         c.has_serial_output_ports = len(c.serial_output_ports) > 0
+        roles = [role for name, ptype, sync, priority, role, max_number in c.output_ports]
+        c.has_time_get = "TimeGet" in roles
+
 
     def initPortIncludes(self, obj, c):
         c.port_includes = list()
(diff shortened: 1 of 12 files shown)


Autocoders/Python/src/fprime_ac/generators/visitors/ComponentVisitorBase.py
```diff
@@ -442,6 +442,9 @@ def initPortFlags(self, obj, c):
         c.has_output_ports = len(c.output_ports) > 0
         c.has_typed_output_ports = len(c.typed_output_ports) > 0
         c.has_serial_output_ports = len(c.serial_output_ports) > 0
+        roles = [role for name, ptype, sync, priority, role, max_number in c.output_ports]
```
Joshua-Anderson, line 445: Optional: use `_` for unused for loop variables.

Joshua-Anderson (commented):
Changes look good, but if it's possible adding an autocoder test case in `Autocoders/Python/test` covering this would be awesome.

Autocoders/Python/src/fprime_ac/generators/visitors/ComponentVisitorBase.py
LeStarch: Using `_` here would be inconsistent with the rest of the file.  So I intentionally avoided that.

Joshua-Anderson: :cry: I'm surprised LGTM isn't complaining about it.

LeStarch:
Unit test added.

Joshua-Anderson (approved):

PR #404, opened 2021-02-21
Ensure the same Python executable is used by subprocess

gershwinlabs:
| | |
|:---|:---|
|**_Originating Project/Creator_**|Bill Allen|
|**_Affected Component_**|GDS|
|**_Affected Architectures(s)_**|Python|
|**_Related Issue(s)_**|None|
|**_Has Unit Tests (y/n)_**|n|
|**_Unit Tests Pass (y/n)_**|n (none available)|
|**_Documentation Included (y/n)_**|n|

---
## Change Description

Changed `"python3"` to use `sys.executable`.  Also removed duplicate import of `sys`.

## Rationale

Calling `"python3"` directly could use the wrong Python executable.  Using `sys.executable` ensures the subprocess calls the same Python executable as the script, and therefore corresponds with the same dependency package installations.

Should be able to spin up GDS as normal.  Also checked with `pylint --disable=R,C`.

## Future Work

Did not find any other instances of `"python3"` being called directly.

### DIFF
--- a/Gds/src/fprime_gds/executables/run_deployment.py
+++ b/Gds/src/fprime_gds/executables/run_deployment.py
@@ -6,7 +6,6 @@
 import os
 import sys
 import platform
-import sys
 import webbrowser
 from pathlib import Path
 from fprime.fbuild.settings import (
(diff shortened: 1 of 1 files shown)


LeStarch (approved):
Great fix.  Thanks!

Issue #419, opened 2021-02-25
fprime-util impl autocoder does not use return type


metzdigital:
The return type used in the interface definition is ignored when using `fprime-util impl`, the autocoder return type is hard coded to always be `void`

![image](https://user-images.githubusercontent.com/2592986/109230394-a6e97d00-7779-11eb-9b3a-55a77334fc37.png)

After manually replacing `void` with `Backplane::CardConfigReturn` I could get this to compile. 

Here is my interface definition:
```
<interface name="CardConfig" namespace="Backplane">
    <comment>
        Port to grab card configuration details used by the backplane manager at initialization
    </comment>
    <import_serializable_type>Backplane/Common/CardConfigReturnSerializableAi.xml</import_serializable_type>
    <args>
        <!-- No args only the return is important -->
    </args>
    <return type="Backplane::CardConfigReturn" pass_by="value"> </return>
</interface>
```

saba-ja:
## Root cause:

The problem is from line 61

https://github.com/nasa/fprime/blob/598f850b0784735acc9a8b6e367a9cd5acdd0972/Autocoders/Python/src/fprime_ac/generators/templates/impl/hpp.tmpl#L53-L64

and 

https://github.com/nasa/fprime/blob/598f850b0784735acc9a8b6e367a9cd5acdd0972/Autocoders/Python/src/fprime_ac/generators/templates/impl/cpp.tmpl#L55-L70

The template always adds `void` for user-defined input port types.

## Fix:

To resolve the issue the following can be done:

In fprime/Autocoders/Python/src/fprime_ac/generators/templates/impl/hpp.tmpl
```c++
  #for $instance, $type, $sync, $priority, $full, $role, $max_num in $typed_user_input_ports:
      //! Handler implementation for $instance
      //!
      #set $return_type = $port_return_type_strs[$instance]
      ${return_type}${instance}_handler(
$emit_port_params([ $param_portNum ] + $port_params[$instance])
      );
```

In fprime/Autocoders/Python/src/fprime_ac/generators/templates/impl/cpp.tmp
```c++
  #for $instance, $type, $sync, $priority, $full, $role, $max_num in $typed_user_input_ports:
  #set $return_type = $port_return_type_strs[$instance]
  ${return_type}${name}ComponentImpl ::
    ${instance}_handler(
$emit_port_params([ $param_portNum ] + $port_params[instance])
    )
  {
  #if $return_type[:4] == "void":    
    // TODO
  #else:
    // TODO return
[...]

PR #420, opened 2021-02-26
Added option to impl templates to check for return type of user defined functions.

saba-ja:
Added option to impl templates to check for return type of user defined
functions.

| | |
|:---|:---|
|**_Originating Project/Creator_**| Fprime |
|**_Affected Component_**|  autocoder impl template|
|**_Affected Architectures(s)_**|  1.5 >|
|**_Related Issue(s)_**|  #419 |

---
## Change Description

Added option to check for the return type of user-defined ports. The previous template only returned `void` without checking for user-defined return types causing the generated template inconsistent with the user definition of the port return type.

## Rationale

Now the fprime-util impl will generate the correct return type for each function. 

Manually tested the change on one of FileDownlink Component which had a port with return type. The result is shown below:

Svc/FileDownlink/FileDownlinkComponentImpl.hpp-template
image

Svc/FileDownlink/FileDownlinkComponentImpl.cpp-template
image

## Future Work

Note: The user still needs to add the return statement manually after the template is generated.

### DIFF
--- a/Autocoders/Python/src/fprime_ac/generators/templates/impl/cpp.tmpl
+++ b/Autocoders/Python/src/fprime_ac/generators/templates/impl/cpp.tmpl
@@ -58,12 +58,17 @@ $emit_non_port_params($params_init_cpp)
   // ----------------------------------------------------------------------
 
   #for $instance, $type, $sync, $priority, $full, $role, $max_num in $typed_user_input_ports:
-  void ${name}ComponentImpl ::
+  #set $return_type = $port_return_type_strs[$instance]
+  ${return_type}${name}ComponentImpl ::
     ${instance}_handler(
 $emit_port_params([ $param_portNum ] + $port_params[instance])
     )
   {
+    #if $return_type[:4] == "void":    
     // TODO
+    #else:
+    // TODO return
+    #end if
   }
 
   #end for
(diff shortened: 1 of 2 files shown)


Autocoders/Python/src/fprime_ac/generators/templates/impl/cpp.tmpl
```diff
@@ -58,12 +58,17 @@ $emit_non_port_params($params_init_cpp)
...
 
   #for $instance, $type, $sync, $priority, $full, $role, $max_num in $typed_user_input_ports:
-  void ${name}ComponentImpl ::
+  #set $return_type = $port_return_type_strs[$instance]
+  ${return_type}${name}ComponentImpl ::
     ${instance}_handler(
 $emit_port_params([ $param_portNum ] + $port_params[instance])
     )
   {
+    #if $return_type[:4] == "void":
```
Joshua-Anderson, line 67: It might be worth rewriting this as `$return_type[:-1] == "void` to avoid edge cases where a type name starts with void.

Joshua-Anderson (commented):
I've been bitten by this before, thanks for fixing it!

Autocoders/Python/src/fprime_ac/generators/templates/impl/cpp.tmpl
saba-ja: Good point! Just updated the code to use 0 to -1 instead of 0 to 4.

Joshua-Anderson (approved):

Issue #356, opened 2021-02-10
Segment Fault after FATAL in Ref/PingReceiver


saba-ja:
| | |
|:---|:---|
|**_F´ Version_**| 1.5 > |
|**_Affected Component_**|  Ref/PingReceiver |
---
## Problem Description

When sending command pingRcvr.PR_StopPings from the GDS GUI, after a few second the FSW client exits with Segmentation fault: 11.

![image](https://user-images.githubusercontent.com/35859004/107449226-08fa7f00-6af8-11eb-8f7f-be8033a1b3be.png)

## How to Reproduce

1. cd fprime/Ref
2. frpime-util generate
3. frpime-util build
4. fprime-gds -n
5. ./build-artifacts/Darwin/bin/Ref -a 127.0.0.1 -p 50000
6. Send Command pingRcvr.PR_StopPings from GDS
![image](https://user-images.githubusercontent.com/35859004/107449050-b91bb800-6af7-11eb-8d83-e41232997578.png)

The Event should look like the following:
![image](https://user-images.githubusercontent.com/35859004/107449075-c20c8980-6af7-11eb-98ea-10db8acc9db6.png)
## Expected Behavior

FATA exit should not cause segment fault corruption.

PR #427, opened 2021-03-02
Change signal raised by fatal event handling

Kemyt:
| | |
|:---|:---|
|**_Affected Component_**| Svc/FatalHandler |
|**_Related Issue(s)_**| [#356](https://github.com/nasa/fprime/issues/356) |

---
## Change Description

Segmentation fault doesn't seem like an appropriate choice for handling Fatal events. SIGSEGV was replaced with SIGTERM.

## Rationale

Raising SIGTERM allows for cleaner exit on fatal event (if signal handler is implemented).

None

## Future Work

None

### DIFF
--- a/Svc/FatalHandler/FatalHandlerComponentLinuxImpl.cpp
+++ b/Svc/FatalHandler/FatalHandlerComponentLinuxImpl.cpp
@@ -27,9 +27,9 @@ namespace Svc {
             FwEventIdType Id) {
         // for **nix, delay then exit with error code
         Fw::Logger::logMsg("FATAL %d handled.\n",(U32)Id,0,0,0,0,0);
+        (void)raise( SIGTERM );
         (void)Os::Task::delay(1000);
         Fw::Logger::logMsg("Exiting.\n",0,0,0,0,0,0);
-        (void)raise( SIGSEGV );
         exit(1);
     }
 
--- a/Svc/FatalHandler/FatalHandlerComponentLinuxImpl.cpp
+++ b/Svc/FatalHandler/FatalHandlerComponentLinuxImpl.cpp
@@ -27,9 +27,9 @@ namespace Svc {
             FwEventIdType Id) {
         // for **nix, delay then exit with error code
         Fw::Logger::logMsg("FATAL %d handled.\n",(U32)Id,0,0,0,0,0);
-        (void)raise( SIGTERM );
         (void)Os::Task::delay(1000);
-        Fw::Logger::logMsg("Exiting.\n",0,0,0,0,0,0);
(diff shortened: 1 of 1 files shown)


timcanham:
SIGSEGV was chosen for Linux so that we could have a core file to look at.

Kemyt:
That makes sense. It could be confusing and lead to some unnecessary debugging. Maybe some additional message could be printed to let the user know that no actual memory access violation occured?

Svc/FatalHandler/FatalHandlerComponentLinuxImpl.cpp
```diff
@@ -27,9 +27,9 @@ namespace Svc {
             FwEventIdType Id) {
         // for **nix, delay then exit with error code
         Fw::Logger::logMsg("FATAL %d handled.\n",(U32)Id,0,0,0,0,0);
+        (void)raise( SIGTERM );
```
LeStarch, line 30: Why was this raise moved before the delay?

Svc/FatalHandler/FatalHandlerComponentLinuxImpl.cpp
```diff
@@ -27,9 +27,9 @@ namespace Svc {
...
         (void)Os::Task::delay(1000);
         Fw::Logger::logMsg("Exiting.\n",0,0,0,0,0,0);
```
LeStarch, line 32: Would it help to add a message here?  e.g. "Exiting with segfault and core file"?

Svc/FatalHandler/FatalHandlerComponentLinuxImpl.cpp
Kemyt: It was done so that if some deployment contained a SIGTERM handler that does some cleanup (similarly to [Ref](https://github.com/nasa/fprime/blob/0e731d14c161a5c5dc37a2ce2e55009353806247/Ref/Top/Main.cpp#L17) it would allow for a cleaner exit. In case of SIGSEGV it doesn't make a difference.

Svc/FatalHandler/FatalHandlerComponentLinuxImpl.cpp
Kemyt: Yes, I think so. Even if someone finds it suspicious, it will be easy enough to find such specific message in code to confirm that it's intentional.

LeStarch: Would you like to make that change to line 32 and submit it here? Otherwise I will open up an issue to correct the message.

LeStarch (approved):
Explains the segfault and keeps the core file for debugging. Also, I have lost many hours to this segfault, I am grateful to the clean message!

PR #432, opened 2021-03-03
fixed handling of redefined messages field types in Gds and GroundInterface

AaronMcDaniel:
| | |
|:---|:---|
|**_Originating Project/Creator_**|GT1/Georgia Tech |
|**_Affected Component_**| Svc/GroundInterface |

---
## Change Description

Edited the GroundInterface component to use the compilation flags defining the Packet Descriptor.
Edited the Gds to use the configuration manager's types when encoding and decoding commands, events, telemetry, parameters, and packet headers.

## Rationale

Fixes bug where configuring message fields in deployments causes the Gds to be unable to communicate with the deployment.
The compilation flags for configuring message fields are necessary on certain platforms with limited memory to reduce memory usage.

A deployment should be made that edits the (compilation flag, config manager field) pairs described in the included Gds/docs/README.md file are changed. This deployment should be able to have successful exchanges of commands, events, parameters, and telemetry. There should also be builds containing individual changes to each flag to ensure that all fields are set correctly.

## Future Work

The `FwEnumStoreType` flag is confirmed to break the communications with the Gds when it is changed from an I32. I do not know how to change the way this type is serialized since it is defined in the `fprime/common/models/serialize/enum_type.py` file which I don't believe has access to the configuration manager from the Gds.
 
There may be additional fields that can be configured with compilation flags, but will still cause issues with encoding and decoding that will need to be handled correctly to avoid broken communications.

### DIFF
--- a/Gds/docs/README.md
+++ b/Gds/docs/README.md
@@ -9,9 +9,9 @@ This tool includes logging capability and can be connected to several adapters t
 
 Certain deployments may alter the fields of messsages for various reasons including memory conservation.
 To make the corresponding changes to the Gds for successful communication, 
-please edit the Gds/src/fprime_gds/common/utils/config_manager.py file accordingly. 
-Each field in the _set_defaults() function corresponds to a specific compile time flag for an FPrime deployment.
-Bellow is a table of corresponding fields that must match between deployment and Gds in order for successful communication to occur.
+please edit the `Gds/src/fprime_gds/common/utils/config_manager.py` file accordingly. 
+Each field in the `_set_defaults()` function corresponds to a specific compile time flag in the `/config/FpConfig.hpp` file for an FPrime deployment.
+Below is a table of corresponding fields that must match between deployment and Gds in order for successful communication to occur.
 
 Gds Field | FPrime Flag
 ----------- | ----------- |
@@ -23,4 +23,4 @@ self.__prop['types']['event_id'] | FwEventIdType
 self.__prop['types']['pkt_id'] | Unkown
 self.__prop['types']['key_val'] | unknown
 
-It is important to note that the default value of all of these fprime fields is U32, which is the behavior when none of these compile time flags are set in any CMakeLists.txt fields.
+It is important to note that the default value of all of these fprime fields is U32, which is the behavior when none of these compile time flags are changed in the `FpConfig.hpp` file.
--- a/Gds/docs/README.md
+++ b/Gds/docs/README.md
@@ -1,0 +1,26 @@
+# Gds Command and Control Tool
+
+## 1. Introduction
+
+This is a tool intended to be used with FPrime dpeloyments to exchange commands, telemetry, events and other messages.
+This tool includes logging capability and can be connected to several adapters to communicate with ground station equipment.
+
+## 2. Message Field Configuration
+
+Certain deployments may alter the fields of messsages for various reasons including memory conservation.
+To make the corresponding changes to the Gds for successful communication, 
+please edit the Gds/src/fprime_gds/common/utils/config_manager.py file accordingly. 
+Each field in the _set_defaults() function corresponds to a specific compile time flag for an FPrime deployment.
+Bellow is a table of corresponding fields that must match between deployment and Gds in order for successful communication to occur.
+
+Gds Field | FPrime Flag
+----------- | ----------- |
+self.__prop['types']['msg_len'] | TOKEN_TYPE
+self.__prop['types']['msg_desc'] | FwPacketDescriptorType
+self.__prop['types']['op_code'] | FwOpcodeType
+self.__prop['types']['ch_id'] | FwChanIdType
+self.__prop['types']['event_id'] | FwEventIdType
+self.__prop['types']['pkt_id'] | Unkown
+self.__prop['types']['key_val'] | unknown
+
+It is important to note that the default value of all of these fprime fields is U32, which is the behavior when none of these compile time flags are set in any CMakeLists.txt fields.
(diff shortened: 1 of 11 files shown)


SterlingPeet:
@AaronMcDaniel : it looks like the CI integration found some issues already, maybe see if you can fix those?

AaronMcDaniel:
I'm not particularly sure why the CI build is failing. There are no new dependencies and I've successfully build and run Ref with modified packet fields and could successfully connect with the Gds. I would appreciate input from anyone who may know what is going on in these CI builds

LeStarch:
From CI:

```
Failed to run 'check --all' in /github/workspace/
```

This means the unit tests are failing.  Run `fprime-util check --all` and ensure they all pass.

LeStarch:
Also we should resolve what the above LGTM error is.

LeStarch:
That LGTM error seems to be a false-positive.

AaronMcDaniel:
This PR is ready for review now. The only issue that was found appears to be a false positive, and this has been tested using Ref with modified lengths of different message fields.

Gds/docs/README.md
```diff
@@ -0,0 +1,26 @@
...
+This is a tool intended to be used with FPrime dpeloyments to exchange commands, telemetry, events and other messages.
+This tool includes logging capability and can be connected to several adapters to communicate with ground station equipment.
+
+## 2. Message Field Configuration
+
+Certain deployments may alter the fields of messsages for various reasons including memory conservation.
+To make the corresponding changes to the Gds for successful communication, 
+please edit the Gds/src/fprime_gds/common/utils/config_manager.py file accordingly. 
+Each field in the _set_defaults() function corresponds to a specific compile time flag for an FPrime deployment.
+Bellow is a table of corresponding fields that must match between deployment and Gds in order for successful communication to occur.
```
LeStarch, line 14: Spelling: Below

Gds/docs/README.md
```diff
@@ -0,0 +1,26 @@
...
+----------- | ----------- |
+self.__prop['types']['msg_len'] | TOKEN_TYPE
+self.__prop['types']['msg_desc'] | FwPacketDescriptorType
+self.__prop['types']['op_code'] | FwOpcodeType
+self.__prop['types']['ch_id'] | FwChanIdType
+self.__prop['types']['event_id'] | FwEventIdType
+self.__prop['types']['pkt_id'] | Unkown
+self.__prop['types']['key_val'] | unknown
+
+It is important to note that the default value of all of these fprime fields is U32, which is the behavior when none of these compile time flags are set in any CMakeLists.txt fields.
```
LeStarch, line 26: Ideally, you'd describe this as "setting values in FpConfig.hpp" as for now that would be how we'd set these.

Gds/src/fprime_gds/common/pipeline/encoding.py
```diff
@@ -50,15 +50,19 @@ def setup_coders(self, dictionaries, distributor, sender):
 
         :param dictionaries: a dictionaries handling object holding dictionaries
         :param distributor: distributor of data to register to
+        :param sender: TODO
```
LeStarch, line 53: `sender` is used to send the bytes created by the command encoder

Svc/GroundInterface/GroundInterface.cpp
```diff
@@ -125,8 +125,23 @@ namespace Svc {
...
-      U32 packet_type = Fw::ComPacket::FW_PACKET_UNKNOWN;
-      m_in_ring.peek(packet_type, HEADER_SIZE);
+      FwPacketDescriptorType packet_type = Fw::ComPacket::FW_PACKET_UNKNOWN;
+
+      //read packet descriptor in size agnostic way
+      U8 packet_descriptor_size = sizeof(FwPacketDescriptorType);
+      U8 packet_type_bytes[packet_descriptor_size];
+      Fw::SerializeStatus stat = m_in_ring.peek(packet_type_bytes, packet_descriptor_size, HEADER_SIZE);
+      //m_in_ring.peek(packet_type, HEADER_SIZE); // this way is only valid for 4byte packet descriptors
+      if(stat == Fw::FW_SERIALIZE_OK)
```
LeStarch, line 135: I'd recommend we refactor this to use our standard serialization code.  This isn't a blocker.

LeStarch (commented):
This works exactly as I'd expect.  Just a few documentation clean ups.

1. Note the C++ settings are configured in a deployment's `FpConfig.hpp`
2. Minor spelling.

Some "future work" items:
1. We should remove TOKEN_TYPE and make it something in FpConfig
2. Improve circular buffer to allow peaking of U16 and U64
3. Add these changes to Framer and Deframer.

LeStarch:
This has been on my list of things I need to get to for a long time.  Thanks!!!!

LeStarch (changes requested):
Changing to "Request Changes" for the few spelling issues.

Gds/docs/README.md
AaronMcDaniel: This spelling mistake has been fixed

Gds/docs/README.md
AaronMcDaniel: I changed the documentation to focus on FpConfig.hpp edits instead of CMakeLists flag defines for changing the fields

Gds/src/fprime_gds/common/pipeline/encoding.py
AaronMcDaniel: I added this to the doc comment

Svc/GroundInterface/GroundInterface.cpp
AaronMcDaniel: Based on the review comment this would involve adding peel methods for U16 and U64 types. Are there any other types that might be needed? Like signed types perhaps?

LeStarch (approved):

LeStarch:
Let's not fix the de-serialization immediately. We have a replacement for the GroundInterface component, and our effort would be better spent porting changes over there.  We can track it as an issue.

LeStarch:
Will merge when the build completes.  The build should complete as the changes are purely in comments/documentation

Issue #422, opened 2021-02-27
Drop Python 2 support


billallen256:
| | |
|:---|:---|
|**_F´ Version_**| current |
|**_Affected Component_**| Autocoder |
---
## Feature Description

Remove work-arounds in the code that are there to support Python 2.

## Rationale

There are some parts of the Autocoder that have work-arounds to support Python 2 and Python 3.  Since Python 2 reached end-of-life at the beginning of 2020, support should be dropped so these work-arounds can be removed.

LeStarch:
We have formally dropped Python 2 support, however; as you have seen we haven't removed all work around.

LeStarch:
We merged the PR here.  I am going to close this.  We can reopen if more are found.

PR #440, opened 2021-03-04
The six package is no longer a dependency.

billallen256:
| | |
|:---|:---|
|**_Originating Project/Creator_**| Bill Allen |
|**_Affected Architectures(s)_**| all |
|**_Related Issue(s)_**| #422 |

---
## Change Description

Removed the `six` package from the dependency listing in `setup.py`.

## Rationale

`six` is no longer referenced in Fw, plus Python 2 support has been dropped.

Should run as before.

## Future Work

No other references to six were found in the project.

### DIFF
--- a/Fw/Python/setup.py
+++ b/Fw/Python/setup.py
@@ -71,10 +71,9 @@ to interact with the data coming from the FSW.
         "Programming Language :: Python :: Implementation :: CPython",
         "Programming Language :: Python :: Implementation :: PyPy",
     ],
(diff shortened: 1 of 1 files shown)


LeStarch:
This looks good.  We'll need #442 in first as it remove a less-than-obvious use of "six".  As a side note, that whole package validation thing is slated for removal as with the creation of Python packages, we no longer need to check that the python packages exist before running CMake.  For now, I am removing "six" so it doesn't break anything.

LeStarch (approved):

Issue #343, opened 2021-01-28
The only command boolean argument value GDS interprets as false is False (not 0, false, etc)


nabcouwer:
| | |
|:---|:---|
|**devel as of 2021-01**| |
|**GDS**|   |
---
## Problem Description

https://groups.google.com/g/fprime-community/c/wdzHHbBXKfU

When I make an F' command with a bool arg, the only value that GDS seems to interpret as false is:

False

(no quotes, capitalized)

0, false, "false", 'false' all get interpreted as true. 

![Screenshot from 2021-01-27 12-55-20](https://user-images.githubusercontent.com/2354310/106071183-bd3ce400-60ba-11eb-959c-e7da6daec360.png)

## How to Reproduce

1. Create a command with an arg of type "bool", add it to a deployment.
2. Try sending the command to the deployment with many of the values one might reasonably assume GDS might interpret as false (0, false, "false", 'false')
3. Look at the GDS command history and see that the arguments got interpreted as true.

## Expected Behavior

GDS would ideally give a true/false dropdown for bool arguments, or interpret entries of 0, false, etc as false.

Thanks!

saba-ja:
The issue is the following:
https://github.com/nasa/fprime/blob/5d631af802664719a2de61bf22cf0a1cf3b2dc9f/Gds/src/fprime_gds/common/data_types/cmd_data.py#L161-L166

To fix, the following can be used in Gds/src/fprime_gds/common/data_types/cmd_data.py:
```
def convert_arg_value(self, arg_val, arg_type):
       ...
        if isinstance(arg_type, BoolType):
            if isinstance(arg_val, bool):
                av = bool(value)
            else:
                value = str(arg_val).lower().replace('"','').replace("'",'').strip()
                if value in ("true", "t", "1", "yes", "y"):
                    av = True
                elif value in ("false", "f", "0", "no", "n"):
                    av = False
                else:
                    raise CommandArgumentException(
                        "Argument value is not a valid boolean"
                    )
            arg_type.val = av
       ...
```
It will support ("true", "t", "1", "yes", "y") for True and ("false", "f", "0", "no", "n") for False. Single or double quotes, trailing spaces, letter case(Lower, Uppercase) will be ignored. Anything else will raise CommandArgumentException

Demo cases:
0 is false
![image](https://user-images.githubusercontent.com/35859004/108590380-198ccf80-7318-11eb-8153-6eb3873ba800.png)
f is false
[...]

PR #441, opened 2021-03-04
update: fix 343 GDS bool type issue

saba-ja:
| | |
|:---|:---|
|**_Originating Project/Creator_**| FPrime |
|**_Affected Component_**|  GDS |
|**_Affected Architectures(s)_**|  1.5 > |
|**_Related Issue(s)_**|  #343 |

---
## Change Description

The fix will make the GDS interpret `true`, `yes` as `True` and `false`, `no` as `False`. Values are not case sensitive. Trailing white spaces will be striped out. Any other value will raise `CommandArgumentException` exception.

## Rationale

In the previous implementation only `False` (case sensitive) was considered as boolean `False` and any other value as boolean `True`. No exception was raised on non-boolean values. This could cause confusion and problems. The new implementation has a limited set of valid boolean values (`true`, `yes`, `false`, `no`). Any other value will be rejected.

Manually verified the change as shown below:

image

## Future Work
Need to add documentation to indicate valid boolean values in FPrime GDS.

### DIFF
--- a/Gds/src/fprime_gds/common/data_types/cmd_data.py
+++ b/Gds/src/fprime_gds/common/data_types/cmd_data.py
@@ -159,10 +159,15 @@ class CmdData(sys_data.SysData):
                 "Argument value could not be converted to type object"
             )
         if isinstance(arg_type, BoolType):
-            if arg_val == "False":
+            value = str(arg_val).lower().strip()
+            if value in ("true", "yes"):
+                av = True
+            elif value in ("false", "no"):
                 av = False
             else:
-                av = True
+                raise CommandArgumentException(
+                    "Argument value is not a valid boolean"
+                )
             arg_type.val = av
         elif isinstance(arg_type, EnumType):
             arg_type.val = arg_val


Joshua-Anderson (approved):

Joshua-Anderson:
Bonus points (and possibly out of scope of this issue), but have you thought about adding frontend validation to the GDS website make sure that users can't submit invalid boolean values? Or maybe make it a dropdown with true/false options for booleans?

saba-ja:
@Joshua-Anderson Good point. A dropdown would be even better option. Just did add it and it looks like this:
image

To see how it looks like I did a small change here:
https://github.com/nasa/fprime/blob/0e731d14c161a5c5dc37a2ce2e55009353806247/Gds/src/fprime_gds/common/utils/jsonable.py#L48-L49

and added the following:
```javascript
  if arg_dict["type"] == "Enum":
      arg_dict["possible"] = arg_spec[2].keys()
  if arg_dict["type"] == "Bool":
      arg_dict["possible"] = ["True", "False"]
      arg_dict["type"] = "Enum"
```
Will look deeper into this and see how it can be achieved in a clear way.

LeStarch:
Let's make that a separate PR.

Issue #241, opened 2020-11-09
Duplicate Enums in Dictionary


Joshua-Anderson:
There doesn't seems to be any deduplication logic when adding enums to the deployment's dictionary.

The Ref deployment, for example, has the `Ref::SignalType` enum in the dictionary a half dozen times.
We should investigate what is causing this.

```
  <enums>
    <enum type="CommandDispatcher::error::ErrorResponse">
      <item name="ERR_INVALID_OPCODE" value="0" description="Invalid opcode dispatched"/>
      <item name="ERR_VALIDATION_ERROR" value="1" description="Command failed validation"/>
      <item name="ERR_FORMAT_ERROR" value="2" description="Command failed to deserialize"/>
      <item name="ERR_EXECUTION_ERROR" value="3" description="Command had execution error"/>
      <item name="ERR_BUSY" value="4" description="Component busy"/>
      <item name="ERR_UNEXP" value="5" description="Unexpected response"/>
    </enum>
    <enum type="CommandDispatcher::Status::CmdSerError">
      <item name="ERR_BUFFER_TOO_SMALL" value="0" description="Buffer too small"/>
      <item name="ERR_BUFFER_FORMAT" value="1" description="Buffer wrong format"/>
      <item name="ERR_SIZE_MISMATCH" value="2" description="Buffer size mismatch"/>
      <item name="ERR_TYPE_MISMATCH" value="3" description="Buffer type mismatch"/>
      <item name="ERR_UNEXP_STAT" value="4" description="Unexpected status"/>
    </enum>
    <enum type="PrmDb::stage::PrmWriteError">
      <item name="PRM_WRITE_OPEN" value="0"/>
      <item name="PRM_WRITE_DELIMETER" value="1"/>
      <item name="PRM_WRITE_DELIMETER_SIZE" value="2"/>
      <item name="PRM_WRITE_RECORD_SIZE" value="3"/>
      <item name="PRM_WRITE_RECORD_SIZE_SIZE" value="4"/>
      <item name="PRM_WRITE_PARAMETER_ID" value="5"/>
      <item name="PRM_WRITE_PARAMETER_ID_SIZE" value="6"/>
      <item name="PRM_WRITE_PARAMETER_VALUE" value="7"/>
      <item name="PRM_WRITE_PARAMETER_VALUE_SIZE" value="8"/>
    </enum>
...
    <enum type="Ref::SignalType">
      <item name="TRIANGLE" value="0" description=""/>
[...]

PR #462, opened 2021-03-12
Refactor topology dictionary generation

Kemyt:
| | |
|:---|:---|
|**_Affected Component_**| Autocoders |
|**_Related Issue(s)_**|  |

---
## Change Description

It gets rid of a lot of code duplication and unifies the way
codegen.py and gds_dictgen generate topology dictionaries.

## Rationale

Similarly to #241, there are duplicate serializable entries in dictionaries, and duplicate removal is done in [gds_dictgen.py](https://github.com/nasa/fprime/blob/12c478bd79c2c4ba2d9f9e634e47f8b6557c54a8/Autocoders/Python/bin/gds_dictgen.py#L215), but not in codegen.py. A lot of code was duplicated and implementations diverged over time. With this change, topology dictionary generation will be unified. In process of extracting common code some potential issues were found. For example, "arg" object [here](https://github.com/nasa/fprime/blob/12c478bd79c2c4ba2d9f9e634e47f8b6557c54a8/Autocoders/Python/bin/codegen.py#L777) and [here](https://github.com/nasa/fprime/blob/12c478bd79c2c4ba2d9f9e634e47f8b6557c54a8/Autocoders/Python/bin/codegen.py#L803) is actually "leaked" from [for loop of another functionality](https://github.com/nasa/fprime/blob/12c478bd79c2c4ba2d9f9e634e47f8b6557c54a8/Autocoders/Python/bin/codegen.py#L704)

Check if changes to generated dictionaries are correct.

## Future Work

None


Joshua-Anderson:
Thanks so much for tackling this! These changes look good to me, but the autocoder is brittle code (it feels a bit like wack a mole, where every autocoder bug I try and fix introduces a different bug), so I want to make sure to take the time carefully review this and make sure we're not introducing any new issues.
 
Thanks again for tackling this and I'll make sure this gets a proper review before our next F' release!

LeStarch:
Ok, I have done some limited testing on this.  I can confirm the following:

1. Dictionary output is consistent with `nasa/devel` with the exception that the duplicate serializable entries have been removed.
2. Gds runs happily with this change.
3. Binary output is unaffected.

I will review the code, but I am satisfied with the testing here!

LeStarch (approved):

LeStarch:
@Kemyt thanks for giving us the time to review this.  Looks good to me.  It passes CI, a diff check, a GDS check, and I reviewed the code as well as I could.

Issue #349, opened 2021-02-04
GDS decoding ENUM events incorrectly


saba-ja:
| | |
|:---|:---|
|**_F´ Version_**| 1.5> |
|**_Affected Component_**|All   |
---
## Problem Description

Originally reported by @bcmetz 

GDS crashes when string_format has `%d`

The enum definition xml
```
<?xml version="1.0" encoding="UTF-8"?>
<?oxygen RNGSchema="file:../../Autocoders/Python/schema/ISF_Type_Schema.rnc" type="compact"?>
<enum namespace = "Backplane" name="PowerState">
    <comment>Enum of different supported signal types.</comment>
    <item name="OFF" value="0"/>
    <item name="ON" value="1"/>
</enum>
```

arg xml for the event:
```
<event id="0" name="POWERED" severity="ACTIVITY_HI" format_string = "Instrument has been powered %d" >
            <comment>
            Power state of the instrument
            </comment>
            <args>
                <arg type="Backplane::PowerState" name="State"> </arg>
	   </args>
</event>
```
When the bp_C4D.POWER, ON command is sent, the telemetry is pushed correctly:
![image](https://user-images.githubusercontent.com/35859004/106947037-be839780-66de-11eb-91a7-9d1e631d6a13.png)

But GDS python code crashes:
```
File "fprime_gds/flask/events.py", line 66, in get
    event.template.format_str % tuple([arg.val for arg in event.args]),
TypeError: %d format: a number is required, not str
```

## Expected Behavior

The Event should be printed correctly in the event table in the GDS GUI

saba-ja:
The above issue is due to using Python string interpolation (string_format % tuple of args) which enforces type validation. When an unexpected type is given a TypeError exception is thrown.
Gds/src/fprime_gds/flask/events.py#L66

event.template.format_str % tuple([arg.val for arg in event.args])
In the case of above issue the format_string template has integer place holder Instrument has been powered %d while the argument associated with the template is string ON:

![image](https://user-images.githubusercontent.com/35859004/106947368-23d78880-66df-11eb-8174-f074fe9664f8.png)


The string interpolation is a common code vulnerability pattern in current FPrime GDS.

saba-ja:
Similar to https://github.com/nasa/fprime/issues/343

LeStarch:
@saba-ja did we fix this?

PR #467, opened 2021-03-18
fix 349: GDS decoding ENUM events incorrectly

saba-ja:
| | |
|:---|:---|
|**_Originating Project/Creator_**| FPrime|
|**_Affected Component_**| events |
|**_Affected Architectures(s)_**| GDS |
|**_Related Issue(s)_**|  #349 |
|**_Builds Without Errors (y/n)_**|  Y |
|**_Unit Tests Pass (y/n)_**| Y  |

---
## Change Description

Old style Python %-formatting string interpolation is removed from the `events`.
GDS Python code could throw an exception when the string format defined in the componentEventAi.xml did not have the expected string format value.
For example, the following event format_string could potentially break the GDS Python code since %-formatting expects an integer, but Enum in the form of string is passed to GDS:

image

## Rationale

Python code should not break if user defines Enum as %d in the event string format.
With this fix using %d as the placeholder for Enum will not break the Python code. The correct string value of Enum will be shown in the event page and event log in GDS.

Note 1: The fix still expects correct Hex format (e.g. %04X) and will raise a ValueError if a string is proved in place of expected Hex value.

Note 2: This function will keep the flags, width, .precision and length of C-string template. It will remove all types so they could be duck-typed by the Python interpreter except for hex type X or x.

Mutated fprime/Ref/SignalGen/Events.xml to have:
```C++
format_string="Set Frequency(Hz) %d, Amplitude %f, Phase %f, Signal Type %d"
```

Tested the output as shown below with expected values:
image

image

The unit test can be found here: fprime/Gds/test/fprime_gds/common/utils/test_string_util.py 

## Future Work

This update will not fix the std out to show the Enum string if %d is used as string format.

If we have the following:
```C++
format_string="Set Frequency(Hz) %d, Amplitude %f, Phase %f, Signal Type %d"
```
The output will still show this:
image

Instead of this:
image

### DIFF
--- a/Gds/src/fprime_gds/common/utils/string_util.py
+++ b/Gds/src/fprime_gds/common/utils/string_util.py
@@ -1,0 +1,44 @@
+"""
+string_util.py
+
+Utility functions to process strings to be used in FPrime GDS
+
+@Created March 18, 2021
+@janamian
+"""
+
+import re
+
+def format_string(format_str, values):
+    """
+    Function to convert C-string style to python format
+    without using python interpolation
+    Considered the following format for C-string:
+    %[flags][width][.precision][length]type
+
+    %:           (?<!%)(?:%%)*%
+    flags:       ([\-\+0\ \#])?
+    width:       (\d+|\*)?
+    .precision:  (\.\*|\.\d+)?
+    length:      ([hLIw]|l{1,2}|I32|I64)?
+    type:        ([cCdiouxXeEfgGaAnpsSZ])
+
+    Note:
+    This function will keep the flags, widht, .percision and length of C-string
+    template. It will remove all types so they could be ducktyped by python 
+    interpreter except for hex type X or x.
+    """
+    def convert(match_obj):
+        if match_obj.group() is not None:
+            template = "{:" + match_obj.group()[1:-1]
+            if match_obj.group()[-1].lower() == 'x':
+                template += match_obj.group()[-1]
+            template += "}"
+            return template
+        return match_obj
+
+    pattern = "(?<!%)(?:%%)*%([\-\+0\ \#])?(\d+|\*)?(\.\*|\.\d+)?([hLIw]|l{1,2}|I32|I64)?([cCdiouxXeEfgGaAnpsSZ])"
+    match = re.compile(pattern) 
+    formated_str = re.sub(match, convert, format_str)
+    result = formated_str.format(*values)
+    return result
--- a/Gds/src/fprime_gds/common/utils/string_util.py
+++ b/Gds/src/fprime_gds/common/utils/string_util.py
@@ -24,8 +24,8 @@ def format_string(format_str, values):
     type:        ([cCdiouxXeEfgGaAnpsSZ])
 
     Note:
-    This function will keep the flags, widht, .percision and length of C-string
-    template. It will remove all types so they could be ducktyped by python 
+    This function will keep the flags, width, .precision and length of C-string
+    template. It will remove all types so they could be duck-typed by python 
     interpreter except for hex type X or x.
     """
     def convert(match_obj):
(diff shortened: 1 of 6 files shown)


Gds/src/fprime_gds/common/utils/string_util.py
```diff
@@ -0,0 +1,44 @@
...
+    without using python interpolation
+    Considered the following format for C-string:
+    %[flags][width][.precision][length]type
+
+    %:           (?<!%)(?:%%)*%
+    flags:       ([\-\+0\ \#])?
+    width:       (\d+|\*)?
+    .precision:  (\.\*|\.\d+)?
+    length:      ([hLIw]|l{1,2}|I32|I64)?
+    type:        ([cCdiouxXeEfgGaAnpsSZ])
```
jsoref, line 24: Personally, instead of adding the items you've added to `expect`, I'd consider adding one of the following to https://github.com/nasa/fprime/blob/devel/.github/actions/spelling/patterns.txt

```
# See Gds/src/fprime_gds/common/utils/string_util.py
\[cCdiouxXeEfgGaAnpsSZ\]
```
This would cause just this strange single instance (or any identical instance) to be ignored.

```
# See Gds/src/fprime_gds/common/utils/string_util.py
^\s*type:\s*\(\[[^]*\]\)$
```
This should cause a line like this to be ignored without explicitly managing the precise contents of the file.

saba-ja: @jsoref, thank you for the comment.
I agree we should not keep adding sequences to the expect.txt for spell checking.
I also think adding specific patterns is not a better strategy. For example in the case above `LIw` is one of the sequences that is getting confuse by the spell checker which is on line 23.
In my opinion we need to define opening and closing pattern where the spell checker should not need to check for spelling. I would suggest using back-tick to hint the spell checker to not check for spelling. 
We can add the following pattern to skip any text between two back-ticks:
https://github.com/nasa/fprime/blob/devel/.github/actions/spelling/patterns.txt
```
`.*?`
```
We also need to add a guideline page for contributors to know how to use these techniques.
We should also reduce noises created by the spell checker since spelling, although important for readability, is not the main purpose of most pull requests.

saba-ja: @jsoref, I added two patterns to escape the inline back-ticks and in-code sequences with special symbols. I was not able to add multiline escape pattern as for some reason spell checker ignores `\m` (multiline flag). The prefer method would be to be able to ignore anything between three back-tick block in the comments similar to the following:

/*
Text to be spell checked

\`\`\` <-------- block start
Text to not be spell checked
\`\`\` <-------- block end

Text to be spell checked
*/

FYI: This is the regex I tried to use for multiline block:
```regex
^[ \t]*`{3}[^\n]*[\s\S]+?^[ \t]*`{3}/m
```

jsoref: Unfortunately I don't have a multi line notation. You could define something like `no-spell-check`

And then have this in `patterns.txt`:
```
^.*no-spell-check.*$
```

You can add a file `advice.txt` next to `patterns.txt` suggesting people add `// no-spell-check` to lines to skip them.

jsoref: Yeah, thinking about multiline patterns is on my list. Currently I read files line at a time. Projects by default check all files, likely including large binary files. I've been a bit worried about forcing it to handle large files as a whole.

I'm currently working on teaching the tool to recognize binary files, but amusingly the last thing I looked at was an `.exe` and the tool rightly said _there are **real misspellings here**, and there are **fewer misspelled items** than **correctly spelled ones**_. (Arguably still the wrong behavior since it'd be better to catch those misspellings at their source... But amusing.) I think the heuristic I wrote should at least guard against non English content even if it fails on "binary" content. I'm thinking about trying a file size vs content heuristic which should hopefully catch the `.exe` case.

Once I've added a heuristic that lets the tool gracefully adjust for binary files, I think I'll be more comfortable trying to handle multiline matching. Although I'm still leaning towards using a distinct file for it.

LeStarch (approved):
This code is a bit cryptic, but I can see the test results.

PR #472, opened 2021-03-30
Fix typo in svc_prm_db_component.html

eltociear:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

occuring -> occurring

### DIFF
--- a/docs/UsersGuide/api/c++/html/svc_prm_db_component.html
+++ b/docs/UsersGuide/api/c++/html/svc_prm_db_component.html
@@ -128,7 +128,7 @@ Svc::PrmDb Component</h1>
 </table>
 <h4><a class="anchor" id="autotoc_md420"></a>
 3.2 Functional Description</h4>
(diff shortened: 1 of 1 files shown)


jsoref:
I was wondering why I hadn't fixed/caught this when I landed, the answer is that the bot is currently ignoring all `/html/` directories: https://github.com/nasa/fprime/blob/9ba5e0e8600823f3d8b9078b89bd3aff43e0beea/.github/actions/spelling/excludes.txt#L2

I think I guessed that they were generated/imported from something/somewhere else. If this is the only home for the content, then it should be checked.

If you're feeling adventurous, you could enable GitHub actions in your fork and try removing that exclusion and fixing everything else it complains about (someone should...).

LeStarch:
Yes, these HTML directories are auto-generated.  We update them when merging to master with the next release candidate.

LeStarch (approved):

Issue #497, opened 2021-04-23
identical condition


maddin200:
[/Drv/SocketIpDriver/SocketHelper.cpp line 231

Code:
  // Error is EINTR, just try again
            if (sent == -1 && errno == EINTR) { // <-- identical
                continue;
            }
            // Error bad file descriptor is a close
            else if (sent == -1 && errno == EBADF) {
                Fw::Logger::logMsg("[ERROR] Server disconnected\n");
                this->close();
                this->m_socketOutFd = -1;
                break;
            }
            // Error returned, and it wasn't an interrupt
            else if (sent == -1 && errno == EINTR)  {   // <-- identical
                Fw::Logger::logMsg("[ERROR] IP send failed ERRNO: %d UDP: %d\n", errno, m_sendUdp);
                break;
            }

Joshua-Anderson:
Fixed in #497. Thanks!

PR #501, opened 2021-04-26
Fix copy/paste error resulting in identical condition

bstaletic:
| | |
|:---|:---|
|**_Originating Project/Creator_**| @LeStarch (I think) |
|**_Affected Component_**| `/Drv/SocketIpDriver/ |
|**_Related Issue(s)_**| #497 |

---
## Change Description

As the comment suggests, the intent was to catch all /other/ errors.

## Rationale

This PR should fix a bug where a repeated condition caused most errors on `send()` and `sendto()` to be ignored and the contition itself could never be true on line 231.

Hopefully I haven't missed anything and this bugfix is as easy as it looks.

## Future Work

None.

### DIFF
--- a/Drv/SocketIpDriver/SocketHelper.cpp
+++ b/Drv/SocketIpDriver/SocketHelper.cpp
@@ -228,7 +228,7 @@ namespace Drv {
                 break;
             }
             // Error returned, and it wasn't an interrupt
-            else if (sent == -1 && errno == EINTR) {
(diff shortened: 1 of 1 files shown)


Joshua-Anderson (approved):
Good catch! I also verified that this bug doesn't affect the IP driver redesign in #460

LeStarch (approved):
Worst case, it reports a potentially shadowed bug!

PR #504, opened 2021-04-28
Fix #502: Add option to allow reregister of opCodes

### LINKED ISSUE #502: Adding deregCommands option to CmdDispatcher
| | |
|:---|:---|
|**_F´ Version_**| 2.0 |
|**_Affected Component_**|  Svc/CmdDispatcher |
---
## Feature Description

F Prime CommandDispacher currently does not have an option to deregister commands. This causes a duplicate command register segment fault crash when a client disconnects and reconnects back to the server.
To better understand the situation, let’s assume we have the following configuration:

image

`Ref` loads before `Backplane` and listens on port `4555`. When `Backplane` loads, it connects to `Ref` as a client, and all `Backplane` CMDs get registered to `Ref` `CmdDispacher` table.
Now the issue is when `Backplane` drops, its CMDs still stay registered in the `Ref` `CmdDispacher` table. When `Backplane` tries to reconnect back to `Ref` since all its CMDs are still registered in `Ref`, it causes FW_ASSERT due to duplicate opcode, causing `Ref` to crash with segment fault error.

## Rationale

It might be a good idea to add `deregCommands` (an opposite of `regCommands`) that could be used to deregister all the CMDs when a client gracefully goes down. This will allow the client to reconnect back to the server without causing duplicate opcode registration issue.

These are two ways that come to my mind to resolve this issue:

1. Check for OpCode before registering it. This could be done by the following to `compCmdReg_handler` shown below:

https://github.com/nasa/fprime/blob/e165c914b0d0240eba8164c52f8856ce90bee3dc/Svc/CmdDispatcher/CommandDispatcherImpl.cpp#L34-L49

Add this to line 35 above
```c++
// search for opcode in dispatch table
        for (U32 entry = 0; entry < FW_NUM_ARRAY_ELEMENTS(this->m_entryTable); entry++) {
            if ((this->m_entryTable[entry].used) and (opCode == this->m_entryTable[entry].opcode)) {
                return;
            }
        }

```

2. Add `deregCommands()` to Svc/CmdDispatcher. This could be achieved by the following steps:

a) 
In 
https://github.com/nasa/fprime/blob/e165c914b0d0240eba8164c52f8856ce90bee3dc/Svc/CmdDispatcher/CommandDispatcherComponentAi.xml#L17-L21

Add `compCmdDereg` port and `OpCodeDeregistered` event:

```xml
        <port name="compCmdDereg" data_type="Fw::CmdReg" kind="guarded_input" max_number = "$CmdDispatcherComponentCommandPorts">
            <comment>
            Command Deregistration Port. max_number should match dispatch port.
            </comment>
        </port>

...

        <event id="2" name="OpCodeDeregistered" severity="DIAGNOSTIC" format_string = "Opcode 0x%04X deregistered to port %d slot %d" >
            <comment>
            Op code deregistered event
            </comment>
            <args>
                <arg name="Opcode" type="U32">
                    <comment>The opcode to deregister</comment>
                </arg>          
                <arg name="port" type="I32">
                    <comment>The deregister port</comment>
                </arg>          
                <arg name="slot" type="I32">
                    <comment>The dispatch slot it was freed</comment>
                </arg>          
            </args>
        </event>
```

b)
Update [CommandDispatcherImpl.cpp](https://github.com/nasa/fprime/blob/devel/Svc/CmdDispatcher/CommandDispatcherImpl.cpp) to handle the requests.

c)
Update [Component Autocoder](https://github.com/nasa/fprime/blob/devel/Autocoders/Python/src/fprime_ac/generators/templates/component/cpp.tmpl) template to autogenerate `deregCommnds` for each component.

saba-ja:
| | |
|:---|:---|
|**_Affected Component_**|  Svc |
|**_Affected Architectures(s)_**| CmdDispatcher |
|**_Related Issue(s)_**|  #502 |

---
## Change Description

This PR will add an option to CmdDispatcher to reregister opCodes.

## Rationale
There are cases in which a client need to reregister its commands with a server. The current implementation will fail on duplicate opCodes. This PR adds `ALLOW_REREGISTRATION` option to config/CommandDispatcherImplCfg.hpp. If true will CmdDispatcher will reregister opcodes otherwise will ASSERT for duplicate opcodes.

Please see #502 for the for more discussion.

Manually ran multiple client drop and reconnect without any issue.

## Future Work

None


Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -32,6 +32,19 @@ namespace Svc {
     }
 
     void CommandDispatcherImpl::compCmdReg_handler(NATIVE_INT_TYPE portNum, FwOpcodeType opCode) {
+
```
LeStarch, line 35: Couldn't this be done in one step by adding and elseif on line 44:

```c++
else if (this->m_entryTable[slot].used and (opCode == this->m_entryTable[slot].opcode) and ALLOW_REREGISTRATION) {
                this->m_entryTable[slot].opcode = opCode;
                this->m_entryTable[slot].port = portNum;
                this->m_entryTable[slot].used = true;
                this->log_DIAGNOSTIC_OpCodeRegistered(opCode,portNum,slot);
                slotFound = true;
}
```

I believe your method should work as well, it just splits the process into 2 halves.

config/CommandDispatcherImplCfg.hpp
```diff
@@ -13,6 +13,7 @@
 enum {
     CMD_DISPATCHER_DISPATCH_TABLE_SIZE = 100, // !< The size of the table holding opcodes to dispatch
     CMD_DISPATCHER_SEQUENCER_TABLE_SIZE = 25, // !< The size of the table holding commands in progress
+    ALLOW_REREGISTRATION = true // !< If true will reregister opcodes otherwise will ASSERT for duplicate opcode
```
LeStarch, line 16: I would default this to `false` to preserve standard behavior.

LeStarch (commented):
Couldn't we add j

LeStarch:
@saba-ja I like the general approach to "allow reregistration".  This cleanly solves the problem where a device went offline and is back.  Unless a customer has more detailed requirements to deregister at the point of failure, then I think this should work.

I would like @timcanham's thoughts on this too.

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
LeStarch: You might also consider making the above emit a "OpCodeReregistered" EVR just so it is clear that the opcode is being reused.

LeStarch:
@saba-ja and @timcanham  this has one notable downside to the deregister approach, which is that it won't catch a re-registration of an opcode between components or deployments at system startup.

I guess this is the fundamental trade: not (necessarily) knowing when to deregister will add a lot of system complexity.  Allowing reregistration reduces the error-checking capabilities.

I am still in favor of allowing reregistration because it can be set to `false` early on in development when things are clean and then switched to `true` when ready to test off-nominal handling after the system topology has settled to the point where mistakes in opcodes are less likely.

saba-ja:
@LeStarch, I will change the default value to false as mentioned.

If desired we can have an EVR whenever a reregistration is happening.

```c++
    void CommandDispatcherImpl::compCmdReg_handler(NATIVE_INT_TYPE portNum, FwOpcodeType opCode) {

        if (ALLOW_REREGISTRATION) {
            // search for opcode in dispatch table
            for (U32 slot = 0; slot < FW_NUM_ARRAY_ELEMENTS(this->m_entryTable); slot++) {
                if ((this->m_entryTable[slot].used) and (opCode == this->m_entryTable[slot].opcode)) {
                    // free the slot that was used for this duplicated opCode
                    this->m_entryTable[slot].used = false;
                    this->log_DIAGNOSTIC_OpCodeDegistered(opCode,portNum,slot); <--------------- Deregister EVENT
                    // assuming only one duplicate in the table break out of the loop
                    break;
                }
            }
        }

        // search for an empty slot
        bool slotFound = false;
        for (U32 slot = 0; slot < FW_NUM_ARRAY_ELEMENTS(this->m_entryTable); slot++) {
            if ((not this->m_entryTable[slot].used) and (not slotFound)) {
                this->m_entryTable[slot].opcode = opCode;
                this->m_entryTable[slot].port = portNum;
                this->m_entryTable[slot].used = true;
                this->log_DIAGNOSTIC_OpCodeRegistered(opCode,portNum,slot); <--------------- Register EVENT
                slotFound = true;
            } else if (this->m_entryTable[slot].used) { // make sure no duplicates
                FW_ASSERT(this->m_entryTable[slot].opcode != opCode, opCode);
            }
        }
        FW_ASSERT(slotFound,opCode);
    }
```

LeStarch:
@saba-ja can the event message capture the context that we are deregistering in preparation for registering the opcode again?   If so, I think a diagnostic EVR would be wise.

saba-ja:
@LeStarch, it is possible to report opcode, port and slot that was released. 

This is a modified example from EVENT `OpCodeRegistered` with only changing the word register to deregister:
        
```xml
<event id="0" name="OpCodeDeregistered" severity="DIAGNOSTIC" format_string = "Opcode 0x%04X deregistered from port %d slot %d" >
            <comment>
            Op code deregistered event
            </comment>
            <args>
                <arg name="Opcode" type="U32">
                    <comment>The opcode to deregister</comment>
                </arg>          
                <arg name="port" type="I32">
                    <comment>The deregistered port</comment>
                </arg>          
                <arg name="slot" type="I32">
                    <comment>The freed dispatch slot</comment>
                </arg>          
            </args>
        </event>
```

Joshua-Anderson:
I'm still a little confused about why we can't make command registration idempotent, but I'm probably missing something 😄 

Right now on line 48 we check if the port is used:

```
} else if (this->m_entryTable[slot].used) { // make sure no duplicates
     FW_ASSERT(this->m_entryTable[slot].opcode != opCode, opCode);
}
```

Couldn't we just change that check to check that if the opcode is already in use, make sure that the port you're registering to the opcode is the same as the port that is currently registered? If the opcode is already in use but you're mapping it to the same port, continue and no-op. If the opcode is already in use but mapped to a different port, FATAL out.

I prefer the approach of no-op allowing identical re-registrations of the same opcode rather than allowing the dynamic rebinding of which opcode is paired to which port, since allowing users to dynamically alter command routing seems like it could be a bit of a dangerous footgun. It would remove a valuable runtime check making sure you don't have two commands in different components mapped to the same opcode. In theory we should catch overlapping command opcodes at topology creation time but I do like having strong runtime protections as well.

saba-ja:
@Joshua-Anderson, I'll try to explain the situation so hopefully it would shed some light on what is the issue.

In Backplane/Top/Topology.cpp each component has to register its commands during the initiation using `regCommands()` as shown below:

Backplane/Top/Topology.cpp
```c++
bp_prmDb.regCommands();
bp_fileDownlink.regCommands();
bp_fileManager.regCommands();
```

The `regCommands()` is an autogenerated function based on component/cpp.tmpl as shown below:
https://github.com/nasa/fprime/blob/e14095fb018ffe5657d41c705adb9026840fb143/Autocoders/Python/src/fprime_ac/generators/templates/component/cpp.tmpl#L211-L223

This is an example of autogenerated regCommands() for a component:

```c++
  // Command registration

  void FileDownlinkComponentBase ::
    regCommands(void)
  {
    FW_ASSERT(this->m_cmdRegOut_OutputPort[0].isConnected());
    this->m_cmdRegOut_OutputPort[0].invoke(
        this->getIdBase() + OPCODE_SENDFILE
    );
    this->m_cmdRegOut_OutputPort[0].invoke(
        this->getIdBase() + OPCODE_CANCEL
    );
    this->m_cmdRegOut_OutputPort[0].invoke(
        this->getIdBase() + OPCODE_SENDPARTIAL
    );
  }
```
Up to here the code does not have any knowledge about if the opcode is already registered or not.

The check for duplicate opcode happens in `CommandDispatcherImpl::compCmdReg_handler()`:
https://github.com/nasa/fprime/blob/e14095fb018ffe5657d41c705adb9026840fb143/Svc/CmdDispatcher/CommandDispatcherImpl.cpp#L34-L49

So Unless I am missing any steps I do not see any other place to check for duplicate opcodes to make the regCommands() idempotent to duplicate opcodes. Please advice if there are any other options.

----

### Why not do this in one step in the `else if` on line 44?

The issue is the `else if` gets called each time a slot is used. So we might get into the `else if` block many times during the iteration over the dispatched table before it finds an unused slot for the requested opcode. The logic in the for loop is looking for empty slot and not the opcode. So if it finds an unused slot it will assign the opcode into it and move on without getting into the `else if` block. So assuming the first slot is empty the the opcode will be assigned to that slot while the same opcode already exist in the table. This will not get discovered until later when we keep iterating the table.

The search for opCode before reregistering it aims to look for the specific `opcode` in the dispatch table first as shown below and mark the slot that was occupied by that opcode as unused so it could be reused.
```c++
if ((this->m_entryTable[slot].used) and (opCode == this->m_entryTable[slot].opcode)) {
                    // free the slot that was used for this duplicated opCode
                    this->m_entryTable[slot].used = false;
```

This code is actually from the following with some slight changes:
https://github.com/nasa/fprime/blob/e14095fb018ffe5657d41c705adb9026840fb143/Svc/CmdDispatcher/CommandDispatcherImpl.cpp#L143-L152

----

### Checking for port registered to given opcode:

The check for the port registration is done in the regCommands() function. As shown below it checks if the output port is actually connected and if not it will ASSERT.
```c++
  void FileDownlinkComponentBase ::
    regCommands(void)
  {
    FW_ASSERT(this->m_cmdRegOut_OutputPort[0].isConnected());
```

If desired we can also double check using 
```c++
if (entryFound and this->isConnected_compCmdSend_OutputPort(this->m_entryTable[entry].port)) {
```
but I think it would be redundant.

Joshua-Anderson:
Ah, I think we misunderstood eachother here. By making command registration idempotent I just meant that a component could call the command registration port multiple times with the same opcode multiple times safely, hence my suggestion of the switching the opcode already in use check to only error if the opcode is already in use and the port is difference.

You're correct in that the table search is looking for an empty slot in the opcode table and if there were gaps in the entry table we wouldn't be able to make the modification I suggested. However, looking at the architecture of command registration, we can see that the entries are allocated starting at the start of the array and incrementing from there. Since there's no way to deallocate entries in the table that I can see, we know that when you try and register a new opcode it has to iterate over all the existing opcodes and check for collisions.


I see three potential solutions to solving the problem of opcode collision when re-connecting a component topology:

1.  Add support for de-registering opcodes: This works, but it's hard to handle edge cases where components don't gracefully exit and de-register their opcodes
2. Allow re-registering opcodes (this PR): This works too, but I'm worried about removing a runtime check verifying that you don't have two components trying to register the same command opcode.  Currently, if two components have colliding command opcodes you get a fatal on startup. With this change two components registering the opcode would silently succeed and the last component to register wins.
3. Idempotent approach (my suggestion): Allow re-registration, but only if the opcode->port binding is identical to what we've done before. It's similar to the approach in this PR, but it prevents attempting to register the same opcode in multiple components. It could probably be implemented by tweaking the current code that FATALS on duplicate opcodes

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
saba-ja: Added `OpCodeReregistered` EVR on CMD reregistration.

config/CommandDispatcherImplCfg.hpp
saba-ja: Set default `ALLOW_REREGISTRATION = false`

saba-ja:
Change summary:
1. Updated Cfg to default ALLOW_REREGISTRATION to false.
2. Added OpCodeReregistered EVR.
3. Changed the reregistration logic to check for both opcode and its assigned port and skip updating the dispatcher table if opcode and its assigned port is already registered in the dispatcher table.
If the config option `ALLOW_REREGISTRATION` is false it will ASSERT on duplicate opcode (original behavior) 
```c++
else if ((this->m_entryTable[slot].used) &&
                (this->m_entryTable[slot].opcode == opCode) &&
                (this->m_entryTable[slot].port == portNum) &&
                (not slotFound) &&
                ALLOW_REREGISTRATION) {
                    slotFound = true;
                    this->log_DIAGNOSTIC_OpCodeReregistered(opCode,portNum,slot);
```

config/CommandDispatcherImplCfg.hpp
```diff
@@ -13,6 +13,7 @@
 enum {
     CMD_DISPATCHER_DISPATCH_TABLE_SIZE = 100, // !< The size of the table holding opcodes to dispatch
     CMD_DISPATCHER_SEQUENCER_TABLE_SIZE = 25, // !< The size of the table holding commands in progress
+    ALLOW_REREGISTRATION = false // !< If true will reregister opcodes otherwise will ASSERT for duplicate opcode
```
Joshua-Anderson, line 16: Do we need to make re-registering the opcode/port combination configurable? I don't think there's any situations where this behavior could be dangerous or undesirable, so we might be able to get away with just making this the new behavior and avoid adding another configuration option.

Svc/CmdDispatcher/CommandDispatcherComponentAi.xml
```diff
@@ -217,6 +217,22 @@
...
+            </comment>
+            <args>
+                <arg name="Opcode" type="U32">
+                    <comment>The opcode reregistered</comment>
+                </arg>          
+                <arg name="port" type="I32">
+                    <comment>The reregistration port</comment>
+                </arg>          
+                <arg name="slot" type="I32">
+                    <comment>The dispatch slot already allocated to the opcode</comment>
```
Joshua-Anderson, line 232: The slot number number feels like an implementation detail to me, is this something we need to expose to users? Unless the entire opcode table mapping is exposed somewhere it doesn't seem like this value has any meaning.

I'd be tempted just to remove the dispatch slot from the EVR.

Joshua-Anderson:
Nice work @saba-ja, I just had a few minor comments. Would it be possible to add some units tests to verify this behavior as well?

Svc/CmdDispatcher/CommandDispatcherComponentAi.xml
saba-ja: I agree. Will remove the slot number from the EVR.

config/CommandDispatcherImplCfg.hpp
saba-ja: @LeStarch, is it OK to remove the Cfg option?

saba-ja:
CmdDispatcher has an old UT. I will refactor the UT and add new test for the added feature.

config/CommandDispatcherImplCfg.hpp
LeStarch: @saba-ja yes, in the case where we check specifically the same opcode and port.  The only danger is wasted cycles if users call register too many times....but the behavior will be correct.

saba-ja:
Update summary:
1. Added CMD reregistration UT
2. Removed GCOV and test output files from UT folder
3. Removed GCOV references from SDD
4. Removed ALLOW_REREGISTRATION config option
5. Removed slot from reregistration EVR

Joshua-Anderson (approved):
Looks great!

LeStarch (approved):

LeStarch:
@saba-ja code looks good, UT looks good, and it passes build.  No one has commented that this would pose risk to their project, and we have a diagnostic command to represent what is going on.  I will mege soon.

PR #507, opened 2021-05-03
I2C return status enumeration into its own EnumAi.xml

metzdigital:
| | |
|:---|:---|
|**_Originating Project/Creator_**|  OWLS / Brandon Metz|
|**_Affected Component_**| LinuxI2cDriver |
|**_Related Issue(s)_**| None. |

---
## Change Description

Broke out the return status enumeration into its own EnumAi.xml, this allows us to use the status returned in EVRs along with all the other benefits. 

## Rationale

Would like to be able to use the i2c return status in EVRs and be able to assign it to a variable for later use. 

None. 

## Future Work

None.

### DIFF
--- a/Drv/I2cDriverPorts/I2cStatusEnumAi.xml
+++ b/Drv/I2cDriverPorts/I2cStatusEnumAi.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?oxygen RNGSchema="file:../../Autocoders/Python/schema/ISF_Type_Schema.rnc" type="compact"?>
+<enum namespace = "Drv" name="I2cStatus">
+    <comment>Enum of the status returns from the I2C Driver</comment>
+    <item name="I2C_OK" value="0" comment="Transaction okay"/>
+    <item name="I2C_ADDRESS_ERR" value="1" comment="I2C address invalid"/>
+    <item name="I2C_WRITE_ERR" value="2" comment="I2C write failed"/>
+    <item name="I2C_READ_ERR" value="3" comment="I2C read failed"/>
+    <item name="I2C_OTHER_ERR" value="4" comment="Other errors that don't fit"/>
+</enum>
+
--- a/Drv/I2cDriverPorts/CMakeLists.txt
+++ b/Drv/I2cDriverPorts/CMakeLists.txt
@@ -7,6 +7,7 @@
 ####
 set(SOURCE_FILES
 	"${CMAKE_CURRENT_LIST_DIR}/I2cPortAi.xml"
+	"${CMAKE_CURRENT_LIST_DIR}/I2cStatusEnumAi.xml"
 )
 
 register_fprime_module()
--- a/Drv/I2cDriverPorts/I2cPortAi.xml
+++ b/Drv/I2cDriverPorts/I2cPortAi.xml
@@ -2,6 +2,7 @@
 <?oxygen RNGSchema="file:../xml/ISF_Type_Schema.rnc" type="compact"?>
 
 <interface name="I2c" namespace="Drv">
+  <import_enum_type>Drv/I2cDriverPorts/I2cStatusEnumAi.xml</import_enum_type>
   <include_header>Fw/Buffer/Buffer.hpp</include_header>
     <args>
         <arg name="addr" type="U32">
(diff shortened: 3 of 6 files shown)


Drv/I2cDriverPorts/I2cStatusEnumAi.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
saba-ja, line 1: Need to include this file in Drv/I2cDriverPorts/CMakeLists.txt as shown below:

```cmake
set(SOURCE_FILES
	"${CMAKE_CURRENT_LIST_DIR}/I2cPortAi.xml"
	"${CMAKE_CURRENT_LIST_DIR}/I2cStatusEnumAi.xml" # <----------- Add this 
)
```

metzdigital: Thank you Saba, my mistake when copying over to my forked version. It's now fixed.

Joshua-Anderson (approved):
Looks great!

saba-ja:
In the CMakeLists.txt shown below:
https://github.com/nasa/fprime/blob/811beae18717f26277c1091feba4ee1db56d481c/Drv/LinuxI2cDriver/CMakeLists.txt#L8-L28

any `Linux` platform, despite having I2C or not, is going to use the actual `LinuxI2cDriverComponentImpl`. This might not be the case for most x86 architectures. 

1) Would that be a correct approach to change the CMakeLists such that the Linux platforms also use  `LinuxI2cDriverComponentImplStub.cpp`?

2) If that is not a good idea could we add a CMake option (e.g `-D USE_STUB_LINUX_I2C_DRIVER`) to enforce the build to specifically use the `LinuxI2cDriverComponentImplStub.cpp` on a Linux platform?

-----
Option 1 implementation:

```cmake
elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux") 
 	set(SOURCE_FILES 
 		"${CMAKE_CURRENT_LIST_DIR}/LinuxI2cDriverComponentAi.xml" 
 		"${CMAKE_CURRENT_LIST_DIR}//LinuxI2cDriverComponentImplStub.cpp"
 	) 
```

Option 2 implementation:

```cmake
if (${USE_STUB_LINUX_I2C_DRIVER}) 
 	add_definitions(-DSTUBBED_LINUX_I2C_DRIVER) 
 	set(SOURCE_FILES 
 		"${CMAKE_CURRENT_LIST_DIR}/LinuxI2cDriverComponentAi.xml" 
 		"${CMAKE_CURRENT_LIST_DIR}/LinuxI2cDriverComponentImplStub.cpp" 
 	) 
```

Joshua-Anderson:
@saba-ja I believe (but don't know for certain!) that any linux system with linux userspace headers should have the i2c header, so any linux system should be able to compile these drivers. For example, the docker container we run use to run CI is able to compile the I2C driver.

The main purpose of the stub I2C driver for me at least is just so that you can compile deployments using the linux i2c driver on Mac OS. I wouldn't recommend trying to actually run the deployment on a Mac OS machine, since all the code actually trying to talk to I2C devices is going to misbehave.

In the same way, trying to run a deployment trying to communicate with I2C devices on not a linux system that isn't the target platform probably isn't going work well either The only benefit I see with using the stub driver on linux would be that there's no way you could unintentionally command I2C devices on your system. To be fair, I could see errant I2C transactions on a desktop linux system causing problems, but generally you can only communicate with I2C buses as root, so somebody would have to run an unknown fprime deployment using I2C as root to have even a potential of experiencing issues.

I'm not super opposed to making the stub driver a configuration option, but if we do that, we should probably revisit all the other drivers too. If we do that I'd probably want to cleanup the way the stub driver configuration works to make it a bit cleaner.

Either way, I think it's probably out of scope of this pull request! If you think this is worth pursing we should open a separate issue for this.

saba-ja:
@Joshua-Anderson, thank you for the helpful clarification!
Regarding the refactor of drivers I believe it should be added to one of future action items.

LeStarch:
@saba-ja I think we are going to need to rethink the drivers package w.r.t linux, darwin, and embedded linux.  I like the idea of using stub drivers...but I agree that it is outside the scope of this PR.

PR #523, opened 2021-05-06
FileDownlink: Don't return an error status if attempting to transfer past EOF

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**| FileDownlink  |
|**_Related Issue(s)_**|  |

---
## Change Description

Return a OK status instead of INVALID when attempting to send a partial chunk past the end of a file.

## Rationale

Often times file sizes aren't known before sequence execution because they're generated in the same sequence as the file downlink. This can make commands like SEND_PARTIAL dangerous because if the file is smaller anticipated, they can return an INVALD error to the command dispatcher, aborting and preventing execution of the remaining sequence.

While attempting to send a chunk past the end of a file is an error, it's very rare that the rest of sequence is dependent on the success of a file downlink (see the FILEDOWNLINK_COMMAND_FAIL_ON_MISSING_FILE option).

Returning an ok command status but raising an WARNING_HI event seems like a decent compromise.

## Future Work

Should this be tied into the FILEDOWNLINK_COMMAND_FAIL_ON_MISSING_FILE to make this behavior configurable? 

I don't think this is a perfect solution. I could definitely anticipate some users who might prefer to abort a sequence in the presence of a file downlink error (like possibly when transferring files between different computers in a hub topology).
I think the idea solution would be a more advanced command sequencer redesign that allows users to indicate that certain commands are allowed to fail. However, I think this fix to make FileDownlink a bit safer is the right move until a command sequencer redesign occurs.


timcanham:
@Joshua-Anderson You beat me to it! :)

timcanham:
Where is the `FILEDOWNLINK_COMMAND_FAIL_ON_MISSING_FILE` option discussed?

saba-ja:
Defined in the following config files:
https://github.com/nasa/fprime/blob/222048e67978f32f591e873c2a7157f6c6add06a/config/FileDownlinkCfg.hpp#L21-L23

Used in the following:

https://github.com/nasa/fprime/blob/222048e67978f32f591e873c2a7157f6c6add06a/Svc/FileDownlink/FileDownlink.cpp#L362-L366

timcanham:
My vote would be to have this behavior use the `FILEDOWNLINK_COMMAND_FAIL_ON_MISSING_FILE` (or a better named one as @LeStarch ) suggests.

Joshua-Anderson:
I'll update this PR to add a common `FILEDOWNLINK_COMMAND_FAILURES_DISABLED` option that people can adjust.

LeStarch (approved):
Looks ok to me, assuming the UTs pass.

config/FileDownlinkCfg.hpp
```diff
@@ -13,11 +13,11 @@ namespace Svc {
...
-    // If this is set to true, if a file is not found the command will fail.
-    // If it is false, the command will still pass.
-    // The false setting is to prevent a sequence from ending with
-    // an error simply because one file name was wrong
-    static const bool FILEDOWNLINK_COMMAND_FAIL_ON_MISSING_FILE = false;
+    // If this is set, errors that would cause FileDownlink to return an error response, such as a
+    // missing file or attempting to send a partial chunk past the end of the file will instead
+    // return success. This is recommended to avoid a non-serious FileDownlink error aborting a
+    // sequence early. These errors will still be logged as events.
+    static const bool FILEDOWNLINK_COMMAND_FAILURES_DISABLED = true;
```
LeStarch, line 20: Do we really want to default this to true?  It seems that the default should be more restrictive, and open it up for projects to discuss.

Joshua-Anderson: It previously defaulted to the less restrictive option as well. Since a file downlink error doesn't impact the sate of the rest of the flight software (at least for most use cases!), I don't think most projects would want an error in file downlink to abort a sequence and end an activity prematurely.

LeStarch:
Merging.

Issue #531, opened 2021-05-07
Yes


lorena1976:
Yes 

__Originally posted by @lorena1976 in https://github.com/nasa/earthdata-search/pull/1369#discussion_r628492574__

lorena1976:
[]()

lorena1976:
![image](https://user-images.githubusercontent.com/79397110/117518620-4d5db580-af55-11eb-9986-13bc25cd6656.jpeg)

PR #532, opened 2021-05-07
Bugfix for `UnboundLocalError` in SerializableType

dwillmer:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Bugfix for `UnboundLocalError` in SerializableType, including unit-test.

## Rationale

The docstring on SerializableType declares that 'the member descriptions can be None'.
However if you actually pass `None` (or don't pass the list at all, given that None is set
as the default kwarg), then you get an `UnboundLocalError: local variable 'new_mem_list' referenced before assignment`, due to the `if mem_list` branch hiding the creation of the `new_mem_list` local variable.

This change adds a default binding as `new_mem_list = []` so that in the case of an empty `mem_list`, or `mem_list=None`, then the object will always be initialised properly.

A check has been added to the existing SerializableType unit test verifying this behaviour. If you remove the change from the class constructor and run the updated test you will get the UnboundLocalError.

Run the python tests, to verify the original bug, comment out the new line in the SerializableType constructor.

## Future Work

N/A

### DIFF
--- a/Fw/Python/test/fprime/common/models/serialize/test_types.py
+++ b/Fw/Python/test/fprime/common/models/serialize/test_types.py
@@ -274,6 +274,10 @@ def test_serializable_type():
     memList = [(a, b, c, None) for a, b, c in memList]
     check_cloned_member_list(mem_list, memList)
 
+    serTypeEmpty = SerializableType("ASerType", [])
+    assert serTypeEmpty.val == {}
+    assert serTypeEmpty.mem_list == []
+
 
 # def test_array_type():
 #    """
--- a/Fw/Python/src/fprime/common/models/serialize/serializable_type.py
+++ b/Fw/Python/src/fprime/common/models/serialize/serializable_type.py
@@ -32,6 +32,8 @@ class SerializableType(ValueType):
         if not isinstance(typename, str):
             raise TypeMismatchException(str, type(typename))
         self.__typename = typename
+
+        new_mem_list = []
         # If the member list is defined, stamp in None for any missing descriptions
         if mem_list:
             new_mem_list = [


Fw/Python/test/fprime/common/models/serialize/test_types.py
```diff
@@ -274,6 +274,10 @@ def test_serializable_type():
     memList = [(a, b, c, None) for a, b, c in memList]
     check_cloned_member_list(mem_list, memList)
 
+    serTypeEmpty = SerializableType("ASerType", [])
+    assert serTypeEmpty.val == {}
```
lorena1976, line 278: Hi

lorena1976: Hi

Fw/Python/test/fprime/common/models/serialize/test_types.py
```diff
@@ -274,6 +274,10 @@ def test_serializable_type():
     memList = [(a, b, c, None) for a, b, c in memList]
     check_cloned_member_list(mem_list, memList)
 
+    serTypeEmpty = SerializableType("ASerType", [])
```
lorena1976, line 277: []()

dwillmer:
Judging by the comments above and the other activity on that account, `lorena1976` appears to be a fairly low-quality bot. 

I think it'll need an owner/someone with write access to report/block that account.

jsoref:
fwiw, it has spammed more than one PR in this repo (and in fact a dependabot PR in another nasa repo).

Here's the approach I'm going to take (I'm not a repo admin / org admin for nasa):
https://support.github.com/contact/report-abuse?category=report-abuse&report=lorena1976&report_type=user

https://github.com/nasa/fprime/pull/532#discussion_r628623104
https://github.com/nasa/edsc-echoforms/pull/94#pullrequestreview-654903093
https://github.com/storybookjs/storybook/pull/14849#issuecomment-835607816
https://github.com/storybookjs/storybook/issues/14867
https://github.com/nasa/fprime/issues/531
https://github.com/mail-in-a-box/mailinabox/issues/1975

LeStarch (approved):
Looks good to me. Good catch!

PR #541, opened 2021-05-11
lestarch: adding options for stubbed drivers and baremetal scheduler

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**| CMake |
|**_Related Issue(s)_**|  |

---
## Change Description

Added switches to turn on/off the baremetal scheduler and stubbed drivers.  In this way these features can be enabled/disabled.  This allows breaks the hard-dependence on platform for drivers.

Documented switches in cmake/options.

## Rationale

"Big" features are nice to be able to throw a switch at the command line when generating.  "BIg" features mean items that have a build-system and C++ component to them.  

Tested on darwin that all the defaulting and switching and error cases work.

### DIFF
--- a/Drv/LinuxGpioDriver/CMakeLists.txt
+++ b/Drv/LinuxGpioDriver/CMakeLists.txt
@@ -5,37 +5,20 @@
 # MOD_DEPS: (optional) module dependencies
 #
 ####
-
-
-if(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
-	set(SOURCE_FILES
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentAi.xml"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplCommon.cpp"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplStub.cpp"
-	)
-elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
-	set(SOURCE_FILES
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentAi.xml"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplCommon.cpp"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImpl.cpp"
-	)
-elseif(${CMAKE_SYSTEM_NAME} STREQUAL "CygWin")
-	set(SOURCE_FILES
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentAi.xml"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplCommon.cpp"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImpl.cpp"
-	)
-elseif(${CMAKE_SYSTEM_NAME} STREQUAL "arm-linux-gnueabihf")
-	set(SOURCE_FILES
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentAi.xml"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplCommon.cpp"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImpl.cpp"
-	)
+if(NOT DEFINED FPRIME_USE_STUBBED_DRIVERS OR FPRIME_USE_STUBBED_DRIVERS)
+    set(SOURCE_FILES
+        "${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentAi.xml"
+        "${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplCommon.cpp"
+        "${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplStub.cpp"
+    )
+elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" OR ${CMAKE_SYSTEM_NAME} STREQUAL "arm-linux-gnueabihf")
+    set(SOURCE_FILES
+        "${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentAi.xml"
+        "${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplCommon.cpp"
+        "${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImpl.cpp"
+    )
 else()
-	set(SOURCE_FILES
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentAi.xml"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplCommon.cpp"
-	)
+    message(FATAL_ERROR "Cannot use ${CMAKE_CURRENT_LIST_DIR} with platform ${CMAKE_SYSTEM_NAME}. Consider using -DFPRIME_USE_STUBBED_DRIVERS=ON")
 endif()
 
 register_fprime_module()
--- a/Drv/LinuxI2cDriver/CMakeLists.txt
+++ b/Drv/LinuxI2cDriver/CMakeLists.txt
@@ -5,26 +5,19 @@
 # MOD_DEPS: (optional) module dependencies
 #
 ####
-if (${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
-	add_definitions(-DSTUBBED_LINUX_I2C_DRIVER)
-	set(SOURCE_FILES
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxI2cDriverComponentAi.xml"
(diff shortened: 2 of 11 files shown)


Drv/LinuxGpioDriver/CMakeLists.txt
```diff
@@ -5,37 +5,20 @@
...
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplCommon.cpp"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImpl.cpp"
-	)
+if(NOT DEFINED FPRIME_USE_STUBBED_DRIVERS OR FPRIME_USE_STUBBED_DRIVERS)
+    set(SOURCE_FILES
+        "${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentAi.xml"
+        "${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplCommon.cpp"
+        "${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplStub.cpp"
+    )
+elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" OR ${CMAKE_SYSTEM_NAME} STREQUAL "arm-linux-gnueabihf")
```
Joshua-Anderson, line 14: Is there a reason we set `CMAKE_SYSTEM_NAME` to `arm-linux-gnueabihf` in the ARM toolchain? Setting it to `Linux` might simplify things so that we only need to handle the Linux and MacOS cases. I can see all these conditional compilation if statements exploding if we add support for more linux cross compilation targets. It's also a bit of problem for out of tree toolchains that want to use these linux drivers in core fprime but can't easily add their own cases without their own custom fork.

Joshua-Anderson (commented):
I like the idea of a big switch to be able to toggle and untoggle features like stubbed drivers. Do you think that we need to support more granularity so that people can stub or unstub a particular driver? Could be useful for enabling a certain driver for testing.

Drv/LinuxGpioDriver/CMakeLists.txt
LeStarch: @Joshua-Anderson  I think this is a bigger issue.  Ideally, the CMAKE_SYSTEM_NAME would be consistent with the compiler (e.g. arm-...) but we should track that it "is a linux" somehow.  We can create an issue to fix this.

LeStarch:
@Joshua-Anderson overall this helps F´ switch between drivers better than it does.  I don't want to tackle turning on/off individual drivers at this point as this will be deeply more in-depth.  We can look into that feature in the future following a similar pattern as demonstrated here.

cmake/platform/Linux.cmake
```diff
@@ -3,5 +3,18 @@
 #
 # Linux platform file for standard linux targets. Merely defers to [./Linux-common.cmake](Linux-common.cmake).
 ####
+
+# Set platform default for stubbed drivers
+if (NOT DEFINED FPRIME_USE_STUBBED_DRIVERS)
+   set(FPRIME_USE_STUBBED_DRIVERS ON)
```
Joshua-Anderson, line 9: Do we really want to default to stubbed drivers on linux? I think all of our linux drivers should compile on standard linux systems. I'd be kinda surprised if I compiled on linux and found the `LinuxGPIODriver` was silently stubbed out.

Joshua-Anderson: I'm a little worried this will present a poor experience for users compiling on their embedded systems, which I think is not an unlikely scenario for systems like the raspberry pi and beaglebone. They'll compile, silently used the stubbed modules, and see the the reference app start but "not work" and actually toggle GPIOs. Personally, I have no concerns compiling the real drivers for desktop systems. Worst case scenario why they try and run the app they'll get permission denied trying to open the hardware, but realistically you shouldn't be trying to actually run most F' deployments on desktop systems. 

It's not a huge deal either way though and if you prefer to leave the drivers stubbed that's okay, we should probably just make sure to explicitly document the configuration option.

cmake/platform/arm-linux-gnueabihf.cmake
```diff
@@ -5,5 +5,17 @@
...
 ##
 
+# Set platform default for stubbed drivers
+if (NOT DEFINED FPRIME_USE_STUBBED_DRIVERS)
+   set(FPRIME_USE_STUBBED_DRIVERS OFF)
+endif()
+
+# Set platform default for baremetal scheduler drivers
+if (NOT DEFINED FPRIME_USE_BAREMETAL_SCHEDULER)
+   set(FPRIME_USE_BAREMETAL_SCHEDULER OFF)
```
saba-ja, line 15: Is the `FPRIME_USE_BAREMETAL_SCHEDULER=OFF` correct here?

Drv/LinuxSerialDriver/CMakeLists.txt
```diff
@@ -5,38 +5,16 @@
...
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxSerialDriverComponentAi.xml"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxSerialDriverComponentImplCommon.cpp"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxSerialDriverComponentImpl.cpp"
-	)
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" OR ${CMAKE_SYSTEM_NAME} STREQUAL "arm-linux-gnueabihf" OR ${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
+    set(SOURCE_FILES
+        "${CMAKE_CURRENT_LIST_DIR}/LinuxSerialDriverComponentAi.xml"
+        "${CMAKE_CURRENT_LIST_DIR}/LinuxSerialDriverComponentImplCommon.cpp"
+        "${CMAKE_CURRENT_LIST_DIR}/LinuxSerialDriverComponentImpl.cpp"
+    )
```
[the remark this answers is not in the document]
LeStarch: It was once used for Darwin, before we added serial support for darwin.

Drv/LinuxGpioDriver/CMakeLists.txt
```diff
@@ -5,37 +5,20 @@
...
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplCommon.cpp"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImpl.cpp"
-	)
-elseif(${CMAKE_SYSTEM_NAME} STREQUAL "arm-linux-gnueabihf")
-	set(SOURCE_FILES
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentAi.xml"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplCommon.cpp"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImpl.cpp"
-	)
+if(NOT DEFINED FPRIME_USE_STUBBED_DRIVERS OR FPRIME_USE_STUBBED_DRIVERS)
```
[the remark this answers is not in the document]
LeStarch: RTEMs support isn't complete, but would use its own driver implementation anyway

cmake/platform/arm-linux-gnueabihf.cmake
LeStarch: yes.  In all in-repo cases, we do not run the scheduler.  We'd use the scheduler on the Arduino/teensyduino packages of fprime-community.

Fw/Cfg/CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ set(SOURCE_FILES
   "${CMAKE_CURRENT_LIST_DIR}/ConfigCheck.cpp"
 )
 register_fprime_module()
-if (BAREMETAL_SCHEDULER)
+if (DEFINED FPRIME_USE_BAREMETAL_SCHEDULER AND FPRIME_USE_BAREMETAL_SCHEDULER)
```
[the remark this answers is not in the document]
LeStarch: Fix incoming.

Drv/LinuxI2cDriver/CMakeLists.txt
```diff
@@ -5,26 +5,19 @@
...
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxI2cDriverComponentAi.xml"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxI2cDriverComponentImpl.cpp"
-	)
-elseif(${CMAKE_SYSTEM_NAME} STREQUAL "arm-linux-gnueabihf")
-	set(SOURCE_FILES
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxI2cDriverComponentAi.xml"
-		"${CMAKE_CURRENT_LIST_DIR}/LinuxI2cDriverComponentImpl.cpp"
-	)
+if(NOT DEFINED FPRIME_USE_STUBBED_DRIVERS OR FPRIME_USE_STUBBED_DRIVERS)
+    add_definitions(-DSTUBBED_LINUX_I2C_DRIVER)
```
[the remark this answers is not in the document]
LeStarch: Different developers, meeting different requirements.

LeStarch:
@saba-ja I think this is ready for the items I will work on in this work unit.

saba-ja (approved):
Changes look good to me. I built for both Ref and RPI.
The RPI was generated RPI with `-D FPRIME_USE_BAREMETAL_SCHEDULE` with no issue.
This PR is fine to be merged into devel.

PR #551, opened 2021-05-14
Always cleanup coverage gcda files before executing unit tests

### LINKED ISSUE #368: Code coverage analysis causes unit test failures on Mac OS
| | |
|:---|:---|
|**_F´ Version_**| 1.5.3 |
|**_Affected Component_**| All |
---
## Problem Description

On Mac OS, the following can occur:

1. Build and run component unit tests successfully. Doing this runs code coverage analysis, which generates .gcda artifacts.
2. Edit the component and/or the unit tests.
3. Build and run the component unit tests again, causing the coverage analysis to be re-run. This time the code coverage analysis gets confused by the old .gcda artifacts, causing it to spew warnings and sometimes fail with a segmentation fault.

A workaround is to manually wipe out the .gcda artifacts in the build cache, but this is inconvenient.

## How to Reproduce

Carry out the steps described above, on any component unit test. Do this on Mac OS. For some reason, the issue arises only on Mac OS and not on Linux.

## Expected Behavior

The code coverage analysis should not spew warnings and should not cause segmentation faults or other apparent test failures.

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  cmake |
|**_Related Issue(s)_**| fixes  #368  |

---
## Change Description

gcda files are generated per object file when executing a binary with coverage enabled. Make sure all existing coverage files are removed before re-running unit test executables

## Rationale

Unit tests are always built with coverage enabled, so every unit test run generates these gcda files. If you recompile an object while a gcda file already exists and re-run unit tests you'll get error messages saying that coverage can't be merged into existing gcda file. As a nice side affect this also means that a full clean is no longer required to get accurate code coverage results.

### DIFF
--- a/cmake/support/Unit_Test.cmake
+++ b/cmake/support/Unit_Test.cmake
@@ -29,7 +29,7 @@ add_custom_target(check_leak COMMAND ${CMAKE_CTEST_COMMAND}
 # Performs registration of the autocoder step for the generation of GTestBase.hpp, GTestBase.cpp,
 # TesterBase.cpp, and TesterBase.hpp. These autocoding steps automate, and keep up-to-date the
 # above files saving time.
-# 
+#
 # - **EXE_NAME:** name of exe (unit test exe)
 # - **SOURCE_FILES:** source files to provide for autocoding
 ####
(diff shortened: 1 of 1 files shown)


LeStarch (approved):

LeStarch (commented):
Thinking about this a bit more, is "checking" one module going to remove the .gdca files for another module?  It looks like it.  Perhaps this behavior is ok, but I'd like clarification.

Joshua-Anderson:
@LeStarch Yes, this cleans up *all* gcda files in your project. The problem is that every dependency file your module uses, like Os/Queue, will also generate gcda files. This contributed to a "feature" where every coverage run of a module would add to the coverage files of other modules, so it was hard to know the true state of your coverage without cleaning between every run.

Rather than try and transitively figure out exactly which modules need to be cleaned, I just clean everything.

I clean up before running the unit tests, so after a unit test run the gcda files are still present and the user can archive or investigate whichever gcda files they were interested in.

LeStarch (approved):

PR #477, opened 2021-04-12
Fix 471: Adding a `MAX` enum item as the last item in the external enum

saba-ja:
| | |
|:---|:---|
|**_Originating Project/Creator_**| FPrime |
|**_Affected Component_**|  External enum_hpp.tmpl |
|**_Affected Architectures(s)_**|  Autocoder |
|**_Related Issue(s)_**|  #471|
|**_Has Unit Tests (y/n)_**|  N|
|**_Builds Without Errors (y/n)_**| Y |
|**_Unit Tests Pass (y/n)_**| NA |
|**_Documentation Included (y/n)_**| N |

---
## Change Description

This PR will add feature requested in #471 by adding a `MAX` enum item as the last item in the external enum.

## Rationale
The fix is adding `${name}_MAX = ${int(item_value) + 1},` after the end of template for loop as shown below:
```c++
    typedef enum {
#for $item_name,$item_value,$item_comment in $items_list:
#if not $item_value == ""
#if not item_comment == ""
        ${item_name} = ${item_value}, //!< ${item_comment}
#else
        ${item_name} = ${item_value},
#end if
#else
#if not item_comment == ""
        ${item_name}, //!< ${item_comment}
#else
        ${item_name},
#end if
#end if
#end for
        ${name}_MAX = ${int(item_value) + 1}, 
    } t;
```

The assumptions are 
1. External enums cannot have empty `value` (See `#if not $item_value == ""` in the code above).
2. External enums cannot have string value (Enforced by compiler).

In the code above the last `item_value` will be incremented by `1` and will be assigned to the last item of the enum with the name of external enum class concatenated to the word `_MAX` as shown in the last line of code above.

## Testing/Review Recommendations

Manually checked the autogenerated code on some external enums and verified that the `_MAX` item is added.
The followings are two examples of the result:

image

image

## Future Work

NA

Issue #471, opened 2021-03-29
Add an ENUM_LENGTH constant to FPrime enums


kbreed:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**| Autocoder |
---
## Automatically add an ENUM_LENGTH constant to F' enums.

Have the autocoder add a constant with the name ENUM_LENGTH to F' enums, similar to how SERIALIZABLE_SIZE works.

## Rationale

Needing to know the length of an enum list is a common need. Now is it up to developers to add a final member to the enum and expect that any maintaining developers would not accidentally add another member after that one.

timcanham:
The autocoder already puts a member `_MAX` to the end of generated enums that serves that purpose. With the C/C++ rules for assigning values to enum members, it should always be the highest value. If none of the members have explicit values assigned, it can be used for the size as well.

saba-ja:
I believe the above request is referring to [enums/enum_hpp.tmpl](https://github.com/nasa/fprime/blob/devel/Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_hpp.tmpl ) rather than [component/hpp.tmpl](https://github.com/nasa/fprime/blob/devel/Autocoders/Python/src/fprime_ac/generators/templates/component/hpp.tmpl)


## Clarification - Part 1:

`${enum_type}_MAX` is implemented in [component/hpp.tmpl](https://github.com/nasa/fprime/blob/devel/Autocoders/Python/src/fprime_ac/generators/templates/component/hpp.tmpl) which adds `_MAX` to the enums defined in `*ComponentAi.xml`

**Example:**

Assuming one has a component named `ActivityManagerComponentAi.xml` with the following command definition:
``` xml
<commands>
        <command kind="guarded" opcode="0" mnemonic="Recording">
            <comment>
            Command to start recording science data to file
            </comment>
            <args>
            <arg type="ENUM" name="ARG1">
                <enum name="rec_start_top">
                    <item name="start"/>
                    <item name="stop"/>
                </enum>
            </arg>
            </args>
        </command>
...
</commands>
```

The autocoder will generate the following `*Ac.hpp` file:
`ActivityManagerComponentAc.hpp`

```c++
...
    typedef enum {
      start, /*  */
      stop, /*  */
      rec_start_top_MAX
    } rec_start_top;

```

In this case, as mentioned above, `rec_start_top_MAX` value can be used as the number of items in the enum.

## Clarification - Part 2:

[...]

PR #553, opened 2021-05-14
Fix external enums "enum-compare" warning issue

saba-ja:
| | |
|:---|:---|
|**_Affected Component_**| Autocoder |
|**_Affected Architectures(s)_**| external enum |
|**_Related Issue(s)_**|  #471|

---
## Change Description

This PR fixes an issue in PR #477.
`NUM_CONSTANTS` was provided as `enum`. This causes `enum-compare` warning when user performs comparison similar to the following:
```
valveId.e >= CommonEnum::ValveId::NUM_CONSTANTS
```

This pr resolves the issue by use const as shown below:
```
const U32 NUM_CONSTANTS = ${len($items_list)};
```

To make the comparison user can write a similar code as following:

```
valveId.e >= valveId.NUM_CONSTANTS
```

## Rationale

Comparison of external enum values with NUM_CONSTANTS should not cause any warning.

Built and run with no issues.
An example output looks like the following:
```
//! Number of items in BpManagerState enum 
const U32 NUM_CONSTANTS = 4;
```

## Future Work

NA

### DIFF
--- a/Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_hpp.tmpl
+++ b/Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_hpp.tmpl
@@ -66,9 +66,11 @@ namespace ${namespace} {
     // ----------------------------------------------------------------------
 
     enum {
-        SERIALIZED_SIZE = sizeof(FwEnumStoreType),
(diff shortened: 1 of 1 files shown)


LeStarch (approved):
This prevents cross-enum comparisons by using a const integer!

PR #591, opened 2021-05-20
lestarch: updating fprime to include versioning information

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |
|**_Documentation Included (y/n)_**|  n/a |

---
## Change Description

Add an fprime version to dictionaries.  Here is how it is found:

1. Use python setuptools_scm to get the version
2. If the above fails because the framework is not store in a repo, use the `version.hpp` header

`version.hpp` needs to be updated with each C++ release.


LeStarch:
@timcanham and @Joshua-Anderson I'd appreciate your thoughts here.

Joshua-Anderson (commented):
If we have a `version.hpp` file, what is the benefit of trying to pull the version from the git repo? I perfer the simplicity of fetching the version from a single approach.

version.hpp
```diff
@@ -0,0 +1,23 @@
+// ======================================================================
+// \title  version.hpp
+// \author mstarch
+// \brief  fprime version information. MUST BE UPDATED for each release.
+//
+// \copyright
+// Copyright 2009-2015, by the California Institute of Technology.
```
jsoref, line 7: If this is a new file, is this block's date really correct?

LeStarch: Old template. I will fix.

LeStarch:
@Joshua-Anderson the ideal solution is to use setuptools_scm because it accounts for "deltas" from the base commit.  This is a good approach as it will ensure that the version information is captured even on `devel`.

`version.hpp` is a fallback for if/when this repo is hosted or used as a tarball or other non-versioned source.  This will ensure we have **any** version information.  Since these use cases would start with a formal tag and would be static, we can safely default to a the `version.hpp`.

When we revisit the delivery process to clean up `documentation`, we can also make version.hpp an autogenerated header that tracks the latest build.

LeStarch:
I fixed the copyright header here...any other thoughts on the versioning and fallback versioning scheme?

Joshua-Anderson:
I'm a little bit concerned about the fprime fork use case. If I was developing a large F' project, I'd probably fork F' and keep a custom branch with our patches. If I did project specific internal F' releases I also probably wouldn't use the same versioning scheme that F' does. For those cases (custom branch, custom version tag) what version does setuptools_scm return?

I somewhat prefer always pulling the version from version.hpp so that F' forks don't accidentally run into issues when they do things like use a non-official tagged version of F'.

timcanham:
A little late to this conversation, but in the old make system there was a single shell to generate a version. That way a project could override it with something else including a non-GitHub scm.

LeStarch:
@Joshua-Anderson and @timcanham I hear what you are saying and I will switch `version.hpp` to a script that projects can override to get the version.  The default will use git scm and then default to a hard-coded version for stock F´.

Joshua-Anderson:
@LeStarch Really all we're using this for the CMake interface version with fprime-util - I don't really want to let users override this unless they really know what they're because otherwise they unintentionally break compatibility with the stock fprime-util - wheras I think a user forking F' and pointing their deployment to use a custom commit is pretty common. I don't know exactly how you intend to make this switch between SCM and the manual file work, but it might be tricky to detect which option to go with - but I suspect you're already thinking about this!

LeStarch:
@Joshua-Anderson this will be used in 2 places: fprime-util and Gds version checks. I don't think projects should be touching this at all.  If a project changes something in F´ substantially enough to need to bump this version, I think that project is "proceeding at their own risk".  

I think it is a dangerous precedent to assume that we need to allow any arbitrary change to F´ to be made by projects and still be supported by our stock tooling.

We **do** need a way to specify a project version in these dictionaries such that projects can track their dictionary version.  However, this is not the intent of this PR.

I will thusly do the following:

1. Add a "version.py" script that can be run to get the version of the framework.
2. Call the above script from the autocoder
3. Call the above script from fprime-util (once we get there)

Autocoders/Python/src/fprime_ac/utils/version.py
```diff
@@ -0,0 +1,14 @@
...
+import os
+from setuptools_scm import get_version
+
+ROOT_PARENT_COUNT=5
+
+def get_fprime_version():
+    """ Gets the fprime version using setuptools_scm  """
+    # First try to read the SCM version
+    try:
+        return get_version(root=os.sep.join([".."] * ROOT_PARENT_COUNT), relative_to=__file__)
```
Joshua-Anderson, line 11: What version number does this generate when you're on the development branch?

Joshua-Anderson, line 11: Optional nitpick: Does get_version need to run in the absolute repository root?  You might be able to get away with not needing to set root.

cmake/support/AC_Utils.cmake
```diff
@@ -99,6 +99,7 @@ function(acwrap AC_TYPE AC_FINAL_SOURCE AC_FINAL_HEADER AI_XML XML_FILE_DEPS MOD
       ${CMAKE_COMMAND} -E env PYTHONPATH=${PYTHON_AUTOCODER_DIR}/src:${PYTHON_AUTOCODER_DIR}/utils BUILD_ROOT="${FPRIME_BUILD_LOCATIONS_SEP}"
       FPRIME_AC_CONSTANTS_FILE="${FPRIME_AC_CONSTANTS_FILE}"
       PYTHON_AUTOCODER_DIR=${PYTHON_AUTOCODER_DIR}
+      FPRIME_FRAMEWORK_PATH="${FPRIME_FRAMEWORK_PATH}"
```
Joshua-Anderson, line 102: Why was this change needed?

Joshua-Anderson (commented):
Looks good overall!

My only remaining concern is how to handle F' project forks that use tags. Right now, they'll grab the project specific tag, which will return an incompatible version in the dictionary, so the GDS won't work with their dictionaries.

Maybe it's worth adding a GDS flag to override the version check?

Autocoders/Python/src/fprime_ac/utils/version.py
LeStarch: yes it does.

cmake/support/AC_Utils.cmake
LeStarch: old change.  I'll revert.

cmake/target/dict.cmake
```diff
@@ -31,6 +31,7 @@ function(dictgen MODULE_NAME AI_XML DEPS)
       ${CMAKE_COMMAND} -E env PYTHONPATH=${PYTHON_AUTOCODER_DIR}/src:${PYTHON_AUTOCODER_DIR}/utils BUILD_ROOT="${FPRIME_BUILD_LOCATIONS_SEP}"
       FPRIME_AC_CONSTANTS_FILE="${FPRIME_AC_CONSTANTS_FILE}"
       PYTHON_AUTOCODER_DIR=${PYTHON_AUTOCODER_DIR}
+      FPRIME_FRAMEWORK_PATH="${FPRIME_FRAMEWORK_PATH}"
```
Joshua-Anderson, line 34: Is this a leftover change too?

LeStarch: Yup. Fixed now.

Joshua-Anderson (approved):
🚀

Autocoders/Python/src/fprime_ac/utils/version.py
```diff
@@ -0,0 +1,14 @@
...
+    # Fallback to a specified version when SCM is unavailable
+    except LookupError:
+        return "1.5.4" # Must be kept up-to-date when tagging
```
Joshua-Anderson, line 14: Is it possible to move this version number to an easy to update place so it's hard for us to forget?

LeStarch: I'd like to leave it here for now.  I have a plan that will make bumping it irrelevant.

LeStarch: Here is the short version of the plan:

1. Create F´ metapackage on pypi
2. Have the above package write out a version file to be included above.
3. Ship the above output as our release tarball.

All automatic, no hand editing.  ....but it is more than I'd like to fit into this PR.

Joshua-Anderson: I'm okay with leaving this as is as long as there is a plan for future automation.

Joshua-Anderson (approved):

PR #593, opened 2021-05-20
lestarch: updating fprime-gds to check dictionary framework_version

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**| GDS |
|**_Related Issue(s)_**|  |
|**_Documentation Included (y/n)_**| y (error message w/ instructions) |

---
## Change Description

This adds version checking to the Gds.  This will error if a dictionary is out of the supported range preventing hidden errors from occurring.

### DIFF
--- a/Gds/src/fprime_gds/common/loaders/xml_loader.py
+++ b/Gds/src/fprime_gds/common/loaders/xml_loader.py
@@ -39,6 +39,7 @@ from fprime.common.models.serialize.numerical_types import (
 from fprime.common.models.serialize.serializable_type import SerializableType
 from fprime.common.models.serialize.string_type import StringType
 from fprime_gds.common.data_types import exceptions
+from fprime_gds.version import MINIMUM_SUPPORTED_FRAMEWORK_VERSION, MAXIMUM_SUPPORTED_FRAMEWORK_VERSION
 
 # Custom Python Modules
 from . import dict_loader
@@ -114,11 +115,21 @@ class XmlLoader(dict_loader.DictLoader):
         xml_parser = etree.XMLParser(remove_comments=True)
 
         with open(path) as fd:
-
             # Parse xml and get element tree object we can retrieve data from
             element_tree = etree.parse(fd, parser=xml_parser)
+        root = element_tree.getroot()
 
-        return element_tree.getroot()
+        # Check version of the XML before continuing. Versions weren't published before 1.5.4.  Only check major minor
+        # and point versions to allow for development versions to be allowed.
+        dict_version_string = root.attrib.get("framework_version", "1.5.4")
+        try:
+            dict_version = dict_version_string.split(".")
+            dict_version = tuple([int(item) for item in dict_version[0:3]])
+        except (ValueError, IndexError):
+                raise UnsupportedDictionaryVersionException(dict_version)
+        if dict_version < MINIMUM_SUPPORTED_FRAMEWORK_VERSION or dict_version > MAXIMUM_SUPPORTED_FRAMEWORK_VERSION:
+            raise UnsupportedDictionaryVersionException(dict_version)
+        return root
 
     @staticmethod
     def get_xml_section(section_name, xml_root):
@@ -389,3 +400,15 @@ class XmlLoader(dict_loader.DictLoader):
             raise exceptions.GseControllerParsingException(
                 "Could not find type %s" % type_name
             )
+
+
+class UnsupportedDictionaryVersionException(Exception):
+    """ Dictionary is of unsupported version """
+    def __init__(self, version):
+        """ Create a dictionary of a specific version """
+        def pretty(version_tuple):
+            """ Pretty print version """
+            return ".".join([str(item) for item in version_tuple])
+        super().__init__("Dictionary version {} is not in supported range: {}-{}. Please upgrade fprime-gds."
+                         .format(pretty(version), pretty(MINIMUM_SUPPORTED_FRAMEWORK_VERSION),
+                                 pretty(MAXIMUM_SUPPORTED_FRAMEWORK_VERSION)))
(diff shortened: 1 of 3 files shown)


Gds/src/fprime_gds/common/loaders/xml_loader.py
```diff
@@ -116,8 +117,14 @@ def get_xml_tree(self, path):
 
         # Parse xml and get element tree object we can retrieve data from
         element_tree = etree.parse(fd, parser=xml_parser)
+        root = element_tree.getroot()
 
-        return element_tree.getroot()
+        # Check version of the XML before continuing. Versions weren't published before 1.5.4.  Only check major minor
+        # and point versions to allow for development versions not "bumped" past a point version.
+        dict_version = root.attrib.get("framework_version", "1.5.4")
+        if dict_version < MINIMUM_SUPPORTED_FRAMEWORK_VERSION or dict_version[:5] > MAXIMUM_SUPPORTED_FRAMEWORK_VERSION:
```
Joshua-Anderson, line 125: I'm a little concerned about using a string comparison to compare a semver version number. Are we sure this won't cause issues if we decide to do a release like `2.10.0`?

Joshua-Anderson (commented):
Is there a dictionary schema we need to update? I didn't see one so I'm guessing the answer is no.

Other than my comment about comparing semver version numbers everything looks good!

Gds/src/fprime_gds/common/loaders/xml_loader.py
LeStarch: Excellent catch.  I can fix this.

Gds/src/fprime_gds/__init__.py
```diff
@@ -1,2 +1,7 @@
 # Python 2/3 compatibility for namespace packages
 __path__ = __import__("pkgutil").extend_path(__path__, __name__)
+
+# Currently the GDS is backwards compatible with all dictionaries. This MUST be bumped with each framework release to
+# ensure compatibility
+MINIMUM_SUPPORTED_FRAMEWORK_VERSION = "0.0.0"
+MAXIMUM_SUPPORTED_FRAMEWORK_VERSION = "1.5.4"
```
saba-ja, line 7: There are different ways to do this:

**Method 1:**
One way to define package version is to define it as `__version__`. See [package-version](https://packaging.python.org/guides/single-sourcing-package-version/)

Define `version.py` in `Gds/src/fprime_gds` with the following content:
```python
__min_version__ = (0, 0, 0)
__max_version__ = (1, 5, 4)
__version__ = '.'.join(map(str, __max_version__))
```
Then import it in Gds/src/fprime_gds/__init__.py as
```python
from .version import __min_version__
from .version import __max_version__
from .version import __version__
```
Now one can get the version of the package by calling 
```python
import fprime_gds
fprime_gds.__version__
```
output would be `'1.5.4'`

to get the version in Gds/src/fprime_gds/common/loaders/xml_loader.py to check the major release we can write
```python
import fprime_gds
dict_ver = tuple(map(int, root.attrib.get("framework_version", "1.2.3").split('.')))[0:2]
min_ver = fprime_gds.__min_version__[0:2]
max_ver = fprime_gds.__max_version__[0:2]

if any([dict_ver < min_ver, dict_ver >  max_ver]):
...
```
Also in the setup.py we can have:
```
exec(open('src/fprime_gds/version.py').read())

setup(
version=__version__,
)
```
This way the setup.py version will be updated when the version.py is updated.

**Method 2:**
Another option is to use importlib.metadata
this library will look at version attribute in setup.py

To get the version in Gds/src/fprime_gds/common/loaders/xml_loader.py one can write:
```
from importlib import metadata
import fprime_gds
if metadata.version('fprime_gds') > "2.2":
...
```

For more resource on defining metadata see [core-metadata](https://packaging.python.org/specifications/core-metadata/)

Gds/src/fprime_gds/common/loaders/xml_loader.py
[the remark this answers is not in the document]
LeStarch: Like @Joshua-Anderson said, there is a problem in my implementation that I am working on.

Gds/src/fprime_gds/__init__.py
LeStarch: I like your suggestion of making these versions tuples.  Easier to handle.  I will do that.

LeStarch:
@saba-ja @Joshua-Anderson should have fixed this to better handle versioning similar to @saba-ja's suggestion.

Note: the "Gds" version is not the same as the  FRAMEWORK version, so I did not update the `setup.py`.  You'll see this when we spin-off fprime-gds.  Cheers.

Gds/src/fprime_gds/common/loaders/xml_loader.py
```diff
@@ -114,11 +115,21 @@ def get_xml_tree(path):
...
-
             # Parse xml and get element tree object we can retrieve data from
             element_tree = etree.parse(fd, parser=xml_parser)
-
-        return element_tree.getroot()
+        root = element_tree.getroot()
+
+        # Check version of the XML before continuing. Versions weren't published before 1.5.4.  Only check major minor
+        # and point versions to allow for development versions to be allowed.
+        dict_version_string = root.attrib.get("framework_version", "1.5.4")
```
saba-ja, line 124: Is it fine to always assume version 1.5.4 for all dictionaries without framework_version?

Joshua-Anderson (approved):
That tuple comparison for the version numbers is really clever!

Gds/src/fprime_gds/common/loaders/xml_loader.py
LeStarch: Yes, as the next version will include a full versioning, and previous versions are all run by the GDS as 1.5.4 anyway.

LeStarch:
Thanks to @saba-ja for the tuple comparisons recommendation.

Issue #519, opened 2021-05-05
GDS: Double Click Repopulation Not Working


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| 1.5.3/devel |
|**_Affected Component_**|  GDS GUI  |
---
## Problem Description

Double clicking on commands in the command history should repopulate the command-send box.

## How to Reproduce

1.  Run GDS/GUI with Ref
2. Send command NO-OP STRING with string arg
3. Send command NO-OP without string
4. Double click on the command history entry for NO-OP-STRING.
5. It **should** repopulate the send box.

## Expected Behavior

Double clicking on a command hisotry item should repopulate the command send box with that command.

LeStarch:
@saba-ja if you have time could you look into this?

saba-ja:
To clarify, the expectation is to click on a row in CMD hist table and make it to be selected in the Mnemonic dropdown. Also if it has args populate them in the arg input boxes? 

![image](https://user-images.githubusercontent.com/35859004/117311914-80ab2200-ae39-11eb-9fc3-11fbb333e0c4.png)

Is this an expected behavior or a feature request?
I do not see any implementation of `clickAction` shown below:
https://github.com/nasa/fprime/blob/97883e50982a1275383a721f6c8a1c4c2aa1433e/Gds/src/fprime_gds/flask/static/index.html#L344-L349

in command.js
https://github.com/nasa/fprime/blob/update/ground-prototype/Gds/src/fprime_gds/flask/static/js/vue-support/command.js

@LeStarch, Please advice if this functionality should be implemented.

Joshua-Anderson:
@saba-ja Exactly! You should be able to click on a row in the command history and re-populate it back up at the top. Makes it easy to re-send previously sent commands.

LeStarch:
@saba-ja the implementation for `clickAction` probably got moved or broke at some point.  This is likely why it stopped working.

Joshua-Anderson:
Fixed in #604!

PR #604, opened 2021-05-22
Fix 519: Repopulate cmd on double click

saba-ja:
| | |
|:---|:---|
|**_Affected Component_**|  GDS |
|**_Affected Architectures(s)_**|  static/js |
|**_Related Issue(s)_**| #519  |
|**_Unit Tests Pass (y/n)_**|  NA |

---
## Change Description

Added option to repopulate previous commands when user double clicks on a row in the command-history table.

## Rationale

> Double clicking on commands in the command history should repopulate the command-send box.

Manually ran the GUI and checked the double click functionality with no issues.

image

## Future Work

NA

### DIFF
--- a/Gds/src/fprime_gds/flask/static/index.html
+++ b/Gds/src/fprime_gds/flask/static/index.html
@@ -155,6 +155,7 @@ Note: Vue.js annotations are used, as Vue is the data-to-view framework of choic
                               :filter-text="filterText"
                               :initial-views="itemsShown"
                               :compact="compact"
+                              :click-action="clickAction"
                     ></fp-table>
                 </div>
             </template>
--- a/Gds/src/fprime_gds/flask/static/js/vue-support/command.js
+++ b/Gds/src/fprime_gds/flask/static/js/vue-support/command.js
@@ -66,6 +66,10 @@ Vue.component("command-item", {
  * Input command form Vue object. This allows for sending commands from the GDS.
  */
 Vue.component("command-input", {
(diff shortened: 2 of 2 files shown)


LeStarch (commented):
@saba-ja does this repopulate the arguments too?

LeStarch:
Reading your comments, I wonder what happens when you run this test:

1. Send SigGen_Settings with [1,2,3, Triangle]
2. Send SigGen_Settings with [4,5,6, SINE]
3. Double click on 1, 1,2,3, Tringle.

I think that it will it will keep the wrong arguments [4,5,6 SINE].  However, your approach is correct, I think you just need to pass in args from "item" and set them into the selected command.

saba-ja:
@LeStarch, double clicking on a row in `command-history` table will repopulate both command and its arguments.

The code is using the `item` object, passed from `fp-row`, which has both command and arguments info.
See the following:
https://github.com/nasa/fprime/blob/e1d095deb0b74430d218d2d82411f7f774e94a80/Gds/src/fprime_gds/flask/static/js/vue-support/command.js#L250-L254
*In the above `full_name` needed to be in the top level of `item` object instead of `template` property. `full_name` is used in `v-select` as the `label` of select option.*

Tested the above scenario. Clicking on the first row will repopulate correct SigType 'Triangle'

image

LeStarch:
@saba-ja thanks for the explanation.  This is good!

LeStarch (approved):

Issue #405, opened 2021-02-22
Publish gds and fw packages to PyPI


billallen256:
| | |
|:---|:---|
|**_F´ Version_**|current|
|**_Affected Component_**|Gds and Fw/python|
---
## Feature Description

Publish `Gds` and `Fw/Python` as `fprime-gds` and `fprime-fw` packages on PyPI.

## Rationale

- Easier installation for users since they can just use `python3 -m pip install fprime-gds`.
- `fprime-gds` package will be able to list `fprime-fw` as a dependency.
- Actual releases.  Users will be using versions of the packages released on PyPI, not random clones from GitHub.

EverLookNeverSee:
@billallen256 
I would like to work on this issue.

Joshua-Anderson:
Right now fprime-util and fprime-gds are pretty tightly tied to the version of F' you're using. Because different F' deployments/projects could use different versions of the F' framework, we recommend creating a venv for each deployment and then directly installing the two packages from the embedded fprime framework submodule. I guess you could install the specific versions you require from pypi instead, but that doesn't seem like much of an improvement.

Long term there has been some discussion about possibly breaking fprime-util and fprime-gds into their own repositories and versioning them separately, at which point I think publishing them to pypi would make sense, since you would then be able to update them separately from the F' framework.

EverLookNeverSee:
I tried to upload version 1.5.3 to test pipy but it returned me error below:
HTTPError: 400 Bad Request from https://test.pypi.org/legacy/
Invalid value for classifiers. Error: Classifier 'License :: OSI Approved :: Apache 2.0' is not a valid classifier.

BastianZim:
According to https://pypi.org/classifiers/ the only classifier with Apache is `License :: OSI Approved :: Apache Software License` but that is something that should probably be definitively answered by the maintainers.

billallen256:
As @Joshua-Anderson pointed out, I put in this ticket before I realized how tightly coupled everything was.  Decoupling things seems like a good direction to move though, as it would help remove some barriers to entry, and allow things to be more idiomatic.

EverLookNeverSee:
@LeStarch Could you please give me some hints in order to solve this issue?

LeStarch:
Given the complexity of this task, I have begun to start and split the packages apart and prepare for this.  Before this can be done officially, we need to merge:

#593, #594, #574

PR #642, opened 2021-05-27
lestarch: changes necessary to host on pypi

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**| Python |
|**_Affected Architectures(s)_**| Python |
|**_Related Issue(s)_**| Closing #405  |
|**_Has Unit Tests (y/n)_**| n  |
|**_Documentation Included (y/n)_**| Not yet (intended) |

---
## Change Description

Changes to setup.py and other python to work towards hosting in PYPI.

## Rationale

See #405

```
pip install fprime fprime-gds
```

## Future work

Correct documentation to use PYPI.  Split off Python as separate repos.  Update CI.


LeStarch:
@jsoref I created this issue 1min ago, yet the spelling bot commented 5mins ago.  How is this possible?

jsoref:
@LeStarch: you probably pushed your commit 5 mins ago and took 4 minutes to create the PR.

The bot will happily spell check your commits before you create a PR so you can fix things w/o people seeing your mistakes.

.lgtm.yml
```diff
@@ -28,8 +28,7 @@ extraction:
             # setup the venv and install packages
             - "python3 -m venv ./fprime-venv"
             - ". ./fprime-venv/bin/activate" 
-            - "pip install ./Fw/Python"
-            - "pip install ./Gds"
+            - "pip install fprime-tools fprime-gds"
```
Joshua-Anderson, line 31: Once these are moved to their own repositories this should be fine, but for now this means that we're not testing changes to fprime-util. Can we install the local versions of fprime-util and the gds until we've broken them out into their own repositories?

LeStarch: The above is the LGTM file.  It isn't testing, just building for code analysis.  It doesn't care about using a different version of the tools.  The build will result in the same output, or we broke the build entirely and have to re-release packages anyway.

As a side note, it is likely easier to separate out the packages than it is to fix the internals of a docker VM to work with the correct versioning scheme.  Since we are very close to separating out the packages I think this is more work than we can justify.

Joshua-Anderson (approved):

ci/bootstrap.bash
```diff
@@ -14,15 +14,12 @@ rm -rf "${USABLE_VENV}"
...
 # install dependencies based on the TEST_TYPE
 if [[ "${TEST_TYPE}" == "STATIC" ]]
 then
     # Only pylama and pylint are needed for STATIC
     pip install -U pylama pylama_pylint radon
 else
     # These are required for all other tests
-    pip install ${FPRIME_DIR}/Fw/Python || fail_and_stop "Failed to install fprime PIP module from ./Fw/Python"
-    pip install ${FPRIME_DIR}/Gds[test-api-xls] || fail_and_stop "Failed to install fprime PIP module from ./Gds[test-api-xls]"
+    pip install fprime-tools fprime-gds || fail_and_stop "Failed to install fprime PIP module from ./Fw/Python"
```
Joshua-Anderson, line 24: Maybe rename to `Failed to install fprime PIP modules`?

LeStarch: I want to get this merged. I can clean this up later.

PR #662, opened 2021-06-01
Remove unused LogAssert

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**| Logger  |
|**_Related Issue(s)_**|  |

---
## Change Description

It has been superseded by the default assert handler, which logs to logMsg as well.

## Rationale

This file was triggering compilation errors when compiling with -Wextra using gcc.


LeStarch:
Which file replaces this?

Joshua-Anderson:
@LeStarch https://github.com/nasa/fprime/blob/devel/Fw/Types/Assert.cpp#L130 -> the SwAssert function defaults to printing the assert if no assert hook is set.

LeStarch:
@Joshua-Anderson this version uses Fw::Logger specifically.  It exists because printing to the console may or may not be supported, but Logger can be used to implement other output (serial console, write to a file or ram buffer).  In this way the assert gets logged with the other log messages, not just dumped to the console.

Joshua-Anderson:
Good point, is there a reason that the default assert handler uses printf instead of logging using Fw:Logger?

LeStarch:
Historically `Fw::Logger` didn't exist only the implementation `Os::Logger`.  We haven't gone through and cleaned up asserting and thus it still uses the original print implementation.

Note: until Fw::Logger was created, `Fw` would avoid calling log as it created a circular dependency with `Os` .  `Fw::Logger` was created to remove this circular dependency.

Joshua-Anderson:
If I migrated the default assert handler to use Fw:Logger would you be okay with removing Fs:LogAssert?

LeStarch:
Closing as log assert handler is part of a large assert system cleanup that needs to happen.  Tracking is separate issue.

PR #665, opened 2021-06-01
Properly bound size when concatenating strings

Joshua-Anderson:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Use the size argument, rather than the end of string null terminator to bound StringBase::appendBuff

## Rationale

GCC flagged this by highlighting that the size argument was unused. Technically `strncat` will stop at the null terminator of the provided string, but respecting size allows handling the use case that a user only wants to append a partial string, as well as not crash when a string buffer isn't null terminated.


LeStarch (approved):

LeStarch:
size + 1 (null) is still bounded by remaining!

PR #666, opened 2021-06-02
Set O_EXCL when opening file with include_excl

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  Os |
|**_Related Issue(s)_**|  |
|**_Documentation Included (y/n)_**| |

---
## Change Description

When `include_excl` is set, set the `O_EXCL` flag. Previously this argument was unused (caught through gcc's unused parameter warning).

⚠️  Warning: This is a breaking change that silently changes default behavior.

## Rationale

If we have an `include_excl` parameter, it should be respected. However, this does change the default behavior of `File::open(const char* fileName, File::Mode mode)`, since it defaults `include_excl` to true. Because `include_excl` was ignored this effectively is equivalent to `include_excl` = false.

I think the default of not overwriting a file if it already exists is wise. Personally, I believe the value of enforcing this safer default is greater than the impact of breaking projects upgrading to F' 2.0 that implicitly depending on the overwrite on create behavior. However, this is a serious breaking change and it may make sense to keep the current unintentional default and update `File::open(const char* fileName, File::Mode mode)` to set the `include_excl` flag to false.

### DIFF
--- a/Os/Linux/File.cpp
+++ b/Os/Linux/File.cpp
@@ -66,6 +66,9 @@ namespace Os {
 #endif
             case OPEN_CREATE:
                 flags = O_WRONLY | O_CREAT | O_TRUNC;
+                if(include_excl) {
+                    flags |= O_EXCL;
+                }
                 break;
             case OPEN_APPEND:
                 flags = O_WRONLY | O_CREAT | O_APPEND;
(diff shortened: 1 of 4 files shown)


LeStarch:
This should be discussed before we approve this change since it changes the behavior of an interface.

LeStarch:
:shipit:

LeStarch (approved):
Since it only affects CREATE and that usage was corrected, I agree with the safer default.  We'll add to upgrade notes to 2.0

PR #668, opened 2021-06-02
Correct warnings raised by clang-analyze

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Corrects warnings raised by clang's static analyzer.

## Future Work

Ideally we could find some way to run clang's static analyzer in CI, but I haven't found any good github action integration options yet.

### DIFF
--- a/Svc/TlmChan/TlmChanImplRecv.cpp
+++ b/Svc/TlmChan/TlmChanImplRecv.cpp
@@ -38,6 +38,7 @@ namespace Svc {
                     FW_ASSERT(this->m_tlmEntries[this->m_activeBuffer].free < TLMCHAN_HASH_BUCKETS);
                     // add new bucket from free list
                     entryToUse = &this->m_tlmEntries[this->m_activeBuffer].buckets[this->m_tlmEntries[this->m_activeBuffer].free++];
+                    FW_ASSERT(prevEntry);
                     prevEntry->next = entryToUse;
                     // clear next pointer
                     entryToUse->next = 0;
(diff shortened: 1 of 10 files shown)


Svc/TlmChan/TlmChanImplRecv.cpp
```diff
@@ -38,7 +38,9 @@ namespace Svc {
                     FW_ASSERT(this->m_tlmEntries[this->m_activeBuffer].free < TLMCHAN_HASH_BUCKETS);
                     // add new bucket from free list
                     entryToUse = &this->m_tlmEntries[this->m_activeBuffer].buckets[this->m_tlmEntries[this->m_activeBuffer].free++];
-                    prevEntry->next = entryToUse;
+                    if(prevEntry) {
```
LeStarch, line 41: If the code assumes this exists, shouldn't this be an assert not an if guard?

LeStarch (approved):

Svc/TlmChan/TlmChanImplRecv.cpp
Joshua-Anderson: Agreed, an assertion is a better approach. The static analyzer is false warning on a prevEntry being null case which isn't smart enough to realize that entryToUse is guaranteed non-null the first time through the loop.

LeStarch: So you'll change this to an assert, or this is still there to appease the static analyzer?

LeStarch (approved):

PR #671, opened 2021-06-03
update docs.py

FrankFeenix:
| | |
|:---|:---|
|**_Originating Project/Creator_**| cmake |
|**_Affected Component_**| docs |
|**_Affected Architectures(s)_**| n |
|**_Related Issue(s)_**| n |

---
## Change Description

a little adjustment


LeStarch:
Can you explain why this is better functionality?  More pythonic?

FrankFeenix:
> Can you explain why this is better functionality? More pythonic?

Those two code examples are equivalent, with the exception that assert statements can be globally disabled with the -O command-line flag. I then tried using AssertionError which might give a more appropriate error message. Thanks for noticing

LeStarch (approved):
Even though this is a small utility script, it makes sense in the context to prevent the `-O` assertion skip problem as recorded in the PR

PR #674, opened 2021-06-04
I2C Driver Update - Repeated Start Functionality

metzdigital:
| | |
|:---|:---|
|**_Originating Project/Creator_**| OWLS / Brandon Metz |
|**_Affected Component_**| LinuxI2cDriver |
|**_Related Issue(s)_**| None. |

---
## Change Description

Adds a new port interface to the LinuxI2cDriver to allow a write and then read with repeated start bit.

I also changed all the printfs as part of the DEBUG_PRINT over to `Fw::Logger::logMsg`

## Rationale

Certain I2C devices (memory chips for example) require writing a set of bytes and then reading a set of bytes instantly with no stop bit in between. This is called a "repeated start". https://www.i2c-bus.org/repeated-start-condition/

Example from the ADS1119 ADC:
image
image

Testing was performed on a Zynq 7020 Processor/FPGA. The I2C device was using the Xilinx AXI-to-I2C IP core and so using the Xilinx Axi I2C linux driver.

Test data was collected via logic analyzer to see the repeated start.

Here is an example of setting a register to `0x0D` and then reading it back with the repeated start:
Last byte in the chain is the value we want to set:
image
Here we are reading it back (`0x0D`), the green dots represent the start bit:
image

Another example of reading two bytes back from the ADS1119 ADC:
image

## Future Work

None.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -1073,6 +1073,7 @@ Nicolich
 ninjaaron
 nio
 NMEA
+nmsgs
 noargport
 NOBUFFERS
 NOCOLOR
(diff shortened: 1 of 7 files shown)


LeStarch (approved):
LGTM, I made some minor fixes:

1. spelling adjustment
2. Removing discontinued schema line

Issue #678, opened 2021-06-07
Resolve Static Analysis Warnings


saba-ja:
| | |
|:---|:---|
|**_F´ Version_**| 1.5 |
|**_Affected Component_**| All  |
---
## Problem Description
There are 444 warnings after running CodeSonar on Ref deployment.
This ticket will be used to track series of PRs resolving these warnings.

Category and significance  of the warnings is listed below:

| significance | count |
| -- | -- |
|Security | 399 |
|Reliability | 10 |
|Redundancy | 35 |

Class | Count
| -- | -- |
Uninitialized Variable | 255
Cast Alters Value | 120
Unreachable Data Flow | 15
Redundant Condition | 11
No Space For Null Terminator | 9
File System Race Condition | 7
Useless Assignment | 7
Coercion Alters Value | 6
Ignored Return Value | 6
Null Pointer Dereference | 3
Integer Overflow of Allocation Size | 1
Dangerous Function Cast | 1
Tainted Buffer Access | 1
Unreachable Computation | 1
Unreachable Conditional | 1

## How to Reproduce

1. Run CodeSonar on Ref deployment

## Expected Behavior

Warnings should be either resolved or descoped with a rational.

LeStarch:
As part of this we need to run static analysis in CI and continue to fix.

matt392code:
Create a C++ static analysis helper class that can be used alongside the existing F´ framework.
[static-analysis-helper.txt](https://github.com/user-attachments/files/18265520/static-analysis-helper.txt)
Example usage code to demonstrate how to use these utilities:
[static-analysis-example.txt](https://github.com/user-attachments/files/18265521/static-analysis-example.txt)
These utilities provide several safety features:
1. BLOCKER (Security/Null Pointers):
- Pointer validation before use
- Assert-based pointer checking
- Templated pointer safety checks

2. CRITICAL (Buffer/Initialization):
- Buffer size validation
- String safety checks with null termination verification
- Initialization tracking for critical variables
- Buffer overflow prevention

3. MAJOR (Cast/Coercion):
- Type-safe casting with range checking
- Signed/unsigned conversion safety
- Floating-point to integer safety checks
- Compile-time type validation

Key Features:
1. Template-based to work with any data type
2. Compile-time checks where possible
3. Runtime assertions for dynamic checks
4. Easy to integrate with existing F´ components

matt392code:
MINOR: Redundancy warnings
[...]

PR #679, opened 2021-06-07
Fix uninitialized variable warnings

saba-ja:
| | |
|:---|:---|
|**_Affected Component_**| Autocoders,  OS, Svc|
|**_Related Issue(s)_**| #678 |

---
## Change Description

This PR resolved some of uninitialized variable security warnings from CodeSonar.

## Rationale

A variables should either be initialized or have a meaningful value assign to it before being used.

Ran all Ref unit tests with no issues.

## Future Work

There are still about 40 more uninitialized variable warnings 

Screen Shot 2021-06-07 at 12 12 05 AM

due to the following lines in the autocoder:

https://github.com/nasa/fprime/blob/9a9fd99a757750b56da1a46930f48dd6cc43536b/Autocoders/Python/src/fprime_ac/generators/templates/component/cpp.tmpl#L1810

https://github.com/nasa/fprime/blob/9a9fd99a757750b56da1a46930f48dd6cc43536b/Autocoders/Python/src/fprime_ac/generators/templates/component/cpp.tmpl#L1891

https://github.com/nasa/fprime/blob/9a9fd99a757750b56da1a46930f48dd6cc43536b/Autocoders/Python/src/fprime_ac/generators/templates/component/cpp.tmpl#L1947

The autocoder does not initialize the variable and it gets passed to the `deserialize()` function ([see here for list of deserialize functions](https://github.com/nasa/fprime/blob/devel/Fw/Types/Serializable.cpp)). If the condition is not correct the `deserialize()` does not assign any value to the variable, and returns a status other than `FW_SERIALIZE_OK` which gets asserted in the next lines. Since the code will not pass the the assert no further action will be done on the uninitialized variable.

There are other instances of the similar pattern in the autocoder so it might be better to resolve this issue in the `deserialize()` or in the autocoder component template.

Please advice if you have any idea how to resolve this issue.

### DIFF
--- a/Ref/RecvBuffApp/RecvBuffComponentImpl.cpp
+++ b/Ref/RecvBuffApp/RecvBuffComponentImpl.cpp
@@ -34,16 +34,16 @@ namespace Ref {
         // reset deserialization of buffer
         buff.resetDeser();
         // deserialize packet ID
-        U32 id;
+        U32 id = 0;
         Fw::SerializeStatus stat = buff.deserialize(id);
         FW_ASSERT(stat == Fw::FW_SERIALIZE_OK,static_cast<NATIVE_INT_TYPE>(stat));
         // deserialize data
-        U8 testData[24];
+        U8 testData[24] = {0};
         NATIVE_UINT_TYPE size = sizeof(testData);
         stat = buff.deserialize(testData,size);
         FW_ASSERT(stat == Fw::FW_SERIALIZE_OK,static_cast<NATIVE_INT_TYPE>(stat));
         // deserialize checksum
-        U32 csum;
+        U32 csum = 0;
         stat = buff.deserialize(csum);
         FW_ASSERT(stat == Fw::FW_SERIALIZE_OK,static_cast<NATIVE_INT_TYPE>(stat));
         // if first packet, send event
--- a/Autocoders/Python/src/fprime_ac/generators/templates/component/cpp.tmpl
+++ b/Autocoders/Python/src/fprime_ac/generators/templates/component/cpp.tmpl
@@ -323,7 +323,7 @@ namespace ${namespace} {
 
     if (this->m_param_${prmname}_valid == Fw::PARAM_VALID) {
     #if $typeinfo == "enum":
-      FwEnumStoreType ${prmname}Val;
+      FwEnumStoreType ${prmname}Val = 0;
       stat = buff.deserialize(${prmname}Val);
       this->m_${prmname} = static_cast<$type>(${prmname}Val);
     #else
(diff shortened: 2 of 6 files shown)


Ref/RecvBuffApp/RecvBuffComponentImpl.cpp
```diff
@@ -34,16 +34,16 @@ namespace Ref {
         // reset deserialization of buffer
         buff.resetDeser();
         // deserialize packet ID
-        U32 id;
+        U32 id = 0;
         Fw::SerializeStatus stat = buff.deserialize(id);
         FW_ASSERT(stat == Fw::FW_SERIALIZE_OK,static_cast<NATIVE_INT_TYPE>(stat));
         // deserialize data
-        U8 testData[24];
+        U8 testData[24] = {0};
```
LeStarch, line 41: @saba-ja what are the C++ standard requirements for this line?  I don't think this works with all compilers, yes?

LeStarch (commented):
Just the one question!

Ref/RecvBuffApp/RecvBuffComponentImpl.cpp
Joshua-Anderson: @LeStarch given that this is supported in C99 I think we generally count on C++ support for this.

LeStarch: It did compile and run below, and the RPI compiler was the one I was worried about.

LeStarch: ...except this is the Ref app, so it didn't.  I'll check locally.

saba-ja: @LeStarch ,

Short answer:
This only works if the value is `{0}` and I believe won't cause an issue in C++98 and higher.
Please advice if there is any better way to initialize this array.

Long answer:
`{0}` sets the first value of the array to `0` as specified by the user and the rest of the elements are zero initialized by the compiler. If the value is anything other than `0` only the first value will be set and the rest of the elements will be set to zero by the compiler.

I believe brace-enclosed initialization has been supported in C lang since C99
see: https://en.cppreference.com/w/c/language/array_initialization

It is supported in C++11 standard (see section 3.9.3 part 5 in page 74 of http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf)

Did not find specification of it in C++98 but ran the following with the following and did not see any warnings:

```
g++ --std=c++98 -pedantic -Wall -Wextra -Wcast-align -Wdisabled-optimization -Winit-self -Werror test.cpp
```

test.cpp
```c++
#include <stdio.h>

int main() {
    
    int a[10] = {0};
    
   for(int i=0; i < 10; i++) {
    printf("%d ", a[i]);     
   }  
    return 0;
}
```
Output:
```
0 0 0 0 0 0 0 0 0 0
```

saba-ja: Compiled on Lionaro ARM GCC 7.5 (gcc-linaro-7.5.0-2019.12-x86_64_arm-linux-gnueabihf) with no warning.

LeStarch: I just confirmed with the RPI tools too.  We can ship this, and initialize differently if it raises issues somewhere.

LeStarch (approved):

PR #681, opened 2021-06-07
XML support for enum default values

rohandkn:
| | |
|:---|:---|
|**_Originating Project/Creator_**|F Prime|
|**_Affected Component_**|Autocoder|
|**_Affected Architectures(s)_**|enum|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|Y|
|**_Builds Without Errors (y/n)_**|Y|
|**_Unit Tests Pass (y/n)_**|Y|
|**_Documentation Included (y/n)_**|Y|

---
## Change Description

Added support for default values for enums in F Prime XML schema. Modified the enum XML schema, enum parser, and enum cpp template to initialize default values.

## Rationale

FPP has support for the default keyword for enums. So, the enum XML schema needs to support it as well.

Modified the existing enum_xml tests to check that the default value is correctly set.

All tests built and run with no issues.

## Future Work

N/A


Joshua-Anderson (commented):
Looks great in general, and adding unit tests for the new behavior is awesome.

What do you think about adding a check within the autocoder itself to make sure the whatever value the user sets to the default actually exists in the enum? I'm a little concerned that a user might accidentally supply an invalid default value and their only feedback would be a compilation error in the autogenerated code.

rohandkn:
Sounds good! I will add that check.

bocchino:
> What do you think about adding a check within the autocoder itself to make sure the whatever value the user sets to the default actually exists in the enum?

I would recommend against this. Going forward, FPP will generate the XML and will make sure all the symbols are valid. So the check is unnecessary. Also, in general the autocoder does not have these checks.

> I'm a little concerned that a user might accidentally supply an invalid default value and their only feedback would be a compilation error in the autogenerated code.

This is how the F Prime XML autocoder works in general. The fix is to use FPP.

bocchino:
Here is an example:
```
> fpp-check
enum E { A, B, C } default D
^D
fpp-check
stdin: 1.28
enum E { A, B, C } default D
                           ^
error: undefined symbol D
```

XMLLim/MathTypes/MathSerializableAi.xml
```diff
@@ -0,0 +1,16 @@
...
+        <member name="op" type="ENUM">
+            <enum name="Operation">
+                <item name="ADD"/>
+                <item name="SUB"/>
+                <item name="MULT"/>           
+                <item name="DIVIDE"/>           
+            </enum>
+        </member>
+    </members>
+</serializable>
```
jsoref, line 16: Can I encourage you to add a newline at this end of file?

rohandkn:
Sorry, that file should not have been included in the PR. I removed it.

LeStarch (approved):

LeStarch:
Looks ok to me.  Given the recommendations from @bocchino, I will merge once the CI passes.

PR #682, opened 2021-06-07
lestarch: reverting enable UTs to OFF and a deprecation warning

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**| UTs |
|**_Related Issue(s)_**|  |

---
## Change Description

Make framework UTs enabled by default.

## Rationale

Projects pointed out that framework UTs "are" UTs and thus should be run by default. Since this is not the best long-term way to handle this, we are also adding a deprecation warning.

### DIFF
--- a/ci/tests/20-fputil.bash
+++ b/ci/tests/20-fputil.bash
@@ -17,10 +17,10 @@ do
         warn_and_cont "RPI tools not installed, refusing to test."
         continue
     fi
-    # Check not RPI and Ref deployment to enable FRAMEWORK UTS
-    if [[ "${deployment}" != */RPI ]] && [[ "${deployment}" != */Ref ]]
+    # Check if RPI or Ref deployment to disable FRAMEWORK UTS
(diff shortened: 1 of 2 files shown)


Joshua-Anderson (commented):
My concern with this patch is that this makes disabling project unit tests, which aren't something most users want to run, an explicit step that users have to know about and remember to run every time while regenerating.

Nothing a user edits in their deployment should break framework unit tests. I kinda consider this akin to running all your python project dependencies unit tests in addition to your own - a little overkill for most projects, but a requirement to high reliability projects.
 
I totally understand that some projects may want to run these unit tests to verify the framework (probably in CI), but most of the projects that have those sort of requirements will likely know that they need to explicitly run the framework unit tests and setup their CI accordingly.

While I personally think this is likely not the best default behavior for F', I have no issue with being overruled here and the implementation looks good.

Joshua-Anderson (approved):

PR #686, opened 2021-06-08
lestarch: refactoring toolchains to properly set and use CMAKE_SYSTEM…

LeStarch:
…_NAME

| | |
|:---|:---|
|**_Originating Project/Creator_**| Infrastruture |
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description


Drv/LinuxGpioDriver/CMakeLists.txt
```diff
@@ -11,14 +11,14 @@ if(NOT DEFINED FPRIME_USE_STUBBED_DRIVERS OR FPRIME_USE_STUBBED_DRIVERS)
...
     )
-elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" OR ${CMAKE_SYSTEM_NAME} STREQUAL "arm-linux-gnueabihf")
+elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
     set(SOURCE_FILES
         "${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentAi.xml"
         "${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImplCommon.cpp"
         "${CMAKE_CURRENT_LIST_DIR}/LinuxGpioDriverComponentImpl.cpp"
     )
 else()
-    message(FATAL_ERROR "Cannot use ${CMAKE_CURRENT_LIST_DIR} with platform ${CMAKE_SYSTEM_NAME}. Consider using -DFPRIME_USE_STUBBED_DRIVERS=ON")
```
Joshua-Anderson, line 21: If you're attempting to compile linux drivers on a non-linux system and don't explicitly note that you want to use stubbed drivers, shouldn't generation error out? For the most part, I think that stubbed drivers should be opt-in, since using them is almost certain going to break the deployment if you try and actually run it. 

Does Ref somehow indirectly wind up pulling in the linux drivers?

Drv/LinuxSerialDriver/CMakeLists.txt
```diff
@@ -5,14 +5,14 @@
 # MOD_DEPS: (optional) module dependencies
 #
 ####
-if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" OR ${CMAKE_SYSTEM_NAME} STREQUAL "arm-linux-gnueabihf" OR ${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" OR ${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
```
Joshua-Anderson, line 8: Does the linux serial driver work on Mac OS?

cmake/platform/platform.cmake.template
```diff
@@ -63,8 +60,12 @@ set(CMAKE_CXX_FLAGS
 
 # STEP 4: Specify that a thread package should be searched in the toolchain
 #         directory. NOTE: when running without threads, remove this line.
-message(STATUS "Requiring thread library")
-FIND_PACKAGE ( Threads REQUIRED )
+#         Here there is a check for the using baremetal scheduler
+if (NOT DEFINED FPRIME_USE_BAREMETAL_SCHEDULER)
```
Joshua-Anderson, line 64: To simplify the toolchain file a bit, can we have the toolchain set `FPRIME_USE_BAREMETAL_SCHEDULE` and then have the OS platform file require Threads for that platform if neccessary?

cmake/toolchain/raspberrypi.cmake
```diff
@@ -11,13 +11,9 @@
 #
 # e.g. should the user install the tools in ``/home/user1` then the environment variable might be set using
 # `export RPI_TOOLCHAIN_DIR=/home/user/tools/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf/`
-#
-# **Note:** this toolchain file sets the system name to `arm-linux-gnueabihf` to be similar to other raspberry pi
-# toolchain files while providing a cleaner name of `raspberrypi`.  Its parallel platform file is therefore named
-# `arm-linux-gnueabihf.cmake` in the platforms directory.
 ####
 # Set system name
-set(CMAKE_SYSTEM_NAME "arm-linux-gnueabihf")
```
Joshua-Anderson, line 20: We may also want to add `set(CMAKE_SYSTEM_PROCESSOR "arm")` so that if we want to add arm specific logic in the future we can.

Drv/LinuxGpioDriver/CMakeLists.txt
LeStarch: This is a generation time issue in that these packages are included.  So although Ref does not use these drivers, it still scans them as possibilities in the make system.  Thus any non-Linux build would error on generation. 

With this change you will get the above message at generation time and a "no target found" if you try to use the above items.

cmake/platform/platform.cmake.template
LeStarch: The use of this scheduler is an OS related item, so it belongs in the platform file per our previous discussion on the new organization.

cmake/toolchain/raspberrypi.cmake
LeStarch: I think this is an adjustment we can make when we need this functionality.

Drv/LinuxSerialDriver/CMakeLists.txt
LeStarch: Yes!

cmake/platform/platform.cmake.template
Joshua-Anderson: Oh I can't read, I thought this was that toolchain template. Don't mind me 😄

Joshua-Anderson (approved):
👍

cmake/toolchain/raspberrypi.cmake
LeStarch: Changed my mind on this.  I'd like to set this.  It seems to make sense to me.

LeStarch:
Note: got an incorrect commit message for the second set of changes.  When I squash them, I will repair that messgae.

LeStarch:
@bocchino ready for your review.

bocchino (approved):
Looks good!

PR #687, opened 2021-06-08
Fix Ignored Return Value warnings

saba-ja:
| | |
|:---|:---|
|**_Affected Component_**| Svc|
|**_Related Issue(s)_**| #678 |

---
## Change Description

This PR resolved some of ignored return value reliability warnings from CodeSonar.
The proposed fixes are up to debate so please advise if there should be any changes.

## Rationale

The outputs of the following functions are not checked. If the returned value can indicate an error, the error will be ignored.

Also quote from CS output:
> The return value of snprintf() is checked or used 100% of the time in this project.
> The return value of Fw::SerializeBufferBase::serialize() is checked or used 99% of the time in this project.
> CodeSonar is configured to enforce Ignored Return Value checks for any function whose return value is checked at least 96% of the time, unless the function is used fewer than 20 times.

Built and ran Ref with no issues. All UTs passed locally.

## Future Work
Output of the following is still not checked:
https://github.com/nasa/fprime/blob/eebef10711194f131128809ef9e6331acad8d733/Os/Linux/File.cpp#L334

### DIFF
--- a/Svc/Deframer/DeframerComponentImpl.cpp
+++ b/Svc/Deframer/DeframerComponentImpl.cpp
@@ -73,7 +73,11 @@ void DeframerComponentImpl ::route(Fw::Buffer& data) {
     serial.setBuffLen(data.getSize());
     // Serialized packet type is explicitly an I32 (4 bytes)
     Fw::SerializeStatus status = serial.deserialize(packet_type);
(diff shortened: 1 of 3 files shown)


LeStarch (approved):

LeStarch:
@saba-ja I like these fixes.

PR #692, opened 2021-06-09
Fixed incorrect scheduler number in Math Component tutorial

aidan-wagner:
| | |
|:---|:---|
|**_Affected Component_**| Math Component Tutorial |
|**_Related Issue(s)_**|  |

---
## Change Description

The number for the scheduler connection was incorrect and has been fixed

## Rationale

This needed to be fixed for the Math Component tutorial to work correctly

To ensure this is correct, use it in the Math Component tutorial and make sure it works.


Joshua-Anderson (approved):
Nice catch!

Confirmed that [rate group 1 sched port 3 is in use](https://github.com/nasa/fprime/blob/devel/Ref/Top/RefTopologyAppAi.xml#L741) and that the active rate group is configured for a max of 10 ports, so port number 4 is under that max.

Issue #689, opened 2021-06-09
Cleaning up Unit Test folders


saba-ja:
| | |
|:---|:---|
|**_F´ Version_**| 1.5 |
|**_Affected Component_**| All  |
---
## Problem Description

Unit test folders currently contain old binary executable UT files, old R&D checklists, and Google test suite files, gcov files, and old make files.
This ticket will be used to track the subsequent PRs cleaning up UT folders.

## How to Reproduce
To list binary files run `find . -type f -name runtest_*`
To list gcov files run `find . -type f -name *.gcov`
etc.

## Expected Behavior
With the new fprime-util tool there is no need to keep the old executable files or include Google test suite files.

AdvaitDhingra:
Are there any more outdated ut test files to remove?

PR #694, opened 2021-06-09
Remove old GTest UT files

saba-ja:
| | |
|:---|:---|
|**_Affected Component_**|  All |
|**_Related Issue(s)_**| #689 |
|**_Builds Without Errors (y/n)_**|  Y |

---
## Change Description

This PR removes all the old GoogleTest unit test files from the UT folders.
The deleted files are one of the following:

- TesterBase.hpp
- TesterBase.cpp
- GTestBase.hpp
- GTestBase.cpp

No other file was deleted in this PR.

Note: GoogleTest files in the following two folders were left for user reference:
https://github.com/nasa/fprime/tree/devel/docs/Tutorials/MathComponent/MathReceiver/test/ut/Handcode
https://github.com/nasa/fprime/tree/devel/docs/Tutorials/MathComponent/MathSender/test/ut/Handcode

Please advise if these files also should be removed.

## Rationale

With the new fprime-util tool there is no need to keep the old GoogleTest files.

Ran `fprime-util check --all` with no issue.

## Future Work

Cleaning up gcov files will be done in separate PRs.


LeStarch:
@saba-ja you'd need to add a CMakeLists.txt entry for those Ref UTs  `set(UT_SOURCE_FILE ...Ai.xml .../test/ut/Tester...); register_fprime_ut()`  then if they pass without the handcode files you could delete them.

Have you seen anywhere where the handcode files were part of a cmake list?  We used to have some that couldn't rely on the autocoded GTest and TesterBase files....but perhaps we fixed them.  If so this is a great cleanup item.

saba-ja:
@LeStarch,

1) Ran the following greps:
grep -inr --include=CMakeLists.txt "TesterBase.hpp" .
grep -inr --include=CMakeLists.txt "TesterBase.cpp" .
grep -inr --include=CMakeLists.txt "GTestBase.hpp" .
grep -inr --include=CMakeLists.txt "GTestBase.cpp" .
 
and did not find any occurrence of these files in CMakeLists.txt files. So there should not be any dependency on the deleted GTest files.

2) Only SignalGen has a nominal UT test in Ref. I directly ran the SignalGen UT from its folder and all the tests passed.
The Ref/CMakeLists.txt is already including SignalGen as subdirectory and Ref/SignalGen/CMakeLists.txt is already including its UT_SOURCE_FILES. Running `fprime-util check --all` from Ref folder executes SignalGen UT as shown below:
Screen Shot 2021-06-10 at 1 26 13 AM

Please advise if there is still any missing information here.

LeStarch (approved):

PR #695, opened 2021-06-09
Move string copyBuff from implmentation to string base class

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Convert string copyBuff from a virtual function that was implemented by string implementations to a method in the base string class.

## Rationale

Removes code duplication of the copyBuff function and reduces the chance of introducing an error when implementing a string class. It also was triggering clang-tidy warnings because calling a virtual function in a constructor is highly discouraged due to the chance of calling an unintended implementation of the virtual function.

### DIFF
--- a/Autocoders/Python/src/fprime_ac/generators/templates/arrays/array_cpp.tmpl
+++ b/Autocoders/Python/src/fprime_ac/generators/templates/arrays/array_cpp.tmpl
@@ -1,7 +1,7 @@
 ##
 ## Template to stamp out serializable arrays .cpp file
 ##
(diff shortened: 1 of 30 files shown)


LeStarch:
@timcanham could you look at this, and speak to why this wasn't done previously?

timcanham:
This is a good change. I think we didn't really notice it because it was mostly hidden in code generation. This will save a bunch of code...

LeStarch (approved):

PR #700, opened 2021-06-10
Wrap raw task routine with an platform specific wrapper

### LINKED ISSUE #661: Os::Task::taskRoutine not compliant with POSIX Thread Entrypoint
| | |
|:---|:---|
|**_F´ Version_**| devel |
|**_Affected Component_**| Os  |
---
## Problem Description

F Prime's threads have no return value, which was likely done to provide greater compatibility across different platforms.
However, POSIX threads are expected to return a `void *`. This causes compilation errors, since returning void when a return value is expected is undefined behavior.  

We should either wrap the user provided taskRoutine with another function that returns 0 or change taskRoutine to expect a return value.

From compiling with `-Wextra`:

```
/fprime/Os/Posix/Task.cpp: In member function ‘Os::Task::TaskStatus Os::Task::start(const Fw::StringBase&, NATIVE_INT_TYPE, NATIVE_INT_TYPE, NATIVE_INT_TYPE, Os::Task::taskRoutine, void*, NATIVE_INT_TYPE)’:
/fprime/Os/Posix/Task.cpp:119:58: error: cast between incompatible function types from ‘Os::Task::taskRoutine’ {aka ‘void (*)(void*)’} to ‘pthread_func_ptr’ {aka ‘void* (*)(void*)’} [-Werror=cast-function-type]
  119 |         stat = pthread_create(tid,&att,(pthread_func_ptr)routine,arg);
```

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**| Os/Tasks |
|**_Related Issue(s)_**| fixes #661 |

---
## Change Description

Instead of calling a user provided task routine directly, call the routine from a platform specific task wrapper function.

## Rationale

Different platforms have different function definitions for task routines.

- Pthreads are `void * task(void *)`
- VxWorks is `int task(int arg1, ..., int arg10)`

F' provides a generic task definition: `void task(void *)`. Right now we're force casting this function to the OS specific function definition. However, casting a function with a void return value to functions with non-void return values is undefined behavior that should be avoided (and is also triggering static analysis warnings). A platform-specific wrapper is used as the task entrypoint, which then calls the F' task entrypoint.

## Future Work

The VxWorks task implmentation should be updated to use this new pattern.
For consistency, we may want to update the bare metal schedule to use the execRoutine function.

### DIFF
--- a/Os/Posix/Task.cpp
+++ b/Os/Posix/Task.cpp
@@ -19,14 +19,20 @@
 
 typedef void* (*pthread_func_ptr)(void*);
 
-//#define DEBUG_PRINT(x,...) Fw::Logger::logMsg(x,##__VA_ARGS__);
-#define DEBUG_PRINT(x,...)
+void* pthread_entry_wrapper(void* arg) {
+    FW_ASSERT(arg);
+    Os::Task::TaskRoutineWrapper *task = reinterpret_cast<Os::Task::TaskRoutineWrapper*>(arg);
+    FW_ASSERT(task->routine);
+    task->routine(task->arg);
+    return NULL;
+}
 
 namespace Os {
-    Task::Task() : m_handle(0), m_identifier(0), m_affinity(-1), m_started(false), m_suspendedOnPurpose(false) {
+    Task::Task() : m_handle(0), m_identifier(0), m_affinity(-1), m_started(false), m_suspendedOnPurpose(false), m_routineWrapper() {
     }
 
     Task::TaskStatus Task::start(const Fw::StringBase &name, NATIVE_INT_TYPE identifier, NATIVE_INT_TYPE priority, NATIVE_INT_TYPE stackSize, taskRoutine routine, void* arg, NATIVE_INT_TYPE cpuAffinity) {
+        FW_ASSERT(routine);
 
         this->m_name = "TP_";
         this->m_name += name;
@@ -37,7 +43,6 @@ namespace Os {
         this->m_name += pid;
 #endif
         this->m_identifier = identifier;
-
         Task::TaskStatus tStat = TASK_OK;
 
         pthread_attr_t att;
@@ -116,7 +121,8 @@ namespace Os {
         }
 
         pthread_t* tid = new pthread_t;
-        stat = pthread_create(tid,&att,(pthread_func_ptr)routine,arg);
+        this->m_routineWrapper = {.routine = routine, .arg = arg};
+        stat = pthread_create(tid,&att,pthread_entry_wrapper,&this->m_routineWrapper);
 
         switch (stat) {
             case 0:
@@ -214,7 +220,6 @@ namespace Os {
         stat = pthread_join(*((pthread_t*) this->m_handle), value_ptr);
 
         if (stat != 0) {
-            DEBUG_PRINT("join: %s\n", strerror(errno));
             return TASK_JOIN_ERROR;
         }
         else {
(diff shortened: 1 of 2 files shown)


Os/Task.hpp
```diff
@@ -48,6 +48,7 @@ namespace Os {
              * Returns the task-handle owned by this task
              */
             POINTER_CAST getRawHandle();
+            void execRoutine(); // !< used by scheduler to start executing task. Do not explicitly call.
```
LeStarch, line 51: I don't think you should define this function here.  Since it is a member function this has 2 issues:

1. Every creator of Os/Task has to define it or there will be undefined symbols.  This is the case even if you don't need this.
2. As you have pointed out, it would be dangerous to call this directly

A cleaner approach (I think) would be to put in a a Functor just in Posix/Task.cpp that might look  like this:

```c++
struct Functor {
     taskRoutine routine; //!< contains the task entrypoint
      void* arg; //!< contains the task entrypoint pointer
    
     void execute() {
          this->routine(this->arg);
    }
}
void* pthread_entry_wrapper(void* arg) {
     FW_ASSERT(arg);
     Functor* executor = (Functor*) arg;
     executor->execute();
     return NULL;
}
```
This could be created on the stack of the start call and doesn't add a bunch of members to Os::Task that other implementations won't use.

LeStarch (changes requested):

Os/Task.hpp
Joshua-Anderson: I thought about this approach, but was worried about stack allocation. Is there any guarantee that pthread_create will execute the entrypoint before the start function returns?

LeStarch: Looks like there aren't guarantees, so the above method won't work.  Is there another way to prevent adding a bunch of stuff to the Task.hpp that other implementors won't use?

Joshua-Anderson: I think using the class itself is probably necessary to have a static location to store the entrypoint and argument. I think nearly every implementer is going to need a wrapper though, so this isn't adding too much overhead. The only exception is the bare metal scheduler, but it could be structured to use this approach as well.

Joshua-Anderson:
@LeStarch updated with feedback from our discussion yesterday.

Os/Posix/Task.cpp
```diff
@@ -19,14 +19,18 @@
 
 typedef void* (*pthread_func_ptr)(void*);
 
-//#define DEBUG_PRINT(x,...) Fw::Logger::logMsg(x,##__VA_ARGS__);
-#define DEBUG_PRINT(x,...)
+void* pthread_entry_wrapper(void* arg) {
+    Os::Task::TaskRoutineWrapper *task = (Os::Task::TaskRoutineWrapper*) arg;
```
LeStarch, line 23: Probably should add an assert that arg is not NULL, and that the members aren't NULL.

LeStarch (changes requested):
Need some safety asserts, otherwise I like this approach.

@timcanham do you have thoughts on this implementation?

Os/Posix/Task.cpp
Joshua-Anderson: I added an assert that routine was non-null. Null task arguments are okay.

Joshua-Anderson: Oh I see what you mean now lol, I'll fix that

LeStarch (approved):
Looks ok to me!

PR #701, opened 2021-06-10
Remove unused const qualifiers

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

These const qualifiers were unused, triggering GCC warnings when compiling with -Wextra.

### DIFF
--- a/Fw/FilePacket/PathName.cpp
+++ b/Fw/FilePacket/PathName.cpp
@@ -1,4 +1,4 @@
-// ====================================================================== 
+// ======================================================================
 // \title  PathName.cpp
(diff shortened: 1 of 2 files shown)


LeStarch (approved):
I think this will come back to us, but if it fixes issues today then we can merge it.

Joshua-Anderson:
This fixes warnings preventing us from enabling `-Wextra`. Let's merge it for now, adding it back in later if we add const to those functions should be easy (and the compiler will warn us).

PR #758, opened 2021-06-22
Remove usages of variable length arrays

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**| |

---
## Change Description

Remove usages of variable length arrays

## Rationale

Variable length arrays are not part of the C++ standard and present a risk of
stack overflows in the case of an codeing error. Replace variable length arrays
with a statically sized array large enough to handle the maximum possible input.

### DIFF
--- a/Fw/FilePacket/PathName.cpp
+++ b/Fw/FilePacket/PathName.cpp
@@ -44,7 +44,7 @@ namespace Fw {
 
     {
       const U8* addrLeft = serialBuffer.getBuffAddrLeft();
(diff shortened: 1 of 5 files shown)


LeStarch (approved):
We should convert the "size write" of buffer logger to use our standard serialization techniques (e.g. use an Fw::ExternalBuffer" and serialize the size....but this is good for now.

PR #759, opened 2021-06-22
lestarch: making CI log archive always run

LeStarch:
| | |
|:---|:---|
|**_Originating Project/Creator_**| Infra |
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description
Log always.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -28,7 +28,7 @@ jobs:
     # Archive the outputs
     - name: 'Archive Logs'
       uses: actions/upload-artifact@v2
-      if: ${{ always() }}
+      if: always()
       with:
         name: ci-logs
         path: ci-logs.tar.gz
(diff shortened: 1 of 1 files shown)


Joshua-Anderson (approved):
🚀

.github/workflows/build-test.yml
```diff
@@ -28,6 +28,7 @@ jobs:
     # Archive the outputs
     - name: 'Archive Logs'
       uses: actions/upload-artifact@v2
+      if: ${{ always() }}
```
jsoref, line 31: Technically the `${{` and `}}` here are optional

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions

> When you use expressions in an `if` conditional, you may omit the expression syntax (`${{ }}`) because GitHub automatically evaluates the `if` conditional as an expression, unless the expression contains any operators. If the expression contains any operators, the expression must be contained within `${{ }}` to explicitly mark it for evaluation. For more information about `if` conditionals, see "Workflow syntax for GitHub Actions."

And yes, the fact that their example uses them is unfortunate.

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#example-7

PR #768, opened 2021-06-24
Cleanup string classes

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Cleanup F' string classes

## Rationale

- Moves common functionality to StringBase, preventing code duplication
- Doesn't call the getCapacity virtual function on a class while it's being constructed, preventing clang tidy warnings
- Passing size of other string to function when calling copyBuffer

## Future Work

I wan't able to cleanup the LogStringArg and TlmString classes yet - they both have setMaxSerial methods that are theoretically used for per-telm channel and per-event string truncation, but we should have a discussion about if that functionality is necessary and if there's better ways to accomplish it.

### DIFF
--- a/Fw/Types/StringUtils.cpp
+++ b/Fw/Types/StringUtils.cpp
@@ -1,7 +1,16 @@
 #include "StringUtils.hpp"
+#include <Fw/Types/Assert.hpp>
 #include "string.h"
 
 char* Fw::StringUtils::string_copy(char* destination, const char* source, U32 num) {
+    // Handle self-copy
+    if(destination == source) {
+        return destination;
+    }
+
+    // Copying an overlapping range is undefined
+    FW_ASSERT(source + num <= destination || destination + num <= source);
+
     char* returned = strncpy(destination, source, num);
     destination[num - 1] = '\0';
     return returned;
--- a/Autocoders/Python/src/fprime_ac/generators/templates/arrays/array_cpp.tmpl
+++ b/Autocoders/Python/src/fprime_ac/generators/templates/arrays/array_cpp.tmpl
@@ -24,6 +24,7 @@
 \#include <stdio.h>
 
 \#include "Fw/Types/Assert.hpp"
+\#include <Fw/Types/StringUtils.hpp>
 
 \#include "${name}ArrayAc.hpp"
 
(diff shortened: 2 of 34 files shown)


Joshua-Anderson:
LGTM is complaining that there are string class copy constructors but not copy assignment overloads. The copy assignment overload is unnecessary because the StringBase class already has a generic copy assignment overload for all strings.

Since string classes already have a copy constructor from StringBase, ideally I'd remove the class copy constructor to get rid of this warning. Unfortunately, removing this class copy constructor triggers compiler warnings that the copy constructor isn't defined, despite the fact a perfectly valid StringBase constructor exists. I'm curious if there's some way I could hint at the compiler to use the StringBase constructor as the class copy constructor as well.

LeStarch:
@Joshua-Anderson you left a comment here is this work finalized or are you still working?

Joshua-Anderson:
@LeStarch This is ready to go... I just wanted to justify the LGTM warnings and see if anybody with more C++ experience has suggestions for better ways to solve this problem.

LeStarch:
@Joshua-Anderson the copy constructor for any class will either be explicit (user supplied) or implicit (compiler supplied) as the compiler must guarantee that the class is fully constructed after the constructor is called. For this reason it cannot use the base-class defined constructor. If an implicit constructor is used, the "default" base class constructor is used.

Joshua-Anderson:
@LeStarch Following our discussion, I've added in copy assignment constructors as well as copy constructors. It seems likely that the compiler is properly calling the parent copy assignment constructor but it's unclear at best. Since C++ best practices recommend defining the copy assignment constructor when a copy constructor is defined, we may as well define the copy assignment constructors.

LeStarch (changes requested):
Discussed this with @timcanham.  We decided it would be best to avoid virtual function use altogether in the constructors....so we should perform the copy with `string_utils::string_copy` in the constructors.

Also we should clean up the base-class assignment operators as they return the wrong type.  Based on the stl::string class a copy constructor should do:

```
const T& T::operator=(const B&)
```

Joshua-Anderson:
I replaced all the calls to copyBuff with a direct call to StringCopy. @LeStarch and I came to the consensus that `T& T::operator=(const B&)` is the C++ best practice so I didn't update the assignment operator signature

Fw/Types/StringUtils.cpp
```diff
@@ -1,7 +1,16 @@
...
 #include "string.h"
 
 char* Fw::StringUtils::string_copy(char* destination, const char* source, U32 num) {
+    // Handle self-copy
+    if(destination == source) {
+        return destination;
+    }
+
+    // Copying an overlapping range is undefined
+    FW_ASSERT(source < destination || destination + num - 1 < source);
```
LeStarch, line 12: I am not sure this is a complete check.  Consider source = destination -1 and num = 16, it will quickly violate the overlapping rule.

Perhaps:
```
FW_ASSERT(source + num <= destination || distination + num <= source);
```

LeStarch (changes requested):

LeStarch:
Spelling fixed in another PR.

PR #787, opened 2021-06-27
Silence cycle slip event and added cycles between slips telemetry

metzdigital:
| | |
|:---|:---|
|**_Originating Project/Creator_**| OWLS / bcmetz |
|**_Affected Component_**| Active Rate Group |

---
## Change Description

Added ability to silence the active rate groups cycle slip WARNING_HI event. Beneficial for when bringing up a new system and working all the kinks out without a constant stream of WARNING_HIs that make tracking other events difficult and slows GDS down to a crawl. 

Added a new telemetry channel which publishes the number of cycles between slips as a way to help determine the period of these slip events.

## Rationale

See change desc.

Tested running locally on laptop and inducing cycle slips. Works as advertised.

## Future Work

None.

### DIFF
--- a/Svc/ActiveRateGroup/ActiveRateGroupImpl.cpp
+++ b/Svc/ActiveRateGroup/ActiveRateGroupImpl.cpp
@@ -26,7 +26,9 @@ namespace Svc {
             m_maxTime(0),
             m_cycleStarted(false),
             m_overrunThrottle(0),
-            m_cycleSlips(0) {
+            m_cycleSlips(0),
+            m_prevSlipCycle(0),
+            m_silenceSlipWarning(false){
         FW_ASSERT(contexts);
         FW_ASSERT(numContexts == static_cast<NATIVE_UINT_TYPE>(this->getNum_RateGroupMemberOut_OutputPorts()),numContexts,this->getNum_RateGroupMemberOut_OutputPorts());
         FW_ASSERT(FW_NUM_ARRAY_ELEMENTS(this->m_contexts) == this->getNum_RateGroupMemberOut_OutputPorts(),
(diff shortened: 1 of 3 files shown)


Svc/ActiveRateGroup/ActiveRateGroupImpl.cpp
```diff
@@ -43,6 +45,10 @@ namespace Svc {
         ActiveRateGroupComponentBase::init(queueDepth,instance);
     }
 
+    void ActiveRateGroupImpl::silenceSlipWarning() {
```
Joshua-Anderson, line 48: The way we typically configure components is by using config headers (Ex: [FileDownlink](https://github.com/nasa/fprime/blob/devel/config/FileDownlinkCfg.hpp)) However, this config is per component type, not per component instance. Are you planning on silencing slip on a per-ActiveRateGroup instance basis?

Svc/ActiveRateGroup/ActiveRateGroupImpl.cpp
```diff
@@ -83,7 +89,13 @@ namespace Svc {
         if (this->m_cycleStarted) {
             this->m_cycleSlips++;
             if (this->m_overrunThrottle < ACTIVE_RATE_GROUP_OVERRUN_THROTTLE) {
```
metzdigital, line 91: @Joshua-Anderson it appears there is already a hook in the system to silence the EVRs after a given number `ACTIVE_RATE_GROUP_OVERRUN_THROTTLE` are dispatched. For some reason with our set to 5 we still get a flood. I do think the delta telemetry channel is useful. I'm going to bang on the system some more and then update my branch.

LeStarch:
This has been abandoned.

PR #803, opened 2021-06-30
lestarch: fixing Fw::Time to allow comparisons across different time …

LeStarch:
…bases

| | |
|:---|:---|
|**_Originating Project/Creator_**| Infra |
|**_Affected Component_**| Fw::Time |
|**_Related Issue(s)_**|  |

---
## Change Description

Fw::Time doesn't assert when comparing across time bases, but returns INCOMPARABLE instead.  This allows comparing against ZERO_TIME.

## Rationale

Comparisons shouldn't crash the system. They should fail instead.

## Future Work

Look into add and subtract methods.  These are left for future work as it is less clear what to do in this case.

### DIFF
--- a/Fw/Time/test/ut/TimeTest.cpp
+++ b/Fw/Time/test/ut/TimeTest.cpp
@@ -90,7 +90,7 @@ TEST(TimeTestNominal,CopyTest) {
 }
 
 TEST(TimeTestNominal,ZeroTimeEquality) {
-    Fw::Time time(TB_DONT_CARE,1,2);
+    Fw::Time time(TB_PROC_TIME,1,2);
     ASSERT_NE(time, Fw::ZERO_TIME);
     Fw::Time time2;
     ASSERT_EQ(time2, Fw::ZERO_TIME);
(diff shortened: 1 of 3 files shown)


LeStarch:
@metzdigital this should fix your issue too.

metzdigital:
> @metzdigital this should fix your issue too.

You are my hero!

Fw/Time/test/ut/TimeTest.cpp
```diff
@@ -89,6 +89,13 @@ TEST(TimeTestNominal,CopyTest) {
 
 }
 
+TEST(TimeTestNominal,ZeroTimeEquality) {
+    Fw::Time time(TB_DONT_CARE,1,2);
```
bocchino, line 93: This code looks odd. We are using TB_DONT_CARE, but then the test says we do care. Maybe we can use TB_PROC_TIME instead? It's already used in a different test.

bocchino (commented):
Looks good, just one comment.

LeStarch:
@bocchino could you rereview?

bocchino (approved):
Looks good!

PR #808, opened 2021-07-01
Remove unnecessary usages of PRIVATE macro

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Remove unnecessary usages of PRIVATE macro in components not white box testing, particularly in the core framework (/Fw)

## Rationale

White box testing should be the exception, not the norm. Using the PRIVATE macro in particular has a few problems:

- Violates test as you fly. Changing method visibility for unit testing can trigger unexpected side affects. For example, SerialBuffer implicit copy constructors were enabled for unit tests only. Usually, they were marked private, preventing the creation of an implicit copy constructor, but using the macro enabled them while unit testing only, causing some unexpected copying of classes that shouldn't be copied.
- Enforces testing the public interface. There were unit tests in our code where we were validating the state of the class through internal variables and not testing the equivalent public getters.
- Breaks clang-format: Since clang format cannot evaluate macros, it improperly formats code that uses the PRIVATE macro.

While I would like to move away from using the PRIVATE macro in classes by default, there are cases where we do need to white box test components without observable side affects, like memory allocators. We should have a larger group discussion about whether we would like to continue using our current PRIVATE technique for testing them, switch to using friend classes, or expose the necessary observable state to properly test these components.

### DIFF
--- a/Fw/FilePacket/test/ut/Main.cpp
+++ b/Fw/FilePacket/test/ut/Main.cpp
@@ -1,5 +1,5 @@
 // ----------------------------------------------------------------------
-// Main.cpp 
+// Main.cpp
 // ----------------------------------------------------------------------
 
 #include <gtest/gtest.h>
@@ -10,33 +10,6 @@
 #include <Fw/Types/Assert.hpp>
 
 namespace Fw {
-
-  // Serialize and deserialize a file packet header
-  TEST(FilePacket, Header) {
-    const FilePacket::Header expected = {
-      FilePacket::T_DATA, // Packet type
-      10 // Sequence number
-    };
-    const U32 size = expected.bufferSize();
-    U8 bytes[size];
-    SerialBuffer serialBuffer(bytes, size);
-    {
-      const SerializeStatus status = 
-        expected.toSerialBuffer(serialBuffer);
-      FW_ASSERT(status == FW_SERIALIZE_OK);
-    }
-    FilePacket::Header actual;
-    {
-      const SerializeStatus status = 
-        actual.fromSerialBuffer(serialBuffer);
-      FW_ASSERT(status == FW_SERIALIZE_OK);
-    }
-    GTest::FilePackets::Header::compare(
-        expected, 
-        actual
-    );
-  }
-
 
   // Serialize and deserialize a start packet
   TEST(FilePacket, StartPacket) {
@@ -51,20 +24,20 @@ namespace Fw {
     Buffer buffer(bytes, size);
     SerialBuffer serialBuffer(bytes, size);
     {
-      const SerializeStatus status = 
+      const SerializeStatus status =
         expected.toBuffer(buffer);
       ASSERT_EQ(status, FW_SERIALIZE_OK);
     }
     FilePacket actual;
     {
-      const SerializeStatus status = 
+      const SerializeStatus status =
         actual.fromBuffer(buffer);
       ASSERT_EQ(status, FW_SERIALIZE_OK);
     }
     const FilePacket::StartPacket& actualStartPacket =
       actual.asStartPacket();
     GTest::FilePackets::StartPacket::compare(
-        expected, 
+        expected,
         actualStartPacket
     );
   }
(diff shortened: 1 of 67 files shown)


LeStarch:
@Joshua-Anderson not sure I agree with this one.  For better or for worse, our pattern of unit testing includes whitebox testing.  This PR implies it is a feature we support, but don't really want to support .  It will make it harder to use the whitebox testing pattern and requires more oversight over each PR to ensure that no unnecessary usages of PRIVATE have been introduced and no unnecessary PRIVATE usages have been created via an unrelated code change.

I am fully behind cleaning up the erroneous usages. e.g. using `PRIVATE` for private constructors is incorrect as this done to disable access in all cases including unit tests.  However, these definitions used to disable access to constructor inserted constructors are a rarity not a norm.

If we decide to remove WHITEBOX testing, then we'll need this PR as a stepping stone in deprecating and removing this functionality.

Fw/FilePacket/test/ut/Main.cpp
```diff
@@ -11,33 +11,6 @@
 
 namespace Fw {
 
-  // Serialize and deserialize a file packet header
```
LeStarch, line 14: Why was this test deleted?

Svc/BufferLogger/test/ut/Health.cpp
```diff
@@ -23,7 +23,7 @@ namespace Svc {
       U32 key = 42;
 
       this->invoke_to_pingIn(0, key);
-      this->dispatchAll();
```
LeStarch, line 26: This looks like a behavioral change. "all" becomes "one"

Svc/CmdSequencer/test/ut/Tester.cpp
```diff
@@ -43,13 +43,12 @@ namespace Svc {
         this->mallocator,
         BUFFER_SIZE
     );
-    this->component.preamble();
```
LeStarch, line 46: Another behavioral change.

LeStarch (changes requested):

Fw/FilePacket/test/ut/Main.cpp
Joshua-Anderson: The header test was using private variables. Since serialization of headers was already tested in the start, data, and end packet tests, I didn't think a standalone header test was particularly useful. If you'd like to keep it around I'm happy to re-write it so that it's not using any private variables.

Svc/BufferLogger/test/ut/Health.cpp
Joshua-Anderson: Yeah, this change resulted from me trying to work around private access to the QueuedComponentBase class and raw component queue access. I decided that reworking that was complicated enough that it should be it's own pull request.

This change isn't required anymore and I can revert it, but I believe it's safe enough. As you can see, the test case only queues up a single port call so I don't think multiple calls to pingOut are necessary.

Svc/CmdSequencer/test/ut/Tester.cpp
Joshua-Anderson: This was cleaning up an unnecessary private method call. The test case was unnecessarily calling preamble, since this class itself didn't have a preamble implementation.

LeStarch:
@Joshua-Anderson the decision to move this forward is to keep the "fixes" (exposed private constructors, failing to test through interface), but to maintain the `PRIVATE` macros that aren't bugs until we have an established path forward.

We'd like to see a test of the "friend class" pattern because this seems viable....but...it has been tried before and failed with the following:

        1. Would have to be done via convention or explicit fiend classes as we need to predict the friend class name in the autocoder
        2. Autocode along with implementation private members must be available to successfully do whitebox testing
        3. “Doable but annoying” — from previous developers who attempted this.

Recommendation: implement a version of the friend pattern whitebox testing in Svc for review.

I will close this PR but hope to see a reduced version with the discovered fixes.

PR #811, opened 2021-07-07
Remove void from parameterless functions

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

F' inconsistently marks some parameterless functions as taking a void parameter (Ex. `function(void)`).

## Rationale

For consistency, parameterless functions should always include the void argument or never include it.
The C++ best practice is not to include the void parameter, so contributions to F' often don't follow this style, leading to an inconsistent coding style. Removing the void argument better matches C++ best practices and brings F' style more in line with the rest of the C++ community.

According to @timcanham:

```
The void was left over from my Rover software days where there was a style guide to include void if there were no arguments, so I did it out of habit.
 
It was to avoid a gnu-ism where it was legal to do this:
 
foo.h:
 
somefunc();
 
foo.c:
 
int somefunc(int x, char* y);
 
bar.c:
 
#include <foo.h>
 
somefunc(“hello”,3.6); // !legal!
 
I think C++ is way more type-safe these days, so you can remove them if that is a standard C++ coding style.
```


LeStarch (commented):
I assume that the result of these changes should have no functional affect and is merely cosmetic w.r.t. the code?

Joshua-Anderson:
Correct, the purely cosmetic / code style modification only.

LeStarch (approved):
Did my best to review through this.  Ref code generated is identical....but in the future we should break large PRs like this into parts as it is nearly impossible to review.

bocchino:
Note: `(void)` is still required in C code, so we need to be careful. The latest version of clang complained about one of these changes, so I reverted it here: https://github.com/nasa/fprime/pull/1952.

I agree that for C++ code `()` is better than `(void)`.

PR #818, opened 2021-07-12
add offset option for rate groups allowing different trigger patterns

csmith608:
| | |
|:---|:---|
|**_Originating Project/Creator_**| LFPS |
|**_Affected Component_**| RateGroupDriver  |
|**_Related Issue(s)_**|  |

---
## Change Description

This PR allows deployments to specify offsets to stagger RG firings, this is useful when a thread scheduler is not present

## Rationale

This was used on the LFPS firmware and cut our rate group times almost in half preventing major rate group slipping.  The image below shows an example of how the current rate group would fire off vs how the proposed change could be used to fire them off.   
Rate Group Offset Visualization

## Future Work

Testing on the Ref/RPI deployments as well as notifying users that they will have to update their rate group driver declarations to account for the new argument.

### DIFF
--- a/RPI/Top/Topology.cpp
+++ b/RPI/Top/Topology.cpp
@@ -22,7 +22,8 @@ Svc::FprimeFraming framing;
 // Rate Group Dividers for 10Hz and 1Hz
 
 static NATIVE_INT_TYPE rgDivs[] = {1,10,0};
-Svc::RateGroupDriverImpl rateGroupDriverComp("RGDRV",rgDivs,FW_NUM_ARRAY_ELEMENTS(rgDivs));
+static NATIVE_INT_TYPE rgOffs[] = {0,0,0};
+Svc::RateGroupDriverImpl rateGroupDriverComp("RGDRV",rgDivs,rgOffs,FW_NUM_ARRAY_ELEMENTS(rgDivs));
 
 // Context array variables are passed to rate group members if needed to distinguish one call from another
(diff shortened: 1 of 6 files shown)


LeStarch:
@csmith608 we need to remove "tutorial support" from PR submissions.  It should run on `devel` when a PR is merged to warn us of an incoming change.

LeStarch:
@csmith608 can you rebase on devel?  We fixed running tutorial support (I think) on devel....but for some reason it is stuck here.

LeStarch (changes requested):
First off, this is great!  Thank you for doing this work.

We have a number of changes we hope to get into F´ and haven't done since we are afraid of breaking code others are running.....but....since you are breaking the code already we were wondering if you could make a few minor changes:

1. Pass in an array of structs {.divisor=X, .offset=Y} so that we don't define so many lose arrays
2. Move this input to a `configure` method rather than the constructor.  We can automate a lot if F´ constructors are empty, so providing the inputs at a separate configure step paves the way to automation.
3. In step 2 make sure to assert that it was configured before it received a port call (with the other asserts)

As I stated, feel free to reach out if I can assist in these change!

LeStarch:
@csmith608 the build/CI issues are fixed now.

LeStarch:
We are on the verge of bringing in `v3.0.0` into devel. This PR hasn't been updated, so I am going to close it. The changes are here should we need them in the future.

PR #821, opened 2021-07-14
Dev/abcouwer/cmake baremetal fixes

nabcouwer:
| | |
|:---|:---|
|**_Originating Project/Creator_**| JPL COLDArm / Neil Abcouwer |
|**_Affected Component_**| Various, ComLogger, FileDownlink, BlockDriver |
|**_Affected Architectures(s)_**| Baremetal |
|**_Builds Without Errors (y/n)_**| y (Ref and our own baremetal build) |

---
## Change Description

- Added MIN and MAX size #defs for integers in Basic types. 
- Use U32_MAX from basic types in FileDownlink, instead of __UINT32_MAX__, which was missing in our limits.h.  
- Change "interrupt" to "interrupt_val" in BlockDriver (prevented build, "interrupt" must be a special name). 
- Guard on system type and don't include Ip socket-using components in baremetal build
- Guard on system type and don't include LinuxTime in baremetal build. 
- Add Baremetal/Mutex.cpp to Os/CMakeLists.txt. 
- Switchup fatal handler files in CMakeLists based on system type. 
- Allow for undefined PATH_MAX in ComLogger. 
- Change all instances of FPRIME_USE_BAREMETAL_SCHEDULE to FPRIME_USE_BAREMETAL_SCHEDULER.

## Rationale

We're making a baremetal, out of source build, and we were having to make some small tweaks, comment out some components in CMakeLists, to get FPrime components building when FPRIME_USE_BAREMETAL_SCHEDULER was set. These changes add some guards in CMakeLists files and make some very small component tweaks to allow building a baremetal build on a TI R5 processor.

Some of the ways that cmakes were guarded are subject to personal preference, Definitely up for making adjustments.

## Future Work

None at this time.


LeStarch (approved):
Looks good to me.  We may need to take an action to make the check for including Ip stuff less restrictive (e.g. other posix oses) but this is a fine solution for now.

PR #822, opened 2021-07-14
In FprimeProtocol.cpp:deframe(), make buffer no larger than requested…

nabcouwer:
| | |
|:---|:---|
|**_Originating Project/Creator_**| JPL COLDArm / Neil Abcouwer|
|**_Affected Component_**| FramingProtocol |
|**_Affected Architectures(s)_**| n/a |
|**_Has Unit Tests (y/n)_**| y (no change to unit tests for this change, but unit tests still run successfully) |

---
## Change Description

When Svc/FramingProtocol/FprimeProtocol.cpp:deframe() calls m_interface->allocate(), set the size of the returned buffer to be the requested size.

## Rationale

When Svc/FramingProtocol/FprimeProtocol.cpp:deframe() calls m_interface->allocate(), some allocators (StaticMemory) can return buffers that are larger than requested. In the case of uplink deframing, this size disparity makes it so commands can't be deframed successfully.

Alternatively, if returning a larger buffer than requested is bad, StaticMemory component could be changed instead. But should probably make allocation requests robust to bigger buffers due to alignment issues, etc.

Worked properly in our deployment with a deframer and static memory component, but wouldn't have an effect on Reference deployment.

## Future Work

None

### DIFF
--- a/Svc/FramingProtocol/FprimeProtocol.cpp
+++ b/Svc/FramingProtocol/FprimeProtocol.cpp
@@ -109,6 +109,10 @@ DeframingProtocol::DeframingStatus FprimeDeframing::deframe(Types::CircularBuffe
         return DeframingProtocol::DEFRAMING_INVALID_CHECKSUM;
     }
     Fw::Buffer buffer = m_interface->allocate(size);
+    // some allocators may return buffers larger than requested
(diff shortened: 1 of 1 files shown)


LeStarch (approved):
Good catch!

Issue #830, opened 2021-07-15
Broken links 404 in several .md files


IvanReznikov:
1. https://github.com/nasa/fprime/blob/e165c914b0d0240eba8164c52f8856ce90bee3dc/docs/UsersGuide/gds/seqgen.md
link simple_sequence.seq is broken

2. https://github.com/nasa/fprime/blob/e165c914b0d0240eba8164c52f8856ce90bee3dc/docs/UsersGuide/cmake/cmake-platforms.md link fprime Platform Template is broken

3. https://github.com/nasa/fprime/blob/e165c914b0d0240eba8164c52f8856ce90bee3dc/docs/UsersGuide/cmake/cmake-toolchains.md link fprime Platform Template is broken

PR #831, opened 2021-07-15
bugfix-830

IvanReznikov:
Bugfix for https://github.com/nasa/fprime/issues/830

Noticed that some of the URLs link to wrong branches.
devel <----> e165c914b0d0240eba8164c52f8856ce90bee3dc


LeStarch (approved):
Made a few minor changes to point to the newest docs.  Thanks for fixing these!

PR #834, opened 2021-07-15
FPP Ref Topology Model

bocchino:
| | |
|:---|:---|
|**_Affected Architectures(s)_**| Ref Topology |

---
## Change Description

This PR adds an FPP model for the Ref topology. It uses fpp-to-xml to generate the RefTopologyAppAi.xml file. It uses fpp-to-cpp to replace the old handwritten Components.hpp and Topology.cpp.

Incidental changes to F Prime:

* Revise the topology XML autocoder to accept hexadecimal base IDs.
* Add a TlmGet port in Fw/Tlm, so that TlmChan.tlmRecv and Tlm.tlmGet have different types (Tlm and TlmGet, respectively). When these two ports have the same type, the FPP connection resolver can't tell which one is receiving telemetry. As of this PR, Fw::TlmGet is structurally identical to Fw::Tlm, but @rohandkn is working on improvements that will give these ports different types (TlmGet will use a mutable reference and Tlm will use a const reference).
* Add a command response input port to the Deframer, to support FPP port matching. The handler implementation is a no-op.

## Rationale

Continuing to model core F Prime in FPP.

None.

## Future Work

* Make the type names and header file names of the component implementations uniform. This will allow us to remove a lot of boilerplate in the Ref topology initializers.
* Factor the Ref topology into two topologies: (1) the generic command and data handling part and (2) the part specific to the Ref example.


Ref/Top/RefTopologyAc.cpp
```diff
@@ -0,0 +1,1336 @@
...
+          ConfigObjects::health::pingEntries,
+          FW_NUM_ARRAY_ELEMENTS(ConfigObjects::health::pingEntries),
+          ConfigConstants::health::WATCHDOG_CODE
+      );
+      uplink.setup(ConfigObjects::uplink::deframing);
+    }
+
+    // Set component base Ids
+    void setBaseIds() {
+      blockDrv.setIdBase(0x100);
```
bocchino, line 309: @LeStarch I just realized that this code should use the symbolic base IDs, instead of the numeric ones. OK to push this change?

LeStarch: @bocchino sure.

bocchino: Thanks, done.

Ref/Top/RefTopologyAc.cpp
```diff
@@ -0,0 +1,1336 @@
+// ======================================================================
```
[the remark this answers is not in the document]
bocchino: Yes, these file are auto-generated. However, if we remove them from the repo, then F Prime won't build without (1) an FPP tool installation and (2) a redo installation. Currently neither (1) nor (2) is required. I have been assuming that we will never require (2) and will require (1) only after we have CMake support. Among other things, that keeps CI working while we are in this interim state.

BTW the same thing goes for all the Ai.xml files that are generated from the FPP model and currently checked in to release/v3.0.0. Because they are checked in, the standard F Prime build process (as of 2.x) works on this repo. To remove them, we will need to amend the build process and CI to run the FPP tools.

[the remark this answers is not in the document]
LeStarch: @bocchino got it. I needed to understand that this effects .cpp files as well.  Thanks!

LeStarch (approved):

PR #865, opened 2021-07-23
Use exceptionless variant of new to properly handle allocation errors

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**| various |
|**_Related Issue(s)_**|  |

---
## Change Description

Switch F' from using `new` to `new (std::nothrow)`

## Rationale

`new` can throw an exception if allocation fails. When an exception is thrown but not handled it causes the current task to immediately halt. Switch to `new (std::nothrow)`, which returns  NULL on allocation failure.

I didn't update tests using `new` since for unit tests halting on memory allocation is the desired behavior.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -1042,6 +1042,7 @@ nostdlib
 notask
 notchars
 NOTFOUND
+nothrow
 NOTYPE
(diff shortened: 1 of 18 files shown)


LeStarch (approved):
Looks good!

PR #866, opened 2021-07-23
Pass pointer rather than array to rategroup

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**| RateGroup and ActiveRateGroup |
|**_Related Issue(s)_**|  |
|**_Builds Without Errors (y/n)_**| y  |

---
## Change Description

Removes linter warning by switching from passing a raw array to passing a point to rategroup drivers.
Switch make rategroup constructor consistent between header and definition - this is technically a breaking change because I switched the passive rategroup from taking a list of I32/NATIVE_INT_TYPE to taking a list of NATIVE_UINT_TYPE.

### DIFF
--- a/RPI/Top/Topology.cpp
+++ b/RPI/Top/Topology.cpp
@@ -21,7 +21,7 @@ Svc::FprimeFraming framing;
 
 // Rate Group Dividers for 10Hz and 1Hz
 
-static NATIVE_INT_TYPE rgDivs[] = {1,10,0};
+static NATIVE_UINT_TYPE rgDivs[] = {1,10,0};
 Svc::RateGroupDriverImpl rateGroupDriverComp("RGDRV",rgDivs,FW_NUM_ARRAY_ELEMENTS(rgDivs));
 
 // Context array variables are passed to rate group members if needed to distinguish one call from another
--- a/Ref/Top/RefTopologyAc.cpp
+++ b/Ref/Top/RefTopologyAc.cpp
@@ -108,7 +108,7 @@ namespace Ref {
       }
 
       namespace rateGroupDriverComp {
-        NATIVE_INT_TYPE rgDivs[Svc::RateGroupDriverImpl::DIVIDER_SIZE] = { 1, 2, 4 };
+        NATIVE_UINT_TYPE rgDivs[Svc::RateGroupDriverImpl::DIVIDER_SIZE] = { 1, 2, 4 };
       }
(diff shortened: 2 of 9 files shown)


LeStarch:
@Joshua-Anderson we are working on a similar change to the rate group in #818....so I am hesitant to merge this as it will conflict. Essentially we want to pass in an array of structs containing each divisor (and other configuration).

We should correct these errors there so as to reduce devel / v3.0.0 conflicts.

If you concur please CLOSE this PR.

Joshua-Anderson:
It looks like #818 doesn't correct the linter error that I'm trying to correct here, which is switching the function definitions to pass pointers rather than arrays. @csmith608 would you mind pulling those fixes? If you pull in those fixes this PR can be obsoleted and closed.

LeStarch:
@Joshua-Anderson the work I was referring to is part of the requested changes to #818 and hasn't been completed yet.  Specifically, the request was to make the input to RateGroup an array of configuration `struct`s each defining a the divisor and the offset....rather than an array of raw numbers.

Also I am not sure changing the arrays to pointers is really what we are after here.  The linter warning suggests using a wrapper class (array and size as members) as opposed to a raw array.  Switching to pointers may remove the arrays-decay confusion and fix the linter error, but it doesn't really fix the underlying problem that arrays are messy.

It does make the interface harder to understand for an average user as now there is type ambiguity in the function prototype....is the pointer intended to be a pointer to a single item (aka used as a reference) or is it intended to be an array of items that start at that pointer.   Here we are trading "user confusion" against possible type confusion by the original developer.  I am not sure that trade is worth it.

Joshua-Anderson:
> is the pointer intended to be a pointer to a single item (aka used as a reference) or is it intended to be an array of items that start at that pointer. 

I feel that's a problem with C/C++ as a whole. :joy: We explicitly provide a length argument, so by inference people should understand that it's a list. I've never seen arrays be used as a method parameter before, but you're right it's mostly a developer, not user risk.

If #818 redoes the interface and gets rid of the array parameters through other changes that okay with me too

csmith608:
Yeah, sorry, I've been really busy so I haven't had much of a chance to work on the requested changes

LeStarch:
We are on the verge of bringing in `v3.0.0` into devel.  This PR hasn't been updated, so I am going to close it.  The changes are here should we need them in the future.

PR #883, opened 2021-08-02
Fix #873, allow nullptr as an input with buffer constructor with an a…

### LINKED ISSUE #873: Fw::Buffer Should Allow `nullptr` as an input
| | |
|:---|:---|
|**_F´ Version_**| devel |
|**_Affected Component_**| Fw::Buffer  |
---
## Problem Description

Fw::Buffer does not allow null pointers as input.  Thus it is not possible to send an empty/blank buffer:

```
Fw::Buffer my_buffer; // Works, nullptr and 0 length
Fw::Buffer my_buffer(nullptr, 0); //Trips an assert
```

This prevents things like SPI write without read.

zachar1a:
| | |
|:---|:---|

|**_Affected Component_**|  |
Fw::Buffer
|**_Related Issue(s)_**|  |
#873 
y
y
y
n

---
## Description

Allows data=nullptr and size=0 in Fw::Buffer

## Rationale
#873

I added an assert inside the Buffer ut

### DIFF
--- a/Fw/Buffer/Buffer.cpp
+++ b/Fw/Buffer/Buffer.cpp
@@ -35,11 +35,15 @@ Buffer::Buffer(const Buffer& src) : Serializable(),
 {}
 
 Buffer::Buffer(U8* data, U32 size, U32 context) : Serializable(),
-    m_serialize_repr(data, size),
+    m_serialize_repr(),
     m_bufferData(data),
     m_size(size),
     m_context(context)
-{}
+{
+    if(m_bufferData != NULL){
+        this->m_serialize_repr.setExtBuffer(m_bufferData, m_size);
+    }
+}
 
 Buffer& Buffer::operator=(const Buffer& src) {
     // Ward against self-assignment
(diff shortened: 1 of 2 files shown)


Fw/Buffer/Buffer.cpp
```diff
@@ -35,11 +35,19 @@ Buffer::Buffer(const Buffer& src) : Serializable(),
 {}
 
 Buffer::Buffer(U8* data, U32 size, U32 context) : Serializable(),
+#if  data == nullptr && size == 0
```
LeStarch, line 38: I don't think you can you `#if` in this way (to access realtime data) and if you can, it would violate the C++ coding practice of keeping compiler directives simple.

I think a cleaner fix would be to always use `m_serialize_repr()` in the initializer list, and then add the following to the constructor code block:

```
{
     if (m_bufferData != NULL) {
        this->m_serialize_repr.setExtBuffer(m_bufferData, m_size);   
    }
}
```

-- or --

```
{
    this->setData(m_bufferData);
}
```

LeStarch (changes requested):
Minor implementation change.

UT looks good!

Fw/Buffer/Buffer.cpp
LeStarch: Looks fixed.  I am letting this run through CI

LeStarch (approved):

PR #885, opened 2021-08-02
Merge devel into release/v3.0.0

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Merge changed from devel branch back into the release/3.0 branch.
Resolves ~20 merge conflicts.


Joshua-Anderson:
@jsoref It looks like the spellchecker already has Netscape and OMG in the expect.txt dictionary. Do you know why the spellcheck is struggling with the contractions? Ex: I would expect `Netscape` to match `Netscape's`

jsoref:
@Joshua-Anderson: In [0.0.17](https://github.com/check-spelling/check-spelling/releases/tag/0.0.17-alpha) when I dealt with `potatos`, I made it generally stricter and figured I'd need more time to figure out what I could safely relax.

The potato 🥔  case is a clear bug. I think there's currently a bug involving `we'd` which will be fixed in the next release (which will probably be `v0.1.0` instead of `v0.0.20`).

Technically there are some really odd rough edges involving possessives. Namely [biblical / Greek names](http://jlong1.sites.luc.edu/apostrophes.htm) end up being `Moses'`/`Jesus'`/`Socrates'` instead of `Moses's`/`Jesus's`/`Socrates's`.

It's possible to add a `patterns.txt` entry which would cause `'s\b` to be ignored generally if you don't care about them.

Joshua-Anderson:
Ahhh I hate english sometimes. So if I'm understanding correctly, because of the edge cases around correctly detecting contractions, in the latest release we're now required to explicitly allow the contractionized version of the word as well? If that's the expected behavior that's totally fine I just wanted to make sure it was expected.

jsoref:
Yep. Note that these are technically possessives and not contractions, but since it's a naïve spell-checker and not a grammar checker, it doesn't know the difference :-).

Ain't English fun? (I should say [phun](https://umdphysics.umd.edu/events/aboutus-outreach/physics-is-phun.html) since this is a Greek thing to go with phonics and fprime kinda relates to physics.)

I'm open to collecting strategies for handling this in a different manner, but the current `reject.txt` couldn't handle this case, so I'd need an extra file, and that's just sad. The other approach would be to enable people to compose dictionary files in something like `aspell` or `hunspell`, but the complexity of that is really too much.

With the next release, [Area dictionaries](https://github.com/check-spelling/check-spelling/wiki/Feature:-Area-dictionaries) would allow me/someone to at least maintain a list of common companies with possessives so people wouldn't have to maintain the lists in each repository.

Joshua-Anderson:
Thanks for the help @jsoref! I updated our dictionary to include the warned possessives.

jsoref:
Sigh. I wonder what you ran into. I do know that there are some annoying edges as noted for possessives (it looks like I haven't actually switched to that rule, but I might at some point, folding turns out to be a mess in the real world), but I don't think that explains what you hit. Sometimes it's an oscillation between lowercase and uppercase.

If someone writes `Microsoft`, it's possible they don't want to write `microsoft` (since that's a spelling error), so the tool will suggest adding `Microsoft` to `expect.txt`. But if `microsoft` is present in a path, it's written in lowercase, and thus an exception needs to be made, and then the tool for compression reasons (which I may abandon, since oscillation is moderately annoying) encourages removing `Microsoft` in favor of `microsoft` (which it will use to tolerate `Microsoft` and `MICROSOFT`). To discourage that, if the exception no longer applies, the tool will nudge you back to uppercase if it no longer finds a lowercase flavor that matches -- at that point, it will also probably not fold in a `Microsoft's`, since it folds based on what it sees, but it should favor `Microsoft`...

That said, it looks like you've actually hit a bug where it's just doing something totally wrong. I've run out of time for my weekend, but I'll definitely give it a look. (You could also try a branch w/ `@prerelease` instead of the pinned version...)

Joshua-Anderson:
@jsoref No worries! It looks like I got it to pass for now by including both the normal and possessive forms of the words. I didn't realize just how much nuance is in these edge cases.

PR #886, opened 2021-08-02
Standardize component interfaces used in the Ref model

bocchino:
| | |
|:---|:---|
|**_Originating Project/Creator_**| F Prime Infrastructure |
|**_Affected Architectures(s)_**| Ref |

---
## Change Description

This PR standardizes the header file names, type names, and init functions of the components used in the Ref topology. The strategy is as follows:

* Where a component name and/or header file name is nonstandard, provide a "standardization header" with a standard name that defines the standard component name. The standardization header includes the primary header. The standard component name is a type alias for the primary component name. **Example:** Header file `Foo/BarComponentImpl.hpp` defines a component class type `BarImpl`. The standardization header `Foo/Bar.hpp` includes `Foo/BarComponentImpl.hpp` and defines `Bar` as a type alias for `BarImpl`.
* Where a component init function lacks an `instanceId` argument, add an optional `instanceId` argument with default value zero.

## Rationale

These changes standardize the interfaces for the components used in the Ref topology. They enable more of the default FPP code generation to work, reducing the amount of manual C++ code in the Ref topology.

These changes provide an alternative to the existing component interfaces. They preserve backwards compatibility for code that depends on the existing interfaces.

None.

## Future Work

Standardize the rest of the core F Prime components.

Fully rename the nonstandard header files and component class names. This change would break backwards compatibility for code that depends on these elements.


Joshua-Anderson:
Is there a reason we're taking the intermediate step of aliasing the componnent names instead of just renaming everything right away?

It seems like it might be nice to get all the renaming over with at once unless there's stuff we can't rename right now.

bocchino:
This way preserves backwards compatibility, so that users have time to switch over to the new names.

Joshua-Anderson:
How long would we want to leave backwards compatibility for?

Would you want to wait for a point release after the 3.0 release?

bocchino:
What if we leave it in long enough for people to switch over to FPP topologies? Then they can let FPP take care of making the changes for them. They won't have to edit the handwritten topology C++ that uses the old symbols.

LeStarch (approved):

PR #893, opened 2021-08-04
Replace usage of 80 Char String with FixLengthString

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Replace 80 character string usage in the F' core framework with FixedLengthString, which allows for compile time adjustment of max string size. Unit tests still using 80 characters strings will be updated in a separate pull request, since unit tests probably should avoid using FixLengthString where possible to avoid potential compilation errors when users adjust the FixLengthString length.


timcanham:
Would it be worth putting in a typedef to not break existing code?

Joshua-Anderson:
@timcanham We won't be getting rid of 80 characters strings right now, just limiting their usage in the core framework. Most methods using string parameters take the generic StringBase argument, so this migration should be mostly transparent to 80 character string users.

bocchino:
@Joshua-Anderson I regenerated the C++, ran git diff, and got this:
```
diff --git a/Ref/Top/RefTopologyAc.cpp b/Ref/Top/RefTopologyAc.cpp
index 56099baa5..f5ca64ae5 100644
--- a/Ref/Top/RefTopologyAc.cpp
+++ b/Ref/Top/RefTopologyAc.cpp
@@ -1212,7 +1212,7 @@ namespace Ref {
       );
       // Initialize socket server if and only if there is a valid specification
       if (state.hostName != NULL && state.portNumber != 0) {
-          Fw::EightyCharString name("ReceiveTask");
+          Os::TaskString name("ReceiveTask");
           // Uplink is configured for receive so a socket task is started
           comm.configure(state.hostName, state.portNumber);
           comm.startSocketTask(
```
Looks like there is one line to fix.

In the future, instead of revising the Ac file manually, you can let FPP do it for you. Just install this tool:

https://github.com/bocchino/redo

And then install FPP:

https://github.com/fprime-community/fpp/tree/main/compiler

And then run `redo` at the top level. It will regenerate everything.

Eventually FPP will be integrated with CMake, and the Ac file won't be checked in.

bocchino:
Can we rename `Fw::FixedLengthString` to `Fw::String`? We are writing FSW; fixed length is the assumption. If and when there is a variable length string, we can call it `VariableLengthString`. It will be easier to read and save typing.

Joshua-Anderson:
@bocchino Thanks for the help with FPP! I'll give it a whirl and try it out next time I wind up touching FPP files.

> Can we rename Fw::FixedLengthString to Fw::String? We are writing FSW; fixed length is the assumption. If and when there is a variable length string, we can call it VariableLengthString. It will be easier to read and save typing.

It's true that in FSW that fixed length strings are the assumption, but for most programers a fixed length string is a foreign concept 😄 . I'd be a little concerned that somebody unfamiliar with F' would try and use something call `Fw::String` it as a variable length string and get surprised by the truncation.

That being said I like saving typing and I don't have a very strong opinion. @LeStarch @timcanham what do you think?

bocchino:
> I'd be a little concerned that somebody unfamiliar with F' would try and use something call Fw::String it as a variable length string and get surprised by the truncation.

It's true that someone might try to write embedded C++ as if it is Python or desktop C++. But that is not likely to turn out well, no matter how we name our variables. I think the right way to guard against this problem is through tutorials, etc.

LeStarch:
My concern is that I'd assume `Fw::String` is a base for all other strings....and it would not be. It is currently a peer of other string types, and `Fw::StringBase` is the base.

I am not deeply opposed to the rename, but it would introduce confusion for a developer like me.

bocchino:
> My concern is that I'd assume Fw::String is a base for all other strings.

OK, but I'm not sure how adding the prefix `FixedLength` helps overall. Every derived class of StringBase is stored in a fixed-size buffer. But we don't say `FixedLengthLogStringArg`, etc. So the naming is inconsistent.

Also: Note that `FixedLengthString` is not, in fact, a fixed-length string. It is a string of variable length, stored in a buffer of fixed size. For example, running this:
```
#include <stdio.h>
#include "Fw/Types/FixedLengthString.hpp"

int main(int argc, char **argv) {
  Fw::FixedLengthString s = "foo";
  printf("length of s=%d\n", s.length());
  s = "foobar";
  printf("now length of s=%d\n", s.length());
  return 0;
}
```
prints out this:
```
length of s=3
now length of s=6
```
Is this not confusing? :^)

Maybe `FixedMaxLengthString` would be accurate...but that's even uglier...

bocchino:
> Every derived class of StringBase is stored in a fixed-size buffer. But we don't say FixedLengthLogStringArg, etc. So the naming is inconsistent.

And if we're imagining people being misled in various ways...doesn't this naming imply that `LogStringArg` is _not_ a "fixed length string"?

bocchino:
> Maybe FixedMaxLengthString would be accurate...but that's even uglier...

`FixedCapacityString` would also be accurate (consistent with the StringBase API). But again, any string that inherits from `StringBase` has this property.

LeStarch:
@Joshua-Anderson lets go with Fw::String.  Nice and simple.

Joshua-Anderson:
Fw::FixedLengthString has been renamed to Fw::String.

LeStarch (approved):
Looks like what I expect.

PR #897, opened 2021-08-05
Don't overwrite CXX flags when settings the default flags

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

This accidentally clobbering user provided C/CXX flags that were provided before options was imported - affects root level framework tests that set additional cflags.


LeStarch (approved):
We could probably drop -DPROTECTED=public -DPRIVATE=public from C flags as these aren't valid constructs in C

PR #904, opened 2021-08-10
Remove usage of GCC switch range extension

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  CmdSeqencer |
|**_Related Issue(s)_**|  |
|**_Unit Tests Pass (y/n)_**|  y |

---
## Change Description

Using a switch with a range case statement is a GCC extension and causes errors when compiling in pedantic mode.

### DIFF
--- a/Svc/CmdSequencer/FPrimeSequence.cpp
+++ b/Svc/CmdSequencer/FPrimeSequence.cpp
@@ -358,19 +358,18 @@ namespace Svc {
   {
     Fw::SerializeBufferBase& buffer = this->m_buffer;
     U8 descEntry;
+
     Fw::SerializeStatus status = buffer.deserialize(descEntry);
-    if (status == Fw::FW_SERIALIZE_OK) {
-      switch (descEntry) {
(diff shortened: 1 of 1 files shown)


LeStarch:
@bocchino you have more experience than I do with the command sequencer.  Want to take a look?

bocchino (commented):
Looks good. Just a couple of minor comments.

I think the comparison `descEntry < Sequence::Record::ABSOLUTE` in line 366 is not necessary (comparison of U8 value for less than zero). I guess it doesn't hurt to leave it in, but don't some compilers warn about unnecessary comparisons? 

The coding style has changed from "single return" to "multiple return." I prefer single return but won't insist on it. If we're doing multiple return, then I suggest limiting the scope of `status` to lines 361-364 and explicitly returning `OK` in line 371. If we passed line 364, then the return status must be `OK` -- but you have to stare at the code a bit to see that.

Joshua-Anderson:
@bocchino Thanks for the review. PR updated with your feedback

LeStarch (approved):

LeStarch:
Confirmed the requested changes were made.

PR #915, opened 2021-08-12
Run cmake tests in CI

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Run the cmake test cases as a github actions parallel job.


LeStarch:
@Joshua-Anderson given we are essentially rewriting the CMake system, I am hesitant to enable this now.  I don't see the benefit of this given it will likely only trap the v3.0.0 stuff and need to be reworked at that time.

Thoughts?

Since it is working and fairly fast, I think it would be reasonable to merge, but I can tell you it is already broken with the v3.0.0 items, and these tests will need to change not the other way around.

LeStarch:
Rethought, the cmake should have tests, and those tests should pass. All this does is make sure the tests are delivered.

LeStarch (approved):

PR #924, opened 2021-08-16
Switch CI to run in VM instead of docker container

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Instead of running CI steps inside of a docker container, run in the github provided 18.04 VM.

## Rationale

CI was the primary user of the fprime docker container and releasing new docker containers was a maintenance burden.
Since we were testing in ubuntu 18.04 anyway the docker container wasn't providing value.

## Future Work

This will enable the ability for us to more easily test against ubuntu 20.04 in the future without needing to release more docker containers.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -12,18 +12,13 @@ on:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
-  # This workflow contains 4 jobs. One for every CI test."
   Framework:
-    # The type of runner that the job will run on
     runs-on: ubuntu-latest
-    container:
-      image: nasafprime/fprime-base:latest
-
-    # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
     - uses: actions/checkout@v2
-    # Runs the CI action internal to the repository
+    - name: Install fprime tools
+      run: pip3 install fprime-tools fprime-gds
     - name: F prime CI step
       run: ./ci/tests/Framework.bash
     # Archive the outputs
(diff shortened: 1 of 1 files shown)


.github/actions/setup-dependencies.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+source /etc/lsb-release
+
+# TODO: Add setup for Ubuntu 20.04
```
LeStarch, line 4: Shouldn't the packages be the same?  I have never known the packages to differ.

LeStarch: Also shouldn't we check for non-ubuntu systems like redhat?

Joshua-Anderson: > Shouldn't the packages be the same? I have never known the packages to differ.

My thought was to give the ability to install extra packages (i.e. clang analysis tools) for 20.04 that we won't be running on 18.04 for the quality check. On the other hand, I could just add that as a step in the quality job so maybe that's overkill.

> Also shouldn't we check for non-ubuntu systems like redhat?

Nope, I only intend for this to be used to setup github actions and I don't think we're planning on testing on redhat lineage systems anytime soon.

LeStarch: @Joshua-Anderson let us add that as an extra step in the quality job.

LeStarch: Could you go ahead and just make this install packages without release checking.  That way we don't need to update this unless the ubuntu package structure changes on us.

Joshua-Anderson:
@LeStarch Ready for a re-review!

LeStarch (approved):

PR #931, opened 2021-08-18
Use ordinary char pointers for assert strings

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

As part of my C++ modernization work, I'm trying to replace all C-style typecasting with C++ casts.

The C++ treats chars, unsigned chars (U8), and signed chars (I8) as three unique types.
The assert subsystem uses both U8* and I8* to hold assert strings. This quickly becomes awkward when trying to work with const strings and the string handling methods. All string inputs to the assert system must be cast from char pointers to signed or unsigned pointers, then cast back to char pointers when displaying the assert. Since I didn't see any benefit to handling assert strings as U8* or I8* type, rather than replace all these C-style casts with C++ style casts, I just changed the assert system to use char* instead.

### DIFF
--- a/Fw/Types/Assert.cpp
+++ b/Fw/Types/Assert.cpp
@@ -7,6 +7,8 @@
 #include <taskLib.h>
 #endif
 
+#define FW_ASSERT_DFL_MSG_LEN 256
+
 #if FW_ASSERT_LEVEL == FW_NO_ASSERT
 
 #else
@@ -19,8 +21,8 @@
 
 namespace Fw {
 
-    void defaultPrintAssert(const I8* msg) {
-        (void)fprintf(stderr,"%s\n",reinterpret_cast<const char*>(msg));
+    void defaultPrintAssert(const CHAR* msg) {
+        (void)fprintf(stderr,"%s\n", msg);
     }
 
     void defaultReportAssert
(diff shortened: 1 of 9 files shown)


timcanham:
Part of the reason I aliased them in the first place is that JPL has a coding guideline that you shouldn't use native types directly, but always through an alias, hence `U32`, `U8`, etc.

Joshua-Anderson:
@timcanham I thought something like that might be the reasoning. We don't currently have an aliased type for char, do we? It varies based on the implementation whether it's signed or unsigned.

Does that coding guideline hold for strings? Since the compiler itself treats string literals as `const char *` that guideline seems to require a large amount of casting, which also isn't ideal. We use char * within all of our string types, for example, rather than deal with all the casting.

timcanham:
`char` can be unsigned or signed? I thought `char` and `unsigned char` were portable. I think there is an error in the original code in that the `FILE_NAME_ARG` is `U8` instead of `I8`. I think that recent C++ sees `typedef` types as unique where C sees the original type, so it wouldn't complain.

Joshua-Anderson:
https://docs.microsoft.com/en-us/cpp/cpp/fundamental-types-cpp?view=msvc-160

> The C++ compiler treats variables of type char, signed char, and unsigned char as having different types.

The C++ standard doesn't define if char is signed or unsigned.


This means that every conversion from U8 and even I8 to char requires casting, specifically the unsafe `reinterpret_cast`

timcanham:
Perhaps we should create a `CHAR` alias then to confirm to the coding standards?

LeStarch:
I like this: "Perhaps we should create a CHAR alias then to confirm to the coding standards?"

Joshua-Anderson:
Update assert to use `CHAR` an alias to `char`.

Svc/AssertFatalAdapter/AssertFatalAdapterComponentImpl.cpp
```diff
@@ -30,7 +30,7 @@ namespace Fw {
             AssertArg arg4,
             AssertArg arg5,
             AssertArg arg6,
-            I8* destBuffer,
+            char* destBuffer,
```
timcanham, line 33: Forgot to make it `CHAR*`

Fw/Types/Assert.cpp
```diff
@@ -128,7 +128,7 @@ namespace Fw {
 
     NATIVE_INT_TYPE SwAssert(FILE_NAME_ARG file, NATIVE_UINT_TYPE lineNo) {
         if (NULL == s_assertHook) {
-            I8 assertMsg[256];
+            CHAR assertMsg[256];
```
timcanham, line 131: We should move this magic array size number to the config header.

timcanham (changes requested):
Minor comments

Fw/Types/Assert.cpp
LeStarch: Different PR?

LeStarch (approved):

Fw/Types/Assert.cpp
LeStarch: This was fixed.

PR #955, opened 2021-08-24
Convert C style casts to C++ style casts

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

## Rationale


LeStarch:
Closing this, as it is being broken up into smaller PRs.

PR #956, opened 2021-08-24
lestarch: compile options added correctly

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

The way compile options were being added forced a rebuild of UTs for every update to the CMake cache.  This adds compile options properly.


cmake/Options.cmake
```diff
@@ -129,22 +129,9 @@ endif()
 ####
 SET(CMAKE_CXX_FLAGS_RELEASE "-std=c++03" CACHE STRING "C++ flags." FORCE)
```
Joshua-Anderson, line 130: At the same time do you mind correctly setting the C++ standard like we're doing in release/3.0 to avoid similar issues? Make sure C++11 is set when the build type is testing.

LeStarch: I wasn't planning on that, as that would force the C++11 issue outside (before) the v3.0.0 branch.

LeStarch: I guess I need to set it, as add compile options will not work with that flag.

cmake/Options.cmake
```diff
@@ -127,11 +127,16 @@ endif()
 #
 # e.g. `-DCMAKE_BUILD_TYPE=TESTING`
 ####
-SET(CMAKE_CXX_FLAGS_RELEASE "-std=c++03" CACHE STRING "C++ flags." FORCE)
-SET(CMAKE_C_FLAGS_RELEASE "-std=c99" CACHE STRING "C flags." FORCE)
+set(CMAKE_CXX_STANDARD 11)
```
Joshua-Anderson, line 130: What about setting the standard to C++03 here, then, if the build type is testing set the standard to C++11 within the same conditional as the additional C++ flags?

Joshua-Anderson (approved):

PR #973, opened 2021-08-27
Remove max and min integer constants in favor of std::numeric_limits

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Replace numeric limit constants with std::numeric_limits, the recommended C++ technique for calculating a max/min value for an integer type.

### DIFF
--- a/Fw/Types/BasicTypes.hpp
+++ b/Fw/Types/BasicTypes.hpp
@@ -102,61 +102,6 @@ typedef char CHAR;
 #define NULL  (0)  //!< NULL
 #endif
 
-
-#ifndef I8_MAX
-#define I8_MAX static_cast<I8>(127)
-#endif
-
-#ifndef I8_MIN
-#define I8_MIN static_cast<I8>(-128)
-#endif
-
-#ifndef U8_MAX
-#define U8_MAX static_cast<U8>(255u)
-#endif
-
-#if FW_HAS_16_BIT
-#ifndef I16_MAX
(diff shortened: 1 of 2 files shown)


LeStarch:
@Joshua-Anderson why not set the constants we are using in BasicTypes.hpp from <limits>?  That way, like our types, we can use them while still referring to the standard header?

Joshua-Anderson:
> why not set the constants we are using in BasicTypes.hpp from?

I thought about that, but I figured we'd probably want to stick with stock C++. If we have these constants, some people will know they exist and use them. Other people won't know that they exist and directly use limits because that's the recommended technique to lookup max value. In fact, std::limits was already used more frequently in fprime than these constants. I figure the more we stick with the stdlib (as appropriate for embedded contexts) the easier time people will have picking up the codebase. I'm not super tied to this view though, the limits syntax is certainly a little ugly.

LeStarch:
Next question: does <U32> in numeric limits reference a template....or is it just some special syntax?

LeStarch:
It is a template....which we have yet to get approved for use in the core framework.

Joshua-Anderson:
@LeStarch Technically a template, but it compiles down to a constant so I don't think the normal template concerns apply here.

LeStarch:
@Joshua-Anderson you are correct.  The documentation says these are fully specified and they are all constant properties.  Thus the reasons behind not using templates aren't valid in this case.

PR #980, opened 2021-08-30
Fw: Convert C style casts to C++ style casts

Joshua-Anderson:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Breaking changes in #955 into smaller chuncks. This PR is just the modifications to the Fw directory.

## Rationale

### DIFF
--- a/Fw/Log/LogString.cpp
+++ b/Fw/Log/LogString.cpp
@@ -28,7 +28,7 @@ namespace Fw {
     }
 
     NATIVE_UINT_TYPE LogStringArg::length() const {
-        return (NATIVE_UINT_TYPE) strnlen(this->m_buf,sizeof(m_buf));
+        return static_cast<NATIVE_UINT_TYPE>(strnlen(this->m_buf,sizeof(m_buf)));
     }
 
     const char* LogStringArg::toChar() const {
@@ -55,9 +55,9 @@ namespace Fw {
         if (stat != FW_SERIALIZE_OK) {
             return stat;
         }
-        return buffer.serialize((U8*)this->m_buf,strSize,true);
+        return buffer.serialize(reinterpret_cast<const U8*>(this->m_buf),strSize,true);
 #else
-        return buffer.serialize((U8*)this->m_buf,strSize);
+        return buffer.serialize(reinterpret_cast<const U8*>(this->m_buf),strSize);
 #endif
     }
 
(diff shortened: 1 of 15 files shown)


Fw/Log/LogString.cpp
```diff
@@ -28,7 +28,7 @@ namespace Fw {
     }
 
     NATIVE_UINT_TYPE LogStringArg::length() const {
-        return (NATIVE_UINT_TYPE) strnlen(this->m_buf,sizeof(m_buf));
+        return strnlen(this->m_buf,sizeof(m_buf));
```
LeStarch, line 31: Here we have an implicit cast from size_t to NATIVE_UINT_TYPE which on some systems can produce warnings about loss of precision.

Fw/Types/BasicTypes.hpp
```diff
@@ -104,56 +104,56 @@ typedef char CHAR;
 
 
 #ifndef I8_MAX
-#define I8_MAX (I8)(127)
+#define I8_MAX I8(127)
```
LeStarch, line 107: I don't understand this syntax.  Is it described anywhere?

LeStarch: "This cast expression is exactly equivalent to the corresponding C-style cast expression."  is what I found in the C++ documentation.  So, how is this better?

Fw/Types/BasicTypes.hpp
```diff
@@ -104,56 +104,56 @@ typedef char CHAR;
...
 
 #if FW_HAS_64_BIT
 #ifndef I64_MAX
-#define I64_MAX (I64)(9223372036854775807)
+#define I64_MAX I64(9223372036854775807)
 #endif
 
 #ifndef I64_MIN
-#define I64_MIN (I64)(-9223372036854775808)
+#define I64_MIN I64(-9223372036854775808)
```
LeStarch, line 152: Based on my understanding, this needs "ll" at the end to ensure that on 32bit and lower systems, this constant doesn't get rolled over before casts.

Also, I think we should add the "u"  to the unsigned.

LeStarch (commented):
Some comments to discuss, but **much** easier to review.  Thank you.

Fw/Types/BasicTypes.hpp
Joshua-Anderson: Apparently all I managed was outsmarting the static analysis 🤦  I made the incorrect assumption that C++ would treat primitive type constructors/functional casts similar to assignment... thanks for bringing this up and making me dive into this more. Why C++ decided to make the behavior identical to a C style cast is a mystery to me.

Fw/Types/BasicTypes.hpp
Joshua-Anderson: We should use `u` for unsigned constants`. To my understanding, compilers will default to typing numeric literals as int, then increase the size to long or long long as necessary to fit the constant. I believe the `ll` is redundant because a 32bit compiler would still size the constant to long long if necessary.

Joshua-Anderson:
Comments resolved, ready for a re-review @LeStarch

LeStarch (approved):

PR #992, opened 2021-08-31
lestarch: reworking posix tasks

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

There were several issues in the Os/Posix/Task.cpp and Os/Posix/TaskRoot.cpp.  These were:

1. Os/Posix/Task.cpp silently dropped priorities, affinity without any indication
2. Os/Posix/TaskRoot.cpp code-rotted and was no longer kept up-to-date
3. All posix implementations accepted invalid priorities, stack sizes, without validation or explanation
4. Ref, RPI deployments both specified invalid stack sizes, invalid priorities, and were incapable of using these settings anyway
5. Permissions prevent TaskRoot.cpp from running unless user is root.

This PR fixes the above issues by:

1. Merging TaskRoot into Task.cpp
2. Clamping out-of-range values to posix defaults (max/min priority, min stack size)
3. Providing argument-free constructors to use task defaults
4. Falling back to priority-free tasks when user has not permissions to run priority tasks
5. Logging warnings when:
  - Out-of-range values are supplied
  - Old function signatures are used
  - Tasks set to use priority and were unable to do so because of permissions

The design of this PR is to keep existing applications running as-is, but produce warnings to indicate corrections that should be made.  In a future PR, these warnings will be removed, and users will be forced to upgrade.

## Rationale

If a user specifies stack size/priority it should work, not silently be discarded.  For users that don't care, a default should be provided that maps to posix defaults.   If the user does not have permission, the program should fail.

## Future Work

Given this is devel and we don't want to break all users, we have provided warnings for the time being.  As we approach release v3.0.0, we need to remove the warnings and replace them with hard-failures forcing users to upgrade.


Joshua-Anderson (approved):
Looks great!

Drv/Ip/SocketReadTask.hpp
```diff
@@ -44,14 +44,14 @@ class SocketReadTask {
...
      * \param reconnect: automatically reconnect socket when closed. Default: true.
-     * \param cpuAffinity: cpu affinity provided to task. See: Os::Task::start.
+     * \param cpuAffinity: cpu affinity provided to task. See: Os::Task::start. Default: -1, don't care
      */
     void startSocketTask(const Fw::StringBase &name,
-                         const NATIVE_INT_TYPE priority,
-                         const NATIVE_INT_TYPE stack,
+                         const NATIVE_INT_TYPE priority = -1,
+                         const NATIVE_INT_TYPE stack = -1,
                          const bool reconnect = true,
```
timcanham, line 55: Maybe put this before priority/stack since this will probably be more commonly modified.

timcanham (changes requested):

Drv/Ip/SocketReadTask.hpp
LeStarch: Are you referring to the reconnect argument?

timcanham (approved):

PR #1017, opened 2021-09-04
lestarch: deframer fix

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes a known bug in the deframer where packets aggregated into a single receive call will not all unpack when the size of each packet decreases from the previous.

Fixed UTs to work with this fix, and put in a randomized test for framer + fprime protocol.

### DIFF
--- a/Os/test/ut/TestMain.cpp
+++ b/Os/test/ut/TestMain.cpp
@@ -38,10 +38,6 @@ TEST(Nominal, QTestPerformance) {
 TEST(Nominal, QTestConcurrentTest) { 
    qtest_concurrent();
 }
-// Disabled for macOS known issue
-//TEST(Nominal, IntervalTimerTest) { 
-//   intervalTimerTest();
-//}
 TEST(Nominal, FileSystemTest) { 
    fileSystemTest();
 }
--- a/Os/test/ut/TestMain.cpp
+++ b/Os/test/ut/TestMain.cpp
@@ -38,9 +38,10 @@ TEST(Nominal, QTestPerformance) {
 TEST(Nominal, QTestConcurrentTest) { 
    qtest_concurrent();
 }
-TEST(Nominal, IntervalTimerTest) { 
-   intervalTimerTest();
-}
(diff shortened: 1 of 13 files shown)


Svc/Deframer/test/ut-unified/DeframerRules.cpp
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * This file specifies Rule classes for testing of the Svc::Deframer. These rules can then be used by the main
+ * testing program to test the code.
+ *
+ *
+ * @author mstarch
+ */
+#include "DeframerRules.hpp"
+#include <STest/Pick/Pick.hpp>
+#include <stdio.h>
```
Joshua-Anderson, line 12: <cstdio> ... I really need to setup clang-tidy in CI

Svc/Deframer/test/ut-unified/DeframerRules.hpp
```diff
@@ -0,0 +1,47 @@
+/**
+ * GroundInterfaceRules.hpp:
```
Joshua-Anderson, line 2: Comment incorrect

Joshua-Anderson (approved):
Other than a few minor comments, looks good!

Os/test/ut/TestMain.cpp
```diff
@@ -38,9 +38,10 @@ TEST(Nominal, QTestPerformance) {
 TEST(Nominal, QTestConcurrentTest) { 
    qtest_concurrent();
 }
-TEST(Nominal, IntervalTimerTest) { 
-   intervalTimerTest();
-}
+// Disabled for macOS known issue
```
Joshua-Anderson, line 41: I've now actually seen this test fail on linux as well. I think that this test should just be deleted... if we want to unit test interval timers we need to rearchitect the design of interval timers

Joshua-Anderson (approved):

PR #1024, opened 2021-09-09
lestarch: installing dependencies where necessary, checking for other…

LeStarch:
… dependencies

| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Installs: 

1. SBT, when needing to build fpp

Checks for (existing installation:
1. fprime-util (for autocoder dependencies)
2. python3
3. java


LeStarch:
@bocchino this automates the SBT install.  It now works as follows:

1. If fpp is found in path, do nothing
2. If fpp not found, but sbt is....build fpp
3. If neither is found, download sbt, then use it to build fpp

I also added checks for:
1. JAVA (required to run sbt and fpp)
2. python3 + fprime-util (required to run python autocoder)

LeStarch:
Added in: 
- UT fixes,
- Removal of broken validation cmake
- Proper python invocation

bocchino:
Looks good! Questions:

1. What triggers the installation? Is it fprime-util configure?
2. Step 1 (if FPP found in the path do nothing) seems dangerous, because the FPP version in the path could be wrong. Can we have the system locally install the correct FPP version unless (a) the correct version is already locally installed or (b) the user asks not to do this?

LeStarch:
1. @bocchino an install is triggered when you run fprime-util configure (or if you are using cmake configure directly).
2. Most every user will not have FPP installed.  This detection is for developers who should know what they are doing. We can add a bunch of extra flags, etc....but that seems overkill for the people ambitious enough to have fpp installed already.

Thoughts?

LeStarch:
@bocchino I should also ask, does fpp have a way of reporting the version?  It would be nice to be able to do:

```
if [ fpp-depends --version == <FPP VERSION SET FOR AUTO INSTALL ]
do
   use fpp as is (version matches)
else
   install correct fpp
done
```

bocchino:
> Most every user will not have FPP installed. 

Good point. I think it's good to optimize for that case.

Can we have command for installing and using the correct FPP, overriding any system FPP? I have the following use case in mind:

1. I do have FPP installed in my PATH.
2. I pull down a branch of F Prime that needs a special version of FPP in order to work.

Now what? Do I have to remove FPP from my PATH and run fpp-configure? Or manually check out the corresponding commit hash and do a system-wide FPP Installation at that version? Both options seem doable, but slightly painful. If there's a way to eliminate the pain at low cost, that would be good.

bocchino:
You can get the version of any FPP tool with the `--help` option:
```
[I] bocchino@MT-306179 ~/J/T/F/f/R/Top (release/v3.0.0)> fpp-depend --help
fpp-depend v1.0.0
Usage: fpp-depend [options] [file ...]

  -d, --direct <file>     write direct dependencies to file
  -g, --generated <file>  write names of generated files to file
  -i, --included <file>   write included dependencies to file
  -m, --missing <file>    write missing dependencies to file
  -h, --help              print this message and exit
  file ...                input files
```

bocchino:
```
[I] bocchino@MT-306179 ~/J/T/F/f/R/Top (release/v3.0.0)> fpp-depend --help | awk 'NR == 1 { print $2 }'
v1.0.0
```

bocchino:
It's good to do that kind of version check. However, for development purposes we need version matching at the granularity of commit hashes.

bocchino:
To automate the commit hash check, you could have CMake save off the last commit hash that it installed.

LeStarch:
@bocchino I can add a flag to the CMake build:

```
fprime-util -DFPRIME_INSTALL_FPP=OFF
```

otherwise it will download and install fpp.  That way, you only deal with the pain of manually installing fpp if and when you turn that feature of.

As for the commit hash version, I think FPP should (at some point) should record the exact hash of what is expected.  I believe the system will properly setup whatever version is specified...but if CMake is speaking in git hashes, then fpp tools should also speak in git hashes, otherwise guaranteeing the tools are consistent is impossible.  To be clear, this is an eventual need, not a "now" need.

bocchino:
That `-D` flag sounds good. I think most of the time it's convenient to have the configuration just do the right thing (install the correct version). Manually turning it off (e.g., to optimize the number of installations required) seems good.

> CMake is speaking in git hashes, then fpp tools should also speak in git hashes, otherwise guaranteeing the tools are consistent is impossible

It's certainly not impossible. Here is an algorithm for installing hash _H_:

1. If file `installed-hash.txt` exists and contains the value _H_ then do nothing.
2. Otherwise
    a. Install hash _H_.
    b. Write _H_ into `installed-hash.txt`.

Can we do it this way for now?

I see how it might be better to have the FPP tools report the hash. This seems tricky to implement on the FPP side, but not impossible.

bocchino:
> It's certainly not impossible.

I guess what's impossible is ensuring that the **system** FPP (if it exists) matches the desired FPP. For that we need FPP to report its own hash.

bocchino:
Wait, I guess since the FPP install is keyed to configure, and configure can't work without purging, my solution won't work, since any info about what was installed last time gets wiped out in the purge. I keep forgetting how CMake works.

bocchino:
Does this mean we also have to purge and reconfigure everything to update FPP?

bocchino:
Given how CMake works, I agree that having FPP report the hash is best. Otherwise there are too many headaches.

bocchino:
For now (in the absence of this feature) we can have CMake conservatively install the correct FPP version every time configure is run, unless the -D option turns off the FPP installation.

bocchino:
It seems this can create another situation like the following:

1. Check out, configure, and build F Prime. Everything works.
2. Check out a later version of F Prime and build. It breaks because you didn't purge and reconfigure first.

This is not essential for closing this issue, but I think we should try to alert users when they pull down changes that require a purge and reconfigure. For example, any change in the CMake rules, any change in the FPP commit hash, etc.

bocchino:
Try pulling the latest from fprime-community/fpp. It now does this:
```
[I] bocchino@MT-306179 ~/J/T/f/compiler (main)> fpp-depend --help
fpp-depend v1.0.0 commit c27c1a7d9b611fe20d276f140ba67de257a94b1c
Usage: fpp-depend [options] [file ...]

  -d, --direct <file>     write direct dependencies to file
  -g, --generated <file>  write names of generated files to file
  -i, --included <file>   write included dependencies to file
  -m, --missing <file>    write missing dependencies to file
  -h, --help              print this message and exit
  file ...                input files
```

LeStarch:
Will do!

LeStarch:
@bocchino I think we should merge this in (once I have the CI passing).  I don't yet have implemented a -D to use FPP tools ON/OFF path, but most of the refactoring and functionality is here.

It would make a good RC to let users start seeing the new functions while I clean-up the last few items.

bocchino:
Looks good, let me know when it's ready for me to give it a try.

LeStarch:
@bocchino I think this is ready for review and merge.  So go ahead.  I doubt the C++ or Quality checks will fail because I didn't touch much C++ code.

bocchino:
OK, thanks, I'll take a look.

bocchino (approved):
Looks good, everything works as expected. Build is slow after adding or deleting an FPP symbol, but this is work to go.

Recommend linking this PR to an open issue with work to go.

PR #1025, opened 2021-09-09
Os: Convert C style casts to C++ style casts

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  Os |
|**_Related Issue(s)_**|  |

---
## Change Description

Breaking changes in #955 into smaller chunks. This PR is just the modifications to the Os directory.

## Rationale

### DIFF
--- a/Os/Linux/File.cpp
+++ b/Os/Linux/File.cpp
@@ -74,7 +74,7 @@ namespace Os {
                 flags = O_WRONLY | O_CREAT | O_APPEND;
                 break;
             default:
-                FW_ASSERT(0,(NATIVE_INT_TYPE)mode);
+                FW_ASSERT(0, mode);
                 break;
         }
 
@@ -226,9 +226,9 @@ namespace Os {
                         break; // break out of while loop
                     } else {
                         // in order to move the pointer ahead, we need to cast it
-                        U8* charPtr = (U8*)buffer;
+                        U8* charPtr = static_cast<U8*>(buffer);
                         charPtr = &charPtr[readSize];
-                        buffer = (void*)charPtr;
+                        buffer = static_cast<void*>(charPtr);
                     }
                     maxIters--; // decrement loop count
                 }
@@ -285,8 +285,8 @@ namespace Os {
                         stat = NO_SPACE;
                         break;
                     default:
-                        DEBUG_PRINT("Error %d during write of 0x0%llx, addrMod %d, size %d, sizeMod %d\n",
-                                    errno, (U64) buffer, ((U64) buffer) % 512, size, size % 512);
+                        DEBUG_PRINT("Error %d during write of 0x%p, addrMod %d, size %d, sizeMod %d\n",
+                                    errno, buffer, static_cast<POINTER_CAST>(buffer) % 512, size, size % 512);
                         stat = OTHER_ERROR;
                         break;
                 }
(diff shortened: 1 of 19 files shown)


Os/Linux/File.cpp
```diff
@@ -286,7 +286,7 @@ namespace Os {
                         break;
                     default:
                         DEBUG_PRINT("Error %d during write of 0x0%llx, addrMod %d, size %d, sizeMod %d\n",
-                                    errno, (U64) buffer, ((U64) buffer) % 512, size, size % 512);
+                                    errno, static_cast<POINTER_CAST>(buffer), static_cast<POINTER_CAST>(buffer) % 512, size, size % 512);
```
LeStarch, line 289: Pointer cast will cause issues with the "ll" flag used here when compiling for systems where the pointer size is not 64bit.

Os/Posix/IPCQueue.cpp
```diff
@@ -68,7 +68,7 @@ namespace Os {
         handle = mq_open(this->m_name.toChar(), O_RDWR | O_CREAT | O_EXCL, 0666, &att);
 
         // If queue already exists, then unlink it and try again.
-        if (-1 == (NATIVE_INT_TYPE) handle) {
+        if (-1 ==  handle) {
```
LeStarch, line 71: Does this compare work cleanly between mqd_t and int?

Os/Posix/IPCQueue.cpp
```diff
@@ -238,23 +238,23 @@ namespace Os {
...
     NATIVE_INT_TYPE IPCQueue::getQueueSize() const {
-        QueueHandle* queueHandle = (QueueHandle*) this->m_handle;
+        QueueHandle* queueHandle = reinterpret_cast<QueueHandle*>(this->m_handle);
         mqd_t handle = queueHandle->handle;
 
         struct mq_attr attr;
         int status = mq_getattr(handle, &attr);
         FW_ASSERT(status == 0);
-        return (U32) attr.mq_maxmsg;
+        return static_cast<U32>(attr.mq_curmsgs);
```
LeStarch, line 247: Why did the member change here?

Os/Posix/Queue.cpp
```diff
@@ -246,23 +246,23 @@ namespace Os {
...
     NATIVE_INT_TYPE Queue::getQueueSize() const {
-        QueueHandle* queueHandle = (QueueHandle*) this->m_handle;
+        QueueHandle* queueHandle = reinterpret_cast<QueueHandle*>(this->m_handle);
         mqd_t handle = queueHandle->handle;
 
         struct mq_attr attr;
         int status = mq_getattr(handle, &attr);
         FW_ASSERT(status == 0);
-        return (U32) attr.mq_maxmsg;
+        return static_cast<U32>(attr.mq_curmsgs);
```
LeStarch, line 255: Same here.

LeStarch (changes requested):
I see some functionality changes, which I did not expect.

Os/Posix/IPCQueue.cpp
Joshua-Anderson: mq_t is a signed value that can return -1 on error.

Os/Posix/IPCQueue.cpp
Joshua-Anderson: Oh yikes, copy and paste error I somehow overlooked. How did this not break the unit tests.

LeStarch (approved):

LeStarch:
Looks good now.

PR #1038, opened 2021-09-15
Various: Convert C style casts to C++ style casts

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**| various |
|**_Related Issue(s)_**|  |

---
## Change Description

Breaking changes in #955 into smaller chunks. This PR contains the remaining changes that didn't fit into previous pull requests.

## Rationale

### DIFF
--- a/Drv/Ip/IpSocket.cpp
+++ b/Drv/Ip/IpSocket.cpp
@@ -69,7 +69,7 @@ SocketIpStatus IpSocket::setupTimeouts(NATIVE_INT_TYPE socketFd) {
     timeout.tv_sec = this->m_timeoutSeconds;
     timeout.tv_usec = this->m_timeoutMicroseconds;
     // set socket write to timeout after 1 sec
-    if (setsockopt(socketFd, SOL_SOCKET, SO_SNDTIMEO, (char*)&timeout, sizeof(timeout)) < 0) {
+    if (setsockopt(socketFd, SOL_SOCKET, SO_SNDTIMEO, reinterpret_cast<char *>(&timeout), sizeof(timeout)) < 0) {
         return SOCK_FAILED_TO_SET_SOCKET_OPTIONS;
     }
 #endif
--- a/Drv/Ip/test/ut/SocketTestHelper.cpp
+++ b/Drv/Ip/test/ut/SocketTestHelper.cpp
@@ -21,7 +21,7 @@ void force_recv_timeout(Drv::IpSocket& socket) {
     struct timeval timeout;
     timeout.tv_sec = 0;
     timeout.tv_usec = 50; // 50ms max before test failure
-    setsockopt(socket.m_fd, SOL_SOCKET, SO_RCVTIMEO, (char *) &timeout, sizeof(timeout));
+    setsockopt(socket.m_fd, SOL_SOCKET, SO_RCVTIMEO, reinterpret_cast<char *>(&timeout), sizeof(timeout));
 }
 
 void validate_random_data(U8 *data, U8 *truth, U32 size) {
@@ -33,7 +33,7 @@ void validate_random_data(U8 *data, U8 *truth, U32 size) {
 void fill_random_data(U8 *data, U32 size) {
     ASSERT_NE(size, 0u) << "Trying to fill random data of size 0";
     for (U32 i = 0; i < size; i++) {
-        data[i] = (U8) STest::Pick::any();
+        data[i] = static_cast<U8>(STest::Pick::any());
     }
 }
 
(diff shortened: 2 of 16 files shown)


Drv/Ip/IpSocket.cpp
```diff
@@ -69,7 +69,7 @@ SocketIpStatus IpSocket::setupTimeouts(NATIVE_INT_TYPE socketFd) {
     timeout.tv_sec = this->m_timeoutSeconds;
     timeout.tv_usec = this->m_timeoutMicroseconds;
     // set socket write to timeout after 1 sec
-    if (setsockopt(socketFd, SOL_SOCKET, SO_SNDTIMEO, (char*)&timeout, sizeof(timeout)) < 0) {
```
LeStarch, line 72: Isn't this cast necessary for VxWorks?

Drv/Ip/test/ut/SocketTestHelper.cpp
```diff
@@ -21,7 +21,7 @@ void force_recv_timeout(Drv::IpSocket& socket) {
     struct timeval timeout;
     timeout.tv_sec = 0;
     timeout.tv_usec = 50; // 50ms max before test failure
-    setsockopt(socket.m_fd, SOL_SOCKET, SO_RCVTIMEO, (char *) &timeout, sizeof(timeout));
```
LeStarch, line 24: Same.

Drv/SocketIpDriver/SocketHelper.cpp
```diff
@@ -156,7 +156,7 @@ namespace Drv {
         timeout.tv_sec = this->m_timeoutSeconds;
         timeout.tv_usec = this->m_timeoutMicroseconds;
         // set socket write to timeout after 1 sec
-        if (setsockopt(socketFd, SOL_SOCKET, SO_SNDTIMEO, (char *)&timeout, sizeof(timeout)) < 0) {
```
LeStarch, line 159: Same.

RPI/RpiDemo/RpiDemoComponentImpl.cpp
```diff
@@ -129,7 +129,7 @@ namespace Rpi {
   {
       // convert incoming data to string. If it is not printable, set character to '*'
       char uMsg[serBuffer.getSize()+1];
-      char* bPtr = (char*)serBuffer.getData();
+      char* bPtr = reinterpret_cast<char*>(serBuffer.getData());
```
LeStarch, line 132: `CHAR*`

LeStarch: ?

Utils/Hash/libcrc/CRC32.cpp
```diff
@@ -33,7 +33,7 @@ namespace Utils {
         FW_ASSERT(data);
         char c;
         for(int index = 0; index < len; index++) {
-            c = ((char*)data)[index];
+            c = static_cast<const char*>(data)[index];
```
LeStarch, line 36: `CHAR*`

Utils/test/ut/LockGuardTester.cpp
```diff
@@ -58,7 +58,7 @@ namespace Utils {
 
     {
       LockGuard guard(data.mutex);
-      stat = testTask.start(name,12,100,10*1024,taskMethod,(void*) &data);
```
LeStarch, line 61: VxWorks?

Drv/Ip/IpSocket.cpp
Joshua-Anderson: Arrrggggg..... you're right, why does VxWorks have to be so quirky. On everything else option_value value is `void *`, but no, VxWorks decides that this is the perfect opportunity to throw in a random `char *`.

RPI/RpiDemo/RpiDemoComponentImpl.cpp
Joshua-Anderson: I think the conversion of `char` to `CHAR` should be done in another PR at some point. If I used CHAR here I'd want to update the entire file to use it.

Utils/test/ut/LockGuardTester.cpp
Joshua-Anderson: I don't think there's anything vxworks related here. You can always implicitly cast from any pointer type to a void pointer. It's also a unit test so I don't think it's ever going to compiled for vxworks.

RPI/RpiDemo/RpiDemoComponentImpl.cpp
LeStarch: Agreed.  I had erroneously thought this change went through already.

Drv/Ip/IpSocket.cpp
LeStarch: I missed a critical detail....this code is not used on vxworks, as it is the timeout code.  We can revert to what you had before.

LeStarch (changes requested):
Apologies for missing that was in a vxworks excluded block.

Joshua-Anderson:
@LeStarch ready for re-review

LeStarch (approved):

PR #1049, opened 2021-09-20
Replace NULL with nullptr

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Use clang-tidy to replace usage of NULL with nullptr


LeStarch:
I see these alot:

```
static_cast<POINTER_CAST>(NULL)
```

Becomes:
```
0
```

I think in this case we still want:

```
static_cast<POINTER_CAST>(nullptr)
```

The reason is this: nullptr is the representation of the null pointer value.  Thus the above code only works if `nullptr` is equivalent to `0`.  It is, but what is the point of having a label if we are going to use `0` part of the time?

Joshua-Anderson:
@LeStarch Part of the point of nullptr is that it cannot be easily converted to an integer... you might be able to reinterpret cast it, but I figured setting integer handles to zero was probably cleaner than that.

LeStarch:
@Joshua-Anderson that is my point.  `0` and `nullptr` are not the same and extra work has to be done to ensure they are comparable.  Here we are ignoring it and writing in zero directly, which the would ask the question: why not use 0 instead of `nullptr` in all cases?

Joshua-Anderson:
It's the difference between initializing a pointer (initialized with nullptr) and initializing a U64 we use to hold a pointer (initialize with integer value 0). C++ really doesn't want people treating nullptr like an integer and this was considered one of the problems with NULL. It's deliberately hard to initialize an integer will nullptr. Using a U64 as a pointer is kinda an antipattern anyway, but I'd argue that initializing a U64 with 0 is slightly cleaner than a reinterpret cast of nullptr to initialize the handler. I don't have a super strong opinion though, I'm happy to try and make nullptr work to initialize these integers.

Is there a reason we use U64 values to hold handles instead of just a void pointer, which plays much better with C++ typechecking?

Joshua-Anderson:
@timcanham we use POINTER_CAST in a few spots, particularly for queue object and task object handles.

https://github.com/nasa/fprime/blob/devel/Os/Queue.hpp#L75
https://github.com/nasa/fprime/blob/devel/Os/Task.hpp#L68

LeStarch:
The reason we use these is because these aren't guaranteed to be pointers.  In some systems they may be file-handles which are integers.  Hence we chose an integer that could hold the pointer representation or the file representation.

Joshua-Anderson:
@LeStarch I've replaced all usages of 0 with `reinterpret_cast<POINTER_CAST>(nullptr)`.

LeStarch (approved):

Issue #945, opened 2021-08-20
Consider Removing use of Variable Length Arrays for Async Serializable Ports


Joshua-Anderson:
Because serialized ports don't have a predefined message size, users declare the maximum message size at component initialization.

When a user enqueues a message in an async port, the autocoder has to prepend the port ID and port number to the message so that the port can be properly handled when dequeued.

The way this is currently implemented is:
 
- Create a message buffer from a variable length array of maximum message size (provided at initialization)
- Add port id and number to message buffer.
- Copy user provided data into message buffer
- enqueue message buffer.

While this is a perfect use case for variable length arrays, use of variable length arrays is discouraged because they're not part of the C++ standard (they're a gcc extension).

Unfortunately, there's no good alternative... this message buffer can't be dynamically heap allocated. We could make the maximum serialized port message size some sort of constant (part of the component model?) rather than a runtime option, but that would likely require users modify the builtin F' components whenever they'd need to adjust this constant, breaking the reusability of F' components.

matt392code:
[fprime-std-array.txt](https://github.com/user-attachments/files/18304253/fprime-std-array.txt)
Several potential solutions to address the variable-length array concern in F' serializable ports are below: 
Key Challenges:
1. Need to handle variable message sizes
2. Must avoid heap allocation during message processing
3. Need to maintain component reusability
4. Must be C++ standard compliant
5. Performance critical for spacecraft control

Proposed Solutions:
1. **std::array with compile-time maximum size**
- Pros:
  - Standard C++ compliant
  - Zero heap allocation
  - Predictable memory usage
- Cons:
  - Requires template parameter
  - May waste memory if maximum size is much larger than typical messages

2. **Compile-time configuration with inheritance**
- Pros:
  - Allows customization per component
  - Standard compliant
  - No runtime overhead
- Cons:
  - Less flexible than runtime configuration
  - Requires recompilation to change sizes

3. **std::vector with reserve()**
- Pros:
  - More flexible
  - Standard compliant
  - Memory efficient
- Cons:
  - Still uses heap allocation (though only once at initialization)
  - Slightly higher memory overhead

4. **Flexible array member**
- Pros:
  - Single allocation at initialization
  - Memory efficient
  - Standard C++ feature
- Cons:
  - More complex implementation
  - Requires careful memory management

Recommendation:
For spacecraft control software, I would recommend Solution 1 (std::array) or Solution 2 (compile-time configuration) because they:
1. Have zero heap allocation
[...]

PR #1050, opened 2021-09-20
Compile test project with pedantic

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Affected Architectures(s)_**|   |
|**_Related Issue(s)_**| #945 |

---
## Change Description

Compile the test project with pedantic to catch non-portable C++. A few of the pedantic warnings had to be disabled but hopefully we can slowly resolve these issues over time. Ideally we'd disable specific compilation warnings only on the affected  files, but it's tricky to do in a portable way.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,19 +9,48 @@ set(FPRIME_FRAMEWORK_PATH "${CMAKE_CURRENT_LIST_DIR}" CACHE PATH "Location of F 
 set(FPRIME_PROJECT_ROOT "${CMAKE_CURRENT_LIST_DIR}" CACHE PATH "Root path of F prime project" FORCE)
 
 # This cmake project is only intended to be used by CI and developers to test F-prime. We enable
-# -Wall and -Wextra on this particular project as a canary to warn about compilation errors without
-# impacting real projects, where a warning from an untested compiler could break the build.
+# -Wall and -Wextra on this particular project as a canary to warn about compilation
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -9,14 +9,34 @@ set(FPRIME_FRAMEWORK_PATH "${CMAKE_CURRENT_LIST_DIR}" CACHE PATH "Location of F
...
+    -Werror
+    $<$<COMPILE_LANGUAGE:CXX>:-Wold-style-cast>
+    -Wno-unused-parameter
+    -Wno-gnu-zero-variadic-macro-arguments
+    $<$<COMPILE_LANGUAGE:CXX>:-Wno-vla-extension>
+    $<$<COMPILE_LANGUAGE:CXX>:-Wno-zero-length-array>
+)
 
 # For this testing cmake project, enable AddressSanitizer, a runtime memory sanitizer, on all unit tests
 set (CMAKE_C_FLAGS_TESTING "${CMAKE_C_FLAGS_TESTING} -fno-omit-frame-pointer -fsanitize=address")
```
LeStarch, line 42: Should probably clean up the way these flags are being set too.

Joshua-Anderson:
@LeStarch I discovered that unlike clang you cannot disable specific pedantic checks in GCC. I've now updated the code to enable pedantic checks on clang only. I did clean up the way test flags were set though.

LeStarch (approved):

PR #1054, opened 2021-09-22
Cleanup telemetry and event string classes

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**| Events/Telemetry  |
|**_Related Issue(s)_**|  |

---
## Change Description

- Fixed broken string truncation for telemetry strings.
- Updated the way event and telemetry strings are truncated so that they don't modify the string class itself, allowing users to pass const strings into events and telemetry.
- Cleanup up telemetry and string classes in general to match other string classes

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -34,7 +34,7 @@ ALLEXTERNALS
 alloc
 Alltypes
 alphanums
-ampc
+ampcs
 amsmath
 aname
 ANamespace
--- a/Autocoders/Python/src/fprime_ac/generators/templates/component/cpp.tmpl
+++ b/Autocoders/Python/src/fprime_ac/generators/templates/component/cpp.tmpl
@@ -889,7 +889,7 @@ namespace ${namespace} {
   #for $ids, $tlmname, $type, $size, $update, $comment, $typeinfo in $channels:
     #set $tlm_type = $type
     #if $type == "string":
-      #set $tlm_type = "Fw::TlmString&"
+      #set $tlm_type = "const Fw::TlmString&"
     #else if $typeinfo == "user"
       #set $tlm_type = "const " + $type + "&"
     #else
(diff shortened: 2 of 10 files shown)


LeStarch:
Looks like sensible changes to me.

LeStarch (commented):
You've delete the AMPCS serialization block:

```
#if FW_AMPCS_COMPATIBLE
```

We need to discuss this further.

Joshua-Anderson:
@LeStarch AMPCS support is already pretty broken. If we want to re-add ampcs support for strings we easily can add it to parent serialize method

LeStarch:
@Joshua-Anderson there is currently a task to get AMPC up and working....so we need to ensure AMPCS support is maintained after that task finishes.

Joshua-Anderson:
@LeStarch AMPCS string serialization was unsupported for most string types. I see two paths forward:

1. We mark AMPCS string serialization support as unsupported and fix it as part of the AMPC support task.
2. We add AMPCS support for serializing all string types by adding ampcs serialization logic to the string base class.

Which approach do you prefer?

LeStarch:
@Joshua-Anderson AMPCs support is only necessary for EVR and TLM (string) types...as AMPCs only dictates tlm and event format.  In general custom types (other string types, serializables, arrays) are not supported on AMPCs and need special support on the ground.

Joshua-Anderson:
> AMPCs support is only necessary for EVR and TLM (string) types

Agreed... we could add custom serialization and deserialization for those strings types. However, it's probably just easier to update the generic string type to support serializing strings in the AMPCS format. I don't see any harm in serializing other strings in the ampcs style, so it may make sense to just encode all strings identically.

LeStarch:
@Joshua-Anderson I see your point, but I think we just want to serialize Tlm and Event strings only.  Otherwise we will have yet another arbitrary string limit (255) imposed by serializing in AMPCS format.  

String serialization is used all over the system, and only strings bound for output to group need that imposition.  Thus, if it were me, I'd specifically serialize these strings for AMPCs and leave other strings alone.

LeStarch (approved):

LeStarch:
Looks correct to me.

PR #1056, opened 2021-09-23
Switch to C++11 static assertions

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Switch custom static assertions to use C++11 builtin static assertions.

### DIFF
--- a/Fw/Cfg/ConfigCheck.cpp
+++ b/Fw/Cfg/ConfigCheck.cpp
@@ -10,17 +10,18 @@
  *
  */
 
+#include <type_traits>
 #include <FpConfig.hpp>
 #include <Fw/Types/BasicTypes.hpp>
 
 // Check that command/telemetry strings are not larger than an argument buffer
 
-FW_CONFIG_ERROR(FW_CMD_STRING_MAX_SIZE <= FW_CMD_ARG_BUFFER_MAX_SIZE,CMD_STRING_TOO_BIG);
-FW_CONFIG_ERROR(FW_LOG_STRING_MAX_SIZE <= FW_LOG_BUFFER_MAX_SIZE,LOG_STRING_TOO_BIG);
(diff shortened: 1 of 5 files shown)


Fw/Cfg/ConfigCheck.cpp
```diff
@@ -10,17 +10,18 @@
  *
  */
```
LeStarch, line 12: With static asserts, can't we get rid of the ConfigCheck.cpp altogether?  I think you can put static asserts into the .hpp and not need a compilation unit just for assertions.

LeStarch (commented):
Otherwise looks good!

Fw/Cfg/ConfigCheck.cpp
Joshua-Anderson: We probably could, but you could already do that with the old system. The reason I thought this was in a cpp file was so that users could tweak FpConfig.hpp as they wished and during compilation we'd compile the ConfigCheck file to verify sure their config was valid. Is there somewhere else you would prefer to locate these checks?

LeStarch: My point is if we put these at the bottom of the FpConfig.hpp header, the checks will happen when the header is included and this `.cpp`, which does nothing, will be removed.

Joshua-Anderson: The only problem with leaving it in the header file is that users copy this header file into their own projects, moving the config validation out of the control of the F' framework itself.

Personally I like the pattern we currently have where the config is in a header file and the framework itself double checks that the provided config is valid.

I don't have a strong opinion - I'm happy to move these validation checks to the header if that's the pattern we want to move to.

LeStarch: I see your point.  We can move-forward with how you have it and reevaluate later.

PR #1071, opened 2021-09-30
Allow removal of text loggers from build with CMake variable.

nabcouwer:
Allow removal of text loggers from build with CMake variable. Otherwise build fails if FW_ENABLE_TEXT_LOGGING == 0.

| | |
|:---|:---|
|**_Originating Project/Creator_**| JPL COLDArm / Neil Abcouwer  |
|**_Affected Component_**| Build Process |
|**_Affected Architectures(s)_**| Any |

---
## Change Description

In https://github.com/nasa/fprime/blob/devel/Svc/CMakeLists.txt , move adding of Text Logger components into a `if (NOT FPRIME_DISABLE_TEXT_LOGGERS)` conditional. 

## Rationale

If FW_ENABLE_TEXT_LOGGING is set to 0 (https://github.com/nasa/fprime/blob/401c142c7bfd2f8d48dc40a8db25830985b97043/config/FpConfig.hpp#L290) , the text logger components, ActiveTextLogger and PassiveConsoleTextLogger, will fail to build, as will the entire build. 

Now, deployments that wish to set FW_ENABLE_TEXT_LOGGING to 0 can add `FPRIME_DISABLE_TEXT_LOGGERS = 1` to their CMakeLists to prevent the adding of these components.

None

## Future Work

There may be a more desirable way to do this.

### DIFF
--- a/Svc/CMakeLists.txt
+++ b/Svc/CMakeLists.txt
@@ -11,7 +11,6 @@ add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/WatchDog/")
 
 # Components
 add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/ActiveLogger/")
-add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/ActiveTextLogger/")
 add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/ActiveRateGroup/")
 add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/AssertFatalAdapter/")
 add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/BufferManager/")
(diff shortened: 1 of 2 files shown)


Svc/CMakeLists.txt
```diff
@@ -40,6 +38,13 @@ add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/Time/")
 add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/TlmChan/")
 add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/TlmPacketizer/")
 
+# Allow and check for disabling of text loggers
+# Building these fails if FW_ENABLE_TEXT_LOGGING == 0
+if (NOT FPRIME_DISABLE_TEXT_LOGGERS)
```
LeStarch, line 43: This is exactly the way I'd disable these from the build.  Two comments:

1. Add an option in `cmake/Options.cmake` this makes the switch formal, and allows you to set a default.
2. Our other switches are named in the positive. e.g. `FPRIME_ENABLE_TEXT_LOGGERS` with an appropriate default.

nabcouwer:
@LeStarch , thanks for the feedback, changes made.

LeStarch (approved):

PR #1075, opened 2021-10-05
Separate string size and array size in Serializable XML members

bocchino:
| | |
|:---|:---|
|**_Originating Project/Creator_**| F Prime infrastructure |
|**_Affected Architectures(s)_**| Various |

---
## Change Description

This PR revises the Serializable XML schema so that string size and array size are separate concepts. Formerly, the `size` attribute meant (1) string size if the type was string; otherwise (2) array size. Now `size` always means string size and `array_size` always means array size.

## Rationale

The new version is more self-consistent and also supports FPP. In FPP, string sizes and array sizes are separate concepts.

No particular recommendations.

## Future Work

Close out release 3.0. This is the last issue in the area of FPP/XML compatibility.


bocchino:
I need to remove some instances of `(void)` that crept back into the autocoder.

Autocoders/Python/src/fprime_ac/parsers/XmlSerializeParser.py
```diff
@@ -169,23 +169,48 @@ def __init__(self, xml_file=None):
...
+                        size = member.attrib["size"]
+                        if not size.isdigit():
                             PRINT.info(
-                                "{}: Member {}: size must be a number".format(
+                                "{}: Member {}: array_size must be a number".format(
                                     xml_file, n
                                 )
                             )
                             sys.exit(-1)
+                        if t != "string":
```
LeStarch, line 205: SHouldn't this block exit with an  as on line 204, or otherwise produce an error?

Autocoders/Python/src/fprime_ac/generators/templates/serialize/publicSerialH.tmpl
```diff
@@ -41,20 +41,20 @@ public:
...
     void set($args_proto); //!< set values
 
-#for name,type,size,format,comment,typeinfo in $members:
-#if $size == None:
+#for name,type,array_size,size,format,comment,typeinfo in $members:
+#if $array_size == None and not ($typeinfo == 'string' or $typeinfo == 'extern'):
     ${type} get${name}() const; //!< get member $name
-#else if $typeinfo == 'string' or $typeinfo == 'extern':
+#else if ($typeinfo == 'string' or $typeinfo == 'extern') and array_size == None:
     const ${type}& get${name}() const; //!< get member $name
```
LeStarch, line 48: Why can't we return a reference to the array type?

LeStarch (changes requested):
I left two comments, the second (returning an error) is most important.  I also don't see the changes to `Ref/SignalGen` that should go along with this.

Autocoders/Python/src/fprime_ac/parsers/XmlSerializeParser.py
bocchino: Yes, good catch. I'll make that fix.

Autocoders/Python/src/fprime_ac/generators/templates/serialize/publicSerialH.tmpl
bocchino: If there's a struct member array, we return a pointer to it in line 50. This is also what the previous code did.

The previous code said if the type is string or extern (i.e., serializable), assume no array and return the single element as a reference. The new code relaxes this assumption. It says if we see string or serializable and there's no struct member array, then return the single element as a reference. Otherwise fall through and return the struct member array as a pointer.

I think it's correct.

bocchino: Except the new logic may be unnecessarily complicated. I think the condition in line 47 should just be 'if array_size == None'.

LeStarch (approved):
Approved, concerns dealt with.

PR #1126, opened 2021-11-29
Revise C++ code gen for task start

bocchino:
* Update the FPP commit hash in CMake
* Conform to the new way of invoking task start


LeStarch (approved):
Looks right, although this Ac file needs not be checked in anymore.  Don't worry, it'll get cleaned up with the final release.

PR #1127, opened 2021-11-30
lestarch: noop cmake target and cleaner memo for fprime-util integration

LeStarch:
@bocchino this fixes two minor issues to make `fprime-util` to `fpp-check` integration cleaner.

1. It renames memo files for `fpp` invocations to make them deterministic to find. This is needed use this file as an "export" of data.
2. Adds a "noop" target to CMake.  This changes the "regenerate cache", which always regenerates, to a "noop" build which will regenerate only if necessary. In the small cases where `fprime-util` needs to regenerate (e.g. adding directory) will now be much more efficient in the "nothing changed" case.  Thus errors come out quickly.

Both are leveraged in the fprime-util updates.

### DIFF
--- a/cmake/FPrime.cmake
+++ b/cmake/FPrime.cmake
@@ -63,6 +63,7 @@ else()
     generate_fpp_locs()
     message(STATUS "[autocode/fpp] Generating fpp locator file -- DONE")
 
+    register_fprime_target(target/noop)
     register_fprime_target(target/build)
     register_fprime_target(target/dict)
     register_fprime_target(target/install)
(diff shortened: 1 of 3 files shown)


LeStarch:
Changes here support this PR:  https://github.com/fprime-community/fprime-tools/pull/25

bocchino:
Sounds great! How do I test it? Do I need to install the updated fprime-util first?

LeStarch:
@bocchino to test this PR specifically:

```
fprime-util generate
cd build-fprime-automatic-native
make noop
```

This should "build noop target".  Then just run a normal build:

```
fprime-util build
```

to ensure I didn't break FPP.  CI has checked it above.

LeStarch:
@bocchino this PR doesn't need test with the update fprime-util as it doesn't depend on that, but the new fprime-util depends on this.  So we can merge this safely if the above tests work.

bocchino (approved):
Looks good!

PR #1136, opened 2021-12-03
lestarch: updating CMake UTs and fixing issues

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Reenable CMake UTs for use with v3.0.0 and FPP!!!!


cmake/test/src/test_feature.py
```diff
@@ -0,0 +1,107 @@
...
+# Basic CMake tests.
+#
+####
+import platform
+
+import cmake
+import settings
+
+
+FEATURE_BUILD = cmake.get_build(
```
Joshua-Anderson, line 13: For the other test cases you set the return value to `_` and used the fixture instead.

cmake/test/src/cmake.py
```diff
@@ -7,158 +7,143 @@
...
+            "source": source_directory,
+            "build": build_directory,
+            "install": install_directory_calc,
+            "cmake": cmake_output,
+            "targets": target_outputs,
+        }
+        # shutil.rmtree(build_directory, ignore_errors=True)
+        # shutil.rmtree(install_directory, ignore_errors=True)
+
+    return fixture_function
```
Joshua-Anderson, line 149: Is there a reason you're using a fixture instead of a function return value? In general I consider dyanamic dependency injection a bit of an antipattern. Dependency injunction can be complicated enough by itself, but dynamically creating dependencies can make tracing the injection process nearly impossible. Can you tell I've gotten burned by some crazy java dependency injection environments in the past? 😂 

This is a very simple use case of dependency injection so I'm not too concerned, but I do wonder if there's a better approach that could be used.

Are you returning the fixture to prevent it from going out of scope?

cmake/test/src/test_feature.py
Joshua-Anderson: This is my personal opinion, so feel free to disagree, but I prefer to avoid side effects when including python files. Would it be possible to kick off the cmake build through some pytest init method instead?

Joshua-Anderson (commented):
The refactor overall looks great! I just had a few minor implementation comments.

cmake/test/src/cmake.py
LeStarch: I am using fixtures because `pytest` will handle the cleanup once all the tests are run.  I don't have to worry about adding `.atexit` calls and whatnot to ensure that the fixture has been cleaned-up.  In addition, this allows the reuse of test fixtures without forcing a separate rebuild.

The only caveat is that the fixture must exist in the module using it....hence the return call.  The fixture is returned such that it exists in the calling module.

cmake/test/src/test_feature.py
LeStarch: yes, good point.

cmake/test/src/cmake.py
Joshua-Anderson: I might be overlooking it, but is there any explicit cmake cleanup occurring right now? Either way I'm not too concerned since the build is occurring in temporary directories. In some ways I actually prefer the lack of explicit cleanup so I can poke around the temporary build directory on test failure.

LeStarch: It is those two lines that are commented out.  I need to reenable them.  Good catch!

LeStarch: Ok, reenabled.

LeStarch:
@Joshua-Anderson it is ready for rereview.

Joshua-Anderson (approved):
Looks good!

PR #1138, opened 2021-12-06
lestarch: adding deployment targets to fpp-locs faux target list

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Bug fix for fpp-locs targets.

### DIFF
--- a/Ref/CMakeLists.txt
+++ b/Ref/CMakeLists.txt
@@ -51,7 +51,6 @@ set(SOURCE_FILES "${CMAKE_CURRENT_LIST_DIR}/Top/Main.cpp")
 set(MOD_DEPS ${PROJECT_NAME}/Top)
 
 register_fprime_deployment()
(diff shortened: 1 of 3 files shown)


bocchino (approved):
Looks good. Can you explain what the fake target is doing?

LeStarch:
@bocchino often users like to add compile flags and other settings to our built targets.  e.g.

```
add_compile_options(Ref --Wall)
```

This fix ensures that a target of that name exists in the `fpp-locs` build.  That way the above code does not break when a user adds it to their CMake files.  The user **could** also write:

```
if (TARGET Ref)
   add_compile_options(Ref --Wall)
endif()
```

But this is verbose, and confusing....since that target should always exist, with the exception of the internal fpp-locs build before this fix takes effect.

PR #1142, opened 2021-12-08
Update/final v2 pull

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Hopefully this is the last pull from `nasa/devel` to `nasa/release/v3.0.0` before we deliver.


LeStarch:
@bocchino this is a merge from the v2.0 devel line into release v3.0.0.  It includes fixes from there and the port of those fixes into FPP.

Svc/SystemResources/SystemResources.fpp
```diff
@@ -0,0 +1,146 @@
...
+
+    @ System non-volatile available in KB
+    telemetry NON_VOLATILE_FREE: U64 id 3 \
+      format "{} KB"
+
+    @ FPP from XML: could not translate format string "%3.2f percent"
+    @ System's CPU Percentage
+    telemetry CPU: F32 id 4
+
+    @ FPP from XML: could not translate format string "%3.2f percent"
```
bocchino, line 77: Some format strings were not translated here. We should update them by hand. It should be something like `format "{.2f} percent"`

Os/Baremetal/SystemResources.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+SystemResources::SystemResourcesStatus SystemResources::getCpuTicks(CpuTicks& cpu_ticks, U32 cpu_index) {
+    // Always 100 percent
+    cpu_ticks.used = 1;
+    cpu_ticks.total = 1;
+    return SYSTEM_RESOURCES_OK;
+}
+
+
+SystemResources::SystemResourcesStatus SystemResources::getMemUtil(MemUtil& memory_util) {
```
bocchino, line 29: Does this function violate the rule against dynamic allocation after initialization? Also, it seems to rely on behavior of the memory allocator that may not be true. The allocation could occur anywhere that is free, not necessarily at the free space boundary.

Os/Linux/SystemResources.cpp
```diff
@@ -0,0 +1,137 @@
...
+                return SYSTEM_RESOURCES_ERROR;
+            }
+            if (i != cpu_index) { continue; }
+
+            if (!(line[0] == 'c' && line[1] == 'p' && line[2] == 'u')) {
+                fclose(fp);
+                return SYSTEM_RESOURCES_ERROR;
+            }
+
+            sscanf(line, "%*s %d %d %d %d", &cpu_data[0],
```
bocchino, line 92: The uses of sscanf and fscanf make me a bit nervous. I think they are OK, because no strings are stored, only numbers. Strings are scanned but discarded. This operation should be memory safe. Maybe say that in a comment.

bocchino (changes requested):
Looks great! Just a few comments.

Os/Baremetal/SystemResources.cpp
LeStarch: This implementation doesn't work as expected anyway.  This code can be removed.

LeStarch:
@bocchino I fixed the issues.

bocchino (commented):
Looks good.

In Baremetal/SystemResources.cpp

* The comment in line 4 should say "cpp file" instead of "hpp file"
* Maybe add a comment to the stub memory function, similar to the comments in the other stubs

LeStarch:
@bocchino fixed the last round of comments.

bocchino (approved):
Looks good!

PR #1143, opened 2021-12-08
Convert math tutorial to Markdown

bocchino:
This PR converts the Math tutorial from AsciiDoc to Markdown. As discussed [here](https://github.com/nasa/fprime/discussions/1114), AsciiDoc has some advantages. However, there are issues with the tooling, e.g., GitHub Pages and Doxygen, that require us to use Markdown for now.

This exercise was an interesting study in working around the limitations of Markdown. To get a document comparable in quality to what I had in AsciiDoc, I had to write scripts to do the following:

1. Automatically insert section numbers
2. Automatically insert HTML tags for internal section cross references
3. Automatically generate a hyperlinked table of contents
4. Automatically check for broken cross references

I also needed a script for including files. But I needed this in the AsciiDoc version too, because the GitHub renderer doesn't support AsciiDoc file inclusion properly. Also, the GitHub AsciiDoc renderer doesn't insert the section numbers. So now the GitHub-rendered version has section numbers (an improvement).

For now all the scripts live in `/Users/bocchino/JPL/Tools/FPrime/bocchino/fprime/docs/Tutorials/MathComponent/md`. I focused on a point solution for getting the math tutorial done. We should discuss whether to put these scripts somewhere else, so they can be available for general use.

As I did the conversion, there was a lot of pushing to GitHub to see the document rendered (hence the large number of commits). I found the pandoc HTML rendering too poor to be very useful.


LeStarch (approved):
Looks good to me.  I would recommend we put the tools somewhere general purpose,  but we need to discuss where.  Probably with the other documentation scripts.

PR #1149, opened 2021-12-10
lestarch: initial implementation of RPI integration tests

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Adds RPI integration tests to `v3.0.0`.

## Subcommits

lestarch: bash syntax error in RPI integration

lestarch: misstyped variable name

lestarch: fixing glob pattern

lestarch: missing test directory

lestarch: a more complete artifact

lestarch: fixing copy

lestarch: copying with permissions

lestarch: fixing RPI CI path

lestarch: fixing RPI integration test

lestarch: more careful killing of binary

lestarch: better RPI running from scratch


bocchino:
@LeStarch Looks good! What does this test? For example

* Does it run the deployment on the RPI? If so, how is that working in a CI context?
* Does it check whether LED is blinking? This seems to require visual inspection

LeStarch:
@bocchino this doesn't specifically test the LED.  This runs integration tests like we have for `Ref`.  The big check here is:

"Does this run on RPI hardware without segfaulting."  

Most of the failures on RPI are compilation failures, and the second category of failures cause segfaults/initialization errors so the binary does not stay running.  This test catches the second case, which is a major win for CI.

It would be future work to try and see if the LED blinks correctly...but again this is not the primary source of failures.

bocchino (approved):
Looks good.

PR #1150, opened 2021-12-11
Fixed BasicTypes preproc checks for pointer type

vietjtnguyen:
| | |
|:---|:---|
|**_Originating Project/Creator_**|CADRE|
|**_Affected Component_**|n/a|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|n/a|
|**_Unit Tests Pass (y/n)_**|n/a|
|**_Documentation Included (y/n)_**|n/a|

---
## Change Description
Changes the preprocessor checks for defining the `POINTER_CAST` type so that they're acceptable by the compiler even if the expected architecture-dependent definitions are not present.

## Rationale
When built in the CADRE cross compilation environment with `-Werror` enabled this produces compilation errors pertaining to checking against undefined preprocessor variables (the originating warning was `-Wundef` if I remember correctly).

Test against software built against architectures covered by these different preprocessor definitions that are checked.

## Future Work
n/a

### DIFF
--- a/Fw/Types/BasicTypes.hpp
+++ b/Fw/Types/BasicTypes.hpp
@@ -49,13 +49,13 @@ typedef unsigned int NATIVE_UINT_TYPE; //!< native unsigned integer type declara
  #else
   #define POINTER_CAST U8
  #endif
-#elif __i386 == 1 // GCC 4.1.2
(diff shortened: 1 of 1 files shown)


LeStarch (approved):
All the changes do is to add a "is defined" check before checking values.  I am ok with this change.

PR #890, opened 2021-08-03
lestarch: fixes an overzealous string copy assert

LeStarch:
| | |
|:---|:---|
|**_Originating Project/Creator_**| |
|**_Affected Component_**|  |
|**_Affected Architectures(s)_**|  |
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|  |
|**_Builds Without Errors (y/n)_**|  |
|**_Unit Tests Pass (y/n)_**|  |
|**_Documentation Included (y/n)_**|  |

---
## Change Description

The destination must be large enough to hold the full copy without colliding with source...however; source is not required to  be that large and still not overlap (constant string on stack have this property).  This fixes the resulting bug.

PR #1164, opened 2021-12-15
fix StringUtils overlapping assert

FabrizioSandri:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

There's an error in the `Fw/Types/StringUtils::string_copy` method. 

The original assertion
`FW_ASSERT((source < destination) || destination + num <= source);`

should be changed to 
`FW_ASSERT(destination < source ||  source + num < destination);`

## Rationale
I explain this assertion change by analyzing the 3 possible cases in which the destination string and the source string overlaps:
- Case 0: the two string completely overlaps(source and destination point to the same memory address). This problem is resolved in the first if clause by not executing the copy and returning the destination pointer.
- Case 1: the destination string overlaps with the left part of the source string(_Left overlap_). This case should be allowed since every character of the source string will be copied to the destination string one by one from the left to the right. No information is lost (this was forbidden in the original assertion).
- Case 2: the destination string overlaps with the right part of the source string(_Right overlap_). This case should be forbidden, in fact `strncpy` starts replacing every character of the destination string which causes the change of some parts of the original one. The overlapping characters of the source string will be replaced by the first characters of the original string. In the end, the last part of the source string will be lost.

In conclusion we should allow only 3 cases:
- destination and source are pointing to the same string (resolved in the first if clause by returning the destination)
- any string with `destination < source`, allowing in the worst case a "Left overlap" (Case 1)
- any string with `source < destination` but asserting that it doesn't overlap (`source + num < destination`)

## Graphical explanation
Suppose we have 2 strings of length 5 (one for the case 1 and the other for the case 2) that overlaps with the source string. The source string starts at address `2`, the Case 1 string starts at `0` and the Case 2 string starts at `5`. The graphical representation of the left and right overlap looks like the one in the following schema:

overlaps

As you can see, the Case 1 should be allowed, but the case 2 should be forbidden: by copying the content of the source string to the destination string for the case 2 you will end up changing some parts of the source string that have not yet been copied.

### DIFF
--- a/Fw/Types/StringUtils.cpp
+++ b/Fw/Types/StringUtils.cpp
@@ -8,7 +8,7 @@ char* Fw::StringUtils::string_copy(char* destination, const char* source, U32 nu
         return destination;
     }
 
-    // Copying a right overlapping range is undefined
+    // Copying an overlapping range is undefined
     U32 source_len = string_length(source, num) + 1;
     FW_ASSERT(source + source_len <= destination || destination + num <= source);
 
--- a/Fw/Types/StringUtils.cpp
+++ b/Fw/Types/StringUtils.cpp
@@ -17,6 +17,6 @@ char* Fw::StringUtils::string_copy(char* destination, const char* source, U32 nu
     return returned;
 }
 
-U32 Fw::StringUtils::string_length(CHAR* source, U32 max_len) {
+U32 Fw::StringUtils::string_length(const CHAR* source, U32 max_len) {
     return strnlen(source, max_len);
 }
--- a/Fw/Types/StringUtils.cpp
+++ b/Fw/Types/StringUtils.cpp
@@ -8,8 +8,8 @@ char* Fw::StringUtils::string_copy(char* destination, const char* source, U32 nu
         return destination;
     }
 
-    // Copying an overlapping range is undefined
-    FW_ASSERT(source < destination || destination + num <= source);
+    // Copying a right overlapping range is undefined
+    FW_ASSERT(destination < source ||  source + num < destination);
 
     char* returned = strncpy(destination, source, num);
     destination[num - 1] = '\0';
--- a/Fw/Types/StringUtils.cpp
+++ b/Fw/Types/StringUtils.cpp
@@ -9,9 +9,14 @@ char* Fw::StringUtils::string_copy(char* destination, const char* source, U32 nu
     }
 
     // Copying a right overlapping range is undefined
-    FW_ASSERT(destination < source ||  source + num < destination);
+    U32 source_len = string_length(source, num) + 1;
+    FW_ASSERT(source + source_len <= destination || destination + num <= source);
 
     char* returned = strncpy(destination, source, num);
     destination[num - 1] = '\0';
     return returned;
+}
+
+U32 Fw::StringUtils::string_length(CHAR* source, U32 max_len) {
+    return strnlen(source, max_len);
 }
--- a/Fw/Types/StringUtils.hpp
+++ b/Fw/Types/StringUtils.hpp
@@ -30,7 +30,7 @@ char* string_copy(char* destination, const char* source, U32 num);
  * \param max_len: the maximum length of the source string
  * \return length of the source string or max_len
  */
-U32 string_length(CHAR* source, U32 max_len);
+U32 string_length(const CHAR* source, U32 max_len);
 
 };      // namespace StringUtils
 };      // namespace Fw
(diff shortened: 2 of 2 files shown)


LeStarch:
From several sources of C++ documentation, I see the phrase:

"The behavior is undefined if the character arrays overlap"

This implies to me that no overlapping is allowed, regardless if it is a left overlap, right overlap or otherwise.  Most implementations do a left to right copy, but I don't think that is strictly guaranteed.  I think that implies that we need the following to prevent any overlap:

```
FW_ASSERT(source + num <= destination || destination + num <= source);
```

@Joshua-Anderson do you have any recollection of this?  I remember we had some issues with the original assert that was in this code, and we changed it.  That change was here, but upon review, I believe the original assert was the correct one.

https://github.com/nasa/fprime/commit/f2c791a07bf7992909ffcc970a3122a0e8753e0b#diff-ea23df732919393f046f1b51f7c971aef3e838d666928af2a658d01b1f686827

Joshua-Anderson:
> This implies to me that no overlapping is allowed, regardless if it is a left overlap, right overlap or otherwise. Most implementations do a left to right copy, but I don't think that is strictly guaranteed.

Yeah, I would expect a right shift to work on most strcpy implementations but given F' targets a wide variety of platforms and compilers that's not something I'd feel conformable depending on.

>  That change was here, but upon review, I believe the original assert was the correct one.

The original assert was broken and we hit CI failures with it. I believe it was copying a string that was smaller than max size length that was next to the destination in memory or something of that sort

FabrizioSandri:
I agree that a possibile soultion is to completely forbid overlaps of any type (left or right overlaps) by using the following assertion:
`FW_ASSERT(source + num <= destination || destination + num <= source);`

However, I don't understand how CI could fail.

LeStarch:
@FabrizioSandri my guess is some of the code somewhere has a dependance on the undefined overlapping behavior.

If you are up for it, want to make the change back to "FW_ASSERT(source + num <= destination || destination + num <= source);" and then we can trigger CI and see where it fails?

Joshua-Anderson:
@LeStarch I'm pretty sure this bug snuck through because it passed the F' core framework CI but broke an internal JPL project - you brought this up as a regression and fixed it in #890. I believe it asserted on topology initialization. So testing this on F' CI probably won't catch the error.

LeStarch:
@Joshua-Anderson good point, thanks for the reminder.  Looking at #890, what we really want to do (conceptually) is this:

```
FW_ASSERT(source + source_len <= destination || destination + num <= source);
```

As this prevents overlap, and also does not assume that source is as long as the destination.  This would make our (revised) function look more like:

```
U32 source_len = strnlen(source, num);
FW_ASSERT(source + source_len <= destination || destination + num <= source);
```

We might also consider implementing a string_length function:

```
U32 Fw::StringUtils::string_copy(CHAR* source, U32 max_len) {
    return strnlen(source, max_len);
}
```

The reason to do this is we want to migrate away from raw "string.h" functions and thus this would help.

Joshua-Anderson:
Is it worth the computational cost of calling strlen just to strictly enforce the no-overlap assertion? At the time, I think we decided the current assertion was "good enough" and would catch most overlap errors. Hopefully the strcpy implementation has it's own internal error handling for overlaps

LeStarch:
@Joshua-Anderson as pointed out, we do have an overlap bug as a source longer than 1 may be clobbered without being caught.

Given that the `strncpy` function documentation specifically points out that the behavior is undefined, it is telling us that we **can not** assume it handle overlaps.  (A C++ mentor of mine once pointed out, "undefined behavior" is up to  implementer who may choose to wipe your hard drive.)

Thus, to do this correctly we have two choices:

1. Push this up to the caller and let them decide
2. Call for string length as proposed.

The reason we have this function at all is that `strncpy` pushed this problems to us. Given this is for high-reliability embedded software, I think it is imperative we choose option 2.  We often choose safety over raw performance.  If we did not, we'd never use asserts, or parameter checks at all.

FabrizioSandri:
@LeStarch 
> We might also consider implementing a string_length function:
> 
> ```
> U32 Fw::StringUtils::string_length(CHAR* source, U32 max_len) {
>     return strnlen(source, max_len);
> }
> ```

By using `strnlen` function we have to consider that this function excludes the terminating null byte. 
From this consideration, I think that a "+1" is also needed
```
U32 source_len = strnlen(source, num) + 1;
FW_ASSERT(source + source_len <= destination || destination + num <= source);
```

LeStarch:
@FabrizioSandri good catch.  Do you want to make the above changes (adding a string_length function and fixing the assert to differentiate source and destination lengths)?

FabrizioSandri:
@LeStarch I added the mentioned changes

LeStarch (approved):

LeStarch:
@FabrizioSandri looks good.  We'll let CI run.  When it passes we can merge.

PR #1167, opened 2021-12-17
Update/final preparations for v3

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Three final changes before `v3.0.0`.  I'd review this by looking at each commit individually, rather than just the changed files, or ignore deletions.

1.  Fix a UT to cap tlm at 16
2. Improve installation and quick installation instructions
3. Delete many files not needed in v3.


bocchino:
I think we should keep `Ref/Top/visual.do`. Otherwise there's no tool for generating the topology graphs. We could turn it into a regular shell script to remove the dependency on redo.

bocchino:
> Otherwise there's no tool for generating the topology graphs.

OK, there is the browser-based visualizer. Maybe we should just direct users that way. Personally I find it convenient to run a script and get the .eps files.

LeStarch:
@bocchino this seems like it should be merged into the fprime-util tool.  I'll add a ticket on that.

.lgtm.js
```diff
@@ -0,0 +1 @@
+function appease_lgtm() {}
```
Joshua-Anderson, line 1: Possibly a silly question, but do we care if LGTM analysis fails for javascript? It fails for all the other lanugages F' doesn't use

bocchino:
Yes, I agree. There should an automated way to generate the JSON layout files. Then one can run the browser-based visualization tool on them. It would be good to generate the .eps files too, if pic is available on the host system.

Here is the JSON script we should automate: https://github.com/nasa/fprime/blob/devel/docs/Tutorials/MathComponent/Tutorial.md#Updating-the-Ref-Deployment_Visualizing-the-Ref-Topology.

Joshua-Anderson (approved):
This PR changed so many files that my web browser freezes on the diff :) I went through and reviewed all the individual commits and then I checked that 97de3c264553a5348ecf6c26d9e30cdc95aa2567 only touched the magicdraw and redo files.

PR #1240, opened 2022-01-26
Clean autocoders code

iw4p:
| | |
|:---|:---|
|**_Originating Project/Creator_**| iw4p |
|**_Affected Component_**| Autocoder  |
|**_Affected Architectures(s)_**| Python |
|**_Has Unit Tests (y/n)_**| exist |

---
## Change Description
Check and lint codes via pylint to find out what's missing or wrong and clean the code.

## Rationale
Fixed bugs.

Built with `cmake ../ -DCMAKE_BUILD_TYPE=TESTING` and `make check`.

## Future Work

Nothing for now.

### DIFF
--- a/Autocoders/Python/bin/gds_dictgen.py
+++ b/Autocoders/Python/bin/gds_dictgen.py
@@ -192,6 +192,7 @@ def main():
     #
     if len(args) == 0:
         print("Usage: %s [options] xml_filename" % sys.argv[0])
+        return
     elif len(args) == 1:
         xml_filename = args[0]
     else:
--- a/Autocoders/Python/bin/gds_dictgen.py
+++ b/Autocoders/Python/bin/gds_dictgen.py
@@ -192,12 +192,10 @@ def main():
     #
     if len(args) == 0:
         print("Usage: %s [options] xml_filename" % sys.argv[0])
-        return
     elif len(args) == 1:
         xml_filename = args[0]
     else:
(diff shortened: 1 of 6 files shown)


Autocoders/Python/bin/gds_dictgen.py
```diff
@@ -192,12 +192,10 @@ def main():
     #
     if len(args) == 0:
         print("Usage: %s [options] xml_filename" % sys.argv[0])
-        return
```
LeStarch, line 195: Return is necessary here.

LeStarch (changes requested):
Many of the returns that were removed were necessary for the code to function.  In all cases where a return is on an error branch they need to be restored.

Autocoders/Python/bin/gds_dictgen.py
iw4p: Oh, I think it's fine now.

Autocoders/Python/bin/implgen.py
```diff
@@ -218,12 +218,10 @@ def main():
     #
     if len(args) == 0:
         print("ERROR: Usage: %s [options] xml_filename" % sys.argv[0])
-        return
```
LeStarch, line 221: This return is still removed, which will change the behavior

iw4p:
> Two more missing returns. Other than that, it looks good!

I think it's fine now, Thank you.

LeStarch (approved):

Issue #1193, opened 2022-01-03
Having trouble generating fprime-util when building the Ref application


arshiuh:
Hello! I am trying to follow the Getting Started section and checking my F' Installation, and it seems that the autocoder did not install properly. When I cd into the Ref folder and enter `fprime-util generate`. It says `fprime-util: command not found`. I am most likely installed it incorrectly, but I am slightly stuck on where to go from here.

**System Versions:**
Running Linux Ubuntu 20.04.3 on VirtualBox

Java Version
`openjdk 11.0.13 2021-10-19
OpenJDK Runtime Environment (build 11.0.13+8-Ubuntu-0ubuntu1.20.04)
OpenJDK 64-Bit Server VM (build 11.0.13+8-Ubuntu-0ubuntu1.20.04, mixed mode, sharing)`

CMake
`cmake version 3.16.3
CMake suite maintained and supported by Kitware (kitware.com/cmake).
`

Bash
`GNU bash, version 4.4.0(1)-release (x86_64-unknown-linux-gnu)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>`

CLang
`clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin`

Python & PIP
`Python 3.8.10
pip 20.0.2 from /usr/lib/python3/dist-packages/pip (python 3.8)
`

ThibFrgsGmz:
Do you have the python virtual environment activated?

LeStarch:
`fprime-util: command not found` is almost always caused by one of two issues:

1. `fprime-util` was installed, but not in path.  On some systems, `pip` places user installed tools in `.local` and this needs to be added to the path.
2. `pip install fprime-tools` was not run.

Point 1 above is not noted in the installation guide.  We'll need to fix that.

arshiuh:
@LeStarch Point 1 seems to have fixed it. Thank you very much!

LeStarch:
We need to add a note in install instructions to add .local to path or run as an admin.

PR #1253, opened 2022-02-01
Add A Note For User-Installed Pip Packages

astroesteban:
| | |
|:---|:---|
|**_Originating Project/Creator_**| @astroesteban |
|**_Affected Component_**| INSTALL.md |
|**_Affected Architectures(s)_**| Docs |
|**_Related Issue(s)_**| #1193  |
|**_Unit Tests Pass (y/n)_**| Y  |
|**_Documentation Included (y/n)_**| Y  |

---
## Change Description

Add a note to users that sometimes pip will install packages in a folder that needs to be added to the user's `$PATH` environment variable.

## Rationale

Per issue #1193 it can be unclear to users why the `fprime-util` command fails. This adds some clarification to the Python package installation step. 

None

## Future Work

None.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -51,6 +51,10 @@ tools package. This is to enable users to choose which tools they'd like to use.
 pip install --upgrade fprime-tools fprime-gds
 ```
 
+> Note: On some systems pip places user-installed tools in the `.local` folder.
+This results in a `fprime-util: command not found` error when trying to run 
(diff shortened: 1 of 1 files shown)


LeStarch:
Good catch.  I think this can happen if you don't run with admin privileges

astroesteban:
It's a pretty non-obvious issue to new pip users too. That little note should help.

LeStarch (approved):

LeStarch:
Documentation only fix and spelling passed.

PR #1272, opened 2022-02-09
Avoid violating `-Wpedantic`

myint:
This cleans up the GCC/Clang compiler warnings like:

    extra ‘;’ [-Wpedantic] }; // namespace StringUtils

| | |
|:---|:---|
|**_Originating Project/Creator_**| JPL COLDArm / Steven Myint |
|**_Affected Component_**| Multiple |
|**_Documentation Included (y/n)_**| no |

---
## Change Description

Remove unnecessary semicolons from `Fw/Types/StringUtils.hpp`, `Os/Posix/TaskId.cpp`, `Os/TaskIdRepr.hpp`, and `Svc/CmdSequencer/CmdSequencerImpl.hpp`.

## Rationale

Without this, we get warnings from GCC and Clang.

None

## Future Work

None

### DIFF
--- a/Fw/Types/StringUtils.hpp
+++ b/Fw/Types/StringUtils.hpp
@@ -32,6 +32,6 @@ char* string_copy(char* destination, const char* source, U32 num);
  */
 U32 string_length(const CHAR* source, U32 max_len);
 
-};      // namespace StringUtils
-};      // namespace Fw
+}       // namespace StringUtils
+}       // namespace Fw
 #endif  // FW_STRINGUTILS_HPP
--- a/Os/Posix/TaskId.cpp
+++ b/Os/Posix/TaskId.cpp
@@ -7,7 +7,7 @@
 
 extern "C" {
 #include <pthread.h>
-};
+}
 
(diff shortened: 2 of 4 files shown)


LeStarch (approved):

myint:
@Joshua-Anderson It's a slightly older commit of FPrime (37b4bfe), but @Kronos3 got `-pedantic` working as part of SSim's FPrime build:

https://github.jpl.nasa.gov/rsvp/ssim-coldarm/blob/8cf77034293afb7a185f21de380137ecfd713046/src/CMakeLists.txt#L12

With this configuration, we get warnings from GCC (4.8.5). Clang can produce a similar warning about semicolons via `-Wextra-semi` (but that happens not to be supported in the aforementioned older GCC 4.8.5).

Joshua-Anderson:
4.8.5 is a pretty old version of GCC, I suspect it doesn't have the same checks that are failing on newer versions of GCC and that may be why that project isn't having any warnings. Hmm, it may be worth adding `-Wextra-semi` to our code quality CI builds to prevent regressions.

Issue #667, opened 2021-06-02
Adding more detail on cross compiling for math component tutorial. 


rawinza555:
| | |
|:---|:---|
|**_F´ Version_**|devel |
|**_Affected Component_**|Math Components Tutorials  |
---
## Feature Description

Expand the math components more on the deploying to RPI or any other board. Basically, have something similar to cross compile section in R Pi tutorial but for math components.

## Rationale

From a newbie perspective, it's easier to understand how to run a GDS on host machine to interact with R Pi with more simple example like math components. It's more likely for people to have a R Pi or other SBC lying around than having that and a GPS module. A GPS module tutorial can then be focused for learning on hardware interaction.

LeStarch:
This was done!

PR #1274, opened 2022-02-11
astroesteban: Add Cross Compilation Section To Math Tutorial

astroesteban:
| | |
|:---|:---|
|**_Originating Project/Creator_**| @astroesteban |
|**_Affected Component_**| Math Component Tutorial |
|**_Related Issue(s)_**|  #667 |

---
## Change Description

Adds a new section to the Math Component tutorial that guides the user through cross-compiling their new Ref topology for the Raspberry Pi. It also provides instructions on how to copy the Ref deployment to the Raspberry Pi and get the `fprime-gds` and Ref binary running on the host and Raspberry Pi respectively.

## Rationale

The jump from the Math Component tutorial to the GPS tutorial might be a bit too steep for some users and using the simpler Ref topology they just created makes it easier. On top of that some users might not have a GPS receiver around but would still like some information on how to get their cool new Ref topology on their Raspberry Pi. I agree with issue #667 that adding a section that walks users through cross-compiling their new Ref topology for the Pi would be helpful in at least getting a taste for cross-compiling their Ref topology and getting their deployment on the target hardware.

- The new steps were completed using a Raspberry Pi 3B+

**Thoughts on adding this section? Feel free to reject it if it doesn't fit.**

## Future Work

We could get as detailed as we'd like with this. We could have a subsection for how to define a new custom toolchain or how to use the `settings.ini` file. For now I've kept it brief.


LeStarch:
@timcanham what are your thoughts on adding a cross-compile section to the math tutorial?

LeStarch:
@astroesteban we are concerned about putting yet more into the MathComponent.  However, we agree that using the MathComponent as a base to learn cross compiling would be useful.  Could we ask for the following refactor:

1. Make this its own tutorial, with the required knowledge to have completed the MathComponent Tutorial
2. Add a link from the Math component tutorial to this new tutorial

That way it is separate and won't make users feel like cross-compilation is a necessary step to the original tutorial, but we maintain this new content for users who need it.

I certainly agree that the GDS tutorial is not the best guide to cross compiling.

astroesteban:
@LeStarch Awesome. That makes total sense to me too and I totally agree with everything. I'll work on updating this pull request accordingly. Thanks for the feedback!

astroesteban:
### Description of New Changes

1. Create a new CrossCompilation folder to hold the new Tutorial.md for cross-compiling to the Raspberry Pi.
2. Remove the Cross-Compilation section from the Math Component tutorial.
3. Add a Conclusion section to the Math Component tutorial to refer the user back to the Tutorials page or to the Cross-Compilation tutorial.
4. Update the Tutorials README to include the new Cross-Compilation section.

LeStarch (approved):
Awesome!!! This is exactly what I was expecting.  I've started CI. Once that passes we can merge!

PR #1275, opened 2022-02-15
Update/fix executable

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes executables.


LeStarch:
@Joshua-Anderson might you find some time to review this?  I think you are our next best CMake expert and this whole PR is cmake.

Joshua-Anderson:
I'll take a look tomorrow

cmake/target/build.cmake
```diff
@@ -108,7 +75,11 @@ function(setup_build_module MODULE SOURCES GENERATED EXCLUDED_SOURCES DEPENDENCI
             add_dependencies(${MODULE} "${DEPENDENCY}")
             list(APPEND TARGET_DEPENDENCIES "${DEPENDENCY}")
         endif()
-        target_link_libraries(${MODULE} PUBLIC "${DEPENDENCY}")
+        # Avoid linking custom targets, and executables
+        is_target_library(IS_LIB "${DEPENDENCY}")
+        if (NOT TARGET "${DEPENDENCY}" OR IS_LIB)
```
Joshua-Anderson, line 80: What are dependencies that aren't targets? I'm trying to understand what this case is trying to cover.

Joshua-Anderson (commented):
Changes look good! I just had a few questions to make sure I fully understood the changes.

cmake/utilities.cmake
```diff
@@ -178,6 +178,43 @@ function(resolve_dependencies OUTPUT_VAR)
...
+# Does this target represent a real library? OUTPUT is set to TRUE when real, and FALSE otherwise.
+#
+# OUTPUT: variable to set
+# TEST_TARGET: target to set
+####
+function(is_target_library OUTPUT TEST_TARGET)
+    set("${OUTPUT}" FALSE PARENT_SCOPE)
+    if (TARGET "${TEST_TARGET}")
+        get_target_property(TARGET_TYPE "${DEPENDENCY}" TYPE)
+        if (NOT TARGET_TYPE STREQUAL "UTILITY" AND NOT TARGET_TYPE STREQUAL "EXECUTABLE")
```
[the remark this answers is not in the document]
LeStarch: There are multiple types of libraries: shared, interface, archive, and possibly more?  Let me review and see if there is another way to approach this.

cmake/target/build.cmake
LeStarch: You can have a dependency on executables to make them build as part of a deployment.  However, these should not be linked.  Utility targets (custom targets) behave similarly.

cmake/API.cmake
```diff
@@ -331,13 +330,13 @@ endfunction(register_fprime_executable)
 # ```
 ####
 function(register_fprime_deployment)
+    get_module_name("${CMAKE_CURRENT_LIST_DIR}")
     if (NOT DEFINED SOURCE_FILES AND NOT DEFINED MOD_DEPS)
         message(FATAL_ERROR "SOURCE_FILES or MOD_DEPS must be defined when registering an executable")
+    elseif(NOT MODULE_NAME STREQUAL PROJECT_NAME)
+        message(WARNING "Project name ${PROJECT_NAME} does not match expected name ${MODULE_NAME}")
```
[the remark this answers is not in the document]
LeStarch: Currently, many things pull from $PROJECT_NAME.  This needs to be consistent with the deployment name.  This is a common source of error.

As a side note, I don't think this works as written.  The second token should be "${PROJECT_NAME}".  This seems to be a CMake quirk contrary to the documentation.

I figured it was easiest to pin the two together, however we could change that in another PR if you think that will help.

cmake/target/build.cmake
Joshua-Anderson: If that's the case, couldn't this check just be `if (IS_LIB)`? What other than libraries are you trying to link against?

LeStarch: @Joshua-Anderson IS_LIB returns FALSE in the case that the TARGET does not exist. However, you can specify link dependencies before actually creating those libraries.  Thus, this code does the following:

1. If the dependency is not yet a TARGET assume it is a library (common case) and link against it.
2. If the dependency is a a library (defined as a library type) also link against it.

LeStarch: Sorry, I misunderstood the reason behind this code.  Here is a clearer explanation.  We need to call `target_link_libraries` on two types of items:

1. Linker flags (e.g. -lrt)
2. Defined libraries

It is the link flags that are not targets and thus pass the NOT TARGET check.  Perhaps this should be implemented as a "starts with `-l`" check

cmake/utilities.cmake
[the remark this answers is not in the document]
LeStarch: @Joshua-Anderson fixed.  I now look for CMake Target types ending in "_LIBRARY" which accounts for all CMake library types as near as I can tell.

cmake/target/build.cmake
LeStarch: I cleaned this up and documented what is going on with a few local comments, better abstraction, and an explanation of caveats with `register_fprime_executable`.

Note: we have one other thing to link

3. Undefined tagets (that will be defined) are assumed to be libraries for backwards compatibility. Thus, `add_dependencies(TARGET EXECUTABLE)` is preferred to MOD_DEPS to build executables as dependencies.

The above is now documented.

LeStarch:
@Joshua-Anderson thanks for the review.  This is ready for another look.

Joshua-Anderson (approved):
LGTM!

LeStarch:
@Joshua-Anderson you rock.  Thanks!

PR #1304, opened 2022-03-04
Improvements to Circular Buffer

bocchino:
|Change|Rationale|
|----|----|
|Revise implementation to use a head index and an allocated size instead of head and tail pointers|Simplifies the implementation. Eliminates the one-byte discrepancy between the physical and logical store sizes.|
|Replace `get_remaining_size(bool)` with `get_allocated_size` and `get_free_size`|The interface is easier to understand.|
|Add `const` to interface functions where possible|C++ coding practice
|Add SDD|Required documentation


bocchino:
@LeStarch CI is failing because of setup dependencies. Can you investigate?

LeStarch:
@bocchino I have to fix CI first.  Could you review: https://github.com/nasa/fprime/pull/1314?  It should fix the problem

bocchino:
I reviewed and approved #1314.

LeStarch (approved):

LeStarch:
Just looked through this and it looks like what I'd expect!  CI passes, so I am going to merge it.

PR #1330, opened 2022-03-21
Revise Circular Buffer unit tests

bocchino:
* Fix memory check warnings in tests
* Improve code coverage to 100%
* Add tests for boundary cases
* Remove `- 1` adjustment from buffer sizes in tests, matching the latest implementation
* Seed the random number generator in the test main function


Utils/Types/test/ut/CircularBuffer/Main.cpp
```diff
@@ -172,7 +172,22 @@ TEST(CircularBufferTests, BasicRotateBadTest) {
...
+    Types::SerializeOkRule serializeOk("serializeOk");
+    serializeOk.apply(state);
+    // Serialize a max size buffer
+    state.setRandom(MAX_BUFFER_SIZE, 0, 0);
+    serializeOk.apply(state);
+}
+
 int main(int argc, char* argv[]) {
     ::testing::InitGoogleTest(&argc, argv);
+    STest::Random::seed();
```
LeStarch, line 191: Without this statement was it always running the same cases?  If so, we should ensure that is called in the other rules-based testing.

LeStarch (approved):

Utils/Types/test/ut/CircularBuffer/Main.cpp
bocchino: Yes, we should call it. I believe if you don't seed the pseudo-random number generator, then you get a default seed.

bocchino: Also, there's a way to capture and repeat a particular seed, but it requires going into the CMake build cache to execute the test. Ideally we should add this capability to the fprime-util interface.

arizvi786 (approved):
looks good to me

PR #1337, opened 2022-03-22
F´ Metapackage, Removal of FPP Installation of CMake, and Actions Changes

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

This changes the "easy install" of fprime tools from various pip packages and other tarballs or  happening within CMake (inducing delay for every build) to installings these tools using `cd fprime; pip install .`. This engages an fprime meta-package that pulls in the other packages.

The specific changes are:
1.  Add setup.py only depending on other packages
2. Remove FPP cmake auto-install
3. Update github actions to use the metapackage


setup.py
```diff
@@ -0,0 +1,40 @@
...
+        "Operating System :: POSIX :: Linux",
+        "Programming Language :: Python",
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+    ],
+    python_requires=">=3.6",
+    install_requires=["fprime-tools==3.0.1", "fprime-fpp==1.0.1", "fprime-gds==3.0.1"],
```
Joshua-Anderson, line 39: I thought part of the goal of breaking fprime-util and fprime-gds into their own packages is that we could upgrade these packages independently. Would it be possible to make this a semver match, so that the tools install 3.0.1 or newer (minus a major version bump).

LeStarch: @Joshua-Anderson they can be versioned independently.  This meets the need of a set "known configuration" that is tested and known to work.  Users who want differing toolset versions would not install this metapackage.

LeStarch:
@Joshua-Anderson are you commenting here because you want to review this?  It needs reviewers.....

LeStarch:
Or did I absentmindedly assign you already?

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,15 @@
...
+inputs:
+  location:
+    required: true
+    default: $GITHUB_WORKSPACE
+runs:
+  using: "composite"
+  steps:
+  - run: git fetch --tags
+    shell: bash
+  - run: pip3 install setuptools_scm urllib3
```
Joshua-Anderson, line 12: Would it be possible to make these dev dependencies of the setup.py file? If we have dependencies we should make sure they're properly documented through a requirements file/setup.py/install instructions rather than hidden in a github action.

.lgtm.yml
```diff
@@ -29,7 +29,8 @@ extraction:
             # setup the venv and install packages
             - "python3 -m venv ./fprime-venv"
             - ". ./fprime-venv/bin/activate"
-            - "pip install fprime-tools fprime-gds"
+            - "pip install urllib3 setuptools_scm"
```
Joshua-Anderson, line 32: Same as my above comment

cmake/required.cmake
```diff
@@ -3,25 +3,28 @@
...
 
 # Check python was found
 if (NOT FPUTIL)
     message(FATAL_ERROR "fprime-util was not found. Please install with 'pip install fprime-tools'")
 elseif (NOT PYTHON)
     message(FATAL_ERROR "python3 was not found. Please ensure python3 is on the path.")
-elseif(NOT JAVA)
-    message(FATAL_ERROR "java was not found. Please ensure java is on the path.")
+elseif(NOT FPP_FOUND)
+    message(FATAL_ERROR "fpp tools where not found. Please install the fpp tool suite.")
```
Joshua-Anderson, line 21: where->were

Joshua-Anderson (commented):
I \*was\* just lurking, but I guess I outed myself 😂.. I love the cleanup of the github actions setup step! Just a few minor comments, but everything looks good in general.

.github/actions/setup/action.yml
LeStarch: No, these dependencies are of the setup.py file, and must exist before pip is run. The fix is to not use them.  That will is already fixed but will come in when we publish fpp version 1.0.2.

.lgtm.yml
LeStarch: No, these dependencies are of the setup.py file, and must exist before pip is run. The fix is to not use them.  That will is already fixed but will come in when we publish fpp version 1.0.2.

.github/actions/setup/action.yml
Joshua-Anderson: Does this need to be documented anywhere for users of this metapackage?

LeStarch: I don't think so.  I am not officially releasing this package until v3.0.1 and v1.0.2.  That is when I'll add it to the install instructions and it should work without that preinstall.

Until the release, I need CI to pass.

PR #1348, opened 2022-03-29
Work/aligned alloc

timcanham:
| | |
|:---|:---|
|**_Originating Project/Creator_**| JPL Internal|
|**_Affected Component_**| n/a |
|**_Affected Architectures(s)_**| all |
|**_Related Issue(s)_**|  |
|**_Documentation Included (y/n)_**| y, header file |

---
## Change Description

This PR adds the `AlignedAllocator` memory allocator type to the `Fw/Types` directory.

## Rationale

Hardware interfaces can require memory buffers aligned on certain boundaries (DMA for example). The normal `MallocAllocator` does not have this ability. The `AlignedAllocator` can be used with `BufferManager` to supply buffers to drivers as needed.

Code review should be sufficient, especially use of the `::aligned_alloc()` call.

## Future Work

None


LeStarch (approved):

LeStarch:
@timcanham minor comment.  You assert that size % alignment == 0, but not that alignment is a power of 2.  Do you need to?

"
The obsolete function memalign() allocates size bytes and returns a pointer to the allocated memory. The memory address will be a multiple of alignment, **which must be a power of two.**

The function aligned_alloc() is the **same as memalign()**, except for the added restriction that size should be a multiple of alignment. 
"

timcanham:
Good point. I'll add an assert for that as well.

PR #1352, opened 2022-03-31
Handle 0 and negative numbers on string_copy

Drosaca:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Avoids  to write '\0' before  the destination pointer if num is 0
which is set by the following line:
> destination[ num - 1] = '\0'

### DIFF
--- a/Fw/Types/StringUtils.cpp
+++ b/Fw/Types/StringUtils.cpp
@@ -3,8 +3,8 @@
 #include <cstring>
 
 char* Fw::StringUtils::string_copy(char* destination, const char* source, U32 num) {
-    // Handle self-copy
-    if(destination == source) {
+    // Handle self-copy and negative numbers
+    if(destination == source || num <= 0) {
         return destination;
     }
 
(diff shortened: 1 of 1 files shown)


ThibFrgsGmz:
The verification of the negative number seems useless because the number passed in parameter is not signed. Right ?

Drosaca:
@ThibFrgsGmz Indeed the check of negative values has been removed, nevertheless when num equals zero '\0' will still be written at destination[-1]

LeStarch (approved):

LeStarch:
I agree with this improvement.  In the case where 0 is supplied as a limiting length, then the destination should return unchanged as the limit does not even account for a terminating `\0`.  This will will remove the error.

LeStarch:
Closed and reopened as I needed to try and re-trigger a complete rerun of CI to fix the python formatting bug.  If it doesn't work, we'll ignore that check as the python was not touched.

PR #1353, opened 2022-04-02
Using TAB escape sequence instead of pure spaces

lnxpy:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

- Using the escape sequences instead of spaces.
- Using [isort](https://isort.readthedocs.io/en/latest/) in order to sort imports in the entire project.

## Rationale

In some cases, like when you want to stream the STDOUT and you are about the stream a huge amount of data like log records, using the pure spaces might break the responsiveness in the CLI. Better approach is to use sequences. In spite, I found some escape sequences like `\n` in the code, but there were pure spaces (4 spaces) instead of `\t`. It also helps for better readability.

Please make sure you run a full test. Despite of isort's power, there still might be some special cases that isort may cause of import errors and stuff. However, I've never faced any of them.


LeStarch (approved):
Looks good to me.  My guess is the python formatter is going to break on the changes where lines have collapsed.... but this is for CI to catch.

In the future, I'd suggest splitting such PRs into two, as the original (tabs) was easy to review, but the larger one at 62 files took much longer to review.

Regardless, thanks for the PR.  We'll see how CI does.

lnxpy:
Thank you. I actually started using escape sequences in only one file from the project to begin this methodology of logging or even printing stuff to the CLI in a better scheme and shape. Almost all those 61 files are being modified because I tried to sort the import lists of the entire project and I highly believe that CI failure must be due to second commit.

LeStarch:
Closing and reopening as that seems to force a full rerun of the failing CI.

LeStarch:
@lnxpy it might be, but we've had a failing CI for a week that was resolved only today.  Thus I am rerunning all things before I take the failures seriously.

LeStarch:
@Inxpy yes, now the formatter is balking because the tool you ran changed the format.  To fix, I recommend you run:

```
pip install click==8.0.4 black==21.6b0
cd fprime
black .
```

That should clean up the files and fix the issue!  It will still keep the sorted imports, but will correct the collapsed lines.  Let me know if you need help running this!

lnxpy:
@LeStarch Would you please close and reopen for another CI rerun?

LeStarch (approved):

LeStarch:
This is great work.  Thanks for getting this prepared!

PR #1390, opened 2022-04-13
lestarch: splitting FileDownlink from types. Fixes: #1165

### LINKED ISSUE #1165: Separate Modules and Ports/Types
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

Ports and types are defined within the same module as components.    The analysis is below, but the result of this analysis is that the `FileDownlink` component needs its types/ports split from its component definition.  The reason: these ports and types are used externally by other components and thus causes an inter-module dependency that is unnecessary.

Must split:
[Svc/FileDownlink/FileDownlink.fpp](https://github.com/nasa/fprime/blob/devel/Svc/FileDownlink/FileDownlink.fpp) : enum struct port component

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes #1165.


LeStarch:
Failure is random and part of the actions systems.  Ignoring.

ThibFrgsGmz (approved):

timcanham (approved):

PR #1408, opened 2022-04-14
Older compiler compatibility changes

timcanham:
| | |
|:---|:---|
|**_Originating Project/Creator_**| JPL Internal |
|**_Affected Component_**| All |
|**_Related Issue(s)_**|  |

---
## Change Description

Modified headers and code to make it more compatible with C++98 gcc compiler

## Rationale

We are required to support this older compiler for a JPL project

None

## Future Work

None


LeStarch (approved):
Works for now, still need to discuss long term solution.

PR #1423, opened 2022-04-28
Fix Doxygen documentation organization and move references into respective subpages

cadomani:
| | |
|:---|:---|
|**_Originating Project/Creator_**| @cadomani - ABEX (Alabama Burst Energetics eXplorer) |
|**_Affected Component_**| Doxygen Documentation |
|**_Affected Architectures(s)_**| N/A |

---
## Change Description

The generated Doxygen C++  documentation doesn't seem to properly nest components inside subpages in some instances, and other components had broken links within them. This PR fixes those references and properly nests components.

A visual reference is shown here:

| Original | Pull Request |
|:---|:---|
|image|image|

## Rationale

C++ documentation is not properly segmented, and I couldn't find any indication why some subpages were in the root and others were in their respective folders, so I assume it wasn't always updated when documentation changed. This makes browsing through the documentation more efficient and readable.

Run Doxygen build process to verify changes are properly reflected and do not break other documentation.

## Future Work

Online references to images are broken since markdown expects them to be under img/ locally, so the reference is turned into an html tag but retains the same path, which does not exist as the images are not copied. I've added a section to generate_docs.bash to copy those images within html/img after the documentation generates to enable image references to function properly and could add that to a separate PR to keep this one isolated.


LeStarch (approved):

LeStarch:
No code changes, failure is understood.  Merging!

Issue #1229, opened 2022-01-18
Add unlock() alias or rename unLock() to unlock() in Os::Mutex to support std::lock_guard


vietjtnguyen:
[`std::lock_guard`](https://en.cppreference.com/w/cpp/thread/lock_guard) is a handy type to lock/unlock a mutex automatically according the lifetime of a `std::lock_guard` instance. It only requires that the provided mutex type satisfy [BasicLockable](https://en.cppreference.com/w/cpp/named_req/BasicLockable) which only requires the existence of a `lock()` and `unlock()` member function.

The least intrusive method is to add an alias function:

```diff
diff --git a/Os/Mutex.hpp b/Os/Mutex.hpp
index 94ebab234..31934015a 100644
--- a/Os/Mutex.hpp
+++ b/Os/Mutex.hpp
@@ -14,6 +14,14 @@ namespace Os {
             void lock(); //!<  lock the mutex
             void unLock(); //!<  unlock the mutex

+            /**
+             * An alias to @ref unLock() that allows the type to satisfy the
+             * BasicLockable requirements
+             * (https://en.cppreference.com/w/cpp/named_req/BasicLockable)
+             * needed to work with @ref std::lock_guard.
+             */
+            void unlock() { this->unLock(); }
+
         private:

             POINTER_CAST m_handle; //!<  Stored handle to mutex
```

connorfuhrman:
Is there a PR associated with this? I can take this assignment if it's still available.

LeStarch:
There is not PR for this.  We should add an alias as opposed to renaming the function.

connorfuhrman:
Working at https://github.com/connorfuhrman/fprime/tree/enh/mutex_basiclockable

Will simply add aliases for lock and unlock and write some basic tests.

LeStarch:
This was done!

PR #1441, opened 2022-05-10
Mutex BasicLockable

connorfuhrman:
| | |
|:---|:---|
|**_Originating Project/Creator_**| Connor Fuhrman|
|**_Affected Component_**|  None |
|**_Affected Architectures(s)_**|  None |
|**_Related Issue(s)_**|  #1229 |
|**_Builds Without Errors (y/n)_**|  y|
|**_Unit Tests Pass (y/n)_**|  y|
|**_Documentation Included (y/n)_**|y |

---
## Change Description

Added `unlock` alias for `Os::Mutex` type.

## Rationale

`Os::Mutex` is now [BasicLoackable](https://en.cppreference.com/w/cpp/named_req/BasicLockable).

Added test in Os folder but is really only a compile-time check. Perhaps there is a better way to assert at compile-time that the BaiscLockable requirements are satisfied. For now a `std::lock_guard` is instantiated with an `Os::Mutex` parameter to show correctness. 

## Future Work

When/if F' C++ standard bumps to C++17 or above Mutex could be upgraded to satisfy the [Lockable](https://en.cppreference.com/w/cpp/named_req/Lockable) requirement.

### DIFF
--- a/Os/CMakeLists.txt
+++ b/Os/CMakeLists.txt
@@ -101,6 +101,7 @@ set(UT_SOURCE_FILES
   "${CMAKE_CURRENT_LIST_DIR}/test/ut/OsTaskTest.cpp"
   "${CMAKE_CURRENT_LIST_DIR}/test/ut/OsFileSystemTest.cpp"
   "${CMAKE_CURRENT_LIST_DIR}/test/ut/OsSystemResourcesTest.cpp"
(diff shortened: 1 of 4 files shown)


LeStarch (approved):

timcanham:
@LeStarch I don't see that it could. It's just another public method, and we won't try to run those unit tests on the target.

LeStarch:
Thanks!

PR #1462, opened 2022-05-17
Fixed bugs found while compiling for baremetal.

kevin-f-ortega:
| | |
|:---|:---|
|**_Originating Project/Creator_**| @kevin-f-ortega |
|**_Affected Component_**|  Autocoder, PolyType, and BufferLogger|
|**_Related Issue(s)_**|  |

---
## Change Description

Discovered a lack of `operator=` for enum in Autocoder's template files. 
A mismatch in function definition with its prototype in BufferLogger. 
An invalid casting in PolyType

## Rationale

Fixes bugs found while compiling for baremetal

Re-run the UTs on your side.

### DIFF
--- a/Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_cpp.tmpl
+++ b/Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_cpp.tmpl
@@ -52,6 +52,34 @@ namespace ${namespace} {
     return *this;
   }
 
+  ${name}& ${name} :: operator=(const t a)
+  {
+#for $item_name,$item_value,$item_comment in $items_list:
+#if not $item_value == ""
+#set $manual_values = True
+#else
+#set $manual_values = False
+#end if
+#end for
+#if $manual_values == True
+    FW_ASSERT(#slurp
+#set $value_iter = 0
+#for $item_name,$item_value,$item_comment in $items_list:
+#if $value_iter == 0
+a == $item_value#slurp
+#else
+ || a == $item_value#slurp
+#end if
+#set $value_iter = $value_iter + 1
+#end for
+, a);
+#else
+    FW_ASSERT(a >= 0 && a <= $max_value, a);
+#end if
+    this->e = a;
+    return *this;
+  }
+
   ${name}& ${name} :: operator=(const NATIVE_INT_TYPE a)
   {
 #for $item_name,$item_value,$item_comment in $items_list:
--- a/Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_cpp.tmpl
+++ b/Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_cpp.tmpl
@@ -54,39 +54,17 @@ namespace ${namespace} {
 
   ${name}& ${name} :: operator=(const t a)
   {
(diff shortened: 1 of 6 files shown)


LeStarch (approved):
Looks good to me

Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_cpp.tmpl
```diff
@@ -52,6 +52,34 @@ namespace ${namespace} {
     return *this;
   }
 
+  ${name}& ${name} :: operator=(const t a)
+  {
+#for $item_name,$item_value,$item_comment in $items_list:
```
bocchino, line 57: This loop logic doesn't look right. It re-assigns true and false, depending on whether a value is specified. For example `enum { A =1 , B, C = 1 }` would assign true, false, true. The net effect is to assign the result based on the last value. So why assign over and over?

Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_cpp.tmpl
```diff
@@ -52,6 +52,34 @@ namespace ${namespace} {
...
+#if not $item_value == ""
+#set $manual_values = True
+#else
+#set $manual_values = False
```
bocchino, line 61: This code doesn't right. It looks copy-pasted from code that checks whether a raw number is a valid enum value. But here we already have a `t` value, so we don't need this checking.

Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_cpp.tmpl
bocchino: I suspect the intent is to assign manual = true if there is any manual value, but this code doesn't do that. Was this code copy-pasted from somewhere, and does that code have the same issue?

bocchino (changes requested):
Looks good overall. Please address comments.

Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_cpp.tmpl
bocchino: I'm pretty sure we should delete lines 57-78 here.

Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_hpp.tmpl
```diff
@@ -96,6 +96,11 @@ namespace ${namespace} {
         const ${name}& other //!< The other object
         );
 
+    //! Assignment operator
+    ${name}& operator=(
+        const t a //!< The integer to copy
```
bocchino, line 101: This is not an integer; it's an enumerated constant.

bocchino (commented):
Looks good, just one comment.

Autocoders/Python/src/fprime_ac/generators/templates/enums/enum_hpp.tmpl
kevin-f-ortega: Good catch.

bocchino (approved):
Looks good!

PR #1467, opened 2022-05-20
Update RateGroupDriver to new constructor and configure() function

timcanham:
| | |
|:---|:---|
|**_Originating Project/Creator_**| JPL Infrastructure work |
|**_Affected Component_**| `Svc/RateGroupDriver` |
|**_Related Issue(s)_**|  |

---
## Change Description

The rate group divider array was moved from the constructor to a new `configure()` function.

## Rationale

This makes the constructor compliant with the FPP code generator for constructors, simplifying the FPP statements

Re-run unit tests and verify Ref app still runs.

## Future Work

None.


LeStarch (approved):
Looks like what I'd expect.  Is there a reason for the "Draft" status?

timcanham:
I make them draft so I can do a final diff in the browser...

RPI/Top/instances.fpp
```diff
@@ -313,18 +313,13 @@ module RPI {
   instance rateGroupDriverComp: Svc.RateGroupDriver base id 1700 \
   {
 
-    phase Fpp.ToCpp.Phases.configObjects """
+    phase Fpp.ToCpp.Phases.configComponents """
```
bocchino, line 316: I don't think we want to move the array declaration from the `configObjects` phase to the `configComponents` phase.

* The `configObjects` phase is for declaring static objects and has program-lifetime scope
* The `configComponents` phase is inside a function and has local function scope

Maybe it's OK to declare the array inside the function, if the array values get copied by the call, but this feels like a change we don't want to do.

bocchino (changes requested):
Looks good, but please review the comment.

bocchino (approved):
Looks good!

timcanham:
The 'configure()' function actually dies copy it, but it makes sense to keep it consistent with 'ActiveRateGroup'. Still needs to be tested, so don't merge it yet.

timcanham:
OK - Linux testing done.

PR #1466, opened 2022-05-19
Revised specification of implementation types in Ref and RPI topologies

bocchino:
This PR simplifies the specification of the implementation types for the `comm` and `linuxTime` components in the Ref and RPI topologies. It moves the specification from handwritten C++ code into the model.

This PR is one in a series of two. The next PR will make more declarations `extern` in the generated C++ topology code, to enable #1460.

Note: This PR updates the version of FPP used by F Prime. CI is currently not able to pick up the new FPP version. CI failures are expected until this issue is fixed.

Issue #1460, opened 2022-05-17
Rework the Reference Example For FPP Refactors


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

FPP Is reworking to not require phasing.  We need to rework the reference example.

PR #1473, opened 2022-05-23
Revise instance definitions in Ref and RPI topologies

bocchino:
* Update the FPP version
* Specify the implementation type in the model where appropriate
* Use new extern declarations in generated C++ code to simplify handwritten C++

Incorporates changes in #1466. Enables #1460.


LeStarch (approved):

LeStarch:
Looks fine to me.

PR #1479, opened 2022-05-27
Update/ref int tests

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Adds in integration tests to test complex types in Signal gen: Array types, Serializable Types, and nested Serializable and Array types.

### DIFF
--- a/Ref/test/int/ref_integration_test.py
+++ b/Ref/test/int/ref_integration_test.py
@@ -150,7 +150,7 @@ class TestRefAppClass(object):
                     value,
                 ],
                 max_delay=0.1,
-                events=events
+                events=events,
             )
             assert self.api.get_command_test_history().size() == count
 
--- a/Ref/test/int/ref_integration_test.py
+++ b/Ref/test/int/ref_integration_test.py
@@ -142,11 +142,17 @@ class TestRefAppClass(object):
         assert self.api.get_command_test_history().size() == 1
         self.assert_command("cmdDisp.CMD_NO_OP", max_delay=0.1)
         assert self.api.get_command_test_history().size() == 2
-    
+
     def test_send_command_args(self):
         for count, value in enumerate(["Test String 1", "Some other string"], 1):
             events = self.api.get_event_pred("cmdDisp.NoOpStringReceived", [value])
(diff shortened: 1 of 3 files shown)


Ref/test/int/ref_integration_test.py
```diff
@@ -134,10 +143,38 @@ def test_send_command(self):
         self.assert_command("cmdDisp.CMD_NO_OP", max_delay=0.1)
         assert self.api.get_command_test_history().size() == 2
 
+    def test_send_command_args(self):
+        for count, value in enumerate(["Test String 1", "Some other string"], 1):
+            events = self.api.get_event_pred("cmdDisp.NoOpStringReceived", [value])
```
Joshua-Anderson, line 148: Are you intending to assert that this event is emitted? To my understanding this event object look unused.

Ref/test/int/ref_integration_test.py
```diff
@@ -134,10 +143,38 @@ def test_send_command(self):
...
+                [
+                    value,
+                ],
+                max_delay=0.1,
+            )
+            assert self.api.get_command_test_history().size() == count
+
+    def test_send_command_args(self):
+        for count, value in enumerate(["Test String 1", "Some other string"], 1):
+            events = self.api.get_event_pred("cmdDisp.NoOpStringReceived", [value])
```
Joshua-Anderson, line 160: Same as above.

Joshua-Anderson (commented):
Changes look great in general, glad to see a more advanced unit test example! Just had a single question.

LeStarch:
@Joshua-Anderson ok, should be ready!

ThibFrgsGmz:
@LeStarch What would you do if you wanted to run a test to verify the software version, CPU usage greater than 1% or memory usage greater than x KB? (that's what I want to do in https://github.com/nasa/fprime/discussions/1531)

LeStarch:
Software version is publish in channels from SystemResources....so I'd do an assert on that telemetry value equal to the expected version (`git describe --tags --always`).   Is this enough to get you started, or do you want more detail?

PR #1481, opened 2022-05-28
Replace interpolations with Fstring [2/n]

ThibFrgsGmz:
| | |
|:---|:---|

---
## Change Description

This PR aims at replacing the use of the ``%`` string interpolation operator with f-strings

## Rationale

According to the [Python documentation](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting), using the ``%`` operator to format strings can lead to "a variety of quirks that lead to a number of common errors."

[PEP 498](https://peps.python.org/pep-0498/) added F-strings to Python in version 3.6. F-strings are a versatile and powerful method of string formatting. Because the code looks more like the output, they make it shorter and more readable.

void

## Future Work

This PR is the second in a series of PRs on the same theme.

### DIFF
--- a/Autocoders/Python/bin/JSONDictionaryGen.py
+++ b/Autocoders/Python/bin/JSONDictionaryGen.py
@@ -73,7 +73,7 @@ def pinit():
         "--path",
         dest="work_path",
         type="string",
-        help="Switch to new working directory (def: %s)." % current_dir,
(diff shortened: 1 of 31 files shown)


ThibFrgsGmz:
Oops I branched in my first branch and not on devel so that includes the modifications of my first PR.... I'll close it and see later.

Issue #1455, opened 2022-05-15
add/generate `requirements.txt` as a "lockfile" for python packages


capsulecorplab:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Add or generate a `requirements.txt`, containing pinned versions for all python packages and dependencies. Such a `requirements.txt` can be generated from `setup.py` using a tool s.a., `pip-compile` from [pip-tools](https://github.com/jazzband/pip-tools/). `requirements.txt` files could also be generated for `fprime-tools`, `fprime-gds`, and/or `fprime-fpp`

## Rationale

A `requirements.txt`, containing pinned versions for all python packages and dependencies, would provide an optional means to ensure reproducible installation of the _working_ versions of all python packages and dependencies, using `pip install -r requirements.txt`, essentially functioning akin to a `package-lock.json` used in NodeJS projects managed with `npm`.

ThibFrgsGmz:
Perhaps [Poetry](https://python-poetry.org/) can also be considered?
From memory, this tool makes it easy to manage and track such dependencies.

I made a small demo applied to FPrime here: https://github.com/nasa/fprime/discussions/1005

capsulecorplab:
@ThibFrgsGmz The reason I suggested [pip-tools](https://github.com/jazzband/pip-tools/) is because it works with `setup.py`. Does Poetry do the same? Looks interesting, nonetheless

ThibFrgsGmz:
@capsulecorplab From what I remember, it handles the setup.py file.
But it's been a while since I've prototyped with this tool, so my memories are a bit hazy.

I see people using this tool in large software production, so I guess it has some advantages.

ThibFrgsGmz:
Note: there are other tools similar to Poetry like [Pipenv](https://pypi.org/project/pipenv/) and [PDM](https://pdm.fming.dev/). 
Poetry and PDM would be better than Pipenv in terms of performance. And PDM would be better than Poetry on time cost and compatible dependency resolution.

LeStarch:
To expand a bit on what we are thinking here.

##On Packages##

`fprime-tools`, `fprime-gds`, and `fprime-fpp` are python packages. In their development form, and as packages these should have the loosest requirements possible.  This allows PIP to pick the newest packages possible that work together. This pattern works for generic packages, and for developers of these packages.

It also allows an easier update path for packages as there aren't firm, possibly conflicting, package version.

##On Projects##

[...]

PR #1491, opened 2022-06-06
Update/fixed versioning

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

This adds a `requirements.txt` file that can be used to specify the versions of **all** PIP packages required by F´. This allows users to install from a known, working, CI-verified working configuration.

Fixes: https://github.com/nasa/fprime/issues/1455

Note: for now, `setup.py` remains as it can be used for developers needing to update the `requirements.txt` file when versions need to be bumped.


docs/INSTALL.md
```diff
@@ -49,7 +49,8 @@ tool installations are described in [Advanced](#Advanced).
 **Installing F´ Python Packages**
 
 ```
-pip install fprime/
+pip install -U setuptools wheel pip
+pip install -U -r fprime/requirements.txt
```
capsulecorplab, line 53: Any particular reason to use the `-U` flag here? If a python dependency makes a breaking change, I feel like that kind of defeats the point of using a `requirements.txt` for reproducible builds.

capsulecorplab, line 53: Any particular reason to use the `-U` flag here? If a python dependency makes a breaking change, I feel like that kind of defeats the point of using a `requirements.txt` for reproducible builds.

capsulecorplab: I guess the `-U` flag is ignored if a specific version is specified pip install -U flag ignored if specific version is specified

LeStarch: Yes, here we are using the -U flag to force the version to that specified.  The use case being: you've install a previous revision's requirements.txt and need to upgrade/downgrade.  The -U was intended to force a change from what you have to what is needed.

Perhaps this is not necessary. I will test.

timcanham (approved):
Looks good!

PR #1497, opened 2022-06-15
Combine values check into tuple

ThibFrgsGmz:
| | |
|:---|:---|

---
## Change Description

This PR changes the way we check if a variable is equal to one of several values. It combines these values into a ``tuple`` and checks if the variable is contained in it instead of checking equality against each of the values.

## Rationale

 This method is faster, less verbose and easier to read.

void

## Future Work

This PR will be followed by other PRs as there are a significant number of changes.

### DIFF
--- a/Autocoders/Python/bin/codegen.py
+++ b/Autocoders/Python/bin/codegen.py
@@ -1141,7 +1141,7 @@ def generate_dependency_file(filename, target_file, subst_path, parser, the_type
             + parser.get_include_enums()
             + parser.get_include_arrays()
         )
-    elif the_type == "assembly" or the_type == "deployment":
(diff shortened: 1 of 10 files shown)


LeStarch:
Somewhere, I lost track where, you asked why such PRs as this take so long to review.   The short answer is that the autocoder code is "delicate".  It has flight heritage, but also old Python 2 heritage. It is slowly being phased-out in preference for the FPP tools, but "just works" with respect to what it needs to do while that phasing happens.

So....we try not to do anything too in-depth with this code as the effort is better invested elsewhere.  Regardless, I will try to review these as best possible, it will just take time.

LeStarch (approved):
This looks right to me.

LeStarch (approved):

PR #1498, opened 2022-06-15
Add items to a dictionary directly during its definition

ThibFrgsGmz:
| | |
|:---|:---|

---
## Change Description

This PR changes the way of adding elements to a dictionary just after its creation, these elements should be added in the dictionary definition itself.

## Rationale

 This method is faster, less verbose and easier to read.

void

## Future Work

 void

### DIFF
--- a/Autocoders/Python/bin/codegen.py
+++ b/Autocoders/Python/bin/codegen.py
@@ -1208,7 +1208,14 @@ def main():
 
     # Configure the logging.
     log_level = opt.logger.upper()
-    log_level_dict = {}
+    log_level_dict = {
+        "QUIET": None,
+        "DEBUG": logging.DEBUG,
+        "INFO": logging.INFO,
+        "WARNING": logging.WARN,
+        "ERROR": logging.ERROR,
+        "CRITICAL": logging.CRITICAL,
+    }
 
     log_level_dict["QUIET"] = None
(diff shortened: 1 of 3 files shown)


Autocoders/Python/bin/codegen.py
```diff
@@ -1208,7 +1208,14 @@ def main():
...
+    log_level_dict = {
+        "QUIET": None,
+        "DEBUG": logging.DEBUG,
+        "INFO": logging.INFO,
+        "WARNING": logging.WARN,
+        "ERROR": logging.ERROR,
+        "CRITICAL": logging.CRITICAL,
+    }
 
     log_level_dict["QUIET"] = None
```
LeStarch, line 1220: I think this code was to be deleted?  You moved it up, but I don't see the removal of the old-form.

LeStarch (changes requested):
Just the one small item.

Autocoders/Python/bin/codegen.py
ThibFrgsGmz: Oops I didn't see that I left the old form just below... I was on another PR at the same time. I'll fix that later tonight.

LeStarch (approved):

LeStarch (approved):

PR #1500, opened 2022-06-15
Remove use of `len(seq) - 1` to get last item of an iterable

ThibFrgsGmz:
| | |
|:---|:---|

---
## Change Description

This PR removes the use of `len(seq) - 1` to get last item of an iterable.

## Rationale

It is not necessary to calculate the length of an iterable in order to get the last element of the iterable.
We can directly give it a negative index `-1` to get the last element. We won't have to iterate over the sequence using `len` to get the last index if the goal is simply to get the last element.

void

## Future Work

 void


LeStarch (approved):

kevin-f-ortega (approved):
I agree. Using -1 is better.

PR #1501, opened 2022-06-15
Use identity comparison with singleton

ThibFrgsGmz:
| | |
|:---|:---|

---
## Change Description

This PR simply changes the way we compare a variable with the singleton objects ``True``, ``False``, and ``None``.
 
## Rationale

Comparisons with singleton objects such as ``True``, ``False``, and ``None`` should be made with identity rather than equality (either ``is`` or ``is not``). [ref](https://stackoverflow.com/questions/2209755/python-operation-vs-is-not)

Identity checks are more efficient than equality checks. In some cases, equality checks can also result in unintended behavior.

Sorry for the code review which requires a certain mental load. Since the complexity of the changes is very low I didn't think it was worth splitting it.

## Future Work

 void

### DIFF
--- a/Autocoders/Python/src/fprime_ac/generators/formatters.py
+++ b/Autocoders/Python/src/fprime_ac/generators/formatters.py
@@ -65,14 +65,14 @@ class CommentFormatters:
 
         for line in line_list:
 
-            if "@code" in line and code_flag is False:
+            if "@code" in line and not code_flag:
                 leading_spaces = line.find("@code")
                 new_line_list.append(line.strip())
                 code_flag = True
-            elif "@code" in line and code_flag is True:
+            elif "@code" in line and code_flag:
                 new_line_list.append(line.strip())
                 code_flag = False
-            elif (not "@code" in line) and code_flag is True:
+            elif (not "@code" in line) and code_flag:
                 new_line_list.append(line[leading_spaces:])
             else:
                 new_line_list.append(line.strip())
@@ -206,9 +206,9 @@ class CommentFormatters:
                 trimwhitespace = False
             elif "@endcode" == line.strip() or r"\endcode" == line.strip():
                 trimwhitespace = True
-            elif trimwhitespace is False and "@code" == line.strip():
+            elif not trimwhitespace and "@code" == line.strip():
                 trimwhitespace = True
-            elif trimwhitespace is False and r"\code" == line.strip():
+            elif not trimwhitespace and r"\code" == line.strip():
                 trimwhitespace = True
 
             if started:
@@ -231,7 +231,7 @@ class CommentFormatters:
             comment_str += "*\n"
             comment_str += "* TYPE: THIS IS A SYNCHRONOUS INTERFACE.\n"
 
-        if check_and_send is True:
+        if check_and_send:
             comment_str += "*\n"
             comment_str += "* Note: This interface uses an IPC check and send call. The return status\n"
             comment_str += "* indicates if the message was sent, or if the message was not sent due to\n"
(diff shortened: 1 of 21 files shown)


LeStarch (approved):
Looks good, but shouldn't we just check true or false rather than `is True` and `is False`?

ThibFrgsGmz:
You are right, especially since [PEP8](https://pep8.org/) also recommends it. This is done now.

Autocoders/Python/src/fprime_ac/utils/XmlParser.py
```diff
@@ -207,7 +207,7 @@ def parse_file(self, filename):
                 # requirement for parsing a second file is that it has a same
                 # named root Element.
 
-                if self.__root is not None and root_name_checked == False:
+                if self.__root is not None and root_name_checked:
```
LeStarch, line 210: This changed logic, I think.  Seems it should read:

```
if self.__root is not None and not root_name_checked:
```

Autocoders/Python/src/fprime_ac/utils/ParseC.py
```diff
@@ -216,7 +216,7 @@ def ParseTypedefEnum(typename, filename, loadfile=True):
     #
     # LJR Added 30 July 2007 to flag missing enum definitions.
     #
-    if (typename in data) == False:
+    if (typename in data):
```
LeStarch, line 219: Same here:

```
if typename not in data:
```

Autocoders/Python/src/fprime_ac/generators/formatters.py
```diff
@@ -1339,7 +1339,7 @@ def subThreadInstanceFirstArg(self, mod_id, args, context_list):
         # if context_list is None:
         #    return args
 
-        if self.subThreadTest(mod_id) == True and len(context_list) > 0:
+        if self.subThreadTest(mod_id) is True and len(context_list) > 0:
```
LeStarch, line 1342: This one should be:

```
if self.subThreadTest(mod_id) and len(context_list) > 0:
```

Autocoders/Python/src/fprime_ac/generators/formatters.py
```diff
@@ -1010,7 +1010,7 @@ def formatFunComment(self, name, args, proto=True, ident=0):
         new_list = []
         for line in type_args_list[:-1]:
             new_list.append(line + ",")
-        if proto == True:
+        if proto is True:
```
LeStarch:
```
if proto:
```

LeStarch (changes requested):
I found a few nits where I think logic changed.  I'd have fixed it, but I want another opinion, so forgive the pass-back.

Autocoders/Python/src/fprime_ac/utils/ParseC.py
ThibFrgsGmz: Yes

Autocoders/Python/src/fprime_ac/generators/formatters.py
ThibFrgsGmz: Yes

Autocoders/Python/src/fprime_ac/generators/formatters.py
ThibFrgsGmz: And yes

Autocoders/Python/src/fprime_ac/utils/XmlParser.py
ThibFrgsGmz: Fixed.

Autocoders/Python/src/fprime_ac/generators/formatters.py
ThibFrgsGmz: Fixed.

Autocoders/Python/src/fprime_ac/generators/formatters.py
ThibFrgsGmz: Fixed.

Autocoders/Python/src/fprime_ac/utils/ParseC.py
ThibFrgsGmz: Fixed.

LeStarch (approved):

PR #1508, opened 2022-06-16
Add sys/time.h include to IpSocket for musl compat

aditsachde:
| | |
|:---|:---|
|**_Originating Project/Creator_**| @aditsachde |
|**_Affected Component_**| IpSocket |
|**_Related Issue(s)_**| #1507 |

---
## Change Description

This PR includes an explicit `<sys/time.h>` include to the IpSocket component. 

## Rationale

Building with musl requires this change, and does not impact building with glibc. 

This PR does not aim to introduce full musl compatibility, so building with musl will fail to compile. For now, testing with just glibc should be fine.

## Future Work

We are using a toolchain that uses musl, and are aiming to upstream proper support for it.

### DIFF
--- a/Drv/Ip/IpSocket.cpp
+++ b/Drv/Ip/IpSocket.cpp
@@ -14,6 +14,7 @@
 #include <Fw/Types/Assert.hpp>
 #include <Fw/Types/BasicTypes.hpp>
 #include <Fw/Types/StringUtils.hpp>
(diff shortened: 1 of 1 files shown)


LeStarch (approved):
Change looks good, I want CI to approve this works across our OSes and platforms. CI will do this for us.

PR #1517, opened 2022-06-20
Only support set_cpu_affinity with glibc for musl compat

aditsachde:
| | |
|:---|:---|
|**_Originating Project/Creator_**| @aditsachde |
|**_Affected Component_**| `Os/Task` |
|**_Related Issue(s)_**| #1507 |

---
## Change Description

Only support set_cpu_affinity when using glibc. Detects glibc in a way that will not break due to glibc updates.

## Rationale

This PR is meant as a discussion and not necessarily merged in as is. We are aiming to deploy fprime on the Onion Omega2, whose toolchain uses musl instead of glibc. This is one of the two changes (#1508) required to support building fprime with musl. 

As mentioned in this discussion (#1507), it would be preferred to be able to set cpu affinity when musl. However, musl only supports setting cpu affinity after starting the thread using the `pthread_setaffinity_np` function, and not before, using `pthread_attr_setaffinity_np`.

It is possible to change the ordering, so that fprime sets CPU affinity after starting the thread. This then raises the question of what to do if starting the thread succeeds, but setting CPU affinity fails. If there is a way to acceptably handle this case, then it would make sense to switch this ordering and use `pthread_setaffinity_np` instead.

I would appreciate any thoughts on the best way this, and on musl compatibility in general.

This PR was opened to as a discussion, as mentioned in rational. For testing, CI should validate the change.

## Future Work

Add a CI pipeline to run build and integration tests on a musl based system.

### DIFF
--- a/Os/Posix/Task.cpp
+++ b/Os/Posix/Task.cpp
@@ -11,6 +11,9 @@
 #include <climits>
 #include <Fw/Logger/Logger.hpp>
 
+#ifdef TGT_OS_TYPE_LINUX 
+#include <features.h>
+#endif
 
 static const NATIVE_INT_TYPE SCHED_POLICY = SCHED_RR;
 
@@ -100,7 +103,7 @@ namespace Os {
 
     Task::TaskStatus set_cpu_affinity(pthread_attr_t& att, NATIVE_UINT_TYPE cpuAffinity) {
         if (cpuAffinity != Task::TASK_DEFAULT) {
-#ifdef TGT_OS_TYPE_LINUX
+#if TGT_OS_TYPE_LINUX && __GLIBC__
             cpu_set_t cpuset;
             CPU_ZERO(&cpuset);
             CPU_SET(cpuAffinity, &cpuset);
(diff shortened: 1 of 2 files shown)


Joshua-Anderson (approved):
Looks great - this has heritage in how we handle MacOS, where we also don't support thread affinity.

It would be great to support thread affinity on musl, but agree with your judgement call that being deterministic by setting affinity prior to thread creation is more important than portability.

Hopefully musl gains some ability to set thread affinity prior to creation at some point.

Os/Posix/Task.cpp
```diff
@@ -100,7 +100,9 @@ namespace Os {
 
     Task::TaskStatus set_cpu_affinity(pthread_attr_t& att, NATIVE_UINT_TYPE cpuAffinity) {
         if (cpuAffinity != Task::TASK_DEFAULT) {
-#ifdef TGT_OS_TYPE_LINUX
+#ifdef TGT_OS_TYPE_LINUX
```
LeStarch, line 103: You should move this include to the top.  Including a header inside code should be avoided.

LeStarch (changes requested):

Os/Posix/Task.cpp
aditsachde: That's a much cleaner way of doing it, thanks

aditsachde:
Including `features.h` isn't strictly needed, since the `__GLIBC__` macro ends up being pulled into scope anyways, but since the macro is need for compilation, it seems right to explicitly include the header file.

LeStarch (approved):
Looks good to me.  @Joshua-Anderson one more once-over?

Joshua-Anderson (approved):
LGTM!

PR #1519, opened 2022-06-22
lestarch: puring 'writers' code

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

The `writers` package was a copy-and-rename of the `visitors` package meaning that any change to a visitor or a writer had to be done twice. The writers variants were called from the broken-out functions of: implgen, testgen, pymod_dictgen.  However, these are not used in the build system as the build system entirely calls codegen for these services.

Thus, we have removed the dead code in order to reduce maintenance burden on F´.


LeStarch:
See #1533

Issue #298, opened 2020-12-02
Add file stat command to `Svc::FileManager`


timcanham:
It is useful to be able to get the size of individual files in a directory since the file list command is limited in the number of files that can be listed. Add a command that takes the file name and issues and event with the size for that file.

thomas-bc:
Completed in above PR

Issue #297, opened 2020-12-02
FileManager updates


timcanham:
Change `FM_RemoveFile` command to not fail if the file doesn't exist. Just send WARNING event. This allows a deletion of a list of files without requiring that they all be present. Alternatively, add an argument to the command to ignore missing or not.

LeStarch:
I would recommend adding an argument to the command to ignore errors.  This allows flexibility.  Sometimes, you'd want to know if a file did not exist and stop a sequence, and other times you do want to stop the sequence.

thomas-bc:
Completed in the above PR

PR #1522, opened 2022-06-23
FileManager updates

codeflight1:
| | |
|:---|:---|
|**_Affected Component_**|Svc::FileManager|
|**_Related Issue(s)_**|#297, #298|
|**_Has Unit Tests (y/n)_**|yes|
|**_Builds Without Errors (y/n)_**|yes|
|**_Unit Tests Pass (y/n)_**|yes|

---
## Change Description

This PR adds the FileSize command and ignoreErrors argument to the RemoveFile command in the FileManager component.

### DIFF
--- a/Svc/FileManager/Commands.fppi
+++ b/Svc/FileManager/Commands.fppi
@@ -37,3 +37,8 @@ async command AppendFile(
                           target: string size 256 @< The name of the file to append to
                         ) \
   opcode 0x05
+
+async command FileSize(
+                        fileName: string size 256 @< The file to get the size of
+                      ) \
+  opcode 0x06
--- a/Svc/FileManager/Commands.fppi
+++ b/Svc/FileManager/Commands.fppi
@@ -20,6 +20,7 @@ async command RemoveDirectory(
 @ Remove a file
 async command RemoveFile(
                           fileName: string size 256 @< The file to remove
+                          ignoreErrors: bool @< Ignore non-existent files
                         ) \
   opcode 0x03
 
--- a/Svc/FileManager/Events.fppi
+++ b/Svc/FileManager/Events.fppi
@@ -171,3 +171,11 @@ event FileSizeError(
   severity warning high \
   id 0x13 \
   format "Failed to get the size of file {}, returned status {}"
+
+@ Checking file size
+event FileSizeStarted(
+                       fileName: string size 256 @< The name of the file
+                     ) \
+  severity activity high \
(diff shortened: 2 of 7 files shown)


LeStarch:
@timcanham will you review this?

Svc/FileManager/Commands.fppi
```diff
@@ -36,3 +37,8 @@ async command AppendFile(
                           target: string size 256 @< The name of the file to append to
                         ) \
   opcode 0x05
+
+async command FileSize(
+                        fileName: string size 256 @< The file to get the size of
+                      ) \
```
timcanham, line 43: To be consistent, I think we should add an `ignoreErrors` argument here as well, or remove them both and just make them warnings (my preference). Failing to remove a file or get a size is rarely an issue, but returning an error can complicate a long cleanup or listing activity.

timcanham (changes requested):

Svc/FileManager/Commands.fppi
codeflight1: Should it then be included on the other commands (`CreateDirectory`, `MoveFile`, `RemoveDirectory`, possibly `ShellCommand`, and `AppendFile`) as well, or would it be better to make it a component-wide parameter?

Svc/FileManager/Commands.fppi
```diff
@@ -20,6 +20,7 @@ async command RemoveDirectory(
 @ Remove a file
 async command RemoveFile(
                           fileName: string size 256 @< The file to remove
+                          ignoreErrors: bool @< Ignore non-existent files
```
LeStarch, line 23: A suggestion here is to make this an enumeration.  That way it is a little more understandable in sequences.  E.g. a sequence:

```
fm.REMOVE_FILE /path/to/file TRUE
```
vs 
```
fm.REMOVE_FILE /path/to/file IGNORE_ERRORS
```
Using an enum self-describes without needing to look-up the command definition to remember which polarity the bool represented.

timcanham: I agree with this approach. This one has gotten a little stale. @codeflight1: Are you willing to update this PR?

codeflight1: I definitely can, but I was waiting for a decision on whether that should be added to each command or made a component-wide parameter (see other requested changes).

timcanham (approved):
The changes are good for a baseline for further considerations. OK to merge.

PR #1523, opened 2022-06-23
Updates to TlmChan to send multiple channels per ComBuffer

timcanham:
| | |
|:---|:---|
|**_Originating Project/Creator_**| JPL Internal |
|**_Affected Component_**| `Svc/TlmChan` |
|**_Related Issue(s)_**|  |

---
## Change Description

The `Svc/TlmChan` component will now send as many channel updates as will fit in a `Fw::ComBuffer` rather than a single channel for each ComBuffer. The multiple implementation files were consolidated into a single `TlmChan.cpp`. The old unit tests were updated to the new unit test approach.

## Rationale

This takes better advantage of the downlink bandwidth by not requiring the overhead of a packet for each channel update.

Tested with unit tests and full system with downlink.

## Future Work

None

Note: This requires this PR on the GDS to work correctly: https://github.com/fprime-community/fprime-gds/pull/70


LeStarch:
This will pass once we are able to publish our python package.  It is blocked due to this issue: https://github.com/pypa/pypi-support/issues/2317

Fw/Tlm/docs/sdd.md
```diff
@@ -17,16 +17,38 @@ The `Fw::Tlm` port has the following port diagram:
...
 
-##### 2.1.2.2 Fw::ComBuffer
+##### 2.1.2.2 Fw::ComPacket 
 
-The `Fw::TlmBuffer` class contains a buffer that holds the serialized value of a telemetry channel. This buffer is passed as an argument to the `Fw::Tlm` port.
+The `Fw::TlmPacket` class is a packet class derived from [`Fw::ComPacket`](../../Com/docs/sdd.md) that provides methods for encoding a telemetry packet. It contains an internal `Fw:ComBuffer` that holds serialized channel values.
+
+To fill a packet with telemetry values, do the following.
+
+1\. Instantiate an instance of the `TlmPacket` class.
```
bocchino, line 30: The changes to the SDD look good. I fixed a broken link.

It looks like you have suppressed the line-byline numbering by writing 1\. Is this intentional? I think it's easier to read if each numbered item starts on a new line.

Fw/Tlm/TlmPacket.hpp
```diff
@@ -23,19 +24,26 @@ namespace Fw {
...
-            // getters
-            FwChanIdType getId();
-            Time& getTimeTag();
-            TlmBuffer& getTlmBuffer();
-
-        PROTECTED:
-            FwChanIdType m_id; // !< Channel id
-            Fw::Time m_timeTag; // !< time tag
-            TlmBuffer m_tlmBuffer; // !< serialized data
+            // add telemetry value
```
bocchino, line 27: Should we have doxygen style comments, i.e., //! ?

Fw/Tlm/TlmPacket.cpp
```diff
@@ -10,91 +10,146 @@
...
+        // The function deserializeBase inherited from ComPacket deserializes a
+        // value of type FwPacketDescriptorType from this->m_tlmBuffer and stores it
+        // into this->m_type.
+        Fw::SerializeStatus stat = this->deserializeBase(this->m_tlmBuffer);
+        if (stat != Fw::SerializeStatus::FW_SERIALIZE_OK) {
+            return stat;
+        }
+        // make sure that this->m_tlmBuffer stores a telemetry packet
+        if (this->m_type != FW_PACKET_TELEM) {
+            return Fw::SerializeStatus::FW_DESERIALIZE_TYPE_MISMATCH;
```
bocchino, line 41: Should we set this->m_type back to FW_PACKET_TELEM here? Otherwise we leave the component in an inconsistent state. Even better would be not to call deserializeBase at all, but do the deserialization into a local variable and check that against FW_PACKET_TELEM. I don't see why we should ever have to write a value to m_type after assigning it in the constructor -- it should always be FW_PACKET_TELEM.

Fw/Tlm/TlmPacket.cpp
```diff
@@ -10,91 +10,148 @@
...
-#if !FW_AMPCS_COMPATIBLE
-        stat = serializeBase(buffer);
-        if (stat != FW_SERIALIZE_OK) {
+    SerializeStatus TlmPacket::resetPktSer() {
+        this->m_tlmBuffer.resetSer();
+        // reset packet count
+        this->m_numEntries = 0;
+        // serialize descriptor
+        // The function serializeBase inherited from ComPacket converts this->m_type
+        // to type FwPacketDescriptorType and serializes the result into this->m_tlmBuffer.
```
bocchino, line 27: Should we check or assert that this->m_type is FW_PACKET_TELEM? Otherwise this->m_tlmBuffer will be in a bad state.

Fw/Tlm/docs/sdd.md
timcanham: At a certain point I got tired fighting the auto-numbering. I couldn't predict when it would restart, so I manually number.

Fw/Tlm/TlmPacket.hpp
timcanham: Added...

Fw/Tlm/TlmPacket.cpp
timcanham: This is not a component, it is a helper class, so my expectation is that it would be created on the stack or somewhere temporarily for use in extracting channels. If we wanted to do anything, it would be to assert since the `deserializeBase()` should get it. I decided not to, since this could be a bad buffer passed in to the class. I thought it best to punt to the user rather than unilaterally assert or hack in the packet type. I'm not sure hacking in the packet type would accomplish anything since the data is probably bad anyway and the user should decide what to do with it. This is a little hand-wavy since we don't really have a use chase for deserializing on-board yet, so I was trying to implement a plausible way to get data out. I would leave the implementation as-is.

Fw/Tlm/TlmPacket.cpp
timcanham: Added an assert. It should never change since we set it in the constructor and don't modify it (we never deserialize from a buffer for downlink), but it can't hurt.

Fw/Tlm/docs/sdd.md
bocchino: Manual numbering is fine, but can we insert a blank line between the enumerated items? Otherwise they all get mashed together in a single paragraph with no line breaks.

timcanham: Yes - I'll fix that. I didn't realize they were mashing together.

Fw/Tlm/TlmPacket.hpp
bocchino: Looks good! For some reason these changes don't appear in the GitHub PR view, but I see them in the repo.

Fw/Tlm/TlmPacket.cpp
bocchino: OK, thanks. Let's leave as-is for now. It's still not clear to me why we need the m_type field in the base class. Maybe it's for run-time type identification? This is a broader F Prime issue though.

Fw/Tlm/TlmPacket.cpp
bocchino: Sounds goo!

Fw/Tlm/docs/sdd.md
bocchino: OK, sounds good.

timcanham: @bocchino Fixed the numbering. Should be ready.

bocchino: Looks good. Again there appears to be some bug in the GitHub PR viewer (at least in my browser) but when I click on "edit file" I see the changes.

bocchino:
@timcanham It appears some merge conflicts crept in and need to be fixed.

timcanham:
@bocchino Fixed the conflicts. Ready to check again.

bocchino (approved):
Looks good! All of my comments have been addressed.

LeStarch:
@timcanham some questions here:
1. static analysis is reporting a bunch of new violations.  Most are irrelevant, but the "check return value" items are likely apt.  Even if we just need to assert on them
2. I see a lot of commented-out-code

Thoughts?

timcanham:
@mstarch Is the "commented out code" you are referring to the code in the unit test?

LeStarch:
@timcanham yes, I had intended to say "commented out code in the UTs".  Is this code necessary anymore?  If so, should we enable it, if not should we delete it?

PR #1528, opened 2022-06-26
Performance++ and code minimization

OfficialAhmed:
| | |
|:---|:---|
|**_Affected Component_**| autocoder |
|**_Related Issue(s)_**|  |

---
## Change Description

The PR aims for better performance by replacing List iteration to Tuple as tuples use less memory and has fixed malloc hence increasing code execution speed [link](https://www.educative.io/answers/tuples-vs-list-in-python). This PR also minimizes old implementation of string formatting for code readably and performance as well

## Rationale

- [x] Performance: Iterating Tuples is faster and takes less memory than Lists
- [x] Performance: Minimized if-else statement checks 
- [x] Minimizing code: Used newest implementation of string formatting (f-string) for code readability and minimization
 

Void

## Future Work

Void


Autocoders/Python/src/fprime_ac/generators/visitors/PortHVisitor.py
```diff
@@ -114,18 +114,17 @@ def _get_args_string(self, obj):
...
             elif TypesList.isPrimitiveType(t) or isEnum:
-                t = t + " "
+                t += " "
             else:
                 t = "const " + t + " &"
 
-            arg_str += "{}{}".format(t, arg.get_name())
-            arg_str += ", "
+            arg_str += f"{t}{arg.get_name()}, "
         arg_str = arg_str.strip(", ")
```
ThibFrgsGmz, line 128: You can return directly here, as you have done on other lines, right?

Autocoders/Python/src/fprime_ac/generators/visitors/PortHVisitor.py
```diff
@@ -183,15 +182,14 @@ def _get_args_sum_string(self, obj):
...
             else:
                 if arg.get_modifier() == "pointer":
                     t = "sizeof(" + t + "*)"
                 else:
-                    t = t + "::SERIALIZED_SIZE"
-            arg_str += t
-            arg_str += " + "
+                    t += "::SERIALIZED_SIZE"
+            arg_str += f"{t} + "
         arg_str = arg_str.strip(" + ")
```
ThibFrgsGmz, line 193: Here too, direct return?

Autocoders/Python/src/fprime_ac/generators/visitors/SerialHVisitor.py
```diff
@@ -187,18 +186,28 @@ def _get_enum_string_list(self, enum_list):
...
-            # No value, With comment
-            elif (e[1] is None) and (e[2] is not None):
-                s = "{},  // {}".format(e[0], e[2])
-            # With value, No comment
-            elif (e[1] is not None) and (e[2] is None):
-                s = "{} = {},".format(e[0], e[1])
-            # With value and comment
-            elif (e[1] is not None) and (e[2] is not None):
-                s = "%s = %s,  // %s" % (e)
+            # Without value
```
ThibFrgsGmz, line 189: There are too many nested statements in my opinion.

Wouldn't something like below be better?

```py
    def _get_enum_string_list(self, enum_list):
        """"""
        enum_tuple = enum_list[0]
        enum_list = enum_list[1]
        enum_str_list = []
        for e in enum_list:
            # No value, No comment
            if (e[1] is None) and (e[2] is None):
                s = f"{e[0]},"
            elif e[1] is None:
                s = f"{e[0]},  // {e[2]}"
            elif e[2] is None:
                s = f"{e[0]} = {e[1]},"
            else:
                s = f"{e} = {e},  // {e}"
            enum_str_list.append(s)

        return (enum_tuple, enum_str_list)
```

ThibFrgsGmz, line 189: There are too many nested statements in my opinion.

Wouldn't something like below be better?

```py
    def _get_enum_string_list(self, enum_list):
        """"""
        enum_tuple = enum_list[0]
        enum_list = enum_list[1]
        enum_str_list = []
        for e in enum_list:
            # No value, No comment
            if (e[1] is None) and (e[2] is None):
                s = f"{e[0]},"
            elif e[1] is None:
                s = f"{e[0]},  // {e[2]}"
            elif e[2] is None:
                s = f"{e[0]} = {e[1]},"
            else:
                s = f"{e} = {e},  // {e}"
            enum_str_list.append(s)

        return (enum_tuple, enum_str_list)
```

Autocoders/Python/src/fprime_ac/generators/visitors/PortHVisitor.py
ThibFrgsGmz, line 128: You can return directly here, as you have done on other lines, right?

Autocoders/Python/src/fprime_ac/generators/visitors/PortHVisitor.py
```diff
@@ -183,15 +182,14 @@ def _get_args_sum_string(self, obj):
...
                 t = "sizeof(" + t + cl
             else:
                 if arg.get_modifier() == "pointer":
                     t = "sizeof(" + t + "*)"
                 else:
-                    t = t + "::SERIALIZED_SIZE"
-            arg_str += t
-            arg_str += " + "
+                    t += "::SERIALIZED_SIZE"
+            arg_str += f"{t} + "
```
ThibFrgsGmz, line 192: Here too, direct return?

Autocoders/Python/src/fprime_ac/generators/visitors/SerialHVisitor.py
OfficialAhmed: This would work too, better than the first implementation

Autocoders/Python/src/fprime_ac/generators/visitors/PortHVisitor.py
OfficialAhmed: that's right

Autocoders/Python/src/fprime_ac/generators/visitors/SerialHVisitor.py
OfficialAhmed: but it'll take more time to execute, because it's checking some statements that are unnecessary.

OfficialAhmed: every time it runs through all the statements even if we only  want the "else" to be  executed. i.e. if it's value and comment then it would run the 3 statements first which increases execution time, whereas my implementation it would ignore the first 3 if-statements and immediately runs the "else".

Its hard to read but easy for the execution

ThibFrgsGmz: Correct me if I'm wrong, but the `else` will never be executed right?

The variable is `None` or it is not, so you can transform the `elif` into `else`.

Once this is done and a final refactoring, you are back to my proposal it seems.

OfficialAhmed: `else` is executed if we have either **value** `not None` or **comment** `not None`. 
So basically you're right instead of `elif` we can replace it with `else`

correct

OfficialAhmed (commented):
reviewed

Autocoders/Python/src/fprime_ac/generators/visitors/SerialHVisitor.py
LeStarch: I am with @ThibFrgsGmz here.  The nested `if`s make this much harder to review.  I like the existing structure and I'd revert to that.  @ThibFrgsGmz's sample code would also be acceptable.

The other fixes in this PR look good, but the changing of the if structures, although performant, make the code header to understand.

Autocoders/Python/src/fprime_ac/generators/visitors/SerialHVisitor.py
```diff
@@ -100,23 +100,22 @@ def _get_args_string(self, obj):
...
             if isinstance(mtype, tuple):
-                arg_str += "{} {}, ".format(mtype[0][1], name)
-            elif mtype == "string" and array_size is None:
-                arg_str += "const {}::{}String& {}, ".format(obj.get_name(), name, name)
-            elif mtype == "string" and array_size is not None:
-                arg_str += "const {}::{}String* {}, ".format(obj.get_name(), name, name)
-                arg_str += "NATIVE_INT_TYPE %sSize, " % (name)
+                arg_str += f"{mtype[0][1]} {name}, "
+            elif mtype == "string":
+                if array_size is None:
```
LeStarch, line 105: There is extra nesting of `if`s here too.

LeStarch:
I am going to close this PR, as it isn't getting a lot of traction over the last 1-2months.  Additionally it is working on the autocoder Python which we try to reserve for bug fixes.

PR #1552, opened 2022-07-01
Replace manual loop counters by `enumerate()` call

ThibFrgsGmz:
| | |
|:---|:---|

---
## Change Description

This PR aims to convert the manual loop counter with a call to the built-in function `enumerate()`.

## Rationale

Using the built-in Python function `enumerate()` allows us to generate an index directly, eliminating two lines of code that would otherwise be used to access a loop counter that would tell us the index of the element used.
With this change, we no longer need to consider the i/inst variable, allowing us to focus on the code that really matters.

void

## Future Work

 void

### DIFF
--- a/Autocoders/Python/src/fprime_ac/generators/visitors/ComponentVisitorBase.py
+++ b/Autocoders/Python/src/fprime_ac/generators/visitors/ComponentVisitorBase.py
@@ -379,12 +379,10 @@ class ComponentVisitorBase(AbstractVisitor.AbstractVisitor):
                 if len(opcodes) == 1:
                     return f"CMD_{mnemonic.upper()}"
                 else:
-                    mlist = []
-                    inst = 0
-                    for opcode in opcodes:
-                        mlist.append("CMD_" + mnemonic.upper() + "_%d" % inst)
-                        inst += 1
-                    return mlist
+                    return [
+                        f"CMD_{mnemonic.upper()}" + "_%d" % inst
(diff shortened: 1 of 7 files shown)


Autocoders/Python/src/fprime_ac/generators/visitors/ComponentVisitorBase.py
```diff
@@ -379,12 +379,10 @@ def g(xxx_todo_changeme3):
...
                     return f"CMD_{mnemonic.upper()}"
                 else:
-                    mlist = []
-                    inst = 0
-                    for opcode in opcodes:
-                        mlist.append("CMD_" + mnemonic.upper() + "_%d" % inst)
-                        inst += 1
-                    return mlist
+                    return [
+                        f"CMD_{mnemonic.upper()}" + "_%d" % inst
```
LeStarch, line 383: Shouldn't this become:

f"CMD_{mnemonic.upper()}_{inst}"

LeStarch (changes requested):
Collapse out the multi string format style, if you would!

Autocoders/Python/src/fprime_ac/generators/visitors/ComponentVisitorBase.py
ThibFrgsGmz: Indeed, thanks!

LeStarch (approved):

PR #1553, opened 2022-07-03
Switch the `Svc/Deframer` documentation to `MermaidJs`

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**| Svc/Deframer/docs |
|**_Builds Without Errors (y/n)_**| NA |
|**_Unit Tests Pass (y/n)_**|   |

---
## Change Description

This PR aims to change the documentation diagrams of the `Svc/Deframer` component to [Mermaid](https://mermaid-js.github.io/mermaid/#/).

## Rationale

At least as of [2022-02-14](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/), GitHub supports viewing Mermaid diagrams in its Markdown viewer.

This eliminates the need to embed graphical images (png) in Git that cannot be submitted to a code review. It reduces the memory size of the directory by removing PNGs and it takes advantage of a very interesting feature of GitHub.

Mermaid is very similar to PlantUML currently used for these document diagrams.

To play with mermaid, you can use the [Live Editor](https://mermaid.live/edit#pako:eNpt0M2qAjEMBeBXidk68wJdKILC1a3bbkJ7dMrtj9YWEfHd7YzjzqwC-U4gebJJFqz4hmtFNNg6OWcJOlKrjXcG_Wq1PKQhKvqD94nGvqMh3Uky6JHq-if-MCNxJDRAMgXMdJz1jfZTpi12n3RH-ykx6RZb_OZ7OgGezhlSZvIt7jggB3G23fQcZ5rLgADNqrVW8r9mHV_N1YuVgp11JWVWJ_E3dCy1pOMjGlYlV3zR_JRZvd5_C2Gl).

## Future Work

If this PR is accepted, I can take care of the diagrams of the other components. The complexity is very low because the PlantUML and Mermaid synthaxes are very similar.


astroesteban:
That's awesome! I've been using Mermaid for my Tiva port: https://github.com/astroesteban/fprime-tiva/blob/main/Example/docs/DESIGN.md

There's even VS Code plugins that render the mermaid diagrams in the Markdown preview.

ThibFrgsGmz:
@astroesteban Wow your work is so cool, I didn't see that! You handled the Mermaid well!

I'm not a big fan of the VSCode plugin because it doesn't support all the features of Mermaid from what I remember.

LeStarch:
@bocchino @arizvi786 I'd like you two to weigh in on this.  Does this style diagram meet your expectations/needs or should we keep to embedding PNGs or other images?

Also is there a way to generate mermaid from FPP modeling in the same way we use the visualizer?

arizvi786:
@LeStarch I like the use of Mermaid especially considering that the diagrams can be updated from within the SDD. It's much better than embedding image files generated elsewhere. Looking at the diagrams output, they look good and I don't see any major issues. Also, there seems to be good documentation available for it. My only concern is if it would be ok for us to rely on this plugin/tool for our documentation which may or may not be supported in the future. We can say quite reliably that embedding diagram images will not break anything; however, can we say the same for the use of Mermaid in our SDDs?

Svc/Deframer/docs/sdd.md
```diff
@@ -386,27 +392,103 @@ Vertical dashed lines represent component code.
...
+    activeComm->>buffMgr: Allocate frame buffer FB
+    buffMgr-->>activeComm: Return FB
+    activeComm->>activeComm: Fill FB with framed data
+    activeComm->>deframer: Send FB[framedIn]
+    deframer->>buffMgr: Allocate packet buffer PB [bufferAllocate]
+    buffMgr-->>deframer: Return PB
+    deframer->>deframer: Deframe FB into PB
+    deframer->>deframer: Copy PB into a command packet C
+    deframer-)cmdDisp: Send C [comOut]
+    deframer->>buffMgr: Deallocation PB [bufferDeallocate]
```
bocchino, line 407: I think this should say `Deallocate PB` and similarly two lines below.

Svc/Deframer/docs/sdd.md
```diff
@@ -386,27 +392,103 @@ Vertical dashed lines represent component code.
...
+    deframer->>buffMgr: Allocate packet buffer PB [bufferAllocate]
+    buffMgr-->>deframer: Return PB
+    deframer->>deframer: Deframe FB into PB
+    deframer-)fileUplink: Send PB [bufferOut]
+    deframer->>buffMgr: Deallocate FB [framedDeallocate]
+    buffMgr-->>deframer: 
+    deframer-->>activeComm: 
+    deactivate activeComm
+    activate fileUplink
+    fileUplink->>-buffMgr: Deallocate PB
```
bocchino, line 439: Is this a typo? `>>` between `-` and `-`.

bocchino (approved):
These changes look good to me.

Svc/Deframer/docs/sdd.md
```diff
@@ -129,7 +129,13 @@ See <a href="#dpi-impl">below</a> for a description of how `Deframer` implements
 
 Here is a class diagram for `Deframer`:
 
-![classdiagram](./img/class_diagram_deframer.png)
+```mermaid
+classDiagram
+    ObjBase <|-- PassiveComponentBase
+    PassiveComponentBase <|-- DeframerComponentBase
+    DeframerComponentBase <|-- DeframerComponentImpl
```
bocchino, line 136: Let's remove the `ComponentImpl` suffix here to bring the diagram in line with the newest version of the implementation. Same on the next line.

Svc/Deframer/docs/sdd.md
ThibFrgsGmz: The `-` is a shortcut to deactivate a thread. But the code contained an error, I deactivated the component at the wrong time.

Svc/Deframer/docs/sdd.md
ThibFrgsGmz: Fixed

Svc/Deframer/docs/sdd.md
ThibFrgsGmz: Fixed

LeStarch:
@ThibFrgsGmz does this work when generating github docs pages?

ThibFrgsGmz:
Apparently this support does not yet extend to GitHub pages. I haven't looked into the GitHub Pages feature yet so I'm not sure how it works (it's on my to-do list).

I see that some people have solved the problem with these solutions:

1. HTML code inclusion: https://stackoverflow.com/a/53893998
2. Or adding the plugin [jekyll-spaceship](https://github.com/jeffreytse/jekyll-spaceship)

PR #1575, opened 2022-07-20
Fixed ExternalSerializeBuffer copy constructor.

awiswasi:
## Change Description

Minor fix since the passed in pointer is never going to call the copy constructor.

## Rationale

Improve quality of C++ code.


LeStarch:
It seems the file was reformatted as well as the changes added.  In addition, `= delete` is one of the C++  11 additions that we are currently trying to avoid because they break on certain compilers.

PR #1578, opened 2022-07-21
Update cross compile toolchain for raspberrypi in cmake & RPI CI

### LINKED ISSUE #1163: Cross-compile toolchain for raspberry pi is deprecated
[Cross-compile toolchain](https://github.com/raspberrypi/tools) in [package installation](https://github.com/nasa/fprime/blob/devel/RPI/README.md#package-installation) for raspberry pi demo is listed as a deprecated project.

capsulecorplab:
| | |
|:---|:---|
|**_Affected Component_**| RPI |
|**_Affected Architectures(s)_**| |
|**_Related Issue(s)_**| #1163  |
|**_Has Unit Tests (y/n)_**| |
|**_Unit Tests Pass (y/n)_**| |

---
## Change Description

- Updates cross compile toolchain for raspberry pi by first checking `RPI_TOOLCHAIN_DIR` environment variable, then by checking PATH, in cmake toolchain for raspberry pi.
- Updates CI build for RPI to use arm toolchain installed from Apt

## Rationale

To resolve #1163 

(Optional: Run following in https://github.com/capsulecorplab/kasm-fprime-workspace)

For building locally:

1. clone & checkout https://github.com/nasa/fprime/pull/1578/commits/c94e28c8cbab14adeec30f39234440ef272ed5d9
2. `cd` into `fprime/RPI/`
3. generate build cache for raspberrypi with `fprime-util generate raspberrypi`
4. build RPI binary with `fprime-util build raspberrypi`

## Future Work

ARM toolchain in RPI CI should be installed from a pinned version/release


cmake/toolchain/raspberrypi.cmake
```diff
@@ -1,41 +1,31 @@
...
 
-# where is the target environment
-set(CMAKE_FIND_ROOT_PATH  "${RPI_TOOLCHAIN}")
+# Set the GNU ARM toolchain
+set(CMAKE_AR            arm-linux-gnueabihf-ar)
+find_program(CMAKE_C_COMPILER NAMES arm-linux-gnueabihf-gcc PATHS ENV RPI_TOOLCHAIN_DIR PATH_SUFFIXES bin REQUIRED)
+find_program(CMAKE_CXX_COMPILER NAMES arm-linux-gnueabihf-g++ PATHS ENV RPI_TOOLCHAIN_DIR PATH_SUFFIXES bin REQUIRED)
+message(STATUS "[arm-linux] C Compiler: ${CMAKE_C_COMPILER}")
+message(STATUS "[arm-linux] CXX Compiler: ${CMAKE_CXX_COMPILER}")
+set(CMAKE_ASM_COMPILER  arm-linux-gnueabihf-as)
```
LeStarch, line 22: Don't we need to run a `find_program` for every tool listed here, not just `g++` and `gcc`?

LeStarch (changes requested):
This looks like what I'd expect, just the one question about how some tools differ from others.

I appreciate that you:

1. Allow a defined path
2. Kept the same environment file
3. Allow system tools when not set.

cmake/toolchain/raspberrypi.cmake
LeStarch: Also, there is one true spelling error. The rest should be added to the expect list.

capsulecorplab: What other tools would we want to run a `find_program` for?

capsulecorplab: How do I resolve the spell check errors?

LeStarch: For example each line should be a find_program:
```
set(CMAKE_ASM_COMPILER  arm-linux-gnueabihf-as)
```
->
```
find_program(CMAKE_ASM_COMPILER NAMES arm-linux-gnueabihf-as PATHS ENV RPI_TOOLCHAIN_DIR PATH_SUFFIXES bin REQUIRED)
```

The reason is that you'd likely want to ensure toy are pulling a consistent set of tools from the same directory.

To fix spelling errors:

```
db
multiarch
objcopy
objdump
```

Can be added to `expect.txt` under `.github/actions/spelling`

```
Raspbery 
```

Appears to be a true misspelling.

capsulecorplab: Sorry for the delay, been out traveling these past few weeks! The spell check errors should be resolved now. Also replaced `set` with `find_program` on the lines where it looks for the ARM toolchain

LeStarch (approved):

LeStarch:
I reviewed this.  It looks correct!  The following settings:

```
set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY   NEVER)
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE   NEVER)
set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE   NEVER)
```

Are setup to ignore any `sysroot` setting which is appropriate for the use of package-installed tools for the raspberry platform.

**Note:** using a toolchain setup like this does require a minimum raspberrypi Os version (e.g. the PI needs to be running an OS as new or newer than the compilation package) or the compiled binary will not run.  This is because libc and other libraries are not forward-compatible.  However, this a small price to pay for non-deprecated non-gcc-4 tooling!

LeStarch:
@capsulecorplab I committed to your branch to remove GDB.  It isn't used in our CMake system anywhere, so the tool shouldn't be needed.  That should fix the RPI build failure we just saw.

LeStarch:
I updated `RPI/README.md` to point to the new way to install tools.  This should be ready to go!

PR #1580, opened 2022-07-21
Revert "Switch the `Svc/Deframer` documentation to `MermaidJs`"

LeStarch:
Reverts nasa/fprime#1553.

This should not have gone into `master` branch, but rather devel


LeStarch:
Documentation only, will leave as is.

PR #1590, opened 2022-07-23
Replace NULL with nullptr

Anirban166:
| | |
|:---|:---|
|**_Affected Component_**| N/A |
|**_Affected Architectures(s)_**| STest |
|**_Related Issue(s)_**| -/- |

---
## Change Description

Change `NULL` to `nullptr`.

## Rationale

Since C++11, using a nullptr over NULL is the de facto standard since:
- The former provides a type-safe representation of an empty/null pointer (both raw and smart pointers), automatically (defined using templates, evaluated during compile-time) deducing the correct type depending upon the type of the instance it is being assigned to.
- The latter might not be valid in some cases depending on where it is used, as it has no return type resolving logic under the hood and is implementation specific at times (this macro is usually defined to be either an integer literal with the value 0, or a pure rvalue of type `std::nullptr_t`; if it has to be a macro for consistency then `#define NULL nullptr` is how they put it so as to align with the former).

void

## Future Work

((void*)0)

### DIFF
--- a/STest/STest/Scenario/ConditionalIteratedScenario.hpp
+++ b/STest/STest/Scenario/ConditionalIteratedScenario.hpp
@@ -61,7 +61,7 @@ namespace STest {
       Scenario<State>* nextScenario_IteratedScenario(
           State& state //!< The system state
       ) {
-        Scenario<State>* scenario = NULL;
+        Scenario<State>* scenario = nullptr;
         if (!this->condition_ConditionalIteratedScenario(state)) {
           this->done = true;
         }


LeStarch (approved):
## Pre-CI Checks: Must perform before launching CI
- [x] No obviously malicious code introduced
- [x] PR is against `devel`. Use `Edit -> change branch` to fix, or close PR.

## Code Review
- [x] Code follows flight software quality guidelines
- [x] Changes are of reasonable length for review

## Procedural Checks
- [x] PR passes CI
- [x] Accept with `squash-and-merge`

PR #1598, opened 2022-07-27
Avoid branching (and thus branch prediction) using bitwise operators on boolean primitives

Anirban166:
| | |
|:---|:---|
|**_Affected Component_**| Fw/Port/PortBase |
|**_Affected Architectures(s)_**| Fw |

---
## Change Description

A more concise way of representing boolean logic with a branchless version (using bitwise operators) for a code segment that operates on `bool` variables with branching.

## Rationale

A more compact form of this:
```cpp
bool do_trace = false;
... // Procure variables to be put under conditionals
// or initialize them, for e.g.: int a{1}, b{0}, c{1};
if (a) { if (b) do_trace = true; }
else if (c) do_trace = true;
```
would be this:
```cpp
if ((a && b) || (!a && c)) do_trace = true;
// if ((this->m_ovr_trace && this->m_trace) || (!(this->m_ovr_trace) && PortBase::s_trace)) do_trace = true;
```
or without the `if` conditional, simply:
```cpp
do_trace = (a && b) || (!a && c);
// do_trace = ((this->m_ovr_trace && this->m_trace) || (!(this->m_ovr_trace) && PortBase::s_trace));
``` 
And even better would be to use the bitwise operators `|` and `&` instead of the logical `||` and `&&` for better performance: (applicable, given that we are dealing with just booleans under the hood)
```cpp
do_trace = (a & b) | (!a & c);
// do_trace = ((this->m_ovr_trace & this->m_trace) | (!(this->m_ovr_trace) & PortBase::s_trace));
```

I think code readability would be the con for some, but otherwise, this would be an optimization given that the branching has been eliminated (no `cmp` instructions), which in turn avoids branch prediction on processors - a factor for performance degradation (although, the severity of it would entirely depend on the frequency that the branch being taken gets mispredicted). Let me know if I missed something!

## Future Work

None

### DIFF
--- a/Fw/Port/PortBase.cpp
+++ b/Fw/Port/PortBase.cpp
@@ -47,13 +47,7 @@ namespace Fw {
     void PortBase::trace() {
         bool do_trace = false;
 
-        if (this->m_ovr_trace) {
-            if (this->m_trace) {
-                do_trace = true;
-            }
-        } else if (PortBase::s_trace) {
-            do_trace = true;
-        }
+        do_trace = ((this->m_ovr_trace & this->m_trace) | (!(this->m_ovr_trace) & PortBase::s_trace));
 
         if (do_trace) {
 #if FW_OBJECT_NAMES == 1


bocchino (commented):
I think we should avoid using bitwise operations & and | on boolean values, for the reasons stated here: https://stackoverflow.com/questions/24542/using-bitwise-operators-for-booleans-in-c. Bitwise operators are for bitwise operations, not logical operations. Using logical operations && and || is OK.

bocchino:
> this would be an optimization given that the branching has been eliminated (no cmp instructions), which in turn avoids branch prediction on processors - a factor for performance degradation 

I would guess that modern optimizing compilers with -O2 would generate the most efficient possible code even without the hand optimization. One could generate the assembly code and check this.

bocchino:
FYI, here are some results from my Mac laptop:
```
[I] bocchino@kirk ~> cat test1.cpp
bool f(bool a, bool b) {
  return a && b;
}
[I] bocchino@kirk ~> g++ -c -O2 -S test1.cpp
[I] bocchino@kirk ~> cat test1.s
	.section	__TEXT,__text,regular,pure_instructions
	.build_version macos, 10, 15	sdk_version 10, 15, 6
	.globl	__Z1fbb                 ## -- Begin function _Z1fbb
	.p2align	4, 0x90
__Z1fbb:                                ## @_Z1fbb
	.cfi_startproc
## %bb.0:
	pushq	%rbp
	.cfi_def_cfa_offset 16
	.cfi_offset %rbp, -16
	movq	%rsp, %rbp
	.cfi_def_cfa_register %rbp
	movl	%edi, %eax
	andl	%esi, %eax
                                        ## kill: def $al killed $al killed $eax
	popq	%rbp
	retq
	.cfi_endproc
                                        ## -- End function
.subsections_via_symbols
```
I don't see any branching in the optimized code. I get similar results when I replace `a && b` with a more verbose sequence of if...else statements.

bocchino:
I think a good general principle is: write the code for clarity, and let the compiler optimize it. I would deviate from this rule if (1) the code is performance-critical; (2) I've profiled the code to identify it as a performance bottleneck; and (3) I have identified a specific optimization that the compiler has missed and that would improve performance if added by hand. An example is auto-vectorization. Sometimes the most straightforward way of writing a loop is not the best for vectorization.

Anirban166:
> I think a good general principle is: write the code for clarity, and let the compiler optimize it. I would deviate from this rule if (1) the code is performance-critical; (2) I've profiled the code to identify it as a performance bottleneck; and (3) I have identified a specific optimization that the compiler has missed and that would improve performance if added by hand. An example is auto-vectorization. Sometimes the most straightforward way of writing a loop is not the best for vectorization.

Agreed.
I totally missed out on the fact that compiler optimizations should be considered - Thanks for pointing this out! 
I get similar results from Clang on my Mac too, and I tested that snippet (`bool f(bool a, bool b) { return a && b; }`) for x86-64 gcc on Godbolt right now and it seems like even the O1 level does it better already:
```
f(bool, bool):
        mov     eax, esi
        and     eax, edi
        ret
```
compared to no compiler optimizations:
```
f(bool, bool):
        push    rbp
        mov     rbp, rsp
        mov     edx, edi
        mov     eax, esi
        mov     BYTE PTR [rbp-4], dl
        mov     BYTE PTR [rbp-8], al
        cmp     BYTE PTR [rbp-4], 0
        je      .L2
        cmp     BYTE PTR [rbp-8], 0
        je      .L2
        mov     eax, 1
        jmp     .L3
.L2:
        mov     eax, 0
.L3:
        pop     rbp
        ret
```
From now on, I'll keep in mind to compare hand optimizations with -O1/2/3 generated assembly.

PR #1425, opened 2022-04-28
Fix broken Doxygen image references

cadomani:
| | |
|:---|:---|
|**_Originating Project/Creator_**| @cadomani - ABEX (Alabama Burst Energetics eXplorer) |
|**_Affected Component_**| Doxygen Documentation |
|**_Affected Architectures(s)_**| N/A |
|**_Related Issue(s)_**| #1424 |
|**_Has Unit Tests (y/n)_**| n |
|**_Builds Without Errors (y/n)_**| y |
|**_Unit Tests Pass (y/n)_**| N/A |
|**_Documentation Included (y/n)_**| n |

---
## Change Description

While generating Doxygen documentation, the relative path references are not updated, and the site expects images to reside locally in an img/ path, but breaks when they cannot be found. This adds in a section at the bottom of the generation script to find all image files in Fw, Svc, and Drv, and copy them into the appropriate Doxygen folder so references are found.

## Rationale

Viewing markdown documentation retains images and diagrams that are helpful in understanding the framework, but the online documentation loses some usability when it does not show images alongside the components and ports it documents.

## Testing/Review Recommendations

Additional logic could be added to ensure that the find command was able to locate images in the provided directories, and check if the output path has received images to ensure copy was successful. 

## Future Work

None

PR #1599, opened 2022-07-27
Fix Autodocumentation (+ ci tweaks)

jsoref:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Fix Autodocumentation and make it easier to test such changes locally

## Rationale

The biggest change is renaming two files named `file.png`:
https://github.com/nasa/fprime/runs/7529492880?check_suite_focus=true#step:5:5430
> cp: will not overwrite just-created '/home/runner/work/fprime/fprime/docs/UsersGuide/api/c++/html/img/file.png' with '/home/runner/work/fprime/fprime/Svc/Framer/docs/img/top/file.png'

The other changes are to enable [`act`](https://github.com/nektos/act) to test the changes 

The expect.txt update is based on the check-spelling 0.0.20 update (in order for the update to get a green check from 0.0.19, the outgoing version, I left in some stale entries).

## Future Work

It might be worth it for the docs build system to eagerly check for conflicting files instead of relying on users to notice from very far away.

There are two other things I trip on in my forks:
* There's a thing that tries to fetch tags (by default I don't have any) -- if this isn't absolutely necessary, having it do something when there aren't any would be nice.
* The autodocumentation workflow absolutely needs that branch -- asking for it earlier (and dying faster) would be helpful


Svc/Deframer/docs/sdd.md
```diff
@@ -376,7 +376,7 @@ assign these numbers.
 **Topology 3: Buffers containing packet data:**
 
 <div>
-<img src="img/top/file.png" width=1000/>
+<img src="img/top/deframer.png" width=1000/>
```
bocchino, line 379: `deframer.png` is not a good choice here. All the diagrams in the Deframer SDD pertain to the deframer. This one pertains to file packets. If necessary we can call it `deframer-file.png`.

At a higher level, can someone explain to me what is going on here? What is "Autodocs"? Why is it apparently requiring that every file in the entire F Prime repository have a unique name, disregarding the path where it is located? This is a severe restriction and IMO not a very reasonable one. Can this restriction be relaxed? If not, how can we avoid the situation where we merge an update to the docs and everything seems fine, and someone later reports that "Autodocs" has failed because of global name collisions?

Svc/Framer/docs/sdd.md
```diff
@@ -199,7 +199,7 @@ The `eventLogger` instance sends event packets to the `framer` instance.
 **Topology 3: File packets:**
 
 <div>
-<img src="img/top/file.png" width=1000/>
+<img src="img/top/framer.png" width=1000/>
```
bocchino, line 202: See my comment above. If necessary we can call this one `framer-file` -- but again, why are we managing all file names in the entire F Prime repository in one global namespace? This is bad.

Svc/Deframer/docs/sdd.md
jsoref: I can only point to the code that does the flattening:
https://github.com/nasa/fprime/blob/1f7b07da365bc23e10df4184c35d83147b4f1a1e/docs/doxygen/generate_docs.bash#L102
It was introduced in #1425.

jsoref: Sorry, here's the entrypoint for this (it was reachable from the log, but there's no reason to force you to jump through that): https://github.com/nasa/fprime/blob/devel/.github/workflows/autodocs.yml

jsoref: To answer the other question, the reason nothing warned you as you developed your PR is that it's only running on pushes to the development branch: https://github.com/nasa/fprime/blob/1f7b07da365bc23e10df4184c35d83147b4f1a1e/.github/workflows/autodocs.yml#L2-L7

If it also ran on pull requests, you'd have noticed.

---

The reason I'm paying any attention to this is because I checked to make sure my PR merged cleanly without fallout and I was surprised to see that it didn't. So I chased back from the error message (which is one I don't particularly recognize – neither busybox nor macOS generate it, it appears to be a feature of gnu-cp).

bocchino: Thanks for that detailed explanation. In the near term, it looks like we should rename the files. Longer term, we should look into making the doc generation more robust.

jsoref: I'm happy to use the names you suggested. I have very limited relation to the project, so I just needed a placeholder to hang a discussion off.

LeStarch (approved):

PR #1605, opened 2022-08-02
Use the original sourcing command '.' for compatibility with shells where 'source' isn't defined

Anirban166:
| | |
|:---|:---|
|**_Affected Component_**| -/- |
|**_Affected Architectures(s)_**| -/- |
|**_Related Issue(s)_**| -/- |

---
## Change Description

Replaced `.` with its alias `source` that is defined when using shells like bash and zsh. 

## Rationale

The `source` command is not recognized while using the default shell, and I found this while I was setting up a virtual python environment for fprime on a Docker container running Linux (Ubuntu 20.04):
```sh
$ sudo python3 -m venv $HOME/fprime-venv
$ source $HOME/fprime-venv/bin/activate
/bin/sh: 47: source: not found
```
The issue resolves if I use `.` instead:
```sh
$ . $HOME/fprime-venv/bin/activate
(fprime-venv) $ 
```
This happens because my shell in the container isn't bash or any other shell implementation with `source` defined to be an alias of `.`:
```sh
$ ps -p $$
  PID TTY          TIME CMD
  832 pts/2    00:00:00 sh
```
It would be better to use just the original shell command for sourcing instead of its alias, so it does not go unrecognized for shells like the default one.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -42,7 +42,7 @@ use the path: `$HOME/fprime-venv`
 
 ```
 python3 -m venv $HOME/fprime-venv
-source $HOME/fprime-venv/bin/activate
+. $HOME/fprime-venv/bin/activate
 pip install -U setuptools setuptools_scm wheel pip
 ```
 
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -46,7 +46,7 @@ python3 -m venv $HOME/fprime-venv
 pip install -U setuptools setuptools_scm wheel pip
 ```
 
-> Note: `source $HOME/fprime-venv/bin/activate` must be run in each new terminal where the user wishes to use the virtual environment.
+> Note: `. $HOME/fprime-venv/bin/activate` must be run in each new terminal where the user wishes to use the virtual environment.
 
 ### Cloning the F´ Repository and Installing F´ Tools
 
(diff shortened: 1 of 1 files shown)


LeStarch (approved):
## Pre-CI Checks: Must perform before launching CI
- [x] No obviously malicious code introduced
- [x] PR is against `devel`. Use `Edit -> change branch` to fix, or close PR.

## Code Review
- [x] Code follows flight software quality guidelines
- [x] Changes are of reasonable length for review

## Procedural Checks
- [x] PR passes CI
- [x] Accept with `squash-and-merge`

Anirban166:
Just as a note to anyone using containers, I'd definitely recommend going for shell implementations like bash for their convenient features. Using `.` will play it safe with all shells for sourcing, but using default ones (dash for e.g.) in general or for the long run is not that ideal - I personally find the absence of certain features such as command history to be inconvenient.

Anirban166:
Most Linux containers would have bash installed, but if not explicitly set up<sup>1</sup> in the running shell environment or CLI, one can invoke it through the system shell:
```sh
$ /bin/sh -c bash
<user/root>@<containerID>: <Path relative to home> $
```

<sup>1</sup>Depends on how one sets up the container - for e.g., some may have set `ENV SHELL=/bin/bash` within their Dockerfile, some may specify to run bash for their `ENTRYPOINT`, or some via CMD (i.e. `CMD ["/bin/bash"]`, `docker run -it <image> /bin/bash`).

PR #1619, opened 2022-08-16
Update/less angry i2c driver

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Currently the Linux I2C driver fails on an FW_ASSERT when the I2C driver fails to open the device and a subsequent call comes in to read/write.  This proves restricting as often the calls to read/write come from a component not involved in the `open` process and are driven out of the static topology (e.g. a rate group).  Asserting almost guarantees the software cannot and will not run in the event that the I2C device fails to open.

This PR reduces that restriction.  Users can still ASSERT using the following pattern:

```
FW_ASSERT(i2cDriver.open() == SUCCESS)

or 

FW_ASSERT(readI2C_out() != I2cStatus::I2C_OPEN_ERR);
```

But should an assert be declined, the driver itself will report an error and not assert.

## Rationale

Failing to open a file is an error that can occur outside of software.  Communicating this failure to components "managing" I2C devices is convoluted.  Asserting on a device that attempts a read after an open failure doesn't make sense in all cases and thus should not be forced by the I2C implementation.

Users who want asserts can easily replicate the behavior as shown above.

Should be put in system reference and ensured to function as expected

## Future Work

None.

### DIFF
--- a/Drv/I2cDriverPorts/I2cDriverPorts.fpp
+++ b/Drv/I2cDriverPorts/I2cDriverPorts.fpp
@@ -16,7 +16,8 @@ module Drv {
     I2C_ADDRESS_ERR = 1 @< I2C address invalid
     I2C_WRITE_ERR = 2 @< I2C write failed
     I2C_READ_ERR = 3 @< I2C read failed
-    I2C_OTHER_ERR = 4 @< Other errors that don't fit
+    I2C_OPEN_ERR = 4 @< I2C driver failed to open device
+    I2C_OTHER_ERR = 5 @< Other errors that don't fit
   }
 
 }
--- a/Drv/LinuxI2cDriver/LinuxI2cDriver.cpp
+++ b/Drv/LinuxI2cDriver/LinuxI2cDriver.cpp
@@ -76,7 +76,9 @@ namespace Drv {
     )
   {
       // Make sure file has been opened
-      FW_ASSERT(-1 != this->m_fd);
+      if (-1 == this->m_fd) {
+          return I2cStatus::I2C_OPEN_ERR;
+      }
 
 #if DEBUG_PRINT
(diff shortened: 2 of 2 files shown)


LeStarch:
@bcmetz any chance I could get your opinion on this?  @bitWarrior this is the solution we discussed.

metzdigital:
Given the fact that there isn't an assert in the open() function I agree that you probably shouldn't then be asserting on the read/write calls. That being said I do think it's strange that it doesn't assert in the open() function. This is done only once at boot and if it fails you're just going to be swimming in a sea of I2C_OPEN_ERRs when rate groups start firing. I guess the question is, should the software assert if it hasn't been configured properly (pointed at a non existent device)? I noticed UART and SPI also do not assert on open. Enlighten me software elders.

timcanham (approved):

LeStarch:
@metzdigital the idea here is to allow more flexibility.  In some projects and use-cases you may want exactly as you described:  assert the software on a missing device to avoid swimming in errors.  However, not all projects wan this.  Some projects may wish to not brick a spacecraft if a dead sensor or device is not coming up.

When the I2C driver asserts itself, only one option is possible.  When the device returns an error, the caller/project can decide what to do.

e.g. Assert on open failure:  (in topology)


```
FW_ASSERT(i2cDriver.open(/path/to/device) == GOOD);
```

This has the same effect as you described.  However, it would allow a project to handle the code differently.

Same with the read/write calls.  The calling code could be:

```
FW_ASSERT(i2cWrite() == Status::OK)
```

Which would have the same effect as the code today.  However, for a less critical I2C device, the code could be:

```
if (i2Cwrite() == Status:OPEN_ERR)
{
    //TURN OF I2C and keep flying
}
```

PR #1607, opened 2022-08-02
Update/type refactor

LeStarch:
| | |
|:---|:---|
|**_Originating Project/Creator_**| |
|**_Affected Component_**|  |
|**_Affected Architectures(s)_**|  |
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|  |
|**_Builds Without Errors (y/n)_**|  |
|**_Unit Tests Pass (y/n)_**|  |
|**_Documentation Included (y/n)_**|  |

---
## Change Description

**This contains breaking changes**

This implements the initial "logical types" design approved in PR #1571.  These changes amount to:

1. Improving `StandardTypes.hpp` to include the new platform type definitions
2. Improving `BasicTypes.hpp` to include _MIN and _MAX definitions
3. Reworking `FpConfig.hpp` to use `typedef`

This includes some "breaking changes".  

1. `StandardTypes.hpp` implementors **must** implement new types, or include `DefaultTypes.hpp` to get default implementations.  See `Fw/Types/Linux/StandardTypes.hpp` for an example
2. `BasicTypes.hpp` no longer includes the types defined in `FpConfig.hpp`.  Users should switch to include `FpConfig.hpp` when needing project-defined types
3. `AssertArg` has been renamed `FwAssertArgType`.  Users must update.

This PR also includes some deprecated features:

1. Use of `NATIVE_INT_TYPE`, `NATIVE_UINT_TYPE` and `POINTER_CAST` is no longer advised.  Projects should update, at minumum, to use `PlatformIntType`, `PlatformUIntType` and `PlatformPointerCastType` respectively.
2. Use of `PlatformIntType`, `PlatformUIntType` is discouraged outside of API calls returning `int` and `unsigned int` respectively. Other uses should be replaced with new "logical types".  e.g. `NATIVE_INT_TYPE` in port calls should be replaced with `FwIndexType`.

Other restrictions:
1. `DefaultTypes.hpp` should only ever be included in a `StandardTypes.hpp` header.  All other users should be using `FpConfig.hpp` and trusting it to setup the types correctly.

## Future Work

Define new "logical types" when identified.
Replace uses of PRI[xud]NN for "logical types" with defined equivalent.

PR #1571, opened 2022-07-18
lestarch: adding numerical types (planned) design

LeStarch:
| | |
|:---|:---|
|**_Originating Project/Creator_**| |
|**_Affected Component_**|  |
|**_Affected Architectures(s)_**|  |
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**| n/a  |
|**_Builds Without Errors (y/n)_**| n/a |
|**_Unit Tests Pass (y/n)_**| n/a |
|**_Documentation Included (y/n)_**| yes |

---
## Change Description

This PR includes an in-draft design proposal for a refactor for the numerical types system.  There are several main items this is designed to address:

1. `NATIVE_INT_TYPE`, `NATIVE_UINT_TYPES` are used often in F´ but aren't clearly defined across platforms
2. The use of some types with some platforms generate compiler warnings or compiler errors
3. Easier to understand int types (e.g. size, index, etc.) don't exist and are often replace with either the native type, or with a type that is overkill (e.g. `U64` for size)

This proposes a new way of handling these types in the framework, and if approved...this will be implemented.

PR #1623, opened 2022-08-18
Update/types refactor as constants

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

See #1607 and #1615.  This is the final implementation of the design specified in #1571.

In short we have two goals for the implementation:

1. Do not increase binary size by allocating space for constants
2. Use type-safe constants instead of `#define`

This PR implements the design in #1571 using `static const` in structs (public-by-default classes):

For example:
```
struct PlatformLimits {
    static const int int_min = 0;
}
```

Allows usage of the int_min constant without allocating storage in the .rodata section of a final binary.


LeStarch:
@kevin-f-ortega I still have to refactor the documentation samples.

kevin-f-ortega (approved):
LGTM. I ran with it on my side.

PR #1629, opened 2022-08-19
FPP CMake integration

tiffany1618:
## Change Description

- Add parameter`GENERATED_SOURCES`, a list of sources generated by previous autocoders, to the `_filter_sources` function in `cmake/autocoder/autocoder.cmake`
- Change the `ai_xml_is_supported` function in `cmake/autocoder/ai_xml.cmake` to also filter sources against previously generated sources. This means that XML autocoder will only receive Ai.xml files whose corresponding Ac.cpp and Ac.hpp files have not already been generated by the FPP autocoder, preventing it from overwriting those files.

## Rationale

This change is necessary to support FPP v1.1.0, which will include C++ code generation for FPP enums, arrays, and structs, intended to replace the XML autocoder's C++ code generation for enums, arrays, and structs.

### DIFF
--- a/cmake/autocoder/ai_xml.cmake
+++ b/cmake/autocoder/ai_xml.cmake
@@ -19,11 +19,16 @@ autocoder_setup_for_individual_sources()
 ####
 function(ai_xml_is_supported AC_INPUT_FILE)
     ends_with(IS_SUPPORTED "${AC_INPUT_FILE}" "Ai.xml")
+    set(PREVIOUSLY_GENERATED)
+    if (ARGC GREATER_EQUAL 2)
+        set(PREVIOUSLY_GENERATED ${ARGV2})
+    endif()
+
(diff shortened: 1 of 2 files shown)


LeStarch (approved):
This looks good to me.  I'd like to test cmake 3.13, and see what happens.

LeStarch:
This code works as expected in CMake 3.21+, but breaks on CMake 3.13-3.16.  The reason is that the code that replicates the `cmake_language CALL` functionality supports a maximum number of arguments.

An easy fix should be to switch from `ARGN` to `ARG2` and pass GENERATED_FILES as a list rather than unrolling it.  I'll go ahead and try this fix.

LeStarch:
The changes worked, so I pushed a commit to this branch.  Once CI is done I will merge!

PR #1653, opened 2022-09-14
Adding modular dependencies to enable compiling with shared libraries

EbenezerA99:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Added modular dependencies to CmakeLists.txt for port type folders in `Drv/` and `Svc/` to enable compilation with the cmake flag `BUILD_SHARED_LIBS` set to ON. 

## Rationale

These additions are necessary for users who need to build with shared libs. 

Built the Ref deployment with and without the `BUILD_SHARED_LIBS` flag, and the deployment builds successfully. I also tested this with an external deployment and it built successfully. However, it is possible that there are additional changes required for other components/ports that were not included in the Ref or my external deployments.

## Future Work

It would be a good idea to include this in the CI workflow so that this functionality is continually supported!

### DIFF
--- a/Drv/DataTypes/CMakeLists.txt
+++ b/Drv/DataTypes/CMakeLists.txt
@@ -10,4 +10,8 @@ set(SOURCE_FILES
   "${CMAKE_CURRENT_LIST_DIR}/DataBuffer.cpp"
 )
 
+set(MOD_DEPS
+    Fw/Port
+)
+
 register_fprime_module()
--- a/Svc/Cycle/CMakeLists.txt
+++ b/Svc/Cycle/CMakeLists.txt
@@ -12,5 +12,6 @@ set(SOURCE_FILES
 )
 set(MOD_DEPS
     Os
+    Fw/Port
 )
 register_fprime_module()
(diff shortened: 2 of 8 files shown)


Joshua-Anderson:
The cmake unit tests have a [shared lib unit test](https://github.com/nasa/fprime/blob/master/cmake/test/src/test_ref_shared.py) that we run as part our CI checks. We should investigate why this test isn't failing

LeStarch:
@EbenezerA99 can you move this work to devel?

LeStarch:
May be an underlying CMake issue.  We'll investigate.

LeStarch:
@EbenezerA99, I just ran the following on my Linux machine and all worked as expected:

```
cd Ref
fprime-util generate -DBUILD_SHARED_LIBS=ON
fprime-uitl build
LD_LIBRARY_PATH=`pwd`/build-artifacts/Linux/lib
```

Can you elaborate the problems you are seeing in the build?  Version you are running?  I am trying to reproduce your issue, but it doesn't seem like an issue on `Linux/devel`

EbenezerA99:
@LeStarch Sorry for not getting back to you sooner!

It also builds for me just fine on Linux, however it is not building on Darwin

I'm using fprime-tools=3.1.1

LeStarch:
Ok, I'll try it on Darwin

LeStarch:
I was able to confirm this breaks on macOS.  Making sure this fixes it.

LeStarch (approved):

PR #1667, opened 2022-09-23
Improve `getMemUtil()` of the OSAL Linux port

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**| OSAL/Linux/`getMemUtil()` |
|**_Affected Architectures(s)_**| OSAL |

---
## Change Description

This PR aims to improve the implementation of `getMemUtil()` of the OSAL Linux port.

## Rationale

To retrieve this type of information it is better to call the syscall `sysinfo()` (see [here](https://man7.org/linux/man-pages/man2/sysinfo.2.html)) than to read `/proc/meminfo` and parse its contents.

I don't know if we should take into account the swap in the calculations and therefore bring up the total and used virtual memory.

If we wanted to have the total amount of virtual memory, it would be :

```cpp
  U64 totalVirtualMem = memory_info.totalram;
  totalVirtualMem += memory_info.totalswap;
  totalVirtualMem *= memory_info.mem_unit;
```

If we wanted to have the total amount of virtual memory used, it would be :

```cpp
  U64 virtualMemUsed = memory_info.totalram - memory_info.freeram;
  virtualMemUsed += memory_info.totalswap - memory_info.freeswap;
  virtualMemUsed *= memory_info.mem_unit;
```

## Future Work

Given the memory footprint and the frequency with which the function is triggered, it would be better to allocate the `memory_info` structure at compile time in the `BSS` segment.

### DIFF
--- a/Os/Linux/SystemResources.cpp
+++ b/Os/Linux/SystemResources.cpp
@@ -9,10 +9,8 @@
 // acknowledged.
 //
 // ======================================================================
-#include <cstdio>              /* fopen() */
-#include <cstdlib>             /* scanf */
-#include <sys/vfs.h>            /* statfs() */
-#include <sys/sysinfo.h>		/* get_nprocs() */
+#include <cstdio>
+#include <sys/sysinfo.h>
 #include <cstring>
 #include <Os/SystemResources.hpp>
 #include <Fw/Types/Assert.hpp>
--- a/Os/Linux/SystemResources.cpp
+++ b/Os/Linux/SystemResources.cpp
@@ -82,31 +82,17 @@ namespace Os {
     }
 
     SystemResources::SystemResourcesStatus SystemResources::getMemUtil(MemUtil &memory_util) {
-        FILE *fp = nullptr;
-        NATIVE_INT_TYPE total = 0;
-        NATIVE_INT_TYPE free = 0;
-        // Fallbacks
-        memory_util.total = 1;
-        memory_util.used = 1;
 
-        fp = fopen("/proc/meminfo", "r");
-        if (fp == nullptr) {
+        struct sysinfo memory_info;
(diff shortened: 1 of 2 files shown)


LeStarch:
Note: I fixed the spelling errors

Os/Linux/SystemResources.cpp
```diff
@@ -82,31 +80,17 @@ namespace Os {
...
-            return SYSTEM_RESOURCES_ERROR;
-        }
-        fclose(fp);
 
-        // Check results
-        if (total < 0 or free < 0 or total < free) {
+        struct sysinfo memory_info;
+        sysinfo(&memory_info);
+        
+        if (memory_info.totalram < memory_info.freeram) {
```
ThibFrgsGmz, line 87: @LeStarch Do you think it is still necessary to do this check? I kept it by mimicry with the old code, but in my opinion it's not useful anymore because theoretically it's the `sys_info()` function that takes care of it.

LeStarch: I cannot see in the man page where it does this.  Thus I'd keep it because it prevents the subtraction later from going off the rails.

Since we could assume that this is true, we could use an FW_ASSERT, but I think it is ok how it is as there is an establish error-return pattern.

LeStarch:
C++ failure is know.  Accepting!

LeStarch (approved):
This reduces our a lot of code!

ThibFrgsGmz:
@LeStarch @sfregoso What are your thoughts on taking the swap into account?

Do I make a second `PR` to put the variable in the `BSS`? It bothers me a bit to allocate on the stack such a "big" structure cyclically.

FYI, I'm also trying to optimize the `getCpuTicks` function.

PR #1678, opened 2022-09-29
added missing cpp file for baremetal

kevin-f-ortega:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Updated CMakeList file to include the IntervalTimer.cpp for bare metal.

## Rationale

This cpp file is needed when building for bare metal.

## Future Work


arizvi786 (approved):
LGTM

LeStarch:
I count 7 baremetal stub files, and 4 listed for compilation.  Should we not be adding more?

LeStarch (approved):
This change seems correct, but I wonder if we need to add more of the missing stubs?

kevin-f-ortega:
> This change seems correct, but I wonder if we need to add more of the missing stubs?

Let me double check.

kevin-f-ortega:
@LeStarch updated cmake list file.

LeStarch (approved):
@kevin-f-ortega this is now "right" or at least better that it was. However, there is a deeper problem here that another project discovered:

What if a user wants to use "BAREMETAL" for threads, but posix files?  Nothing specifically prevents this except for the structuring of this cmake file.  Same with queues etc.

As we expand the use of `BAREMETAL` we really ought to consider what it means.

As we look to alternative OS implementations, we really out to make structuring them easier.

Still, these problems are only loosely affiliated, so this PR should be `approved`

LeStarch:
@kevin-f-ortega squash and merge, you have the honors.

Issue #1681, opened 2022-09-29
Modules With No Dependencies are Stuck With No Dependencies


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

Autocoder models producing no `MODULE_DEPENDENCIES` in the autocoder/cmake system result in modules that are never scanned to look for new `MODULE_DEPENDENCIES`.

In order to work properly, the autocoder implementer needs to tell us if a file will result in the above change (i.e. that file needs to be watched) or not.  There is no way to automatically detect if a file needs to be watched.

PR #1682, opened 2022-09-29
lestarch: cleaned-up cmake autocoder handling. Fixes: #1431, #1680, and #1681

### LINKED ISSUE #1431: Unanticipated Regenerations
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

Ref occasionally regenerates the build-cache between the build and the installation step.  Why?

Reports of regeneration of AC files if they are touched.  Why?

### LINKED ISSUE #1680: Empty Models Break CMake
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

If a file results in no generated files.  It will break the CMake build. Empty files should be allowed as a placeholder and at no point is the autocoder required generate files for every module.

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

This PR addresses three inconsistencies in the build system:

1. #1431, caused by attempting to mark the wrong files automatically as CMAKE_CONFIGURE_DEPENDS
2. #1681, caused by attempting to automatically mark any file as CMAKE_CONFIGURE_DEPENDS
3. #1680, similar to #1680 but causing empty .fpp files to break builds due to the same fundamental flawed assumption about models.

`CMAKE_CONFIGURE_DEPENDS` is necessary to cause CMake to look for new dependencies between a given module target and other module targets. If done incorrectly it results in #1431.  In cleaning this up another bug (#1681) was discoverd as the the autocoder was erroneously trying to mark files as `CMAKE_CONFIGURE_DEPENDS` only when there was a non-zero number of dependencies already detected.  Hence, #1680 was also discovered as an empty module has no dependencies but is valid.

This PR fixes #1431 by cleaning up #1681 and corrects the few outstanding issues needed to resolve #1680 that weren't caught by #1681.


kevin-f-ortega (approved):
LGTM. Nice work and good catches!

PR #1694, opened 2022-09-30
Revise file ID type for FSW assertions

bocchino:
The correct type for numeric file IDs is U32, because the file ID is a 32-bit hash. The code being replaced breaks when NATIVE_UINT_TYPE has a bit width smaller than 32.

@LeStarch The other issue we looked at (format of printing assert arguments) is already fixed in devel. The version we looked at was behind the head of devel.

### DIFF
--- a/Fw/Types/Assert.hpp
+++ b/Fw/Types/Assert.hpp
@@ -8,7 +8,7 @@
 #else // ASSERT is defined
 
 #if FW_ASSERT_LEVEL == FW_FILEID_ASSERT
-#define FILE_NAME_ARG NATIVE_UINT_TYPE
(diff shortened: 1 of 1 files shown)


bocchino:
> The other issue we looked at (format of printing assert arguments) is already fixed in devel. The version we looked at was behind the head of devel.

Actually, strike that. I'm seeing some more issues regarding assertion arguments. Still working on it.

bocchino:
The problem I see is that there are still %d format specifiers in Assert.cpp. I'll scrub those and add the fix to this PR.

bocchino:
Closed in favor of #1697.

PR #1697, opened 2022-09-30
Fix type specifications in framework

bocchino:
* Correct inconsistencies in uses of types
* ~~Remove NATIVE_INT_TYPE and NATIVE_UINT_TYPE from the FPP model. These are not serializable types, because they have a platform-variable size. They should not appear in FPP models.~~
* ~~Change the use of NATIVE_UINT_TYPE in Svc/Sched to U16. This seems reasonable for now. We can use FwIndexType here, but this will require an update to the FPP code generator. (We should update the FPP code generator to support the new types anyway.)~~

Supersedes #1694.

### DIFF
--- a/Fw/Types/Assert.hpp
+++ b/Fw/Types/Assert.hpp
@@ -8,7 +8,7 @@
 #else // ASSERT is defined
 
 #if FW_ASSERT_LEVEL == FW_FILEID_ASSERT
-#define FILE_NAME_ARG NATIVE_UINT_TYPE
+#define FILE_NAME_ARG U32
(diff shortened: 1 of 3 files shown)


bocchino:
OK, it looks like I forgot to change the schedIn handlers that use NATIVE_UINT_TYPE. We may want to hold off on this change until we have the FPP update, because it's forcing a change in user code. For now I'll back out the NATIVE_UINT_TYPE change and resubmit.

LeStarch (approved):
Looks good to me!

kevin-f-ortega (approved):
This looks good to me.

PR #1700, opened 2022-09-30
Make `Ref/Main.cpp` agnostic of F´

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**| `Ref/Main.cpp` |

---
## Change Description

This PR aims to make the `Ref/Main.cpp` of the F' framework as agnostic and standalone as possible.

## Rationale

In the event that a unit supplier delivers a product developed with F´ to an integrator, it must be able to deliver it as a third-party library.

The older version of `Ref/Main.cpp` is tightly coupled to the framework and may put off the user for fear of introducing unnecessary dependencies to the end integrator.

With this PR, the `Ref/Main.cpp` deployment shows the user how to wrap the F´ deployment using a minimalist API. 
This API has three functions:
- `Initialize()` : which is responsible for gathering all the initialization functions of the library.
- `Deinitialize()` : performs the reverse operation of Initialize.
- `run_one_cycle()` : allowing the user to cycle the whole library.

I took the liberty of cleaning up the `Ref/Top` folder by deleting files that seemed unnecessary for the deployment and that added noise to the user.

## Future Work

The next step would be to gather all the static libraries from the deployment into a single archive for easy integration into a client's code base. (see https://github.com/nasa/fprime/discussions/1657)


Anirban166 (approved):

LeStarch:
I don't understand the advantage of making `Ref` a library.  The `Ref` app is a sample deployment.  Projects are expected to use their own custom deployment on their project. 

I can see the value of making the F´ project a library (everything except `Ref` and `RPI`) as this might make linking or building F´ more efficient.

If your project really needs `Ref` as a library, perhaps create a separate repository that sets this up?

ThibFrgsGmz:
@LeStarch The interest in putting `Ref` in the library is to meet a need of the unit suppliers.

Today, each equipment manufacturer must be able to deliver its material in at least two forms.
- Either in a library that can be integrated into the customer's computer (in a digital twin version or in a flight version).
- Or in a standalone version where the unit supplier provides a processing unit to host its processing.

In my experience this is a recurring need of Unit Suppliers, so I think that integration mode should be present in the reference deployment and maintained as part of the project. We can see this as a best practice.

I'm not against doing it in another repo but it might get lost because we don't have the `awesome-fprime`  yet that lists all the community projects.

codeflight1:
> I'm not against doing it in another repo but it might get lost because we don't have the `awesome-fprime` yet that lists all the community projects.

Speaking of which, I've been working on a PR for fprime-community/fprime-contributions to bring it more in line with other awesome- repos, but perhaps it would be better to separate those two?

LeStarch:
@codeflight1 you should start a new repository if you want to make `fprime-awesome`.   I can set that up for you.  `fprime-contributions` is a grab bag of stuff we needed to find a place for in the early days of open source before we figured out a better model.  It contains a few random components and other things.

LeStarch:
@ThibFrgsGmz in the long term, we intend to move `Ref` and `RPI` out of the base repo.  This is because it represent the old project structure and it can be jarring to users to be presented one model to initially learn fprime and then be forced into another model as intermediate users.

If the hesitation for doing this kind of work in a new repository is the lack of an `fprime-awesome` repository, let's get one setup.  This is something I'd love for the community to take initial ownership of because we have limited direct experience here.

codeflight1:
That would be awesome, I'd be happy to maintain it.

LeStarch:
@codeflight1 done.  I'd request three things:

1. Highlight the system reference (https://github.com/fprime-community/fprime-system-reference)
2. Let me know if you need support
3. Indicate that the NASA fprime team isn't specifically endorsing any specific projects or services.  Especially for flight.

Thanks!

PR #1703, opened 2022-10-04
Make UTs work with FW_ASSERT_LEVEL == FW_FILEID_ASSERT

bocchino:
Some unit tests were broken in this configuration.

### DIFF
--- a/Fw/Types/test/ut/TypesTest.cpp
+++ b/Fw/Types/test/ut/TypesTest.cpp
@@ -795,7 +795,11 @@ void AssertTest() {
 
         private:
 
+#if FW_ASSERT_LEVEL == FW_FILEID_ASSERT
+            FILE_NAME_ARG m_file = 0;
+#else
             FILE_NAME_ARG m_file = nullptr;
+#endif
             NATIVE_UINT_TYPE m_lineNo = 0;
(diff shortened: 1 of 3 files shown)


timcanham:
Part of the reason for having it a macro is that if you leave the file name part in, it ends up embedding a bunch of strings in the code.

timcanham:
I agree that we should have builds that exercise the scaling features like this. Things break if we don't.

bocchino:
> I agree that we should have builds that exercise the scaling features like this. Things break if we don't.

Right. One idea is to have a CI script that rewrites the config file and then rebuilds and reruns UTs, over two or more configurations. We can't test all combinations of config settings -- there are too many -- but we could pick a few and test them. For example, if the config settings are x, y, and z, and the possible values are { 0, 1 }, { 0, 1 }, { 0, 1 }, and the default values are { 0, 0, 0 } we could have a script that compiles and runs tests with the defaults and then compiles and runs tests with { 1, 1, 1 }.

LeStarch (approved):

LeStarch:
@timcanham did you want to review the code, or shall I merge this?

timcanham (approved):

PR #1734, opened 2022-10-24
Fix autogeneration of header guards and tests

SMorettini:
| | |
|:---|:---|
|**_Affected Component_**| Autocoder  |
|**_Related Issue(s)_**|  |

---
## Change Description

Added the support to multi level namespace to enum and array.

## Rationale

If an array or an enum is created under multiple level namespace, the header guards generated are of the form:
```
#ifndef Namespace::namespace_nameobject_HPP_
```
This form cannot compile because of the usage of `::`.

I fix it by using the approach already used in the port templates.


LeStarch:
@bocchino would you review?  It looks correct to me and it qualifies as a bug fix.

bocchino (changes requested):
This is a welcome improvement.

This change fixes the header guards but not the namespace declarations. In the hpp file, I think we will still get code of the form
```
namespace A::B {

}
```
which doesn't compile under C++11. The way that components and ports generate the namespace declarations should work here as well.

Also, we should add at least one unit test for each of { enum, array } to make sure that the fix works as intended.

LeStarch:
@SMorettini are you up for working through these requests, or would you like us to take ownership of this work and carry it through?

SMorettini:
@LeStarch I would like to work on it. 

@bocchino I'l change the namespace generation as you said. About 

> Also, we should add at least one unit test for each of { enum, array } to make sure that the fix works as intended.

I don't know how the unit test should be made but I'll try.

LeStarch:
@SMorettini is this work still in progress?  Otherwise this PR is fairly old and I'd like to close it.

SMorettini:
Sorry for the delay. I finally implemented all the requested changes. 

Now I changed the namespace to make them work with C++11, here is a snippet from the tests:
```
#ifndef EXAMPLE_SUBNAMESPACE_ARRAYTYPE_HPP_
#define EXAMPLE_SUBNAMESPACE_ARRAYTYPE_HPP_

...

namespace Example {
namespace SubNameSpace {
```

@bocchino are the changes you asked for?

bocchino (approved):
Looks good!

LeStarch:
Passes CI!  Thanks!

PR #1735, opened 2022-10-24
Fix tests failing because of variable optimized out

SMorettini:
| | |
|:---|:---|
|**_Affected Component_**| Drv/TcpClient - Drv/Udp  |
|**_Related Issue(s)_**|  |

---
## Change Description

The unit tests of Drv/TcpClient and Drv/Udp failed if some optimization flag where used. The cause of the failure was that the `m_spinner` in `while (not m_spinner) {}` was optimized out and so the test was not waiting. Using `std::atomic` m_spinner is never optimized and the test work for any cases.

### DIFF
--- a/Drv/TcpClient/test/ut/Tester.hpp
+++ b/Drv/TcpClient/test/ut/Tester.hpp
@@ -115,7 +115,7 @@ namespace Drv {
       Fw::Buffer m_data_buffer;
       Fw::Buffer m_data_buffer2;
       U8 m_data_storage[SEND_DATA_BUFFER_SIZE];
-      bool m_spinner;
+      std::atomic<bool> m_spinner;
 
   };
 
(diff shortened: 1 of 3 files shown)


Joshua-Anderson:
@LeStarch what is our C++11 philosophy for unit tests? Unit tests already require full C++11 support for gtest, so I'm not sure if there's any harms to requiring std::atomic for unit tests.

LeStarch:
@Joshua-Anderson you are absolutely right that this is probably ok for unittest code.  I was trying to determine if the `volatile` would work because then I would not have to clarify "probably".

bocchino:
The `volatile` keyword ensures that the compiler won't optimize away the memory access, but it doesn't prevent a hardware-level data race (unguarded concurrent read/write). We should avoid all hardware-level data races by using std::atomic, lock/unlock, or some other concurrency control mechanism.

LeStarch:
@bocchino in this case this isn't a hardware-level race condition....it is just inter-thread synchronization.

bocchino:
The point is that on modern processors you should not use plain reads and writes to do inter-thread synchronization. You need to use C++ atomic, or an OS mutex, or an assembly-language memory barrier, or something.

`volatile` will prevent the compiler from deleting the access, but a volatile read/write is not enough.

LeStarch (approved):

LeStarch:
Users who are concerned would not run UTs on hardware anyway.

PR #1736, opened 2022-10-25
Add tests for the `Time` class in Fw

gazayas:
| | |
|:---|:---|
|**_Originating Project/Creator_**| Gabriel Zayas |
|**_Has Unit Tests (y/n)_**| Yes |
|**_Unit Tests Pass (y/n)_**| Yes |

---
## Change Description

I added a few tests for the `Time` class in the `Fw` component.

## Rationale

There was a comment indicating the tests should check the `add` function, but no tests had been implemented.

https://github.com/nasa/fprime/blob/91c578ee37c01ba92e556dbca0287d82f9f18b95/Fw/Time/test/ut/TimeTest.cpp#L32-L42

The tests are pretty straight-forward, but I particularly gave attention to how I commented each test since I felt it could've used a little cleaning up.

## Future Work

Apart from `add` I added tests for checking comparison logic, but if there are any other parts of the class that aren't covered here, it might be a good idea to test those parts as well.

### DIFF
--- a/Fw/Time/test/ut/TimeTest.cpp
+++ b/Fw/Time/test/ut/TimeTest.cpp
@@ -24,24 +24,41 @@ TEST(TimeTestNominal,MathTest) {
     Fw::Time time1;
     Fw::Time time2;
 
-    // Try various operations
+    // Comparison
+    time1.set(1000,1000);
+    time2.set(1000,1000);
+    ASSERT_TRUE(time1 == time2);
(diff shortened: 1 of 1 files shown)


LeStarch (approved):

LeStarch:
This is failing on an unrelated problem that has already been fixed in devel.  I am not sure why CI is unhappy.  I am going to merge it and keep an eye on the post-merge build.

Issue #1693, opened 2022-09-30
Switch to Github Actions Code Scanner


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

Switch to Github actions code scanner to replace LGTM.

PR #1739, opened 2022-10-27
Switch to Github Actions Code Scanner

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Affected Architectures(s)_**| Github Actions |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/1693 |
|**_Unit Tests Pass (y/n)_**| |

---
## Change Description

Adds a Github Actions workflow that runs the same queries as the current LGTM set up.

## Rationale

[lgtm.com](lgtm.com) will be shut down at the end of the year.

## Future Work

We may want to remove the LGTM configuration files once it's fully shut down.

## Additional comments
I currently have it set up to run on each commit and pull request to `master` and `devel` as I see this seems to be the convention in the workflows of this repo, but we could restrict to only running when source files are modified.

Here's the result of the queries [on Github Actions results](https://github.com/thomas-bc/fprime/security/code-scanning?query=is%3Aopen+branch%3Agh-code-scanner). Comparing to [the LGTM results](https://lgtm.com/projects/g/nasa/fprime/alerts/?mode=tree), we see the same warnings with an additional 2 on the Github Actions side.

### DIFF
--- a/.github/actions/codeql/security-pack.yml
+++ b/.github/actions/codeql/security-pack.yml
@@ -1,0 +1,19 @@
+name: "CodeQL security and quality"
+
+queries: 
+  - uses: security-and-quality
+
+query-filters:
+  - include: 
+      id: cpp/incorrect-not-operator-usage
+  - include: 
+      tags contain: correctness
+  - include: 
+      tags contain: reliability
+
+paths-ignore:
+  - docs/
+  - cmake/docs/
+  - cmake/test/
+  - Autocoders/Python/src/fprime_ac/utils/DiffAndRename.py
+  - Autocoders/Python/src/fprime_ac/utils/pyparsing.py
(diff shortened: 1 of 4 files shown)


LeStarch (changes requested):
Looks good to me.  We should fix the one error, so that doesn't count against us.  Then we'll merge.

thomas-bc:
I added in the fix for the Python error. Turns out the if statement isn't needed at all, the behavior would be the same for any list length >0. There is no check to ensure that the list isn't empty though, but I didn't want to risk to break things in the scope of this PR so I didn't add that. It also looks like these functions aren't used anyways.

LeStarch (approved):
Approved!

LeStarch:
This fix looks right to me.  Also `.split` should never return an empty list as it would always return back a list of at least one element.

PR #1741, opened 2022-10-27
lestarch: adding Svc::ComQueue component

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

This adds a cleaned-up version of `Svc::ComQueue`, which is running in the `SystemReference`


LeStarch:
@bocchino you volunteered to review this one.

bocchino:
Looks great! This is a big PR, so I'll add my comments in stages.

## Comments on the SDD

The SDD is very clear and well structured overall. Here are a few comments:

1. The communication adapter links are broken.
2. The spelling of types is inconsistent. Sometimes it's A::B, sometimes it's A:B, and sometimes it's A.B. A:B is not correct. In each context we should should choose one of A::B (C++) or A.B (FPP).
3. The component diagram is incomplete. I recommend using the diagram below.
4. The port list is incomplete. I recommend using the port list below.
5. The description of the run port in s. 4.6.4 is unclear. Looking at the implementation, I see that it reports and then resets the high-water marks for the two sets of queues. This could be stated more clearly.
6. There's no description of the retryReturn handler. Also, I don't see the new retry logic described in the SDD. 
7. I recommend adding sections that summarize the telemetry and event interfaces. For telemetry, you can have a table of rows with (name, type, description). For events, you can have a table of rows with (name, description).

ComQueue

| Kind | Name | Port Type | Usage |
|------|------|-----------|-------|
| `output` | `comQueueSend` | `Fw.Com` | Fw::ComBuffer output port |
| `output` | `buffQueueSend` | `Fw.BufferSend` | Fw::Buffer output port |
| `async input` | `comStatusIn` | `Fw.SuccessCondition` | Port for receiving the status signal |
| `async input` | `comQueueIn` | `[ComQueueComPorts] Fw.Com` | Port array for receiving Fw::ComBuffers |
| `async input` | `buffQueueIn` | `[ComQueueBufferPorts] Fw.BufferSend` | Port array for receiving Fw::Buffers |
| `async input` | `run` | `Svc.Sched` | Port for scheduling telemetry output |
| `sync input` | `retryReturn` | `Fw.BufferSend` | Port for receiving Fw.Buffers back for retry |
| `output` | `retryDeallocate` | `Fw.BufferSend` | Port for deallocating Fw.Buffers after potential retry |
| `event` | `Log` | `Fw.Log` | Port for emitting events |
| `text event` | `LogText` | `Fw.LogText` | Port for emitting text events |
| `time get` | `Time` | `Fw.Time` | Port for getting the time |
| `telemetry` | `Tlm` | `Fw.Tlm` | Port for emitting telemetry |

Svc/ComQueue/ComQueue.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+void ComQueue::buffQueueIn_handler(const NATIVE_INT_TYPE portNum, Fw::Buffer& fwBuffer) {
+    const NATIVE_INT_TYPE queueNum = portNum + COM_PORT_COUNT;
+    // Ensure that the port number of buffQueueIn is consistent with the expectation
+    FW_ASSERT(portNum >= 0 && portNum < BUFFER_PORT_COUNT, portNum);
+    FW_ASSERT(queueNum < TOTAL_PORT_COUNT);
+    this->enqueue(queueNum, QueueType::BUFFER_QUEUE, reinterpret_cast<const U8*>(&fwBuffer), sizeof(Fw::Buffer));
+}
+
+void ComQueue::comStatusIn_handler(const NATIVE_INT_TYPE portNum, Fw::Success& condition) {
```
bocchino, line 138: This implementation of `comStatusIn_handler` does not agree with what the SDD says. See https://github.com/LeStarch/fprime/blob/update/com-queue/Svc/ComQueue/docs/sdd.md#463-comstatusin.

Svc/ComQueue/ComQueue.cpp
```diff
@@ -0,0 +1,308 @@
...
+    FW_ASSERT(this->m_state == READY);
+    this->m_comRetry = comBuffer;
+    this->comQueueSend_out(0, comBuffer, 0);
+    this->m_state = WAITING;
+}
+
+void ComQueue::sendBuffer(Fw::Buffer& buffer) {
+    FW_ASSERT(this->m_state == READY);
+    this->m_lock.lock();
+    // Buffer ownership transferred to downstream component. Clear local reference.
```
bocchino, line 227: Why are we clearing out the retry buffer here? We may or may not be sending from the retry buffer.

Svc/ComQueue/ComQueue.cpp
```diff
@@ -0,0 +1,308 @@
...
+    FwIndexType priorityIndex = 0;
+    FwIndexType sendPriority = 0;
+    // Check that we are in the appropriate state
+    FW_ASSERT(this->m_state == READY);
+
+    // Clean-up retry buffer when sending something new
+    this->m_lock.lock();
+    if (this->m_bufferRetry.getData() != nullptr) {
+        this->retryDeallocate_out(0, this->m_bufferRetry);
+        this->m_bufferRetry = Fw::Buffer(nullptr, 0, 0);
```
bocchino, line 261: Same question as above: Why clear the retry buffer here, when we may not be sending from it? Shouldn't we clear it if and only if we send from it?

bocchino:
## Comments on Implementation

Implementation looks good, see inline comments.

bocchino:
## Comments on Unit Tests

Looks great! Excellent test structure and code coverage.

This completes my round of review comments.

Svc/ComQueue/ComQueue.cpp
```diff
@@ -0,0 +1,308 @@
...
+    // Check that we are in the appropriate state
+    FW_ASSERT(this->m_state == READY);
+
+    // Retry the last message based on type
+    if (this->m_lastIndex < COM_PORT_COUNT) {
+        this->sendComBuffer(this->m_comRetry);
+    } else {
+        // Ensure the retry buffer ownership has been returned before resending it.
+        this->m_lock.lock();
+        FW_ASSERT(this->m_bufferRetry.getData() != nullptr);
```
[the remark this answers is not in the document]
LeStarch: Yes, we could move the clear here.  Clearing it above is a superset as it clears every time a buffer is sent, not just retry.  Since the send of any buffer should result in a new retry buffer it is safe to clear retry in preparation for the callback.

Svc/ComQueue/ComQueue.cpp
LeStarch: Here we are selecting a new item and are not retrying.  Thus the stored retry buffer is sent for final deallocation.  We clear the local reference as the data is now invalid.

Svc/ComQueue/ComQueue.cpp
bocchino: My concern is on its face the code is sending buffer and clearing bufferRetry, and there is no obvious relationship between these two variables. What if we hit this code and the retry buffer is not null and buffer is not the retry buffer? Then we will send buffer and clear the retry buffer, resulting in a lost send. Maybe this can't happen, but I think it would be clearer to write the code so it's obvious this can't happen. Or maybe you could add an assertion against this case.

Svc/ComQueue/ComQueue.cpp
LeStarch: Theoretically we can reduce to this case (READY -> no retry -> deallocate) and the the retry case (RETRY -> clear -> callback).

Clearing on every buffer send means we'd clear twice. in this case.

Svc/ComQueue/ComQueue.cpp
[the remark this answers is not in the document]
bocchino: I think if we cleared the retry buffer here, then it's more obvious that we should be clearing it, and we won't have a lost buffer.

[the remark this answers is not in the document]
bocchino: > Since the send of any buffer should result in a new retry buffer it is safe to clear retry in preparation for the callback.

I think this assumes that when we send a buffer B, either we are sending the retry buffer(i.e., B == m_retryBuffer) or there is no retry buffer to send (i.e., m_retryBuffer == NULL). If neither of these conditions holds, then I think we will have a lost buffer.

Svc/ComQueue/ComQueue.cpp
bocchino: OK. But why are we deallocating the retry buffer here at all? Shouldn't we send it and let it be deallocated downstream? If we're deallocating the buffer here, then it means it will never be sent.

bocchino: The way I would expect it to work is that we send a buffer B, the send either succeeds or fails. If it succeeds, we never see B again. If it fails, then B comes back for retry and we retry by resending. I'm not sure why we would ever deallocate B here.

bocchino: Unless there's some sort of timeout and we give up and deliberately drop a buffer?

bocchino: I think it would help to see the retry algorithm written out in the SDD.

LeStarch: Ownership is received from upstream.  It is transferred downstream and handed back for the purposes of retry.  If retry is not necessary it is deallocated.  Otherwise it is transfered downstream and handed-back again.

Svc/ComQueue/ComQueue.cpp
LeStarch: I went ahead and cleaned this up.

Svc/ComQueue/ComQueue.cpp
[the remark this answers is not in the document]
LeStarch: Agreed. I have fixed this.

LeStarch:
@bocchino I renamed the buffer as discussed. I'd request a final review from you.

bocchino:
The latest changes look good! We discussed a few small changes to the state machine diagram offline. For reference, below is my summary of the algorithm. The bold parts are the ones that don't quite correspond to the diagram.

algorithm

bocchino:
Here is the asciidoc, also for reference:
```
. while (true)
.. block and wait for a message _M_. On receipt of _M_:
... *if _M_ is a returned buffer _B_, then*
.... *assert _returnedBuffer == EMPTY_*
.... *set _returnedBuffer = B_*
... else if _M_ is an incoming buffer _B_ then
.... if state is _READY_ then
..... Send _B_
..... Go to _WAITING_ state
.... else add _B_ to _buffQueue_ `// state is WAITING or RETRY`
... else if _M_ is _SUCCESS_ status then
.... Assert that _returnedBuffer =/= EMPTY_
.... if state is _WAITING_ then
..... Deallocate _returnedBuffer_
..... if buffQueue is empty then go to _READY_ state
..... else dequeue _B_ from buffQueue and send _B_
.... else if state is _RETRY_ then
..... *send _returnedBuffer_*
..... go to _WAITING_ state
.... else assert failure (should not happen in _READY_ state)
.... set _returnedBuffer == EMPTY_
... else `// M is FAILURE status`
.... if state is _WAITING_ then go to _RETRY_ state
.... else if state is _RETRY_ then do nothing
.... else assert failure (should not happen)
```

LeStarch:
@bocchino fixed diagram, spelling, and added the missing assert.

Svc/ComQueue/ComQueue.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+void ComQueue::buffQueueIn_handler(const NATIVE_INT_TYPE portNum, Fw::Buffer& fwBuffer) {
+    const NATIVE_INT_TYPE queueNum = portNum + COM_PORT_COUNT;
+    // Ensure that the port number of buffQueueIn is consistent with the expectation
+    FW_ASSERT(portNum >= 0 && portNum < BUFFER_PORT_COUNT, portNum);
+    FW_ASSERT(queueNum < TOTAL_PORT_COUNT);
+    this->enqueue(queueNum, QueueType::BUFFER_QUEUE, reinterpret_cast<const U8*>(&fwBuffer), sizeof(Fw::Buffer));
+}
+
+void ComQueue::comStatusIn_handler(const NATIVE_INT_TYPE portNum, Fw::Success& condition) {
```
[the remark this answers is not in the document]
LeStarch: FW_ASSERT is an assert.

LeStarch:
Note: the mac build error are now fixed in `devel`

LeStarch:
@bocchino per further review, retry belongs with the radio not in the priority queue.  This update has been made.  It should now be ready for final review.

bocchino (approved):
Looks great! I regenerated the component diagram and fixed a typo.

PR #1763, opened 2022-11-04
lestarch: add communication adapter interface handling to Svc::Framer

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

This adds in handling to `Svc::Framer` for the communication adapter interface.  In short it forwards statuses through such that the communication adapter isn't disconnected.

This change comes with several fixes:

1. Statuses are forwarded through the `Svc::Framer`
2. In the case of no `send` no buffer is sent to the communication adapter interface and thus no status would be triggered. In this case, `Svc::Framer` emits success to trigger itself the next packet.

> Note theses changes check for port connection, so without connecting these ports, `Svc::Framer` behavior doesn't change.


LeStarch:
@bocchino and @timcanham I formatted the code and added the newly required `this->` too.  In order to make it easier to review, I did the separate commit for the formatting change.  Thus you can review commit-by-commit and skip the formatting commit (which is all automated).

Svc/Framer/docs/sdd.md
```diff
@@ -20,6 +20,12 @@ in each frame; typically it is a frame header, a data packet, and a hash value.
 You can use the standard F Prime downlink protocol implementation.
 This implementation works with the F Prime ground data system (GDS).
 
+`Svc::Framer` is designed to act alongside instances of the
+[communication adapter interface](https://nasa.github.io/fprime/Design/communications-adapter-interface.html) and thus
+forwards statuses from any downstream communication adapter. In order to not violate this interface protocol,
+`Svc::Framer` will emit a success status anytime a buffer is received but no framed packet is sent (e.g. in the case of
```
bocchino, line 26: I think this should say "any time a buffer is received, regardless of whether a corresponding framed packet is sent". The current wording implies that a status is sent if and only if a buffer is received and no framed packet is sent.

bocchino: We may also want to say that we always return success here because any downstream failures are handled downstream.

I'm also a little confused as to why the communication adapter protocol requires us to emit a status if the status is always success. Maybe it's to preserve the option of hooking up the ComQueue to something that could return a failure status? We may also want to say that.

Svc/Framer/docs/sdd.md
```diff
@@ -20,6 +20,12 @@ in each frame; typically it is a frame header, a data packet, and a hash value.
 You can use the standard F Prime downlink protocol implementation.
 This implementation works with the F Prime ground data system (GDS).
 
+`Svc::Framer` is designed to act alongside instances of the
+[communication adapter interface](https://nasa.github.io/fprime/Design/communications-adapter-interface.html) and thus
```
bocchino, line 24: I see a couple of issues with this explanation:

1. It refers to the communication adapter interface page, but the diagram there has the Communication Adapter sending status to the Com Queue. It looks like the intent of the new design is to have the Communication Adapter send status to the Framer, which forwards the status to the Com Queue. We should update the diagram to match the new design.
2. It says that the Framer must not violate the communication adapter protocol. However, the protocol applies to communication adapters, and Framer is not a communication adapter. I think we need to revise the protocol. For example, we could rename it to Communication Adapter Status protocol and say that it must be followed by any Communication Adapter or any component that sends data to and forwards status from a Communication Adapter.

Svc/Framer/docs/sdd.md
```diff
@@ -28,31 +34,33 @@ This implementation works with the F Prime ground data system (GDS).
...
+| SVC-FRAMER-002   | `Svc::Framer` shall accept file packets stored in `Fw::Buffer` objects.                                                                | `Svc::FileDownlink` emits packets as `Fw::Buffer` objects.                                                                                                              | Unit test           |
+| SVC-FRAMER-003   | `Svc::Framer` shall use an instance of `Svc::FramingProtocol`, supplied when the component is instantiated, to wrap packets in frames. | The purpose of `Svc::Framer` is to frame data packets. Using the `Svc::FramingProtocol` interface allows the same Framer component to operate with different protocols. | Unit test           |
+| SVC-FRAMER-004   | `Svc::Framer` shall emit a status of `Fw::Success::SUCCESS`  when no framed packets were sent in response to incoming buffer.          | When `Svc::Framer` does not emit any packets it must not violate the Communication Adapter interface protocol.                                                          | Unit Test           |
```
bocchino, line 42: I recommend updating the protocol as stated above and amending this requirement to state that the Framer must obey the protocol.

bocchino (commented):
Implementation and unit test look good. I noted some issues with the design description.

Svc/Framer/docs/sdd.md
bocchino: For item 2, maybe it's better to add a new Framer protocol governing the way the framer sends packets and sends back status. I think the Framer protocol itself is nontrivial and should be described. For example:

1. It seems the intent is that if n packets are aggregated into 1 frame, Framer sends back n - 1 success status messages, then sends out a frame, and ComAdapter sends back a status message which is forwarded by Framer, for a total of n status messages. This procedure should be spelled out (if it is correct).
2. What if Framer receives one packet and emits n frames? Do n status messages go back in this case? Is ComQueue equipped to handle n status messages in response to a single packet?

bocchino: Or in item 2 above, should Framer wait for n status messages from ComAdapter and forward one message back to ComQueue? This behavior would be more symmetrical.

LeStarch:
@bocchino I did the three things we agreed:
1. Added an assert to ensure framer only emits 0 or 1 frames
2. Noted this constraint in the framing protocol documentation
3. Added description of other protocols that work with the communication adapter interface and the protocol there.

Ready for rereview.

docs/Design/communication-adapter-interface.md
```diff
@@ -43,21 +46,51 @@ This port carries a status of `Fw::Success::SUCCESS` or `Fw::Success::FAILURE` t
...
 ## Communication Adapter Protocol
 
-The `comStatus` port must obey a specific protocol to indicate to F´ applications the status of outgoing transmissions.
-A communication status is one of two possibilities:
+The communication adapter must obey a specific protocol to indicate to F´ application components the status of outgoing
+transmissions. This is done with the `comStatus` port. A communication status is one of two possibilities:
 
 | Status               | Description                                                                       |
 |----------------------|-----------------------------------------------------------------------------------|
 | Fw::Success::SUCCESS | *Communication adapter* transmission succeeded*  and is ready for more data.      |
```
bocchino, line 85: I think there is a stray asterisk after "succeeded."

bocchino (commented):
Looks great, just one comment.

docs/Design/communication-adapter-interface.md
LeStarch: Fixed!

bocchino (approved):
Looks good!

PR #1768, opened 2022-11-10
Update/internal to cmake settings

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Adds processing for basic `settings.ini` to CMake. CMake will now load the `settings.ini` file and process it into CMake variables as-if they were passed into CMake via `fprime-util`. This approach has several advantages:

1. CMake builds are independent.  They do not require CLI arguments to be setup correctly to build standard F´ projects.
2. CMake has more insight into changing variables that drive the project.

As a by-produce of these changes, the variable passing to "prescan" has been updated to ensure that all variables pass in correctly.

## Rationale

These changes are important for several reasons:

1. Plain `cmake` and IDE builds work without a slew of poorly-understood CLI variables
2. `settings.ini` variables changes can be detected within CMake
3. (Future) `fprime-util` need not pass in a default set of CLI arguments simplifying the interchange
4. (Future) Regenerating CMake build cache due to changes in `settings.ini` could be avoided entirely.


LeStarch:
@thomas-bc some things to consider:

1. the _CLI_ variables exist to detect changes to _CLI_ variables and warn about them. This is used to detect if F´ util changed something.  Perhaps we do not need these checks because:
  1. `fprime-util` checks for changes to `settings.ini` itself and fails before calling CMake
  2. Users are not allowed to touch the "required" command line variables.
  3. Therefore these variables should not be able to change without detection before reaching CMake and thus the CMake checks should never flag.
  4. Removing these checks would allow plain CMake and IDE builds to change CLI args without producing endless warnings.
2. Should we add a "skip ini loading" flag for prescan, since we pass in all settings anyway?

I don't think any of these are critical items.

cmake/settings/ini.cmake
```diff
@@ -0,0 +1,113 @@
...
+    "FPRIME_LIBRARY_LOCATIONS"
+    "FPRIME_PROJECT_ROOT"
+    "FPRIME_ENVIRONMENT_FILE"
+    "FPRIME_AC_CONSTANTS_FILE"
+    "FPRIME_CONFIG_DIR"
+    "FPRIME_INSTALL_DEST"
+)
+
+####
+# init_to_cache:
```
thomas-bc, line 31: Typo here that may be worth fixing if this generates docs somewhere.

cmake/prescan.cmake
```diff
@@ -12,22 +12,43 @@
...
-    FPRIME_PROJECT_ROOT
-    FPRIME_FRAMEWORK_PATH
-    CMAKE_TOOLCHAIN_FILE
-    CMAKE_BUILD_TYPE
-    CMAKE_DEBUG_OUTPUT
-    FPRIME_USE_STUBBED_DRIVERS
-    FPRIME_USE_BAREMETAL_SCHEDULER
-    FPP_TOOLS_PATH
-    BUILD_TESTING
+set(EXCLUDED_CACHE_VARIABLES
```
thomas-bc, line 15: Out of curiosity, how did you select which variables had to be excluded from the cache?

thomas-bc (commented):
The CMake logic looks good to me. Although I'm not super familiar with the `fprime-util` source just yet to fully understand the implications of the interaction with the __CLI_ variables.

I would suggest changing the note (below) in [fprime/docs/UsersGuide/user/settings.md](https://github.com/nasa/fprime/blob/devel/docs/UsersGuide/user/settings.md#settingsini-build-settings-configuration) to reflect the changes introduced here.
>Note: settings.ini only adjusts the build when running through fprime-util.

cmake/prescan.cmake
LeStarch: Ideally I wanted to pass everything.  However, CMake emitted warnings about unneeded/unused variables. I added those to the list along with the ones we explicitly set in the call below.

cmake/settings/ini.cmake
LeStarch: Fixed.

thomas-bc (approved):

PR #1775, opened 2022-11-17
adding config directory to build

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Two minor CMake/FPP integration clean-up items:

1.  Makes `config` directory a proper FPP model.
2. Removes FPP `includes ...` as drivers of MODULE_DEPENDENCIES

## Rationale

1. `config` must be an FPP target as it contains FPP files
2. includes force targets to exist that may not exist


LeStarch:
There is a file that requests modification in this scenario.

LeStarch:
@thomas-bc this is another CMake change.  It moves config from a special module, to a normally built module.  It also fixes an issue where including an FPP from another module added a dependency on the build of that module.  This was unnecessary as includes are "injected" in and are not directly autocoded.

Would you review?

PR #914, opened 2021-08-12
Fix autodocs github action

Joshua-Anderson:
| | |
|:---|:---|
|**_Originating Project/Creator_**| |
|**_Affected Component_**|  |
|**_Affected Architectures(s)_**|  |
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|  |
|**_Builds Without Errors (y/n)_**|  |
|**_Unit Tests Pass (y/n)_**|  |
|**_Documentation Included (y/n)_**|  |

---
## Change Description

Attempts to fix errors executing the autodocs workflow.

PR #1786, opened 2022-11-29
auto-generate connect ports option. Fixes #934

### LINKED ISSUE #934: Autogenerate separate `Tester::connectPorts` source file
| | |
|:---|:---|
|**_F´ Version_**|2.0 |
|**_Affected Component_**|  |
---
## Feature Description

Currently, `Tester::connectPorts()` is generated by the `generate impl --ut` and placed in the `Tester.cpp` file. This becomes a maintenance annoyance and causes asserts if it's not corrected if a new port is added.

The proposal would be to generate this function into a separate .cpp file that would be treated like the other unit test files and generated each time the test is compiled. 

## Rationale

It makes maintaining unit tests a lot easier, and it isn't intuitive to know you have to go update this function when you add a port.

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**| Fixes #934 |

---
## Change Description

Auto-generates `connectComponents` and `initComponents` helper functions into a `TesterHelpers.cpp` file. To use, a user sets the variable `set(UT_AUTO_HELPERS ON)`.

It also moves the UT `#define` settings to `Tester.hpp` so it is available to  this new helper file, while being available for users to set.


LeStarch:
@thomas-bc this fixes #914 in a minimally invasive way. Would you review the code/cmake fixes?  We can then discuss the approach with the team.

LeStarch:
Make the `#define` into proper constants.

Ref/SignalGen/test/ut/Tester.hpp
```diff
@@ -21,6 +21,8 @@ namespace Ref {
   class Tester :
     public SignalGenGTestBase
   {
+      // Maximum size of histories storing events, telemetry, and port outputs
+      static const NATIVE_INT_TYPE MAX_HISTORY_SIZE = 10;
```
thomas-bc, line 25: Since `TEST_INSTANCE_ID` and `TEST_INSTANCE_QUEUE_DEPTH` are not `#ifndef` in TesterHelper.cpp anymore (with your latest commit), we probably should define them here just like if it was generated - no?

thomas-bc, line 25: Since `TEST_INSTANCE_ID` and `TEST_INSTANCE_QUEUE_DEPTH` are not `#ifndef` in TesterHelper.cpp anymore (with your latest commit), we probably should define them here just like if it was generated - no? Otherwise it won't compile out of the box.

LeStarch: yes.  Not sure why the compile worked for me.....

LeStarch: Fixed!

thomas-bc (approved):
Looks good to me!

PR #1787, opened 2022-11-30
Fixes #1725

### LINKED ISSUE #1725: Unittest Autocoder Fails on Namespace
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

`TestMain.cpp` is created with a stub unit test that hardcodes the namespace `Ref` rather than detecting it from the model.

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Corrects the cheetah template to fix #1725.

### DIFF
--- a/Autocoders/Python/src/fprime_ac/generators/templates/test/test_main.tmpl
+++ b/Autocoders/Python/src/fprime_ac/generators/templates/test/test_main.tmpl
@@ -6,7 +6,11 @@
 
 #for $test_name, $method_name in $test_cases
 TEST(Nominal, ${test_name}) {
-    Ref::Tester tester;
+    #if $namespace_list != None
+    #echo "::".join($namespace_list)#::Tester tester;
+    #else
+    Tester tester;
+    #end if
     tester.${method_name}();
 }
 #end for


LeStarch:
New output in Svc::Framer:

```
    Svc::Tester tester;
```

Autocoders/Python/src/fprime_ac/generators/templates/test/test_main.tmpl
```diff
@@ -6,7 +6,11 @@
 
 #for $test_name, $method_name in $test_cases
 TEST(Nominal, ${test_name}) {
-    Ref::Tester tester;
+    #if $namespace_list != None
+    #echo "::".join($namespace_list)#::Tester tester;
+    #else
+    Tester tester;
```
thomas-bc, line 12: Are there legitimate cases where the namespace_list could be empty/none? Not that the handling should be different, just wondering.

thomas-bc (approved):
I also tested with the Imu in SystemReference, looks good!

Autocoders/Python/src/fprime_ac/generators/templates/test/test_main.tmpl
LeStarch: I don't believe it is possible, but it is correct to have it implemented in the case I am mistaken.

PR #1788, opened 2022-11-30
fixing cppcheck detected issues

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes the issues detected by CPP check (see #1782 ).  Note the following categories of checks are ignored as they are false-positives:

- Uninitialized Variables of the form U8 buffer[SOME_SIZE]:  it is inefficient to zero a buffer that will just be overwritten
- To many format parameters: it isn't picking up the PRIu32 style format codes
- No assigned in =: we delegate to a `set` method and it cannot follow
- Missing copy constructor/=:  Singleton objects are not copied.
 
We could consider adding `=delete` constructors...to prevent that last one.

### DIFF
--- a/Drv/Ip/SocketReadTask.cpp
+++ b/Drv/Ip/SocketReadTask.cpp
@@ -19,7 +19,7 @@
 
 namespace Drv {
 
-SocketReadTask::SocketReadTask() : m_stop(false) {}
+SocketReadTask::SocketReadTask() : m_reconnect(false), m_stop(false) {}
 
 SocketReadTask::~SocketReadTask() {}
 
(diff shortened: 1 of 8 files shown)


LeStarch:
@gcgandhi I fixed it here with a set of other minor fixes detected by an upcoming CI check.  Would you care to review?

gcgandhi:
@LeStarch looks good to me

thomas-bc (approved):
Looks good to me too - trusting your judgement on the `__u64` matter.

LeStarch:
The `__u64` is what the header uses, so it is the correct type.

PR #1794, opened 2022-12-03
Add Cpplint Static Code Analysis to CI Workflow

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**| CI |

---
## Change Description

This PR aims to add a new workflow "Cpplint Scan" that runs the [Google's Cpplint](https://github.com/cpplint/cpplint) linter on the framework c++ code base.
The results are displayed in the workflow summary in the Actions tab (md converter) and in GitHub Code Scanning Alerts (SARIF converter).

## Rationale

Add more static analysis to detect code style issues, best practice flaws or hidden bugs that other scanners would not find.
This linter is notably implemented in the workflow of the SpaceROS framework (https://github.com/ament/ament_lint/tree/rolling/ament_cmake_cpplint)

The different [filters](https://github.com/cpplint/cpplint/blob/develop/cpplint.py#L291-L360) to be activated/deactivated are to be discussed according to their relevance. I have not made any error feedback if there are any because for the moment the linter detects a lot of errors that are not necessarily critical and are more about the coding style.

## Future Work

Rework/refactor `cpplint_to_cppcheckxml.py`


ThibFrgsGmz:
@thomas-bc The unrecognized words (cppcheck, linelength, multitool, npx, xsltproc) should partly be integrated when merging https://github.com/nasa/fprime/pull/1782

ThibFrgsGmz:
@thomas-bc Don't worry if you don't want to merge. Cpplint may not be relevant, and/or the extra maintenance burden is not worth it.

LeStarch:
@ThibFrgsGmz do you have a sample output?  I am curious what kinds of things this finds.

ThibFrgsGmz:
If I run Cpplint on all source files (cpp,hpp,c,h) of the project and with all filters enabled, I get these error counters:

```
Category 'build/c++11' errors found: 1
Category 'build/header_guard' errors found: 664
Category 'build/include' errors found: 10
Category 'build/include_alpha' errors found: 352
Category 'build/include_order' errors found: 272
Category 'build/include_subdir' errors found: 148
Category 'build/include_what_you_use' errors found: 7
Category 'build/namespaces' errors found: 7
Category 'build/namespaces_headers' errors found: 3
Category 'build/storage_class' errors found: 2
Category 'legal/copyright' errors found: 337
Category 'readability/alt_tokens' errors found: 131
Category 'readability/braces' errors found: 127
Category 'readability/casting' errors found: 26
Category 'readability/check' errors found: 5
Category 'readability/multiline_comment' errors found: 3
Category 'readability/namespace' errors found: 541
Category 'readability/todo' errors found: 29
Category 'runtime/arrays' errors found: 19
Category 'runtime/explicit' errors found: 111
Category 'runtime/indentation_namespace' errors found: 418
Category 'runtime/int' errors found: 58
Category 'runtime/printf' errors found: 15
Category 'runtime/references' errors found: 600
Category 'runtime/threadsafe_fn' errors found: 3
Category 'whitespace/blank_line' errors found: 1919
Category 'whitespace/braces' errors found: 1158
Category 'whitespace/comma' errors found: 2460
Category 'whitespace/comments' errors found: 2410
Category 'whitespace/empty_loop_body' errors found: 2
Category 'whitespace/end_of_line' errors found: 263
Category 'whitespace/ending_newline' errors found: 2
Category 'whitespace/indent' errors found: 815
Category 'whitespace/line_length' errors found: 342
Category 'whitespace/newline' errors found: 108
Category 'whitespace/operators' errors found: 63
Category 'whitespace/parens' errors found: 1677
Category 'whitespace/semicolon' errors found: 19
Category 'whitespace/tab' errors found: 1022
Total errors found: 16149
```

The `build/c++11` raises a warning about C++ features that were not in the original C++11 specification (and therefore may not be well supported).

```bash
Os/test/ut/OsMutexBasicLockableTest.cpp:6:  <mutex> is an unapproved C++11 header.
```

The `build/header_guard` checks the consistency of the include guard styles.

```bash
CFDP/Checksum/Checksum.hpp:13:  #ifndef header guard has wrong style, please use: CFDP_CHECKSUM_CHECKSUM_HPP_  [build/header_guard]
```

The `build/header_alpha` checks that the header files are included in alphabetical order.

```bash
Drv/Ip/IpSocket.cpp:15:  Include "FpConfig.hpp" not in alphabetical order
```

The `runtime/printf` checks that not sprintf but snprintf is used.

```bash
Svc/ActiveTextLogger/test/ut/Tester.cpp:147:  Never use sprintf. Use snprintf instead.  [runtime/printf] 
```

The `runtime/threadsafe_fn` ensures that thread-satefy primitives are used:

```bash
Svc/BufferManager/test/ut/Tester.cpp:246:  Consider using rand_r(...) instead of rand(...) for improved thread safety.
```

The `whitespace` category ensures that we put spaces that improve readability and sting the eyes when reading when they are absent.

ThibFrgsGmz:
As I was not very excited about the tool at first, I talked this weekend with a friend who does C++ in the railway industry to see if he knew `Cpplint`.

He told me that this tool is used extensively on their projects and that it forces to keep a style consistency in the whole code base. Apparently, they would enable most of the filters.

So I went back to the tool to see if there was a better way to implement it on F´. I saw that it would be better to use a `CPPLINT.cfg` configuration file to more efficiently manage the tool's configuration.

I will modify the PR to include this file.

ThibFrgsGmz:
Here it is, I added the `Cpplint` configuration file!

I added/removed the filters according to the `CPPLINT.cfg` files I saw in some open source projects.

Currently the return of the analysis never fails the workflow. In fact, if we take into account `Cpplint` errors, the workflow will inevitably fail due to the huge number of style errors it returns.

In my opinion, we should try to activate as many rule filters as possible in order to ensure the greatest style consistency in the project.

But the project will have to adopt this policy and gradually conform to these rules.

Once the policy rules are established and the code is compliant, we could fail the workflow if parsing errors are found.

----

If you want to test the tool locally:

- Install the tool: `pip install cpplint`.
- Customize the `CPPLINT.cfg` file to your liking. To enable all rules, comment out all filters.
- Run the following command in the root of fprime: `cpplint --counting=detailed --quiet --recursive . > output.txt 2>&1`

*Note: This command redirects stdout and stderr to an output file.*

 - Read the output file comfortably with a hot chocolate in your hand.

If you want to see the result of the workflow, run the fprime CI , otherwise come and see the one of my fork: https://github.com/ThibFrgsGmz/fprime/actions/runs/3749494761

thomas-bc (changes requested):
We're starting to use clang-format to format our C/C++ code so let's deactivate formatting errors like `runtime/indentation_namespace` and `whitespace/comma` as they clutter the warning list pretty heavily right now. Let's also get rid of `build/header_guard`. Otherwise this looks good to me! Good call not making the check fail if it finds errors.

What did you mean by 
>**Future Work**
>Rework/refactor cpplint_to_cppcheckxml.py

ThibFrgsGmz:
@thomas-bc 
> Future Work
> Rework/refactor cpplint_to_cppcheckxml.py

I wrote this because I'm not a fan of this script, it's not very clean code (the main function is too long in my opinion).
I revised it a bit to use `set` to check the membership of a collection of literals.

.github/workflows/cpplint-scan.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: npx "@microsoft/sarif-multitool" convert "cpplint_cppcheck_result.xml" --tool "CppCheck" --output "cpplint_cppcheck_result.sarif" --force
+
+      - name: Convert cpplint results to Markdown & Integrate them in the workflow summary
+        run: xsltproc .github/scripts/cpplint-xml2text.xslt cpplint_cppcheck_result.xml | tee $GITHUB_STEP_SUMMARY cpplint_cppcheck_result.txt
+
+      - name: Upload SARIF file to GitHub Code Scanning Alerts
+        uses: github/codeql-action/upload-sarif@v2
+        with:
+          sarif_file: ${{ github.workspace }}/cpplint_cppcheck_result.sarif
+          checkout_path: ${{ github.workspace }}/source
```
thomas-bc, line 48: I should have caught that sooner but the SARIF upload  is currently failing, the last `/source` is probably from another repo structure and simply needs to be removed here. But I’m not sure what checkout_path is for in this context, I’ll let you find that out.

CPPLINT.cfg
```diff
@@ -0,0 +1,71 @@
...
+
+# `build` rules
+# Enable a warning about C++ features that were not in the original
+# C++11 specification (and so might not be well-supported).  In the
+# case of F´, the minimum supported platforms are potentially not
+# new enough to afford to do without this warning.
+filter=+build/c++11
+
+# F´ uses `#ifndef FOO_H` guard not the `#pragma once`.
+filter=+build/header_guard
```
thomas-bc, line 28: This `+build/header_guard` line should also be removed, some warnings still make it through.

.github/workflows/cpplint-scan.yml
ThibFrgsGmz: OK, I'll investigate next week!

ThibFrgsGmz: I removed this line, and also the same in the Cppcheck workflow.

.github/workflows/cpplint-scan.yml
```diff
@@ -0,0 +1,59 @@
...
+        run: sudo apt-get install xsltproc -y
+
+      - name: Install sarif tool
+        run: npm install @microsoft/sarif-multitool
+
+      - name: Run cpplint & export output to cppcheck format
+        run: cpplint --counting=detailed --quiet --recursive . 2>&1 | python3 .github/scripts/cpplint_to_cppcheckxml.py &> cpplint_cppcheck_result.xml
+
+      - name: Convert cpplint results to SARIF
+        run: npx "@microsoft/sarif-multitool" convert "cpplint_cppcheck_result.xml" --tool "CppCheck" --output "cpplint_cppcheck_result.sarif" --force
```
thomas-bc, line 39: `--tool` should be CppLint, not CppCheck

.github/workflows/cpplint-scan.yml
```diff
@@ -0,0 +1,59 @@
...
+
+on:
+  push:
+    branches: [master, devel]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master, devel]
+
+jobs:
+  cppcheck:
```
thomas-bc, line 11: This identifier doesn't seem to show up in the logs/UI anywhere but let's also rename it cpplint to avoid confusion

thomas-bc (changes requested):
@ThibFrgsGmz apologies for the late and split up review. I've thoroughly looked at it one last time and I have 2 very last, very minimal changes to request. I'll approve as soon as it's done.

.github/workflows/cpplint-scan.yml
ThibFrgsGmz: Agreed and done!

.github/workflows/cpplint-scan.yml
thomas-bc: Ah I see, my bad, I thought this was just a way to define the tool name in the results as it appears in the Security tab on GitHub.

thomas-bc (approved):
Looks good to me! 

@LeStarch As discussed, this new workflow is set up not to make any check fail on PRs - it will just produce alerts. 

<hr>

_Note for posterity:_  The generated alerts are stamped as if they were generated by CppCheck, even though it's really CppLint generating them. It's not a big deal, but it does restrict the capability to filter the issues separately in the GitHub Security tab. 
@ThibFrgsGmz can correct me if I'm wrong, but it seems that this is because CppLint isn't able to generate SARIF files, so we're instead converting the CppLint output to a CppCheck output style, and then using that output to generate a SARIF with [sarif-sdk]( https://github.com/microsoft/sarif-sdk/blob/main/docs/multitool-usage.md#supported-converters). And sarif-sdk thinks it's a CppCheck report, so it stamps it that way. A way to hack around this would be to change the tool name in the SARIF file right before uploading it (e.g. I was able to do that [here](https://github.com/thomas-bc/fprime/commit/53b95282c92ca058190326060691f8b3e614f562#diff-65ad7340341313c10a0f6510f825cfbe07cc0da537e3fd001afa9d74f21c653b)), but this might be brittle and I wouldn't recommend implementing.

ThibFrgsGmz:
@thomas-bc Exactly, I personally like your solution - and I wouldn't have done it any differently!

I vote "yes" to incorporating your workaround.

LeStarch:
I say ship the work around.  It affects CI so we should know if it breaks.

ThibFrgsGmz:
@thomas-bc Do you want to push it directly onto my branch or do you want me to incorporate your modification?

thomas-bc (approved):
Ok I pushed the commit to expedite the process, thank you @ThibFrgsGmz !

@LeStarch ready to merge

LeStarch:
@thomas-bc we likely should develop a plan on how to handle all these reports and data.  We collect a lot, and likely need to wrangle it lest it all becomes white noise.

LeStarch:
Ignoring RPI-CI failure, known fault.

PR #1803, opened 2022-12-07
Revise Linux platform types

### LINKED ISSUE #1791: Svc::ComQueue fails to compile
| | |
|:---|:---|
|**_F´ Version_**|547197194acff1f99d86e6e441fa9b5b5bd2d60b |
|**_Affected Component_**|  `Svc::ComQueue`  |
---
## Problem Description

When compiling for bare metal, I get this error

```
~/myproject/fprime/Svc/ComQueue/ComQueue.cpp: In member function 'void Svc::ComQueue::configure(Svc::ComQueue::QueueConfigurationTable, NATIVE_UINT_TYPE, Fw::MemAllocator&)':
~/myproject/fprime/Svc/ComQueue/ComQueue.cpp:94:76: error: cannot bind non-const lvalue reference of type 'NATIVE_UINT_TYPE&' {aka 'short unsigned int&'} to an rvalue of type 'NATIVE_UINT_TYPE' {aka 'short unsigned int'}
   94 |     this->m_allocation = this->m_allocator->allocate(this->m_allocationId, totalAllocation, recoverable);
      |                                                                            ^~~~~~~~~~~~~~~
In file included from ~/myproject/fprime/Svc/ComQueue/ComQueue.hpp:14,
                 from ~/myproject/fprime/Svc/ComQueue/ComQueue.cpp:8:
~/myproject/fprime/Fw/Types/MemAllocator.hpp:57:39: note:   initializing argument 2 of 'virtual void* Fw::MemAllocator::allocate(NATIVE_UINT_TYPE, NATIVE_UINT_TYPE&, bool&)'
   57 |                     NATIVE_UINT_TYPE &size,
      |                     ~~~~~~~~~~~~~~~~~~^~~~
make[2]: *** [F-Prime/Svc/ComQueue/CMakeFiles/Svc_ComQueue.dir/build.make:132: F-Prime/Svc/ComQueue/CMakeFiles/Svc_ComQueue.dir/ComQueue.cpp.obj] Error 1
```

## How to Reproduce

1. Compile with a bare metal compiler
2.
3.

## Expected Behavior

I can compile for bare metal.

bocchino:
* Simplify the code; similar to #1800
* Add an explicit error for the "should not happen" case that `__SIZEOF_POINTER__` is not in { 8, 4, 2, 1 }. Before the code was falling through to 1 without checking that 1 was correct.
* Incidentally fix a regression in the test autocoder. See comment below.
* Incidentally fix #1791

### DIFF
--- a/Autocoders/Python/src/fprime_ac/generators/templates/test/cpp.tmpl
+++ b/Autocoders/Python/src/fprime_ac/generators/templates/test/cpp.tmpl
@@ -1114,7 +1114,7 @@ $emit_cpp_params([ $param_const_timeTag, $param_val ])
 
     fprintf(
         file,
-        "EVENT: (%" PRI_FwEventIdType ") (%" PRI_FwTimeBaseStoreType ":%" PRIu32 ",%" PRIu32 ") %s: %s\n",
+        "EVENT: (%" PRI_FwEventIdType ") (%d:%" PRIu32 ",%" PRIu32 ") %s: %s\n",
         e.id,
         const_cast<TextLogEntry&>(e).timeTag.getTimeBase(),
         const_cast<TextLogEntry&>(e).timeTag.getSeconds(),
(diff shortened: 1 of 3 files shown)


bocchino:
I also fixed a regression in the test autocoder that somehow slipped through CI before. TimeBaseStoreType and TimeBase are not the same. So we can't use the TimeBaseStoreType format specifier to print TimeBase values.

bocchino:
Since TimeBase is an enum, I think %d is correct. We could also cast the argument to TimeBaseStoreType and use the TimeBaseStoreType format specifier.

bocchino:
> We could also cast the argument to TimeBaseStoreType and use the TimeBaseStoreType format specifier.

On second thought, this is better; I changed it. C/C++ style %d gives me a headache. I think it should be avoided as much as possible in portable code.

LeStarch:
Thanks for fixing the regression.  It seems to fail randomly...and I was just looking into it.

PR #1814, opened 2022-12-14
Remove comSplitter component

LMPS97:
| | |
|:---|:---|
|**_Originating Project/Creator_**|Luca Soares|
|**_Affected Component_**|comSplitter|
|**_Affected Architectures(s)_**| |
|**_Related Issue(s)_**| |
|**_Has Unit Tests (y/n)_**|n|
|**_Unit Tests Pass (y/n)_**|y|
|**_Documentation Included (y/n)_**|y|

---
## Change Description

Remove redundant component comSplitter from F Prime.

## Rationale

The comSplitter component was added before the genericRepeater component was created. With the genericRepeater component, the comSplitter component becomes redundant and can be removed.

Insure that comSplitter component is not required anymore.


LeStarch:
A project looked into it, and GenericRepeater didn't work as a clean com buffer splitter....but com splitter does.  Thus if we are to remove it, we need to make sure ComSplitter's full functionality works before removing it.

LeStarch:
We decided that the GenericRepeater is the component that should be removed.  Thus closing this one and awaiting the next.

PR #1816, opened 2022-12-19
Remove LGTM badges from README.md

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**| README.md |
|**_Affected Architectures(s)_**| N/A |

---
## Change Description

Removes LGTM badges from the project homepage. Currently the badget shows "lgtm invalid".

## Rationale

See:
- https://github.com/nasa/fprime/pull/1739
- https://github.com/nasa/fprime/pull/1744
- https://github.com/nasa/fprime/pull/1745

I spotted some instructions for LGTM in the code base like :

https://github.com/nasa/fprime/blob/91c578ee37c01ba92e556dbca0287d82f9f18b95/Fw/Types/Serializable.cpp#L60
https://github.com/nasa/fprime/blob/91c578ee37c01ba92e556dbca0287d82f9f18b95/Fw/Types/StringType.cpp#L105
https://github.com/nasa/fprime/blob/91c578ee37c01ba92e556dbca0287d82f9f18b95/Svc/FileDownlink/FileDownlink.cpp#L152
https://github.com/nasa/fprime/blob/91c578ee37c01ba92e556dbca0287d82f9f18b95/Svc/FileDownlink/FileDownlink.cpp#L153

I don't know if we should delete them because they would be deprecated, i.e. I don't know if CodeQL takes them.

## Future Work

None

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,5 @@
 # F´: A Flight-Proven, Multi-Platform, Open-Source Flight Software Framework
 
-[![Language grade: C++](https://img.shields.io/lgtm/grade/cpp/g/nasa/fprime.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/nasa/fprime/context:cpp)
-[![Language grade: Python](https://img.shields.io/lgtm/grade/python/g/nasa/fprime.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/nasa/fprime/context:python)
-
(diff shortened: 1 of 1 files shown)


LeStarch (approved):

thomas-bc:
Inline alert suppression using `//lgtm` or similar is not **currently** supported by GitHub Code Scanning. However the comment will show when the alert displays the code -- so it's still meaningful to the reviewer. Also it may become a feature in future versions of GitHub Code Scanning, or any other tool we might adopt in the future. So I would argue that keeping the comments in is a good choice.

PR #1822, opened 2022-12-20
Remove Generic Repeater Component from Svc

LMPS97:
| | |
|:---|:---|
|**_Originating Project/Creator_**| Luca Soares |
|**_Affected Component_**| Generic Repeater |
|**_Affected Architectures(s)_**| Devel |
|**_Related Issue(s)_**|  |

---
## Change Description

Removal of the generic repeater component from Svc.

## Rationale

The generic repeater component passes a reference of the serialize data to its output ports. This means that when the first output port gets read, the serialize data gets internally marked as read. This causes the other ports to have no data. Thus, any component that uses the other ports will receive no data.

Insure that there is no generic repeater code left and that nothing was broken with its removal.

## Future Work

A mechanism should be added to the F' infrastructure to make it possible to have a generic repeater module in the future.


LeStarch (approved):

LeStarch:
To extrapolate a bit: the GenericRepeater has too many caveats to exist in core.  It cannot copy/clone memory as that would require:
1. Access to dynamic memory
2. Knowledge of the generic data is clone-able (no internal pointers)

It also cannot easily be used as a copy-free repeater because:
1. Downstream users may alter the uncloned memory effecting other users.

It also depends on the "port serialization flag" to be turned on.

Given all these usage constraints, lack of customers, and difficulty of expressing how to use this component effectively, we have recommended that it be pulled from core.

PR #1833, opened 2023-01-05
F' documentation latest framework label

jefferyjohn:
## Change Description

Copied v1.5 documentation notice of later version availability to v2.0 and v3.0.0 series.

## Rationale

This change prevents multiple documentation entries from being labeled as the 'latest'.


LeStarch (approved):
This is a good fix!

PR #1834, opened 2023-01-05
Update check-spelling in documentation

jefferyjohn:
## Change Description

Clear pending warnings from check-spelling GitHub CI/CD actions.

## Rationale

This will allow for further documentation to be added while clearing CI/CD checks.

Additions to expect.txt list: COMPARGS, DPLATFORM, HDR, magicdraw

Modification of /mod.mk: Handcode -> Handcoded

## Future Work

Further spelling corrections will be made through this branch as needed


LeStarch:
This PR asks an interesting question: do we want to retroactively fix spelling issues in old documentation.  My instinct is that we shouldn't as it would cause a divergence from the tagged branch and documentation about that branch and I am not sure we'd want that....but I would love to hear from the community in this regard.

We can open up a discussion if we don't get much feedback here.

LeStarch:
The F´ team discussed this and it is changing historical documentation.  The decision was that the code this was generated from has the spelling errors and thus so should the versioned documentation to ensure there is consistency between what is published and what that published version is describing.

We appreciate your motivation to help, especially cleaning up wit the other documentation PR!

PR #1836, opened 2023-01-07
Refactor `Os/Linux/SystemResources.cpp`

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**| SystemResources |
|**_Related Issue(s)_**| None  |
|**_Unit Tests Pass (y/n)_**| N/A|

---
## Change Description

This PR aims to refactor `Os/Linux/SystemResources` to improve the understanding of the logic. It is close to the Clean Code rules with smaller functions responsible for one thing and keeping a relatively homogeneous level of abstraction.

In `SystemResources::getMemUtil`, the return of the `sysinfo(&memory_info)` call is now checked.

I move the data buffer needed to store a file line into the .bss section. This way, we don't need to allocate it recurrently on the stack.
If we use `Linux/SystemResources`, we will always need this buffer to retrieve information about the processor, so we might as well put it in the BSS.

## Rationale

See above.

Best error handling of `sysinfo` call and best memory usage of large buffer.

No regression observed when running `fprime-gds`.

## Future Work

Errors must be handled much more finely. Errors are either ERR or OK, it is not possible to know where the error occurred.

### DIFF
--- a/Os/Linux/SystemResources.cpp
+++ b/Os/Linux/SystemResources.cpp
@@ -142,7 +142,7 @@ namespace Os {
         struct sysinfo memory_info;
         int sysinfo_result = 0;
 
-        if (sysinfo_result = sysinfo(&memory_info) != 0) {
+        if ((sysinfo_result = sysinfo(&memory_info)) != 0) {
             return SYSTEM_RESOURCES_ERROR;
         }
 
--- a/Os/Linux/SystemResources.cpp
+++ b/Os/Linux/SystemResources.cpp
@@ -140,9 +140,8 @@ namespace Os {
 
     SystemResources::SystemResourcesStatus SystemResources::getMemUtil(MemUtil& memory_util) {
         struct sysinfo memory_info;
-        int sysinfo_result = 0;
 
-        if ((sysinfo_result = sysinfo(&memory_info)) != 0) {
+        if (sysinfo(&memory_info) != 0) {
             return SYSTEM_RESOURCES_ERROR;
         }
 
--- a/Os/Linux/SystemResources.cpp
+++ b/Os/Linux/SystemResources.cpp
@@ -122,9 +122,10 @@ namespace Os {
         const FwSizeType memory_unit = static_cast<FwSizeType>(memory_info.mem_unit);
 
         // Check for casting and type errors
-        if (((total_ram <= 0) || (static_cast<unsigned long>(total_ram) != memory_info.totalram)) ||
-            ((free_ram <= 0) || (static_cast<unsigned long>(free_ram) != memory_info.freeram)) ||
-            ((memory_unit <= 0) || (static_cast<unsigned int>(memory_unit) != memory_info.mem_unit))) {
+        if ((total_ram <= 0) || (free_ram < 0) || (memory_unit <= 0) ||
+            (static_cast<unsigned long>(total_ram) != memory_info.totalram) ||
+            (static_cast<unsigned long>(free_ram) != memory_info.freeram) ||
+            (static_cast<unsigned int>(memory_unit) != memory_info.mem_unit)) {
             return SYSTEM_RESOURCES_ERROR;
         }
 
--- a/Os/Linux/SystemResources.cpp
+++ b/Os/Linux/SystemResources.cpp
@@ -82,8 +82,10 @@ namespace Os {
     SystemResources::SystemResourcesStatus SystemResources::getMemUtil(MemUtil &memory_util) {
 
         struct sysinfo memory_info;
-        sysinfo(&memory_info);
-        
+
+        if (sysinfo(&memory_info) != 0) {
+            return SYSTEM_RESOURCES_ERROR;
+        }       
 
         const FwSizeType total_ram = static_cast<FwSizeType>(memory_info.totalram);
         const FwSizeType free_ram = static_cast<FwSizeType>(memory_info.freeram);
--- a/Os/Linux/SystemResources.cpp
+++ b/Os/Linux/SystemResources.cpp
@@ -19,6 +19,8 @@
 #define PROC_STAT_PATH "/proc/stat"
 #define READ_ONLY "r"
 #define LINE_SIZE 256
+std::array<char, LINE_SIZE> proc_stat_line;
+
 namespace Os {
 
     SystemResources::SystemResourcesStatus SystemResources::getCpuCount(U32 &cpuCount) {
@@ -27,7 +29,6 @@ namespace Os {
     }
 
     SystemResources::SystemResourcesStatus SystemResources::getCpuTicks(CpuTicks &cpu_ticks, U32 cpu_index) {
-        std::array<char, LINE_SIZE> line = {0};
         FILE *fp = nullptr;
         std::array<U32, 4> cpu_data = {0};
         U32 cpuCount = 0;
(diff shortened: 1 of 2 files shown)


LeStarch (changes requested):
Generally looks good, would like some comments in the CPU stuff as I struggle to follow and we need to remove the dependence on STL's `std::array`.

Definitely an improvement.

ThibFrgsGmz:
@LeStarch I remember that it was not allowed to use templates and STL but I confess I don't understand the reason.

You allow the use of STL with `std::numeric_limits`, so I don't understand why not other types like arrays at least. 

If I look at the C++ conferences, I see that the embedded C++ conferences do not hesitate to use STL templates extensively and even strongly recommend them :

* [C++ in the World of Embedded Systems - Vladimir Vishnevskii - CppCon 2022](https://www.youtube.com/watch?v=HFgq6EtVbDM&t=2812s&ab_channel=CppCon)

* [Modern C++: C++ Patterns to Make Embedded Programming More Productive - Steve Bush - CppCon 2022](https://www.youtube.com/watch?v=6pXhQ28FVlU&t=1261s&ab_channel=CppCon). Particularly interesting because the speaker does compile-time driven lookup table generation.

* [Modern Software Needs Embedded Modern C++ Programming - Michael Wong - CppCon 2020](https://youtu.be/885TI3jnB7g?t=2389).

* [CppCon 2018: Michael Caisse "Modern C++ in Embedded Systems - The Saga Continues"](https://www.youtube.com/watch?v=LfRLQ7IChtg&ab_channel=CppCon)

* [Code size optimization and compilation techniques for embedded systems - Aditya Kumar - CppCon 2021](https://www.youtube.com/watch?v=JOBs3l1jAkw&ab_channel=CppCon) - advises the use of explicit template instantiations to reduce code size and compilation time.

* [Introduction to Hardware Efficiency in Cpp - Ivica Bogosavljevic - CppCon 2022](https://www.youtube.com/watch?v=Fs_T070H9C8&ab_channel=CppCon)

So please, enlighten me! 😊

Os/Linux/SystemResources.cpp
```diff
@@ -9,107 +9,159 @@
...
+    }
 
-        if (fgets(line, sizeof(line), fp) == nullptr) { //1st line.  Aggregate cpu line.
+    SystemResources::SystemResourcesStatus getCpuData(U32 cpu_index, std::array<U32, 4>& cpu_data) {
+        FILE* fp = nullptr;
+        if (openProcStatFile(fp) != SystemResources::SYSTEM_RESOURCES_OK ||
+            readProcStatLine(fp, proc_stat_line) != SystemResources::SYSTEM_RESOURCES_OK ||
+            getCpuDataLine(fp, cpu_index, proc_stat_line) != SystemResources::SYSTEM_RESOURCES_OK ||
+            parseCpuData(proc_stat_line, cpu_data) != SystemResources::SYSTEM_RESOURCES_OK) {
             fclose(fp);
```
ThibFrgsGmz, line 85: Note: I have to create a `closeProcStatFile` function to have the same level of abstraction as its antagonist `openProcStatFile`.

astroesteban:
> 

I agree that liberal use of the STL is not recommended for embedded systems. However, in the case of `std:array` it's pretty much a wrapper around a C-style array. See the [official docs](https://en.cppreference.com/w/cpp/container/array).

Given that it's an aggregate type it  can be used almost like the fundamental type. The benefits of the `std::array` are the better value semantics, has zero-overhead, and avoids the dangerous array-to-pointer decay that happens with C-style arrays.

The compiler treats it as a regular old C array:
https://godbolt.org/z/GavzEvG7a
https://godbolt.org/z/MWTxff51a

Perhaps the specific use of `std::array` can be revisited or reevaluated down the line?

LeStarch:
@timcanham can you weight in here on the use of STL and specifically `std::array`?

timcanham:
I'm still opposed to it, since it can be a slippery slope when it comes to sub-setting the std library and fitting all the dependencies into tiny MCUs. We made a highly contested exception for `std::numeric_limits` because the product was a constant that we needed. I had a feeling we'd get some "what-abouts" from it...

astroesteban:
A reasonable and insightful objection. Thanks!

Keeping the use of STL in the FPrime core to a minimum is for the best. 

Nothing stops us from using the STL all we want in our own components though 😉

timcanham:
I completely agree that user projects and components are free to use whatever they like.

ThibFrgsGmz:
@Joshua-Anderson I am afraid that we fall into the [Ratchet Effect](https://en.wikipedia.org/wiki/Ratchet_effect), where once you accept the principle of compromise on a basic rule, even as an exception, it is very, very difficult to go back. On the other hand, it is very easy to gradually expand the exceptions.

If you accept the `array` only for the case you're talking about, it opens the door to introducing it into the rest of the code. This is the case of `std::numeric_limits` with `template` and `STL`.

So either we accept the use of `arrays`, or we should clearly do without them unfortunately. 😥

In the worst case, I can propose the implementation of a `ByteArray` class which is a wrapper of a C-array, and then keep Clean Code principles.

ThibFrgsGmz:
As an alternative to `std::array` we can have a `ByteArray` class, something like:

```cpp
class ByteArray {
public:
    ByteArray(size_t size) : size_(size) {
    }

    uint8_t operator[](size_t i) const {
        assert(i < size_);
        return data_[i];
    }

    uint8_t& operator[](size_t i) {
        assert(i < size_);
        return data_[i];
    }

    size_t size() const {
        return size_;
    }

private:
    uint8_t data_[MAX_SIZE];
    size_t size_;
};
```

Here is a list of advantages and disadvantages of a `C-style array` compared to this `ByteArray` class (correct me if I'm wrong, I'm not a C++ expert so I may be missing some info):

- Advantages of a `C-style array`:

    - **Memory size**: a C-style array generally does not contain memory overhead to handle the class mechanisms, which can be an advantage for applications that need to handle large amounts of data.
    - **Performance**: C-style arrays can be faster at accessing elements because they do not require bounds checks or memory management mechanisms.
    - **Simplicity**: C-style arrays are simpler to use and understand because they do not use advanced class mechanisms.

- Disadvantages of a `C-style array`:

    - **Safety**: C-style arrays typically do not check index boundaries when accessing elements, which can lead to out-of-bounds access errors that can cause unpredictable behavior or segmentation errors.
    - **Readability**: Code that uses C-style arrays can be more difficult to understand because it uses raw array access operations rather than methods or operators.
    - **Reusability**: C-style arrays cannot be reused easily because they do not have class mechanisms to handle advanced functionality.

ThibFrgsGmz:
I know that AUTOSAR is not applicable to the project but just for info I found this in the "[Guidelines for the use of the C++14 language in critical and safety-related systems](https://www.autosar.org/fileadmin/standards/adaptive/18-03/AUTOSAR_RS_CPP14Guidelines.pdf)" p327:

"
Rule A18-1-1 (required, implementation, automated)
      C-style arrays shall not be used.

Rationale:
      A C-style array is implicitly convertible to a raw pointer and easily loses informationabout its size.
      This construct is unsafe, unmaintainable, and a source of potential errors. 
      For fixed-size, stack-allocated arrays, std::array is supposed to be used instead. 
      This type is designed to work with standard library algorithms.   
"

LeStarch (approved):

ThibFrgsGmz:
@LeStarch, the linters saw this on the code you had recently introduced:

```
/home/runner/work/fprime/fprime/Os/Linux/SystemResources.cpp:127:47: error: comparison of unsigned expression in ‘< 0’ is always false [-Werror=type-limits]
  127 |         return ((total_ram <= 0) || (free_ram < 0) || (memory_unit <= 0) ||
      |                                      ~~~~~~~~~^~~
```

What do you think? Should we remove these controls? I'm not sure why you needed to add them, since they are basically unsigned.

LeStarch:
I would leave them for now.  Currently those types can be signed if the `PlatformTypes.hpp` is configured with signed types there-in.  Likely we should add some static asserts to make sure that file is consistent.

LeStarch:
You can also see that I fixed this with a `<=0` check for free RAM.  Given that free ram of 0 is valid, we should just take the check out.

LeStarch:
Looks like there are two isses:

1. The result is assigned but not used.  We could just remove that variable, yes?
```
       if ((sysinfo_result = sysinfo(&memory_info)) != 0) {
            return SYSTEM_RESOURCES_ERROR;
        }
```
2. A few of the multiplications go from small type to larger type (FwSizeType -> U64).  Technically speaking we should remove those U64s and make the API speak in FwSizeType....but this is a larger change.

ThibFrgsGmz:
There are a lot of errors and warnings spotted, it's a shame I didn't notice it earlier, sorry!

Regarding the unsigned/signed type checking, in C++ this seems to be done through templates, too bad they are forbidden (cf https://en.cppreference.com/w/cpp/types/is_unsigned).

Regarding the errors you mentioned. Indeed, for the first one, it is better to delete the variable.

For the second one, it is much more serious, besides the criticality is high according to CodeQL. I realize that what I did is disgusting, there are implicit castings everywhere. Right now I'm a bit tired to think calmly about solving the problem.

ThibFrgsGmz:
OK, I think that's it.

For static assertions about the sign of types, in this case `FwSizeType`, where do you put them?
You should add the following lines for example:

```cpp
#include <type_traits>
static_assert(std::is_unsigned<FwSizeType>::value, "Type FwSizeType must be an unsigned type.") ;
```

LeStarch:
@ThibFrgsGmz there are several places for those checks:  FpConfig.hpp and PlatformTypes.hpp.  However, I'd like to think through that a bit more as it needs to fit well with those type designs.

LeStarch:
@ThibFrgsGmz I am going to merge this as-is and worry about those asserts in another PR.  You ready to merge this?

ThibFrgsGmz:
Yeah. You wanted to add one comment but it did not have time to look at it...sorry

PR #1852, opened 2023-01-17
FPP v1.2.0 Integration

tiffany1618:
## Change Description

This PR will integrate FPP v1.2.0, which adds support for the C++ generation of ports, with F Prime.

Unit tests for the FPP port autocoder have been added to the `FppTest` deployment.

Note: The F Prime coverage tool doesn't seem work with generated port C++ files (it either reports 0% coverage for those files or fails to pick up the generated files at all). As a temporary workaround, a few scripts have been added to the FPP port test directory, `FppTest/port/`, that will generate accurate coverage results for the generated port C++ files.

## Changes to Generated C++ Code for Ports

- Port buffer classes (for serialization) are only generated for ports without return types, since those ports cannot be connected to serial ports
- Port string classes are wrapped the namespace `[port type name]PortStrings` to avoid polluting the global namespace
- As a temporary fix for compatibility with the Python component autocoder, typedefs are generated to support string type names derived from the string parameter names, e.g. `typedef StringSize[size] [param name]String;`
- Code generated in the `invoke` function of the `OutputPort` class for ports with return types has been fixed. In the Python port autocoder, access to `m_port` is guarded by an assertion that either `m_port` or `m_serPort` is non-null when port serialization is turned on. In this case, if `m_port` is null and `m_serPort` is non-null, then accessing `m_port` would result in a seg fault.
- All instances of `FW_ASSERT(ptr)` have been replaced with `FW_ASSERT(ptr != nullptr)`, since C++ does not guarantee that `nullptr == 0`


bocchino:
Great work! I think we still need to merge one change into FPP, make a new FPP pre-release, and bump the FPP version here.

FppTest/port/build-exe
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
```
LeStarch, line 1: Is this file left over from something?  Shouldn't these items be done in the CMake system?

FppTest/port/clean
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
LeStarch, line 1: Same here.  This file likely should be deleted.

FppTest/port/gen-cov
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
```
LeStarch, line 1: Same here.

FppTest/port/gen-cov-all
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
LeStarch, line 1: And here.

FppTest/port/gen-cov-html
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
```
LeStarch, line 1: This one too.

LeStarch (changes requested):
This looks correct, however; there are a bunch of build helper scripts that likely need to be deleted.

FppTest/port/build-exe
bocchino: These scripts are to work around issues in the code coverage analysis. I just opened an issue on this: https://github.com/nasa/fprime/issues/1862.

As discussed let's leave these scripts in so we have a way to generate code coverage until #1862 is resolved. Also the scripts may help point the way to resolving #1862.

bocchino (commented):
Looks great! Before merging, we should (1) figure out why the RPI integration build is failing; (2) tag FPP at v1.2.0; and (3) update the FPP version to point to 1.2.0.

LeStarch:
We discussed my concerns and decided to leave as-is until we have port coverage functionality.  It is on me to figure out the CI crash.

LeStarch (approved):

LeStarch:
@tiffany1618 @bocchino this is a real assertion failure.   I am still trying to get to the bottom of the cause.

LeStarch:
The issue seems to be this.

An assert fails in **ByteStreamReadyPortAc.cpp**:
```c++
143 #if FW_PORT_SERIALIZATION
144     FW_ASSERT((this->m_port != nullptr) || (this->m_serPort != nullptr));  # This line is failing
```
Then later  (same file) this code fails on a hard segfault:
```c++
153       _status = this->m_serPort->invokeSerial(_buffer);
```

This behavior (specifically) the second segfault fails in the new autocoded files and does not fail on the old autocoded files.

LeStarch:
There are several fixes that need to happen:
1. LinuxUartDriver needs to be updated to add an `is_connected` check before calling `ready_out`
2. Asserts (even at start-up) need to exit the program.  For some reason execution returns from the assert in the old and new cases.
3. (Not blocking) Segfaults should be handled to at least flush the output such that we don't get empty logs
4. (Not blocking) Asserts cannot produce a segfault in the nominal case as this seems to often "mask" real segfaults happening around ASSERTS.

I'd also like to understand why the old autocode continued to run correctly.  This assert is happening in the other RPI CI setups too, but not causing the program to crash.

bocchino:
> I'd also like to understand why the old autocode continued to run correctly. 

It looks like the old autocode had a guard `if (this->m_serPort)` which is logically unnecessary, given that the only way the condition can be false is that an assertion has already failed.

bocchino:
> Asserts cannot produce a segfault in the nominal case as this seems to often "mask" real segfaults happening around ASSERTS.

I agree! Assertion failures should catch bugs in the code logic, and seg faults should happen only if there is a bug in the assertion mechanism. Assertion failures should never trigger seg faults.

bocchino:
> For some reason execution returns from the assert in the old and new cases.

Per the investigation we did, it looks like the assertion failure occurs after the assert hook is installed, but before connection of the ports used by the assert hook. The fix is to have the assert hook check for this condition and faul back to the default assertion behavior.

bocchino:
Looks like CI passed! Good work. I'll make the FPP v1.2.0 release and update the PR.

LeStarch (approved):

PR #1857, opened 2023-01-21
Add compile-time check to ensure IEEE-754 compliance of the float

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**| None |

---
## Change Description

Add compile-time check to ensure IEEE-754 compliance of the float.

## Rationale

According to [AUTOSAR C++14 rules](https://www.autosar.org/fileadmin/standards/adaptive/18-03/AUTOSAR_RS_CPP14Guidelines.pdf):

```
Rule A0-4-1 (required, infrastructure / toolchain, non-automated)
Floating-point implementation shall comply with IEEE 754 standard.

Rationale:
Floating-point arithmetic has a range of problems associated with it. Some of these can
be overcome by using an implementation that conforms to IEEE 754 (IEEE Standard
for Floating-Point Arithmetic).
Note that the rule implies that toolchain, hardware, C++ Standard Library and C++
built-in types (i.e. float, double) will provide full compliance to IEEE 754 standard in
order to use floating-points in the project.
```
See also:

```
- MISRA C++ 2008 [6]: Rule 0-4-3 Floating-point implementations shall comply
with a defined floating-point standard.
- JSF December 2005 [7]: AV Rule 146 Floating point implementations shall
comply with a defined floating point standard.
```

None

## Future Work

None


LeStarch:
It looks ok to me.  @timcanham I'd appreciate your eyes on this.

For my own edification, are there C/C++ implementations out there that are not compliant with the above?  This reads as `static_assert(C/C++ is self consistent, ...)`.

LeStarch (approved):

LeStarch:
@ThibFrgsGmz in looking at this more, @timcanham suggested we put this code behind a `#ifndef`.  That way, on the off change we run into a platform where standard limits does not support these checks, a platform engineer can set that off and remove it.

**E. g.**
```
#ifndef SKIP_FLOAT_IEEE_754_COMPLIANCE
static_assert(   (std::numeric_limits<float>::is_iec559    == true)
              && (std::numeric_limits<float>::radix        ==    2)
              && (std::numeric_limits<float>::digits       ==   24)
              && (std::numeric_limits<float>::max_exponent ==  128),
              "The 32-bit floating point type does not conform to the IEEE-754 standard.");
#endif
```

We run across many platforms with (often dubious) C++11 support and thus giving options to side-step these issues often proves useful.  Otherwise, we very much appreciate the extra level of check that this does.

As an aside, you might merge/rebase on `devel` as the RPI checks should be working again. However, your PR is still failing because git isn't recognizing the new workflow fixes in `devel`.  It doesn't matter that much because the error is understood and will self-correct upon merge.

LeStarch (changes requested):
See comment in conversation tab.  Responded on behalf of @timcanham .

ThibFrgsGmz:
@LeStarch @timcanham Indeed, you are right, it is smarter, I should have thought about it before. I add it and an explanation in the documentation.

ThibFrgsGmz:
I'm closing the PR, I messed up with the rebasing/merging...

PR #1869, opened 2023-01-26
changing SIGSEGV for SIGABRT to remove segfault without core dump

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

SIGSEGV -> SIGABRT in asserts.

## Rationale

Seg faulting in assert handling code was presumably done to ensure a core-dump file. However, multiple sources show that this is done also in the case of a SIGABRT.  We using SIGSEGV, however; real errors causing segmentation faults are masked as the assert is expected to force a false seg fault.

Use SIGABRT, it is functional without confusing.


LeStarch:
@timcanham you should approve this change.  SIGABRT is documented to cause a core dump.  Is there any other reason to use SEGV?  It makes debugging much harder.

bocchino (approved):
Looks good! I made one small change to the logger output (segfault -> abort signal).

LeStarch:
@bocchino good catch!

timcanham (approved):

PR #1870, opened 2023-01-26
Add `IEEE 754` compliance of the floating point arithmetic

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**| None |

---
## Change Description

Add compile-time check to ensure IEEE-754 compliance of the float.

> **Note** See previous PR attempt: https://github.com/nasa/fprime/pull/1857

## Rationale

According to [AUTOSAR C++14 rules](https://www.autosar.org/fileadmin/standards/adaptive/18-03/AUTOSAR_RS_CPP14Guidelines.pdf):

```
Rule A0-4-1 (required, infrastructure / toolchain, non-automated)
Floating-point implementation shall comply with IEEE 754 standard.

Rationale:
Floating-point arithmetic has a range of problems associated with it. Some of these can
be overcome by using an implementation that conforms to IEEE 754 (IEEE Standard
for Floating-Point Arithmetic).
Note that the rule implies that toolchain, hardware, C++ Standard Library and C++
built-in types (i.e. float, double) will provide full compliance to IEEE 754 standard in
order to use floating-points in the project.
```
See also:

```
- MISRA C++ 2008 [6]: Rule 0-4-3 Floating-point implementations shall comply
with a defined floating-point standard.
- JSF December 2005 [7]: AV Rule 146 Floating point implementations shall
comply with a defined floating point standard.
```

I took the opportunity to improve a little the readability of the `configure-fprime.md` file - I put the new Note/Warning visualization:

> **Note** this is a not

> **Warning** this is a warning

## Future Work

None

### DIFF
--- a/cmake/platform/types/PlatformTypes.hpp
+++ b/cmake/platform/types/PlatformTypes.hpp
@@ -38,10 +38,11 @@
 #include <cinttypes>
 #include <limits>
 
-#define FW_HAS_64_BIT    1 //!< Architecture supports 64 bit integers
-#define FW_HAS_32_BIT    1 //!< Architecture supports 32 bit integers
-#define FW_HAS_16_BIT    1 //!< Architecture supports 16 bit integers
-#define FW_HAS_F64    1 //!< Architecture supports 64 bit floating point numbers
+#define FW_HAS_64_BIT 1  //!< Architecture supports 64 bit integers
+#define FW_HAS_32_BIT 1  //!< Architecture supports 32 bit integers
+#define FW_HAS_16_BIT 1  //!< Architecture supports 16 bit integers
+#define FW_HAS_F64 1     //!< Architecture supports 64 bit floating point numbers
+#define SKIP_FLOAT_IEEE_754_COMPLIANCE 0  //!<  Check IEEE 754 compliance of floating point arithmetic
 
 // Section 1: Logical Types
 //    fprime requires platform implementors to define logical types for their
--- a/Fw/Types/BasicTypes.hpp
+++ b/Fw/Types/BasicTypes.hpp
@@ -49,9 +49,23 @@ typedef char CHAR;
   typedef uint64_t       U64; //!< 64-bit unsigned integer
 #endif
 
-typedef float   F32; //!< 32-bit floating point
+typedef float F32;  //!< 32-bit floating point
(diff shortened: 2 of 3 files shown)


cmake/platform/types/PlatformTypes.hpp
```diff
@@ -38,10 +38,11 @@
...
 
-#define FW_HAS_64_BIT    1 //!< Architecture supports 64 bit integers
-#define FW_HAS_32_BIT    1 //!< Architecture supports 32 bit integers
-#define FW_HAS_16_BIT    1 //!< Architecture supports 16 bit integers
-#define FW_HAS_F64    1 //!< Architecture supports 64 bit floating point numbers
+#define FW_HAS_64_BIT 1  //!< Architecture supports 64 bit integers
+#define FW_HAS_32_BIT 1  //!< Architecture supports 32 bit integers
+#define FW_HAS_16_BIT 1  //!< Architecture supports 16 bit integers
+#define FW_HAS_F64 1     //!< Architecture supports 64 bit floating point numbers
+#define SKIP_FLOAT_IEEE_754_COMPLIANCE 0  //!<  Check IEEE 754 compliance of floating point arithmetic
```
LeStarch, line 45: This will skip the check, as `#ifndef` checks for definitions not value of the definition.  We have two options:

1. (Easy) remove this definition.  The check will run.
2. (More clear) Changes the check in `BasicTypes.hpp` to: `#if !defined(SKIP_FLOAT_IEEE_754_COMPLIANCE) || !SKIP_FLOAT_IEEE_754_COMPLIANCE`

LeStarch: The behavior is of the easy option is:

1. Value undefined: run check
2. Value defined: skip check

The behavior of the clear option is:
1. Value undefined: run check
2. Value defined: run check when not true

In either case, we want the undefined behavior to check and only allow the check to skip when specifically turned off.  The more clear option allows users to define the field but still run the check with a false.

ThibFrgsGmz: "Clear option" done!

LeStarch: Looks good now.  Configuration allows setting, unsetting, and defaulting to the check!

LeStarch (approved):

PR #1897, opened 2023-02-17
Updates to BasicTypes.hpp

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Several changes to basic types and platform types:

1. All non GCC/Clang compilers with override flag allows users of other compilers to get by anyway
2. `BasicTypes.cpp` and `PlatformTypes.hpp` are now C compliant.

## Future Work

We really don't need to define limit structs for all our types.  We use `typedef` which compilers translate to the underlying types.  Thus the following code is perfectly valid:

```
std::numeric_limits<PlatformIntType>::min()
```

We should remove our structs and the defining documentation to ease maintainance.


LeStarch:
Decided to fix FpConfig.hpp too!

bocchino:
This seems like it will work, but the code structure is a bit hard to follow. I suggest refactoring `BasicTypes.hpp` into `BasicTypes.hpp` and `BasicTypes.h`. `BasicTypes.h` can include all and only the C stuff. `BasicTypes.hpp` can include it wrapped in `extern "C"`. Similarly for `PlatformTypes.hpp`.

This would require a small change to existing C code (include `BasicTypes.h` instead of `BasicTypes.hpp`) but I think it would be a lot clearer.

If we keep the existing structure, then in `BasicTypes.hpp` starting at line 88 we should move the `#include` directive to the top of the file (appropriately guarded) and remove the line indentation starting at line 89.

timcanham:
After some thought, I think I agree with @bocchino - we should probably break it up into a `.hpp` and .`h`.

LeStarch:
@bocchino @timcanham I refactored the setup.  We now have:

1. `PlatformTypes.h`: defining all platform named integers in C-style.
2. `BasicTypes.hpp` and `BasicTypes.h`: C++ and C includes for BasicTypes
3. `FpConfig.h`: C compliant configuration header
4. `FpConfig.hpp`: backwards compatibility header that just wraps `FpConfig.h`

Note all usages of FpConfig::<something>_MAX should now use `std::numeric_limits<type>::max()` as our own definitions were not needed.

Also note, I need to update the documentation.  This is sufficient for reviewing the code and patterns.

bocchino (commented):
Looks great! Just one comment. While we are at it, we should fix the following issue in Assert.cpp. This code

https://github.com/LeStarch/fprime/blob/e1f113cfc11e3287e2e5e4e358e2cc3528fda334/Fw/Types/Assert.cpp#L353-L356

does not conform to the JPL coding rules, because it declares an extern function that is also declared in the public header file `CAssert.h`. We should delete these lines and add
```
extern "C" {
#include "Fw/Types/CAssert.h"
}
```
to the top of `Assert.cpp`.

Also `CAssert1` is declared in the header but never defined in `Assert.cpp`. So I see a linker error waiting to happen. Maybe this issue is out of scope for this PR. If so we could open an issue on it.

LeStarch:
@bocchino a quick attempt to fix this reveals several issues:

1. Inconsistent redefined `#define` between CAssert and Assert headers
2. Missing implementations (CAssert1)
3. Code replication

I have been intending to revist `Assert.hpp` for a while, I propose we keep as-is for this PR and fix with a revisit to Assert.hpp.

bocchino:
Sounds good!

bocchino (approved):
Looks good!

PR #1905, opened 2023-02-24
Add pip check Workflow

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Adding a workflow that, on changes to `requirements.txt`, will try to install the requirement file for all supported versions of Python. 

## Rationale

This is because the latest upgrade of requirements.txt (https://github.com/nasa/fprime/commit/8787fe82518be3e1ac0afcc8f7d69bb5444b2ba9) broke the install for Python3.7 and we did not notice it until later.

### DIFF
--- a/.github/workflows/pip-check.yml
+++ b/.github/workflows/pip-check.yml
@@ -3,13 +3,13 @@
 
 name: Python Dependency Check
 
-on: [push]
-  # push:
-  #   paths:
-  #     - 'requirements.txt'
-  # pull_request:
-  #   paths:
-  #     - 'requirements.txt'
+on:
+  push:
+    paths:
+      - 'requirements.txt'
+  pull_request:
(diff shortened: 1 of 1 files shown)


thomas-bc:
The workflow I'm introducing here is not running as part of the checks of this PR because I locked it to run only on changes to `requirements.txt`. However, I ran it in my branch and it _successfully failed_ for Python 3.7 only. Here's the [workflow run](https://github.com/thomas-bc/fprime/actions/runs/4258632603), not sure if you can access it.

Let me know if you want to throw in Python 3.11 in there as well.

LeStarch (approved):

LeStarch:
Like it.  For expediency, I added 3.11

PR #1916, opened 2023-03-01
name consistency in project and topology, install correct dictionary. Fixes #1444

### LINKED ISSUE #1444: F´ FPP Module and CMake Module Should be kept In-Sync
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

It is entirely possible to define a project in CMake (e.g. Ref) that is inconsistent with the FPP module name in `topology.fpp`. We should check this and WARN when inconsistent.

Dictionary file generation must use the FPP module not the CMake module.

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes #1444 by performing the following changes:

1. Removes `dict` target, it added nothing, does nothing, and doesn't provide a build entrypoint.
2. Installing the proper dictionary regardless of the name of the project
3. Produces WARNING message if project and topology have inconsistent names.


LeStarch:
Once model tool is available switch to use that.

cmake/autocoder/deployment_validation.cmake
```diff
@@ -0,0 +1,42 @@
...
+#
+# Required function, look for .fpp files defining a topology block and ensure the following equivalence:
+#     ${PROJECT_NAME} == topology name
+####
+function(deployment_validation_setup_autocode AC_INPUT_FILES)
+    foreach(AC_FILE IN LISTS AC_INPUT_FILES)
+        file(READ "${AC_FILE}" FILE_TEXT)
+        # Does this file match the pattern (module { topology {} })
+        if (NOT FPRIME_SKIP_PROJECT_NAME_VALIDATION AND
+            FILE_TEXT MATCHES "\n?[^\n#@]* topology ([a-zA-Z0-9_]+)" AND
```
thomas-bc, line 32: Is the intent here **not** to match commented out topologies? I don't think this is what's happening since the leading newline is optional. I confirmed that adding a 
```# topology Refff {}``` somewhere in an FPP file will get caught and throw a warning.  Although, as long as it catches the actual topology, it's not that big a deal if it throws false positives.

File: cmake/autocoder/deployment_validation.cmake [hunk 2]
```diff
@@ -0,0 +1,42 @@
+####
+# autocoder/deployment_validation.cmake
+#
+# An autocoder that performs validation steps on the deployment to ensure that it is well-formed. This will produce
```
thomas-bc, line 4: It could be good to add a note to clarify that this is not actually going to generate any files, and that it's simply using the autocoder logic in order to be fed all the .fpp files. I can see a new developer look at this and wonder what's happening and why this is an autocoder.

cmake/autocoder/deployment_validation.cmake
LeStarch: Good catch!

I'd like to avoid false positives. I'll attempt resolution.

cmake/autocoder/deployment_validation.cmake
LeStarch: That is wise.  I will fix this.

cmake/autocoder/deployment_validation.cmake
```diff
@@ -0,0 +1,42 @@
...
+#     ${PROJECT_NAME} == topology name
+####
+function(deployment_validation_setup_autocode AC_INPUT_FILES)
+    foreach(AC_FILE IN LISTS AC_INPUT_FILES)
+        file(READ "${AC_FILE}" FILE_TEXT)
+        # Does this file match the pattern (module { topology {} })
+        if (NOT FPRIME_SKIP_PROJECT_NAME_VALIDATION AND
+            FILE_TEXT MATCHES "\n?[^\n#@]* topology ([a-zA-Z0-9_]+)" AND
+            NOT CMAKE_MATCH_1 STREQUAL "${PROJECT_NAME}")
+            message(WARNING
```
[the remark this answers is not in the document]
LeStarch: I'd prefer to keep this as a WARNING because it uses stderr like the FATAL_ERROR message does.  That way it will appear if a user does:

```
fprime-util generate > /dev/null
```

Thoughts?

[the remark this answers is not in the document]
thomas-bc: Sounds good to me!

thomas-bc (approved):

PR #1932, opened 2023-03-14
Fix #1931 - `microsoft/sarif-multitool` usage changes

### LINKED ISSUE #1931: CI CppCheck/CppScan failing
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

See https://github.com/nasa/fprime/actions/runs/4390265240/jobs/7688586298. CI is failing on CppLint and CppCheck scans because the `--force` option of Microsoft's SARIF-Multitool converter is not available anymore. Probably an update of the convert tool.

Fix is likely going to be to simply remove the option (or use the older version of the tool?), but we should also check what it used to be doing to be sure we're not losing any functionality.

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Fix #1931 

## Rationale

See ticket description

Investigate CI logs

## Future Work

None

### DIFF
--- a/.github/workflows/cppcheck-scan.yml
+++ b/.github/workflows/cppcheck-scan.yml
@@ -29,7 +29,7 @@ jobs:
         run: sudo apt-get install cppcheck xsltproc -y
 
       - name: Install sarif tool
-        run: npm i @microsoft/sarif-multitool@4.0.0
+        run: npm i -g @microsoft/sarif-multitool
 
       # With a CMake-based project, we get the list of files by setting up a build with CMAKE_EXPORT_COMPILE_COMMANDS=ON and
       # referencing the compile_commands.json file produced by the tool.  This will capture the correct include paths and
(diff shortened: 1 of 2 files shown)


ThibFrgsGmz:
Interestingly, the regression is even older, it seems. Either we drop a major version (3.x) or we remove the deprecated attribute completely.

I will try to find more time during the week. At the moment I did the PR very quickly without testing due to lack of time.

Or do you want to do it if you have the time @thomas-bc?

thomas-bc:
@ThibFrgsGmz yes I was tinkering with it I hope you don't mind. I'm ok dropping the `--force` altogether since we're in CI and the environment is fresh and clean every run. 
From the [usage](https://github.com/microsoft/sarif-sdk/blob/main/docs/multitool-usage.md#common-arguments):

> --force: Force overwrite of output file if it exists.


I reverted back to the default install (so the newer v4.1.0 which doesn't have --force) and removed `--force`.

thomas-bc (approved):

ThibFrgsGmz:
Awesome @thomas-bc 😁 !

PR #1950, opened 2023-03-30
Feature/no init

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

The component `init` function is now called automatically from within the topology autocoded files. This means that `init` must have a fixed signature.  In typically use cases, `init` only calls the parent's implementation of the `init` function. e.g.

```c++
void Component ::init(...) {
    ComponentComponentBase::init(...);
}
```

This means that end-users are expected to maintain this function, but not touch it.  Current recommendations for generic setup of components is to define a separate `setup()` or `config()` function and call it specifically in the topology with whatever arguments needed.

This implies that the `init` function should entirely be pushed into the autocoded layer.

This PR does several things:
1. Define the autocoded `init` as `public` such that the topology may call this function
~~2. Set `init` as virtual such that the user may safely override it, should it be necessary~~
3. Strip `init` from the component implementation templates such that, by default, the component does not override the `init` function

Item 2 scrubbed as it is could be a potentially breaking change.

### DIFF
--- a/Autocoders/Python/src/fprime_ac/generators/templates/component/hpp.tmpl
+++ b/Autocoders/Python/src/fprime_ac/generators/templates/component/hpp.tmpl
@@ -244,6 +244,7 @@ namespace ${namespace} {
         const char* compName = "" $doxygen_post_comment("Component name")
     );
 
+  public:
     //! Initialize a ${name}ComponentBase object
     //!
     void init(
@@ -259,6 +260,7 @@ namespace ${namespace} {
 #end if
     );
 
+  PROTECTED:
     //! Destroy a ${name}ComponentBase object
     //!
     virtual ~${name}ComponentBase();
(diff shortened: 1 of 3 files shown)


Autocoders/Python/src/fprime_ac/generators/templates/component/hpp.tmpl
```diff
@@ -244,6 +244,7 @@ namespace ${namespace} {
         const char* compName = "" $doxygen_post_comment("Component name")
     );
 
+  public:
     //! Initialize a ${name}ComponentBase object
     //!
     void init(
```
timcanham, line 250: Didn't you want to make this `virtual`?

timcanham (changes requested):

timcanham (approved):

thomas-bc (approved):

Issue #1847, opened 2023-01-16
Clickable Assert Messages (FSW)


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Make FW_ASSERT messages clickable (e.g. `<file>:<line>`).

## Rationale

This worked well for UT asserts.

LeStarch:
We are still discussing this, it is not ready for implementation.

LeStarch:
This has been approved!  We can implement!

SMorettini:
I would be happy to implement it :)

LeStarch:
@SMorettini did you still want to look into implementing this?

SMorettini:
Just created a pull request to implement it.

PR #2022, opened 2023-05-13
Made assert clickable (#1847)

SMorettini:
| | |
|:---|:---|
|**_Affected Architectures(s)_**| Fw/Assert and Fw/Logger |
|**_Related Issue(s)_**| #1847  |

---
## Change Description

Made assert message clickable

### DIFF
--- a/Fw/Types/Assert.cpp
+++ b/Fw/Types/Assert.cpp
@@ -12,7 +12,7 @@
 #if FW_ASSERT_LEVEL == FW_FILEID_ASSERT
 #define fileIdFs "Assert file ID 0x%08" PRIx32 ": Line: %" PRI_PlatformUIntType
 #else
-#define fileIdFs "Assert file \"%s\": Line: %" PRI_PlatformUIntType
+#define fileIdFs "Assert: \"%s:%" PRI_PlatformUIntType "\""
 #endif
 
 namespace Fw {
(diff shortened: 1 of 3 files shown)


Fw/Types/Assert.cpp
```diff
@@ -12,7 +12,7 @@
 #if FW_ASSERT_LEVEL == FW_FILEID_ASSERT
 #define fileIdFs "Assert file ID 0x%08" PRIx32 ": Line: %" PRI_PlatformUIntType
```
LeStarch, line 13: Should we change the ID asserts (used on constrained downlink, does a CRC32 ID of the filename) something like:

```
Assert: 0xABCD1234:123...
```

LeStarch (commented):
Looks good to me, but was wondering your thoughts on the 32bit file ID messages.

Fw/Types/Assert.cpp
SMorettini: I never use the file id so I don't have any strong opinion in favour or against your suggestion. I didn't change it just thinking that maybe since the assert will contain two numbers `id_file:line_number` maybe could be not clear the meaning. But if you want me to change it let me know and I'll do it.

SMorettini:
I cannot understand why the CI is failing, in my local branch everything works fine, I cannot reproduce the error in the CI.

LeStarch:
@SMorettini our hardware cluster got powered off and thus none-of-the HITL tests worked.

Fw/Types/Assert.cpp
LeStarch: I would request you add that update.  That way our assert messages remain consistent.

SMorettini: I update it as requested

LeStarch (approved):

thomas-bc (approved):

PR #2025, opened 2023-05-15
Update fprime cli documentation

SMorettini:
| | |
|:---|:---|
|**_Affected Component_**| Documentation |
|**_Related Issue(s)_**|  |

---
## Change Description

Update the documentation of the fprime-cli.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -1669,3 +1669,4 @@ XYZZY
 yacc
 yacgen
 yyyymmdd
+zmq
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -935,6 +935,7 @@ NOCTTY
 nodemon
 nogen
 nolog
+non-existent
 nonblock
 noncomma
 NONINFRINGEMENT
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -935,7 +935,6 @@ NOCTTY
 nodemon
 nogen
 nolog
-non-existent
 nonblock
 noncomma
 NONINFRINGEMENT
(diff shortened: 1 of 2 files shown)


SMorettini:
The spell-checking is failing for `non-existent`, should I add it in the exceptions or changed in the documentation? It comes from running `fprime-cli --help`

thomas-bc (commented):
If that's outputted by fprime-cli it's probably best to add it to the expected list. Thanks for updating this!

thomas-bc:
**Food for thought:**
this pattern of displaying in the docs what the output of a `--help` command outputs seems questionable to me. For example, the next release of `fprime-gds` will yet again modify this output, replacing `-r` with `-d` (see [here](https://github.com/fprime-community/fprime-gds/commit/f700190125d578d3f082beb399936a5b3a68ddab#diff-20f9d3d050b68198cff1c8cc88f0baa2cfd8f9ef1f3ba661d076134625ef1927R207)). Are we going to have to remember to update this again then? Most likely, we will forget. Should these things be scrubbed? Could we auto-generate them somehow?

SMorettini:
I updated the exception to solve the spelling issue. About 

> **Food for thought:** this pattern of displaying in the docs what the output of a `--help` command is seems questionable to me. For example, the next release of `fprime-gds` will yet again modify this output, replacing `-r` with `-d` (see [here](https://github.com/fprime-community/fprime-gds/commit/f700190125d578d3f082beb399936a5b3a68ddab#diff-20f9d3d050b68198cff1c8cc88f0baa2cfd8f9ef1f3ba661d076134625ef1927R207)). Are we going to have to remember to update this again then? Most likely, we will forget. Should these things be scrubbed? Could we auto-generate them somehow?

I agree, I think that if we can autogenerate it would be very interesting but I'm not an expert in CI setup so I don't know how to do it. I think it can also make sense to remove it from the documentation since the user can find it by itself. If you want me to remove it I can do it directly in this pull request. Let me know what you prefer.

LeStarch:
There is reason to keep this around, but I'd argue in a different format. Like how online Linux man-pages are setup.  Autogeneration would be awesome, and it should be relegated to API like documentation rather than inside the user guide.

This is a nice stop-gap, so I'll merge.

PR #2024, opened 2023-05-14
Adds telemetry packet definition step.

NoahBraasch:
| | |
|:---|:---|
|**_Originating Project/Creator_**| |
|**_Affected Component_**|  |
|**_Affected Architectures(s)_**|  |
|**_Related Issue(s)_**|  #2020 |
|**_Has Unit Tests (y/n)_**|  |
|**_Builds Without Errors (y/n)_**|  y |
|**_Unit Tests Pass (y/n)_**|  |
|**_Documentation Included (y/n)_**|  |

---
## Change Description

Modifies MathComponent tutorial to provide a fix with a current bug.

## Rationale

Provides bug workaround in tutorial.

## Testing/Review Recommendations

Verify MathTutoial looks nice.

## Future Work

Bug will have to be fixed and section removed when that happens.

PR #2032, opened 2023-05-18
History sync (#2024)

thomas-bc:
| | |
|:---|:---|

---
## Change Description

Merging master into devel to fix a PR that was merged into master... 

## Future work

Lock `master` branch to avoid such mistakes in the future. We may also consider making `devel` the default branch, since people users should not need to clone the `fprime` repo directly anymore because we now have `fprime-util new --project`

### DIFF
--- a/docs/Tutorials/MathComponent/Tutorial.md
+++ b/docs/Tutorials/MathComponent/Tutorial.md
@@ -2015,6 +2015,24 @@ instance mathReceiver
 These lines add the `mathSender` and `mathReceiver`
 instances to the topology.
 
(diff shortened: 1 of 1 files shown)


LeStarch (commented):
LGTM, already approved in master.

LeStarch (approved):

NoahBraasch:
Sorry my mistake!

LeStarch:
@NoahBraasch , mostly mine.  I forgot to check the branch before merging.  We have now locked master, so that merging is impossible unless specifically unlocked, so these errors will go away!

Thanks for the contribution!

Issue #2000, opened 2023-05-03
Add Build Repository With Latest CI Action


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Add a reusable action for Github CI that will build a repository with a parameterized version, defaulting to `devel`.  That way running CI on a specific repo is straight forward.

Steps:
- [x] Clone repo (w/ submodule update)
- [x] Update fprime version to devel
- [x] Generate, build, unit-test, and integration test

thomas-bc:
v0 can be found here: https://github.com/fprime-community/project-builder

This is used in this sample-project: https://github.com/thomas-bc/sample-fprime-project, see the [workflow files](https://github.com/thomas-bc/sample-fprime-project/tree/main/.github/workflows) and the [Actions runs](https://github.com/thomas-bc/sample-fprime-project/actions).

I created both an Action and a Reusable Workflow. My reasoning is that a Reusable Workflow is a little more easy to interact with and debug. An Action will throw **all** the output (so generate, build, check etc...) in a single step, which is a **lot** of text and harder to debug. However a Reusable Workflow can only be used as is, so if projects want to just build their project and go on with other tasks to run in the CI, they can use the Action.

I would recommend projects to start early with the Reusable Workflow, and if they want to get more involved in the CI, they can use the Action instead, and customize/run additional things.

thomas-bc:
I was going to say "how do you want to go about Integration Testing", and I just thought, I'm going to add an optional last step to the Action. 
The users should be able to optionally specify a shell script to execute, post-build. This will allow for integration testing, or any other small things they want to be running.

thomas-bc:
- [ ] remove postbuild_script 
- [ ] either run UT or build in the Action, not both
- [ ] add URL config for being able to build a foreign repository

thomas-bc:
Split into 2 objectives:
- reusable workflow specifically for F´ core CI (building external repositories): see #2049 
- reusable workflow + Action for community usage: https://github.com/fprime-community/project-builder

thomas-bc:
https://github.com/fprime-community/project-builder should be ready for community usage

PR #2049, opened 2023-05-26
Build fprime-community projects as part of CI checks

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**| GitHub Actions |
|**_Affected Architectures(s)_**| CI |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2000    🎉 🍾  |

---
## Change Description

Introduces a [Reusable Workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) (`reusable-builder.yml`) that allows to build and run tests on an external F´ project, with the current revision of F´. This allows, for example, to build and run UTs on [fprime-community/fprime-workshop-led-blinker](https://github.com/fprime-community/fprime-workshop-led-blinker) as part of a CI check on PRs within F´ core.

## Rationale

This is going to help in many ways:
- more testing, that's always good
- will prevent our examples/references from breaking
- improved traceability in changes. If a change is made that breaks an example: first of all we will know (not always obvious), and we can strongly recommend (or better, enforce) that the fprime-community project should be updated.

That could even lead to some sorts of "Migration references" - with our references evolving over time to catch up with the F´ changes - we can diff between releases to show what has changed in how you use F´.

"Let the CI fly!"
   \- LeStarch 

## Future Work

- Introduce workflows for other projects such as HelloWorld, SystemReference when it's ready, and other projects (incoming BaremetalReference? 👀 )
- Pull out `Ref/` and `RPI/` out of nasa/fprime ?

### DIFF
--- a/.github/workflows/reusable-builder.yml
+++ b/.github/workflows/reusable-builder.yml
@@ -19,11 +19,6 @@ on:
         required: false
         type: boolean
         default: true
-      run_integration_tests:
-        description: "Run an additional job in parallel to run unit tests."
-        required: false
-        type: boolean
-        default: false
       fprime_location:
         description: "Relative path from the external project root to its F´ submodule"
         required: false
@@ -59,16 +54,15 @@ jobs:
         run: |
           pip3 install -r ${{ inputs.fprime_location }}/requirements.txt
         shell: bash
-      - name: "Build Project"
+      - name: "Generate build cache"
         working-directory: ${{ inputs.build_location }}
         run: |
-          fprime-util generate ${{ inputs.target_platform }}
-          fprime-util build ${{ inputs.target_platform }}
+          fprime-util generate ${{ runner.debug == '1' && '--verbose' || '' }} ${{ inputs.target_platform }}
         shell: bash
-      - if: ${{ inputs.run_integration_tests }}
-        name: "Run Integration Tests (Pytest)"
+      - name: "Build"
         working-directory: ${{ inputs.build_location }}
-        run: pytest 
+        run: |
+          fprime-util build ${{ runner.debug == '1' && '--verbose' || '' }} ${{ inputs.target_platform }}
         shell: bash
 
   runUT:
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-builder.yml
```diff
@@ -0,0 +1,100 @@
...
+          submodules: recursive
+          path: ${{ inputs.fprime_location }}
+      - name: "Install requirements.txt"
+        run: |
+          pip3 install -r ${{ inputs.fprime_location }}/requirements.txt
+        shell: bash
+      - name: "Build Project"
+        working-directory: ${{ inputs.build_location }}
+        run: |
+          fprime-util generate ${{ inputs.target_platform }}
```
LeStarch, line 65: Should we break these steps out for greater clarity? e.g.

Generate build cache

Build Project

.github/workflows/reusable-builder.yml
```diff
@@ -0,0 +1,100 @@
...
+          fprime-util build ${{ inputs.target_platform }}
+        shell: bash
+      - if: ${{ inputs.run_integration_tests }}
```
LeStarch, line 68: I don't think this will work, as you have yo have a running GDS and binary at this point....but that does ask the question if we should wrap that all in a single invocation.

LeStarch (commented):
This is amazing!  A few comments/questions

.github/workflows/reusable-builder.yml
thomas-bc: Good point, I'll split that up

.github/workflows/reusable-builder.yml
thomas-bc: Removing for the time being

thomas-bc:
Enabled verbose with a Debug Run: https://github.com/nasa/fprime/actions/runs/5095135148/jobs/9159854919
I also confirmed that the overlaying of the F´ revision is happening as expected by tweaking with my fprime fork and witnessing that the CI build would fail because of it.

LeStarch (approved):

LeStarch:
@thomas-bc ready to merge?

thomas-bc:
@LeStarch ready to merge!

PR #2076, opened 2023-06-08
Disable some CI checks on docs updates

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Affected Architectures(s)_**| CI |
|**_Related Issue(s)_**|  |

---
## Change description

Disable the build/test/static code analysis checks on PRs that only touch `docs/**` and `**.md` files.

## Rationale

We often get PRs on documentation, which trigger the entire CI check suite. I propose we disable some of those checks on these occasions, which are computationally intensive and quite irrelevant. They also slow down the review process for PRs that could be merged in a matter of minutes.

### **A case _against_ this PR:**
Sometimes checks will fail on documentation PRs for reasons that are unrelated (e.g. the RPI agents are unavailable, something broke upstream/fprime-tools etc...).
The case could be made that we should know as soon as possible when this happens, and therefore we should not disable those checks. Which I don't disagree with tbh. This is up for debate, totally fine if we don't want to accept this change.

### DIFF
--- a/.github/workflows/python-format.yml
+++ b/.github/workflows/python-format.yml
@@ -6,9 +6,6 @@ on:
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [master, devel]
-    paths-ignore:
-      - 'docs/**'
-      - '**.md'
 
 jobs:
   format:
--- a/.github/workflows/python-format.yml
+++ b/.github/workflows/python-format.yml
@@ -1,6 +1,15 @@
 name: Format Python
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [master, devel]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master, devel]
+    paths-ignore:
+      - 'docs/**'
+      - '**.md'
+
 jobs:
   format:
       name: Format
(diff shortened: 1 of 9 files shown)


.github/workflows/python-format.yml
```diff
@@ -1,6 +1,15 @@
...
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [master, devel]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master, devel]
+    paths-ignore:
+      - 'docs/**'
```
LeStarch, line 10: I might not ignore this particular action.  I think there is python code under `docs/**` and if we don't catch formatting errors here....then we might have a failure manifest somewhere completely different once merged.

LeStarch (changes requested):
I *love* this!  

Another case against this PR could be "what if someone adds a check (e.g. broken link scan) to an existing action.  It might not get triggered".

I don't but that argument nor the one you presented.  I think, however; we should *always* run python and spelling checks....see inline comment.

.github/workflows/python-format.yml
thomas-bc: Fixed

LeStarch (approved):

thomas-bc:
Agreed. If we get to modifying workflows down the line, the restriction can always be reversed. Workflows should be tightly scoped though, adding URL checks within a build workflow would -in my humble opinion- be a questionable design.

PR #2078, opened 2023-06-10
Adding telemetry and event automatic processing for GenericHub

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

GenericHub's difficult in deployment comes from primarily from the specific wrangling of standard types across the hub.  This PR adds in dedicated telemetry and event processing  such that the hub can be used with FPP pattern specifiers when connecting components to it.

### DIFF
--- a/Svc/GenericHub/GenericHub.fpp
+++ b/Svc/GenericHub/GenericHub.fpp
@@ -4,6 +4,22 @@ module Svc {
   passive component GenericHub {
 
     # ----------------------------------------------------------------------
+    # Mimic Ports
+    # ----------------------------------------------------------------------
+
+    @ Telemetry input port for mimicking an event processor
+    guarded input port TlmRecv: Fw.Tlm
+
+    @ Event input port for mimicking an event processor
+    guarded input port LogRecv: Fw.Log
+
+    @ Cross-hub event output port
+    output port LogSend:  Fw.Log
+
+    @ Cross-hub telemetry output port
+    output port TlmSend: Fw.Tlm
+
+    # ----------------------------------------------------------------------
     # General Ports
     # ----------------------------------------------------------------------
 
--- a/Svc/GenericHub/GenericHub.fpp
+++ b/Svc/GenericHub/GenericHub.fpp
@@ -8,10 +8,10 @@ module Svc {
     # ----------------------------------------------------------------------
 
     @ Telemetry input port for mimicking an event processor
-    guarded input port TlmRecv: Fw.Tlm
+    sync input port TlmRecv: Fw.Tlm
 
     @ Event input port for mimicking an event processor
-    guarded input port LogRecv: Fw.Log
+    sync input port LogRecv: Fw.Log
 
     @ Cross-hub event output port
     output port LogSend:  Fw.Log
@@ -24,13 +24,13 @@ module Svc {
     # ----------------------------------------------------------------------
 
     @ Input array of generic ports to shuttle to the other side of the hub connection
-    guarded input port portIn: [GenericHubInputPorts] serial
+    sync input port portIn: [GenericHubInputPorts] serial
 
     @ Output array of generic ports shuttled from the other side of the hub connection
     output port portOut: [GenericHubOutputPorts] serial
 
     @ Input array of generic ports shuttling in copy-free buffers form external sources
-    guarded input port buffersIn: [GenericHubInputBuffers] Fw.BufferSend
+    sync input port buffersIn: [GenericHubInputBuffers] Fw.BufferSend
 
     @ Output array of generic ports shuttling in copy-free buffers form external sources
     output port buffersOut: [GenericHubOutputBuffers] Fw.BufferSend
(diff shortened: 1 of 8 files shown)


Svc/GenericHub/GenericHub.fpp
```diff
@@ -3,6 +3,22 @@ module Svc {
   @ A generic hub component
   passive component GenericHub {
 
+    # ----------------------------------------------------------------------
+    # Mimic Ports
+    # ----------------------------------------------------------------------
+
+    @ Telemetry input port for mimicking an event processor
+    guarded input port TlmRecv: Fw.Tlm
```
timcanham, line 11: I guess I never looked at any of these closely - why are any of the input ports guarded? There's no private data being shared.

Svc/GenericHub/GenericHubComponentImpl.cpp
```diff
@@ -96,13 +94,85 @@ void GenericHubComponentImpl ::dataIn_handler(const NATIVE_INT_TYPE portNum,
...
         buffersOut_out(port, fwBuffer);
+    } else if (type == HUB_TYPE_EVENT) {
+        FwEventIdType id;
+        Fw::Time timeTag;
+        Fw::LogSeverity severity;
+        Fw::LogBuffer args;
+
+        // Deserialize tokens for events
+        status = incoming.deserialize(id);
+        FW_ASSERT(status == Fw::FW_SERIALIZE_OK, static_cast<NATIVE_INT_TYPE>(status));
```
timcanham, line 111: Are you assuming that the upstream device reliably decodes the data, and that it will never pass you a partial packet? If no, then you should assert, as that makes the software assert on a potentially unreliable communication channel.

Svc/GenericHub/GenericHubComponentImpl.cpp
```diff
@@ -96,13 +94,85 @@ void GenericHubComponentImpl ::dataIn_handler(const NATIVE_INT_TYPE portNum,
...
 
+void GenericHubComponentImpl ::LogRecv_handler(const NATIVE_INT_TYPE portNum,
+                                  FwEventIdType id,
+                                  Fw::Time& timeTag,
+                                  const Fw::LogSeverity& severity,
+                                  Fw::LogBuffer& args) {
+    U8 buffer[sizeof(FwEventIdType) + Fw::Time::SERIALIZED_SIZE + Fw::LogSeverity::SERIALIZED_SIZE + FW_LOG_BUFFER_MAX_SIZE];
+    Fw::ExternalSerializeBuffer serializer(buffer, sizeof(buffer));
+    serializer.resetSer();
+    serializer.serialize(id);
```
timcanham, line 153: Why aren't you checking status?

timcanham, line 153: Also, to be symmetric with the decode function, why aren't you serializing the port type (`HUB_TYPE_EVENT`,etc)

timcanham: Status in this case should be asserted since you should have allocated enough room on the stack.

Svc/GenericHub/GenericHubComponentImpl.cpp
```diff
@@ -96,13 +94,85 @@ void GenericHubComponentImpl ::dataIn_handler(const NATIVE_INT_TYPE portNum,
...
+}
+
+void GenericHubComponentImpl ::TlmRecv_handler(const NATIVE_INT_TYPE portNum,
+                                  FwChanIdType id,
+                                  Fw::Time& timeTag,
+                                  Fw::TlmBuffer& val) {
+    U8 buffer[sizeof(FwChanIdType) + Fw::Time::SERIALIZED_SIZE + FW_TLM_BUFFER_MAX_SIZE];
+    Fw::ExternalSerializeBuffer serializer(buffer, sizeof(buffer));
+    serializer.resetSer();
+    serializer.serialize(id);
```
timcanham, line 169: Same comments as events.

Svc/GenericHub/GenericHubComponentImpl.cpp
LeStarch: Yes, I am making that assumption.  GenericHub is designed to sit on top of framer and deframer and those two pieces ensure proper packet transmission.

Svc/GenericHub/GenericHubComponentImpl.cpp
LeStarch: That I will do!

LeStarch: I am, it it done in the `send_data` helper.  This level function is just for wrangling the specific data into a buffer that function could serialize.

Svc/GenericHub/GenericHub.fpp
LeStarch: Made them `sync` as there is no member data.  I think these were originally `guarded` to force an order between accesses and the outgoing data, but that isn't a good argument.

timcanham (approved):

thomas-bc (approved):

PR #2095, opened 2023-06-21
Migrated HelloWorld tutorial to F' community

hAkselS:
This PR removes the hello world tutorial from NASA/fprime repo and references the hello world on fprime-community in all NASA/fprime READMEs


thomas-bc (commented):
Something I realize just now is that by doing this (removing the entire folder), the https://nasa.github.io/fprime/Tutorials/HelloWorld/Tutorial.html link will die, however that's the link that's in `master`'s README and there's not much we can do about this until next release.

hAkselS:
Hmm yeah I see the problem there, but won' the old tutorial live there
until the next release too?

thomas-bc:
The docs are built out of `devel`, so I expect the page would die immediately if we remove the folder from there.

LeStarch (approved):
The tutorial rework looks good to me.  @thomas-bc ready for merge?  If we break the README as shown on github.com, perhaps we can change that link in a dedicated PR?  or we can move the "main branch" to `devel` since we have a tool that pulls the correct version down.

Also, we should probably rework that readme too....

thomas-bc (approved):

PR #2097, opened 2023-06-23
Add SPI Mode Enum to LinuxSpiDriver

mohitsingh999:
SPI Modes specify the clock polarity for each transaction. The default Fprime LinuxSpiDriver assumed SPI Mode 0. This should be generalized to account for SPI devices with different clock polarity and phases.

| | |
|:---|:---|
|**_Originating Project/Creator_**|SRH/Mohit Singh |
|**_Affected Component_**| Drv::LinuxSpiDriver |
|**_Affected Architectures(s)_**| SRH MPPT Tester |
|**_Related Issue(s)_**|  |

---
## Change Description
Add a SPI Mode Enumeration to the LinuxSpiDriver Component Implementation, and include the SpiMode as a formal parameter in the `LinuxSpiDriverComponentImpl::open()` function . 

## Rationale
This allows for SPI devices with a different clock phase and polarity to be properly integrated with F-Prime, without having to modify the component implementation. Previously, the SPI Mode for any SPI device was assumed to be SPI mode 0. This implies a clock phase=0 and a clock polarity=0, which is not always true. 

The implementation has been tested with SPI Mode 0 and SPI Mode 1 devices and works correctly. 

## Future Work
In addition to SPI Modes, the Linux SPI implementations include several additional parameters, such as `SPI_CS_HIGH`, `SPI_LSB_FIRST`, and the bits per word for each SPI Transaction. These parameters generally do not change for most SPI devices, but it might make sense to give the user ability to configure these parameters when opening a SPI device in FPrime.

### DIFF
--- a/Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
+++ b/Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
@@ -100,6 +100,10 @@ namespace Drv {
         /*
          * SPI Mode 0, 1, 2, 3
          */
+
+        //Assert that the device SPI Mode is in the correct range
+        FW_ASSERT(spiMode >= SpiMode::SPI_MODE_0 || spiMode <= SpiMode::SPI_MODE_3);
+
         U8 mode = spiMode; // Mode Select (CPOL = 0/1, CPHA = 0/1)
         ret = ioctl(fd, SPI_IOC_WR_MODE, &mode);
         if (ret == -1) {
--- a/Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
+++ b/Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
@@ -115,7 +115,7 @@ namespace Drv {
                 mode = SPI_MODE_3;
             default:
                 //Assert if the device SPI Mode is not in the correct range
-                FW_ASSERT(spiMode >= SpiMode::SPI_MODE_0 || spiMode <= SpiMode::SPI_MODE_3);                
+                FW_ASSERT(0, spiMode);                
                 break;
         }
 
--- a/Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
+++ b/Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
@@ -70,7 +70,8 @@ namespace Drv {
 
     bool LinuxSpiDriverComponentImpl::open(NATIVE_INT_TYPE device,
                                            NATIVE_INT_TYPE select,
-                                           SpiFrequency clock) {
+                                           SpiFrequency clock,
+                                           SpiMode spiMode) {
 
         this->m_device = device;
         this->m_select = select;
@@ -97,9 +98,9 @@ namespace Drv {
 
         // Configure:
         /*
-         * SPI Mode 0
+         * SPI Mode 0, 1, 2, 3
          */
-        U8 mode = SPI_MODE_0; // Mode 0 (CPOL = 0, CPHA = 0)
+        U8 mode = spiMode; // Mode Select (CPOL = 0/1, CPHA = 0/1)
         ret = ioctl(fd, SPI_IOC_WR_MODE, &mode);
         if (ret == -1) {
             DEBUG_PRINT("ioctl SPI_IOC_WR_MODE fd %d failed. %d\n",fd,errno);
--- a/Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
+++ b/Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
@@ -101,10 +101,24 @@ namespace Drv {
          * SPI Mode 0, 1, 2, 3
          */
 
-        //Assert that the device SPI Mode is in the correct range
-        FW_ASSERT(spiMode >= SpiMode::SPI_MODE_0 || spiMode <= SpiMode::SPI_MODE_3);
+        U8 mode; // Mode Select (CPOL = 0/1, CPHA = 0/1)
+        switch(spiMode) {
+            case SpiMode::SPI_MODE_0:
+                mode = SPI_MODE_0;
+                break;
+            case SpiMode::SPI_MODE_1:
+                mode = SPI_MODE_1;
+                break;
+            case SpiMode::SPI_MODE_2:
+                mode = SPI_MODE_2;
+            case SpiMode::SPI_MODE_3:
+                mode = SPI_MODE_3;
+            default:
+                //Assert if the device SPI Mode is not in the correct range
+                FW_ASSERT(spiMode >= SpiMode::SPI_MODE_0 || spiMode <= SpiMode::SPI_MODE_3);                
+                break;
+        }
 
-        U8 mode = spiMode; // Mode Select (CPOL = 0/1, CPHA = 0/1)
         ret = ioctl(fd, SPI_IOC_WR_MODE, &mode);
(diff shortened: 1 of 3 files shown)


Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.hpp
```diff
@@ -59,7 +77,8 @@ namespace Drv {
             //! Open device
             bool open(NATIVE_INT_TYPE device,
                       NATIVE_INT_TYPE select,
-                      SpiFrequency clock);
+                      SpiFrequency clock,
+                      SpiMode spiMode);
```
LeStarch, line 81: Should we add a default parameter here such that old code still works?

`spiMode = SPI_MODE_0`

LeStarch (approved):

LeStarch (changes requested):
We should add the default to that parameter to preserve old code.

Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
```diff
@@ -70,7 +70,8 @@ namespace Drv {
 
     bool LinuxSpiDriverComponentImpl::open(NATIVE_INT_TYPE device,
                                            NATIVE_INT_TYPE select,
-                                           SpiFrequency clock) {
+                                           SpiFrequency clock,
+                                           SpiMode spiMode) {
```
LeStarch, line 75: We should assert that spiMode is in the correct range:

```
FW_ASSERT(spiMode >= SPI_MODE_0 || spiMode <= SPI_MODE_3);
```

LeStarch (changes requested):
Two changes, assert mode is in-range and add a default.

LeStarch:
@timcanham want to look at this too?

timcanham:
@LeStarch Glad to look at it after @mohitsingh999 makes the changes.

mohitsingh999:
@LeStarch @timcanham I just made the changes to 1) Have a default SPI Mode parameter value and 2) Assert that the SPI Device mode is in range. Let me know if anything else needs fixing!

LeStarch (approved):

LeStarch:
@timcanham I'll leave you to do a final approval and run.

Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
```diff
@@ -97,9 +98,9 @@
 
         // Configure:
         /*
-         * SPI Mode 0
+         * SPI Mode 0, 1, 2, 3
          */
-        U8 mode = SPI_MODE_0; // Mode 0 (CPOL = 0, CPHA = 0)
+        U8 mode = spiMode; // Mode Select (CPOL = 0/1, CPHA = 0/1)
```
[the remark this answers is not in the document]
timcanham: For small enums, it's better IMHO to use a switch statement to cover the values (that pass) and a default clause with the assert. Thoughts @LeStarch?

timcanham (changes requested):

Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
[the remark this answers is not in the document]
LeStarch: Switch to a switch statement e.g. :
```
switch (spiMode) {
    case SpiMode::SPI_MODE_0:
         mode = SPI_MODE_0;
         break;
...
    default:
        FW_ASSERT(...)
       break;
}
ret = ioctl(fd, SPI_IOC_WR_MODE, &mode);

```

Does get around the hard-coded values in the enum definition.  It would also protect us from changes to the kernel #defs, although I don't see how that would happen.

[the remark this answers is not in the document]
LeStarch: I am ok with making it a switch statement.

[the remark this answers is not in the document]
mohitsingh999: @timcanham @LeStarch I pushed a commit which utilizes a switch statement.

Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
```diff
@@ -97,9 +98,27 @@ namespace Drv {
...
+            case SpiMode::SPI_MODE_1:
+                mode = SPI_MODE_1;
+                break;
+            case SpiMode::SPI_MODE_2:
+                mode = SPI_MODE_2;
+            case SpiMode::SPI_MODE_3:
+                mode = SPI_MODE_3;
+            default:
+                //Assert if the device SPI Mode is not in the correct range
+                FW_ASSERT(spiMode >= SpiMode::SPI_MODE_0 || spiMode <= SpiMode::SPI_MODE_3);
```
LeStarch, line 118: You should just assert zero in this case.

```
FW_ASSERT(0, spiMode);
```

This will force-fail the assert and print the invalid mode.

LeStarch (commented):
One minor nit.

Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
mohitsingh999: @LeStarch Made the fix

Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
```diff
@@ -97,9 +98,27 @@ namespace Drv {
...
+        U8 mode; // Mode Select (CPOL = 0/1, CPHA = 0/1)
+        switch(spiMode) {
+            case SpiMode::SPI_MODE_0:
+                mode = SPI_MODE_0;
+                break;
+            case SpiMode::SPI_MODE_1:
+                mode = SPI_MODE_1;
+                break;
+            case SpiMode::SPI_MODE_2:
+                mode = SPI_MODE_2;
```
timcanham, line 113: Just noticed! Missing `break` statements.

timcanham (changes requested):

Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp
mohitsingh999: Thanks! Just added the missing statements

Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.hpp
```diff
@@ -33,6 +33,30 @@ namespace Drv {
...
+     * SPI Mode Select
+     *
+     * Defines the SPI Clock Polarity and Phase for each SPI Transaction.
+     * 
+     * SPI Clock Polarity(CPOL): Defines clock polarity as idle low (CPOL = 0) or idle high(CPOL = 1)
+     * SPI Clock Phase(CPHA): Defines if data is shifted out on the rising clock edge and sampled 
+     *                        on the falling clock edge(CPHA = 0) or if data is shifted out on the 
+     *                        falling clock edge and sampled on the rising clock edge(CPHA=1)
+     * 
+     * SPI Mode 0: (CPOL = 0, CPHA = 0)
```
timcanham, line 46: I would take these SPI comments out and annotate the `SpiMode` enum with a description of what each means, since the "SPI Mode X" values are now hidden.

timcanham (changes requested):

timcanham (approved):

LeStarch (approved):

LeStarch (approved):

PR #2103, opened 2023-06-26
FPP v1.3.0 integration

tiffany1618:
## Change Description

This PR will integrate FPP v1.3.0, which adds support for the C++ generation of component base and implementation classes, with F Prime.

Unit tests for the FPP component autocoder have been added to the FppTest deployment. Unit tests for the FPP port autocoder have been combined with the new component tests, since the port tests are a subset of the component tests.

### Component base classes

- Fix issue where the implementations for callback functions for command ports are not generated when a component contains command ports but no commands
- Fix return type for handler, handler base, and callback functions for ports that return a string type to be the string type defined in the corresponding generated port file instead of string
- Reorganize class members
    - All constants appear before function declarations in the header file
    - All member variables appear after function declarations in the header file
    - Member functions are first organized by access specifier in the following order:
        - `public`
        - `PROTECTED`
        - `PRIVATE`
    - Within each access specifier, member functions are organized by category in the following order:
        - Component functions (constructor, destructor, initialization, etc.)
        - Input/output port functions
        - Internal port functions
        - Command functions
        - Time get function
        - Event functions
        - Telemetry functions
        - Parameter functions
- Add an assertion to check the port number in handler base functions for serial ports

Example:
https://github.com/fprime-community/fpp/blob/22dc4207987e0f30ac8c75a6c91fd9fd14ccb3e7/compiler/tools/fpp-to-cpp/test/component/ActiveSerialComponentAc.ref.cpp#L2849-L2853 

- Add static assertion for components containing serial ports to make sure FW_PORT_SERIALIZATION is on

Example:
https://github.com/fprime-community/fpp/blob/22dc4207987e0f30ac8c75a6c91fd9fd14ccb3e7/compiler/tools/fpp-to-cpp/test/component/ActiveSerialComponentAc.ref.hpp#L39-L42

- Remove connector functions for serial input to typed output port connections, where the typed output port has a return type, since ports of these types should not be connected  
- Remove private parameter getter function, `getParam()`, which does not appear to be used by either the component base class itself or any of the F Prime framework classes

### Component implementation templates

- Add handler implementations for internal interfaces
- Remove `init` function
- Change file extension of template files from `.{hpp,cpp}-template` to `.template.{hpp,cpp}`
- Fix formatting of Doxygen comments

## Todo before this is ready to merge

- [x] Add component implementation class generation to FPP
- [x] Create FPP pre-release to test PR
- [ ] Create FPP release v1.3.0
- [ ] Update fprime-fpp version


LeStarch:
@tiffany1618 @bocchino how much work is the component implementations? I heard a suggestion of a point release, however; it will increase the CMake integration effort as we'll still need AiXML files generated....but specifically to disable AI XML autocoding.

We should also track somewhere the CMake integration for this work.

Lastly, this is sooooo fantastic!

tiffany1618:
Thank you! I have the code generation part of component implementations done (it was simpler than expected), so all that is left is to integrate it with the fprime tools, which might require some CMake work.

LeStarch:
Fantastic!  Please clue us in when you need anything from our side (e.g. CMake).

bocchino:
This is great! Thanks, Tiffany.

> it will increase the CMake integration effort as we'll still need AiXML files generated....but specifically to disable AI XML autocoding.

This is already working for C++ component base classes. When you run the build, FPP generates the component XML and the component C++. CMake detects that the component C++ is there and compiles it.

As Tiffany said, we still need to integrate the generation of C++ component implementation skeletons into the F Prime tools.

tiffany1618:
@LeStarch @bocchino Could this be approved so CI can run? I think everything has been completed for this PR on the FPP side.

LeStarch (approved):

gallegosdamian505 (commented):
*_

PR #2136, opened 2023-07-17
Fix bug in Fw::Buffer

bocchino:
The copy constructor for `Fw::Buffer` was causing an assertion failure when copying an invalid buffer. I noticed this because I was writing unit tests that passed invalid buffers around, and I saw the assertion failure.

This patch fixes the issue.

Note: For correctness, I had to replace `FW_ASSERT(p)` with `FW_ASSERT(p != nullptr)` in one place. Otherwise we are assuming that `nullptr == 0` which is not guaranteed to be true. We should make this change consistently throughout the framework, but I made the minimal change for now.

### DIFF
--- a/Fw/Buffer/Buffer.cpp
+++ b/Fw/Buffer/Buffer.cpp
@@ -28,11 +28,15 @@ Buffer::Buffer(): Serializable(),
 {}
 
 Buffer::Buffer(const Buffer& src) : Serializable(),
-    m_serialize_repr(src.m_bufferData, src.m_size),
+    m_serialize_repr(),
     m_bufferData(src.m_bufferData),
     m_size(src.m_size),
     m_context(src.m_context)
-{}
+{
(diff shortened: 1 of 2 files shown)


timcanham:
I guess this means that the new buffer is now invalid?

bocchino:
I think so. I think a buffer should be considered invalid if the pointer is `nullptr` or the size is zero.

Note the SDD implies that if `pointer == nullptr` then `size` should be zero. We could add an assertion for that.

bocchino:
But then it should be an invariant of `Fw::Buffer` that's checked everywhere, not just in this one place.

bocchino:
For this PR I made the minimal change to avoid an assertion failure that I thought shouldn't be happening.

LeStarch:
This change seems correct to me.

LeStarch (approved):

PR #2137, opened 2023-07-18
Update STest

bocchino:
Update STest to the latest.

* Fix static analysis warnings.
* Fix a corner-case bug where `Pick::lowerUpper(lower, upper)` could return `upper+1` in rare cases.
* Other minor fixes.


LeStarch:
Shouldn't we just update to using the STest package as opposed to copying in updated files?

bocchino:
Yes, that would be an improvement.

bocchino:
But it requires changes to the F Prime build infrastructure. This PR is a minimal change to get the latest version.

LeStarch:
Agreed.  However, we make exactly the same argument each time we update files in this area.

bocchino:
Is there an issue open to make STest a subrepo? If not, let's open an issue and get it in the priority queue.

PR #2150, opened 2023-07-22
Update/framing deframing context

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

This pull-request makes several changes to `Framer` and `Deframer`. 

1. Framer now accepts buffers w/ context.  This allows for frame protocols that need to metadata to assist in framing. For example: an APID may be passed-in alongside the data.
2. Framer file buffer type adjustments are handled through context rather than side-band data
3. Deframer may now create context
4. Deframer routing functionality has been moved to a new component as only projects using F´  protocol require this routing functionality.

Work:
- [x] Framer changes
- [x] File buffer refactor
- [x] Deframer changes
- [x] Protocol changes
- [x] FprimeRouter implementation
- [x] Topology update (Ref)
- [ ] Topology update (RPI)
- [x] Existing UTs
- [ ] Router UTs
- [ ] Framer documentation update
- [ ] Deframer documentation  update
- [ ] New router documentation
- [ ] Update "new deployment" and tutorials

## Rationale

Many projects need additional contextual metadata when framing, but wish to leverage existing Framer and Deframer code. This PR allows this.  Many projects also handle routing and don't need routing code commingled with Deframer.


LeStarch:
@LMPS97 @arizvi786 here are the changes you were interested in for Framer/Deframer.  You can leverage this work immediately as:

- [x] Documentation is not required for you to leverage the implementation
- [x] UTs pass for the components that concern you
- [x] FprimeRouter is not needed in your project.

LeStarch:
@pcrosemurgy this PR breaks out routing from Deframer. It is possible that you could use these updates as you were specifically disrupted by the combined deframing and routing.

pcrosemurgy:
> @pcrosemurgy this PR breaks out routing from Deframer. It is possible that you could use these updates as you were specifically disrupted by the combined deframing and routing.

Thank you! I'll take a closer look at this!

LMPS97:
I am getting an error compiling `FprimeProtocol.cpp`. Please add `#include <limits.h>` to it.

Svc/Framer/Framer.hpp
```diff
@@ -39,11 +39,6 @@ class Framer : public FramerComponentBase, public FramingProtocolInterface {
     Framer(const char* const compName /*!< The component name*/
     );
 
-    //! Initialize object Framer
-    //!
-    void init(const NATIVE_INT_TYPE instance = 0 /*!< The instance number*/
```
LMPS97, line 44: Why was the init function deleted on `framer`? Also, why only on `framer` and not `deframer`?

LMPS97:
I ran into a slight issue with the `Framer`. Sometimes, I need to send a package out with an empty data buffer. This is because the information is carried out by the header which is created from the context buffer instead. Because of that, the empty buffer is deallocated to the buffer manager which throws an EVR  that the buffer size is 0. A solution would be to add a check on the `Framer` before deallocating the buffer to see if it is a valid buffer in the first place.

LMPS97:
I just ran into an other issue when unit testing my code. Apparently, if you try to create a `Fw::Buffer` from a non valid  `Fw::Buffer` using the following constructor causes an assert:
```
Buffer::Buffer(const Buffer& src) : Serializable(),
    m_serialize_repr(src.m_bufferData, src.m_size),
    m_bufferData(src.m_bufferData),
    m_size(src.m_size),
    m_context(src.m_context)
{}
```
I encountered this issue when a function of mine returns a non-valid `Fw::Buffer`.

Svc/FramingProtocol/FprimeProtocol.cpp
```diff
@@ -19,11 +19,11 @@ FprimeFraming::FprimeFraming(): FramingProtocol() {}
...
-void FprimeFraming::frame(const U8* const data, const U32 size, Fw::ComPacket::ComPacketType packet_type) {
-    FW_ASSERT(data != nullptr);
+void FprimeFraming::frame(const Fw::Buffer& data, const Fw::Buffer& context) {
+    FW_ASSERT(data.getData() != nullptr);
     FW_ASSERT(m_interface != nullptr);
     // Use of I32 size is explicit as ComPacketType will be specifically serialized as an I32
-    FpFrameHeader::TokenType real_data_size = size + ((packet_type != Fw::ComPacket::FW_PACKET_UNKNOWN) ? sizeof(I32) : 0);
+    FpFrameHeader::TokenType real_data_size = context.getSize() + data.getSize();
     FpFrameHeader::TokenType total = real_data_size + FpFrameHeader::SIZE + HASH_DIGEST_LENGTH;
     Fw::Buffer buffer = m_interface->allocate(total);
```
LMPS97, line 28: There should be a check here to see if the buffer is empty since the buffer manager can run out of buffers. I don't think the check should be an assert since it wouldn't be appropriate to assert flight software on a lack of buffer manager buffers. Therefore, I think it should do an early return with an error code. This error code could be use by the framer and an evr could be outputted.

Svc/FramingProtocol/FprimeProtocol.cpp
```diff
@@ -128,7 +128,8 @@ DeframingProtocol::DeframingStatus FprimeDeframing::deframe(Types::CircularBuffe
     buffer.setSize(size);
```
LMPS97, line 128: Similar to the `frame` function, this check should be a assert but a early return with an error code.

Svc/Framer/Framer.hpp
LeStarch: `init` functions are provided by the component base class and are, in older component implementations, improperly shadowed.  Thus, we are removing them.

Why not deframer?  Oversight.  We will remove it there too before merging.

LeStarch:
> is deallocated to the buffer manager which throws an EVR that the buffer size is 0. A solution would be to add a check on the `Framer` before deallocating the buffer to see if it is a valid buffer in the first place.

We should deallocate it when the buffer `isValid()` and not dealocate otherwise.  While the "zero length" EVR is annoying, it would be worse to never return a buffer to buffer manager even when it is zero length.  Make sure the "i don''t use this buffers" are invalid and I agree, we should guard the deallocation.

Svc/FramingProtocol/FprimeProtocol.cpp
LeStarch: Agreed wholeheartedly.  Although we should check for `.isValid()` in the latest F´  because when buffer manager has no buffers, an invalid buffer is returned.

Svc/FramingProtocol/FprimeProtocol.cpp
LeStarch: Agreed!

LeStarch:
I am going to close this PR.  We are working on breaking apart deframer and as such this PR will continue to code-rot.

Issue #2135, opened 2023-07-17
Add offset option for rate groups allowing different trigger patterns


csmith608:
| | |
|:---|:---|
|**_F´ Version_**| baremetal |
|**_Affected Component_**| RateGroupDriver  |
---
## Feature Description

Addition of an offset option when specifying rate group dividers. The offsets would be used to stagger RG firings, this is useful when a thread scheduler is not present

## Rationale

This was used on the LFPS firmware and cut our rate group times almost in half preventing major rate group slipping. The image below shows an example of how the current rate group would fire off vs how the proposed change could be used to fire them off.
This is an example of how the rate groups fired before and after adding the offsets. 
![image](https://github.com/nasa/fprime/assets/63731123/1557ace3-9b48-4e20-b3de-5b25f325947c)

It was implemented [here](https://github.com/nasa/fprime/pull/818) for fprime 3.0.0, but should have used a struct for the rate group divider and offset. However, the logic changes are there and can be used pretty much as is. [Here](https://github.com/nasa/fprime/pull/818#pullrequestreview-706694500) are the relevant requested changes by @LeStarch 

@ethancheez

SMorettini:
Hello @csmith608 , I need this feature too, are you already implementing it? Otherwise, I'll work on it next week.

csmith608:
Hi @SMorettini, I'm excited that this feature is immediately useful. I don't have the bandwidth right now to do the implementation. I was hoping @ethancheez was going to have a chance, but I'm not sure, @LeStarch?

SMorettini:
@csmith608 I create the [pull request](https://github.com/nasa/fprime/pull/2166)  since I had to implement it on my private repository as soon as possible.

thomas-bc:
Merged! 🚀

PR #2166, opened 2023-08-01
Implementation of the offsets in the RateGroupDriver

SMorettini:
| | |
|:---|:---|
|**_Affected Component_**| Svc.RateGroupDriver  |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2135  |

---
## Change Description

This pull request adds the possibility to define some offset in the RateGroupDriver. This pull request was first created [here](https://github.com/nasa/fprime/pull/818) but never merged. An issue was reopened [here](https://github.com/nasa/fprime/issues/2135). 

I apply the changes suggested in the previous pull request with the [suggested changes](https://github.com/nasa/fprime/pull/818#pullrequestreview-706694500) from the reviewer @LeStarch .


csmith608:
@SMorettini upon initial inspection this looks good, thanks so much for (hopefully) taking it across the finish line

LeStarch:
@SMorettini you should make sure to update the RPI deployment as well.  This is the source of the code break. 

Also, "No raw arrays in interfaces" is caused because you created an array of `Dividers`.  To fix you could create a type called `DividerSet`, which is:

```c++
struct DivisderSet {
   Divider dividers[Svc::RateGroupDriver::DIVIDER_SIZE];
}
```

I'd prefer it if you'd make that adjustment as this was what we were trying to adjust last time.

LeStarch (changes requested):
Made some comments in the discussion.  Let me know what you'd do and we'll pick-up the rest.

SMorettini:
@LeStarch I added the changes requested, let me know if some other improvements are required.

I also created a pull request to update the Led blinker tutorial to work with the new RateGroupDriver dividers:  https://github.com/fprime-community/fprime-workshop-led-blinker/pull/41

LeStarch:
I made some changes here:
1. Added constructors to DivisorSet to guarantee when not explicitly initialized, that that pair was set to zero
2. Removed argument "numDivisors" as when we transitioned to a set, this resulted in it being a constant.

Need to let CI run, but otherwise ready to go!

Svc/RateGroupDriver/RateGroupDriver.hpp
```diff
@@ -43,17 +68,10 @@ namespace Svc {
             //!
             RateGroupDriver(const char* compName);
 
-            //!  \brief RateGroupDriver initialization function
-            //!
-            //!  The init() function initializes the autocoded base class
-
-            void init(NATIVE_INT_TYPE instanceId = 0);
-
             //!  \brief RateGroupDriver configuration function
```
timcanham, line 71: What happened to the init() call? Is the new pattern to call the base class init()?

timcanham (approved):
Approve but with a general question about init() calls...

Svc/RateGroupDriver/RateGroupDriver.hpp
LeStarch: Yes, we've been slowly removing `init` calls for some time.  The base-class method is sufficient. Here is why:

1. `init` is called in autocode by FPP.  Thus custom configuration is discouraged here.  Devs may choose to use the constructor or a dedicated `configure` method that takes in the necessary parameters.
2. `init` is not implemented with `override` and thus often leads to shadowing of the `init` method.

LeStarch:
@thomas-bc: core component, would you be a second set of 👀 ?

Svc/RateGroupDriver/RateGroupDriver.cpp
```diff
@@ -12,29 +12,30 @@ namespace Svc {
...
                 static_cast<NATIVE_INT_TYPE>(FW_NUM_ARRAY_ELEMENTS(this->m_dividers)),
                 this->getNum_CycleOut_OutputPorts());
-        // clear table
-        ::memset(this->m_dividers,0,sizeof(this->m_dividers));
         // copy provided array of dividers
-        for (NATIVE_INT_TYPE entry = 0; entry < numDividers; entry++) {
-            this->m_dividers[entry] = dividers[entry];
+        for (NATIVE_UINT_TYPE entry = 0; entry < RateGroupDriver::DIVIDER_SIZE; entry++) {
+            // A port with an offset equal or bigger than the divisor is not accepted because it would never be called
+            FW_ASSERT((this->m_dividers[entry].offset==0)||(this->m_dividers[entry].offset < this->m_dividers[entry].divisor),
```
thomas-bc: Should this not assert `dividersSet.dividers`, which is the passed-in argument, rather than `this->m_dividers` which is initialized to 0 and is going to be overwritten down below anyways?

LeStarch: Yes.

timcanham (approved):

thomas-bc (approved):
🚀

LeStarch (approved):
Approved!  Thanks for the fixes!

PR #2180, opened 2023-08-03
A note on sudo for virtual environments

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Minor changes from the tutorial run-through.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -31,8 +31,7 @@ The ecosystem of tools supporting F´ is installed as python packages available 
 pip install fprime-tools
 ```
 
-> Python is used by many operating systems. To prevent problems users are encouraged to run F´ python from within a [virtual environment](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/). This is set up when creating a new F´ project.
-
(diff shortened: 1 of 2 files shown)


docs/INSTALL.md
```diff
@@ -31,8 +31,7 @@ The ecosystem of tools supporting F´ is installed as python packages available
 pip install fprime-tools
 ```
 
-> Python is used by many operating systems. To prevent problems users are encouraged to run F´ python from within a [virtual environment](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/). This is set up when creating a new F´ project.
-
+> Python is used by many operating systems. To prevent problems you are encouraged to run F´ python from within a [virtual environment](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/). If these tools are not installed in a virtual environment, you may need to use `sudo` to run the above command.
```
thomas-bc, line 34: Is there a motivation for removing the `This is set up when creating a new F´ project.` ?
I feel like there may be confusion here for users that come in and have never used F´ prior to v3.3, and have therefore not installed requirements.txt themselves.

>The ecosystem of tools supporting F´ is installed as python packages available via PIP. `pip install fprime-tools`

I would read that as "this is all that's needed". We may want to clarify,
- either get `fprime-tools`, then have `new --project` setup a virtual environment
- or setup your virtual environment yourself, but you'll need to `pip install -r requirements.txt`

thomas-bc (commented):
I changed a dead link to HelloWorld, and a comment:

PR #2181, opened 2023-08-03
FPP v2.0 integration

tiffany1618:
## Change Description

This PR integrates FPP v2.0, which adds support for C++ generation of unit test harness classes, with F Prime.

Unit tests for the FPP unit test autocoder have been added to the existing tests for FPP components.

### General
- Prepend the component name to the generated file names, e.g. `[ComponentName]TesterBase.cpp` instead of `TesterBase.cpp`. This allows the generation of test harness classes for multiple components within the same directory.

### TesterBase classes
- Reorganize class members (follows the same pattern as the reorganization of class members for component base classes, see #2103)
- Change `typedef struct {...} [name];` to `struct [name] {...};`
- Move initialization of `m_param_[paramName]_valid` variable members from the `init` function to initializer lists in the constructor
- Fix issue where `[portName]_static` and dispatch function implementations for special ports are not generated when the corresponding special port constructs do not appear (e.g. a component contains command ports, but no commands)
- Change `sizeof([serializableType])` to `[serializableType]::SERIALIZED_SIZE` where appropriate in the `dispatchEvents` function

## To do before merging

- [x] Merge unit test code generation into FPP
- [ ] Integrate unit test code generation with fprime tools
- [ ] Create FPP pre-release to test PR
- [ ] Create FPP release v2.0

### DIFF
--- a/FppTest/component/active/test/ut/Tester.hpp
+++ b/FppTest/component/active/test/ut/Tester.hpp
@@ -197,8 +197,6 @@ class Tester : public ActiveTestGTestBase {
                                                       */
     );
 
-    void cmdResponseIn(const FwOpcodeType opCode, const U32 cmdSeq, const Fw::CmdResponse response);
-
   PRIVATE:
     // ----------------------------------------------------------------------
     // Handlers for serial from ports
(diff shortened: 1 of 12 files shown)


bocchino:
The top-level comment sets out the procedure for integrating FPP 2.0 into F Prime. However, the changes in this PR just improve code coverage of existing tests. I think this PR can be reviewed and merged now. Perhaps we should move the top-level comments and checklist to another issue focused on FPP 2.0 integration.

CI passes except for one failure that seems unrelated to this PR.

thomas-bc (approved):

bocchino (commented):
Looks good!

PR #2205, opened 2023-08-17
Add missing include

SMorettini:
| | |
|:---|:---|
|**_Affected Component_**|  config |

---
## Change Description

Added missing include.

## Rationale

Some static analysis tools like `clang-tidy` are creating a warning because they cannot find the declaration of `FwOpcodeType`. Moreover, any other config file that uses Fprime types already includes `FpConfig.hpp`.


LeStarch (approved):
Looks right to me.

PR #2217, opened 2023-08-18
Add support for additional baud rates

SMorettini:
| | |
|:---|:---|
|**_Affected Component_**| Drv.LinuxUartDriver |
|**_Related Issue(s)_**|  |

---
## Change Description

Added support for additional baud rates.


Drv/LinuxUartDriver/LinuxUartDriver.cpp
```diff
@@ -173,6 +173,30 @@ bool LinuxUartDriver::open(const char* const device,
...
+        case BAUD_1152K:
+            relayRate = B1152000;
+            break;
+        case BAUD_1500K:
+            relayRate = B1500000;
+            break;
+        case BAUD_2000K:
+            relayRate = B2000000;
+            break;
+        case BAUD_2500K:
```
LeStarch, line 189: According to the termios man page:
       Due to differences between architectures, portable applications
       should check if a particular Bnnn constant is defined prior to
       using it.

Thus, for the last 4 (250K) - (400K) could we add in:

```
#ifdef B2500000:
        case BAUD_2500K:
            relayRate = B2500000;
            break;
#endif
```

and the same in the .hpp

Other than that, this is a good change!

LeStarch (changes requested):

Drv/LinuxUartDriver/LinuxUartDriver.cpp
SMorettini: I made the requested change. The pull request is back to review

LeStarch:
@thomas-bc could you look through this too?

thomas-bc (approved):

PR #2279, opened 2023-09-20
Refactor code

PointlessUser:
| | |
|:---|:---|
|**_Affected Component_**|  |
Mostly Autocoders/python/bin
|**_Related Issue(s)_**|  |
code is a bit dated
y
y
y
y

---
## Change Description
Refactored some python code mostly in Autocoders/python/bin
I also ran "black ." to use black format on all the files

## Rationale
I was bored and the code is a bit dated so yeh

make sure Autocoders/python/bin doesn't fail which it probably wont

## Future Work
Might refactor more stuff

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,4 @@ depend
 /venv/
 
 Packet-Views
+*.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -84,4 +84,3 @@ depend
 /venv/
 
 Packet-Views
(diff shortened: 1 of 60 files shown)


LeStarch:
Per our contributing guidelines, we aren't accepting PRs against the autocoder python package as it is actively being replaced. https://github.com/nasa/fprime/blob/devel/CONTRIBUTING.md

PR #2305, opened 2023-10-10
Split fpp-input-list to import and source

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Split fpp-input-list to import and source files.

## Rationale

`fpp-to-cpp` needs to differentiate between source and import files to generate the correct set of files. The set of files is passed to fprime-tools using the `fpp-input-list`, which used to not differentiate between import and source. Splitting this in 2 different files allows for distinction to be made.


LeStarch (approved):
Minor comment:  does this need a cmake change to support this feature?

PR #2317, opened 2023-10-13
Convert Svc.Time into An Interface

### LINKED ISSUE #2256: Svc::Time Interface
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Make `Svc::Time` and interface.  See #2255.

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes #2256


Svc/Interfaces/Time.fppi
thomas-bc, line 1: Food for thought - `.fppi` is kind of our convention to say "this will be included somewhere else", right?

Can we see a scenario where it will get confusing for users if we go down the road of defining all ***i***nterfaces with `.fppi`, meaning users may come across `.fppi` first and think it means FPP-***I***nterface, but then some components also define their own `Events.fppi` etc... which have nothing to do with interfaces.

Should we have different file extensions to differentiate those, or are we happy to keep `.fppi` meaning "this will get included"

thomas-bc (commented):
Looks good to me, just one minor comment

Svc/Interfaces/Time.fppi
LeStarch: We decided that `.fppi` should become `.interface.fppi`.  However, I'd like to fix that in a future PR as this effects other interfaces too.

LeStarch: @thomas-bc if you concur, mark approval, otherwise, post back.

thomas-bc (approved):

LeStarch:
Merge it merge it!

PR #2360, opened 2023-11-07
Refactor locs, dependencies into generic sub build

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Creates generic sub-builds and makes locs/dependencies into a sub-build.  This replaces the hardcoded prescan function.

## Rationale

Sub-builds are a standard pattern in CMake and this exposes this functionality in a generic form to users.  Reworking pre-scan into this has the following advantages:

1. "Rebuilding" loc/dependencies does not take substantial time after the first run
2. Proper dependency tracking and rebuilds work
3. `-j` will work for sub/builds and fpp-depend isn't mapped to the whole CPU.

## Future Work

## Review fixes tracked as check-list:
- [x] Remove commented out FPrime-Code
- [x] Remove comment link to pre-scan
- [ ] Remove Fprime-Code.cmake
- [x] Change [prescan] to [subbuild] or [subbuild-name]
- [x] Always print executing step


LeStarch:
Note: although structurally much nicer, this slows down the initial generate time.  Here are some numbers:

```
Original (Ref): initial generation
-- Configuring done (14.0s)
-- Generating done (0.6s)

This PR (Ref): initial generation w/ 16 jobs
-- Configuring done (17.7s)
-- Generating done (0.6s)

Original (Ref): re-generation
-- Configuring done (12.7s)
-- Generating done (0.5s)

This PR (Ref): initial re-generation w/ 16 jobs
-- Configuring done (6.0s)
-- Generating done (0.6s)
```

Notice that the re-generation (changed a model/CMakeLists.txt) is substantially faster.

Ref/CMakeLists.txt
```diff
@@ -31,7 +30,11 @@ project(Ref VERSION 1.0.0 LANGUAGES C CXX)
 ##
 include("${CMAKE_CURRENT_LIST_DIR}/../cmake/FPrime.cmake")
 # NOTE: register custom targets between these two lines
-include("${CMAKE_CURRENT_LIST_DIR}/../cmake/FPrime-Code.cmake")
+#include("${CMAKE_CURRENT_LIST_DIR}/../cmake/FPrime-Code.cmake")
```
thomas-bc: Are you leaving these commented out on purpose?

Ref/SignalGen/CMakeLists.txt
```diff
@@ -8,9 +8,20 @@
 set(SOURCE_FILES
   "${CMAKE_CURRENT_LIST_DIR}/SignalGen.fpp"
   "${CMAKE_CURRENT_LIST_DIR}/SignalGen.cpp"
+
+
 )
 
 register_fprime_module()
+
+
```
thomas-bc, line 17: Not that it matters, but pointing out the extra spaces here. Testing artifacts? 👀

cmake/sub-build/sub-build.cmake
```diff
@@ -0,0 +1,115 @@
+####
+# sub-build.cmake:
+#
+# CMake occasionally requires sub-builds in order to generate certain types of information.  The sub-build mechanics are
+# implemented in this file. This includes the ability to setup targets, builds, and handle information passing back and
+# forth between the modules.
+#
+# This file runs CMake on fprime, but registering only a single target: (e.g. [prescan](./target/prescan.cmake) that
```
thomas-bc: comment `e.g. [prescan](./target/prescan.cmake` is out-of-date here

cmake/sub-build/sub-build.cmake
```diff
@@ -0,0 +1,115 @@
...
+            continue()
+            # Exclude internal cache values
+        elseif("${CACHE_TYPE}" IN_LIST TYPES_DISALLOWED_LIST)
+            continue()
+        endif()
+        # Add escaping for list type variables (for pass to function call and then again to another function call)
+        string(REPLACE ";" "\\\\;" PROP_VALUE "${${PROPERTY}}" )
+        # Check for debugging output
+        if (CMAKE_DEBUG_OUTPUT)
+            message(STATUS "[prescan] Adding cache variable: '${PROPERTY}=${PROP_VALUE}'")
```
thomas-bc: Should this say `sub-build` or use `SUB_BUILD_NAME` instead of `[prescan]`

cmake/FPrime-Code.cmake
thomas-bc, line 1: I'm assuming this file is kept in for backwards-compatibility only? Why not remove it and require projects to change their setup as a (pretty simple) breaking change? Or maybe mark it for deprecation and remove it in the release after the upcoming one? e.g. `message(WARNING "[Deprecation] ...")`

cmake/sub-build/sub-build.cmake
```diff
@@ -0,0 +1,115 @@
...
+            if(CPU_COUNT GREATER 0)
+                set(ENV{MAKEFLAGS} "--jobs=${CPU_COUNT}")
+            endif()
+        endif()
+        foreach (TARGET IN LISTS ARGN)
+            get_filename_component(TARGET_NAME "${TARGET}" NAME_WE)
+            if (CMAKE_DEBUG_OUTPUT)
+                message(STATUS "[sub-build] Executing: ${SUB_BUILD_NAME} with ${TARGET_NAME}")
+            endif ()
+            execute_process_or_fail("[sub-build] Failed to execute: ${SUB_BUILD_NAME}/${TARGET_NAME}"
```
thomas-bc, line 65: Would it be worth logging this regardless of DEBUD_OUTPUT? It's not a lot of output but does give an indication on progress - my laptop seems to be a little slow on executing the subbuilds

thomas-bc (commented):
A couple of nits/comments, but overall looks good to me!

Ref/CMakeLists.txt
LeStarch: The commented out call should be removed.  It is all done by the function call.

Ref/SignalGen/CMakeLists.txt
LeStarch: Not sure, but I cleaned it up in a subsequent PR....so let's catch that there.

cmake/sub-build/sub-build.cmake
LeStarch: That should be fixed!

cmake/sub-build/sub-build.cmake
LeStarch: Aye, it should!

cmake/FPrime-Code.cmake
LeStarch: Yes I suppose it was kept for backwards compatibility.  I didn't think to delete it, but you are right it can be!

cmake/sub-build/sub-build.cmake
LeStarch: Sure!

cmake/FPrime-Code.cmake
LeStarch: For now I decided to keep backwards compatibility.  We can fix this when I get the "find_package(fprime)" stuff working.   That way users don't need to fix one, then the other at separate times.

thomas-bc (approved):
🚀

PR #2366, opened 2023-11-10
Update/sloc cache

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Builds off #2360.  This adds generation of two files in each modulke: `module-info.txt` and `module-ut-info.txt` (BUILD_TESTING only). These track sources, generated outputs, etc.  This allows external introspection of the modules for things like SLOC counting.


Ref/SignalGen/CMakeLists.txt
```diff
@@ -9,24 +9,21 @@ set(SOURCE_FILES
   "${CMAKE_CURRENT_LIST_DIR}/SignalGen.fpp"
   "${CMAKE_CURRENT_LIST_DIR}/SignalGen.cpp"
 
-
+)
+set(HEADER_FILES
```
thomas-bc, line 13: We may want to document this and UT_HEADER_FILES in some kind of advanced/optional section?

thomas-bc (approved):
LGTM! My only recommendation would be to add some documentation if we want to advertise this functionality.

Ref/SignalGen/CMakeLists.txt
LeStarch: I agree, not sure where to put it.   I'll think on it.

LeStarch: I am going to merge this.  We can add the documentation in with the tool that leverages this feature.

PR #2367, opened 2023-11-10
Removing join after task failed creation. Fixes #1955

### LINKED ISSUE #1955: create_pthread() crashes if pthread_create() failed
Look here:

https://github.com/nasa/fprime/blob/91c578ee37c01ba92e556dbca0287d82f9f18b95/Os/Posix/Task.cpp#L161

If `pthread_create()` fails, fprime checks for failure. But if it did fail, it tries to `pthread_join()`. THIS causes a segfault. There is no valid thread (because we couldn't create it), so any operation on it will crash.

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

When pthread_create fails, the contents of `*thread` is undefined, however; we erroneously pass it to a `pthread_join` after a failure.  This removes that erroneous line.

### DIFF
--- a/Os/Posix/Task.cpp
+++ b/Os/Posix/Task.cpp
@@ -182,7 +182,6 @@ namespace Os {
         }
         (void)pthread_attr_destroy(&att);
         if (stat != 0) {
-            (void)pthread_join(*tid, nullptr);
             delete tid;
             tid = nullptr;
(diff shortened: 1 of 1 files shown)


LeStarch:
Fixes #1955

thomas-bc (approved):

timcanham (approved):
LGTM!

PR #2386, opened 2023-11-28
Fix static analysis warning in Serializable.cpp (#2385)

### LINKED ISSUE #2385: Static Analysis Warning in Serializable.cpp
According to static analysis, this code may not have the intended behavior:

https://github.com/nasa/fprime/blob/ec1a9b9f3eccec94863b7498a64b845edd451b89/Fw/Types/Serializable.cpp#L481

It casts the incoming `void*` value to a reference, which is undefined.

I believe the correct code should be something like this:
```c++
SerializeStatus SerializeBufferBase::deserialize(void*& val) {
  POINTER_CAST pointerCastVal = 0;
  const SerializeStatus stat = this->deserialize(pointerCastVal);
  if (stat == FW_SERIALIZE_OK) {
      val = reinterpret_cast<void*>(pointerCastVal);
  }
  return stat;
}
```

Note: This may technically be a static analysis false positive (casting a reference to a reference does seem to be allowed in C++). However, the cast from void*& to POINTER_CAST& apparently confused the static analyzer, so I think the proposed code is an improvement in any event.

Also: @LeStarch points out that the original code assumed that POINTER_CAST and void* have the same size, which may not be true. We just require sizeof(POINTER_CAST) >= sizeof(void*).

bocchino:
* Fix the coding issue.
* Add unit test coverage for the code.

Closes #2385.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -785,6 +785,7 @@ ptf
 pthread
 ptmcg
 pton
+ptrt
 puml
 punc
 pushd
--- a/Fw/Types/Serializable.cpp
+++ b/Fw/Types/Serializable.cpp
@@ -479,7 +479,7 @@ namespace Fw {
 
     SerializeStatus SerializeBufferBase::deserialize(void*& val) {
         // Deserialize as pointer cast, then convert to void*
-        POINTER_CAST pointerCastVal = 0;
+        PlatformPointerCastType pointerCastVal = 0;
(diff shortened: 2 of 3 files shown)


LeStarch:
Since spelling needs to be appeased anyway, could we rename `POINTER_CAST` to `PlatformPointerCastType`?  This is a migration we need to get to anyway.

bocchino:
Done!

LeStarch (approved):

thomas-bc (approved):

PR #2412, opened 2023-12-07
Fixing Two Minor CMake Bugs

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|  ||

---
## Change Description

There are two minor CMake bugs:
1. A set of variables was cleared rather than unset
2. CMAKE_SYSTEM_NAME was used instead of FPRIME_PLATFORM after creation of that variable

## Rationale
Bug fixes!


thomas-bc (approved):

LeStarch:
Fixes #2413 and #2414

PR #2430, opened 2023-12-12
Fix check targets name mismatch

### LINKED ISSUE #2424: _check Targets Miss-Named
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

`fprime-util check` fails on modules because the `_check` target is miss-named.

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2424 |

---

## Rationale

Fixes https://github.com/nasa/fprime/issues/2424
Also removes the `check_leak` target

## Future Work

Remove check_leak in fprime-tools ( `fprime-util check --leak` )


cmake/target/check.cmake
```diff
@@ -1,3 +1,5 @@
+include(target/ut)
```
thomas-bc, line 1: Needed to retrieve (and keep in a single place) the hardcoded `"ut_exe"` suffix

cmake/target/check.cmake
```diff
@@ -59,12 +61,17 @@ function(check_add_module_target MODULE_NAME TARGET_NAME SOURCE_FILES DEPENDENCI
     # Protects against multiple calls to fprime_register_ut()
     if (NOT BUILD_TESTING OR NOT MODULE_TYPE STREQUAL "Unit Test")
         return()
-    elseif (NOT TARGET ${MODULE_NAME}_${TARGET_NAME})
+    endif()
+    # UTs MODULE_NAME defaults to <FPRIME_MODULE_NAME>_ut_exe
+    # The below handling gives CHECK_TARGET_NAME = <FPRIME_MODULE_NAME>_check
+    string(REGEX REPLACE "_${UT_TARGET}$" "" CHECK_TARGET_NAME "${MODULE_NAME}")
+    string(APPEND CHECK_TARGET_NAME "_${TARGET_NAME}")
```
thomas-bc, line 68: Doing regex replace _then_ append so that it also works with UTs that register themselves with a hand-coded name, e.g. https://github.com/nasa/fprime/blob/95c42a0496ad5f0172da28ee23d0a962a0f3ff7d/Drv/Ip/CMakeLists.txt#L44

LeStarch (approved):

PR #2436, opened 2023-12-17
Fix pedantic warning on assert macro

JohanBertrand:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/discussions/2409 |

---
## Change Description

Fix the pedantic warning due to empty variadic variables in the assert macro.

Also cleaned some of the code in `Assert.cpp` to avoid repeated code.

## Rationale

The new definition is fully compatible with the previous implementation and should solve the issue mentioned in https://github.com/nasa/fprime/discussions/2409

### DIFF
--- a/Fw/Types/Assert.cpp
+++ b/Fw/Types/Assert.cpp
@@ -165,162 +165,10 @@ namespace Fw {
         s_assertHook = this->previousHook;
     }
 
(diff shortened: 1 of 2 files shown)


timcanham (approved):
Looks good to me. We'll put multiple eyes on it since it is core code, but good work!

thomas-bc (approved):

LeStarch (approved):

PR #2441, opened 2023-12-19
Fix pedantic warnings

JohanBertrand:
| | |
|:---|:---|

---
## Change Description

Fixes some of the pedantic warnings in the code base. To be merged/checked once https://github.com/nasa/fprime/pull/2436 is merged.
I activated the pedantic flag on the Ref project, I am not sure if it is something that was expected.
I kept it deactivated on the root CMakeLists.txt.

Still open to be pedantic-compliant is the declaration of 0 sized array, which I guess, would have to be fixed from the fpp side.

```
build-fprime-automatic-native/F-Prime/Svc/CmdSequencer/CmdSequencerComponentAc.cpp:41:60: error: ISO C++ forbids zero-size array ‘seqCancelInPortSize’ [-Werror=pedantic]
   41 |       BYTE seqCancelInPortSize[Svc::InputCmdSeqCancelPort::SERIALIZED_SIZE];
      |                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
```

There are also some warnings on the unit tests, but I guess it will need its own PR.

---
Main changes:
- Changed `%d` to `%u` for printing unsigned integer
```
error: format ‘%d’ expects argument of type ‘int*’, but argument 4 has type ‘U32*’ {aka ‘unsigned int*’} [-Werror=format=]
```
- Removed semicolon after closure of namespaces
```
error: extra ‘;’ [-Werror=pedantic]
```
- Changed designated initializers to sequential initialisation
```
error: C++ designated initializers only available with ‘-std=c++2a’ or ‘-std=gnu++2a’ [-Werror=pedantic]
```
- Changed variadic macro definition to have at least one argument
```
error: ISO C++11 requires at least one argument for the "..." in a variadic macro [-Werror]
```


SMorettini:
> Still open to be pedantic-compliant is the declaration of 0 sized array, which I guess, would have to be fixed from the fpp side.

Should be solved by https://github.com/fprime-community/fpp/pull/368

Drv/LinuxI2cDriver/LinuxI2cDriver.cpp
```diff
@@ -174,25 +174,23 @@ namespace Drv {
...
-            .len = static_cast<U16>(readBuffer.getSize()),
-            .buf = readBuffer.getData()
-        }
-    };
-
-    struct i2c_rdwr_ioctl_data rdwr_data = {
-        .msgs = rdwr_msgs,
-        .nmsgs = 2
-    };
+    struct i2c_msg rdwr_msgs[2];
```
LeStarch, line 177: I don't understand why this needed to be refactored.  We use this style initialization in multiple places.  Is this a violation of `-pedantic`?

LeStarch (commented):
Just the one question, otherwise looks good!

Drv/LinuxI2cDriver/LinuxI2cDriver.cpp
JohanBertrand: From what I have seen, it has only been officially supported by c++20, before that it was just an extension of GCC, but not part of the official c++11 standard, hence the pedantic warning.

```
error: C++ designated initializers only available with ‘-std=c++2a’ or ‘-std=gnu++2a’ [-Werror=pedantic]
```

I checked the code base using some regex to look for this type of initialization, and I did not spot any other places, other than in the shared folder for some tests.

LeStarch: Thanks for looking into this!  I will re-review!

LeStarch (approved):

thomas-bc (approved):

PR #2444, opened 2023-12-20
Add FPRIME_PROJECT_ROOT to CMAKE_MODULE_PATH. Fixes #2443

### LINKED ISSUE #2443: FPRIME_PROJECT_ROOT/cmake Not Added to CMake Path
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

FPRIME_PROJECT_ROOT/cmake  is expected to be added to the CMAKE_MODULE_PATH just like fprime libraries are.  However, it is not.

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes #2443

### DIFF
--- a/cmake/FPrime.cmake
+++ b/cmake/FPrime.cmake
@@ -12,6 +12,11 @@ include(options)
 include(sanitizers) # Enable sanitizers if they are requested
 include(required)
 
+# Add project root's cmake folder to module path
+if (IS_DIRECTORY "${FPRIME_PROJECT_ROOT}/cmake")
+    list(APPEND CMAKE_MODULE_PATH "${FPRIME_PROJECT_ROOT}/cmake")
+endif()
+
 # Setup fprime library locations
 list(REMOVE_DUPLICATES FPRIME_LIBRARY_LOCATIONS)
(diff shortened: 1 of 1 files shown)


cmake/FPrime.cmake
```diff
@@ -12,6 +12,11 @@ include(options)
 include(sanitizers) # Enable sanitizers if they are requested
 include(required)
 
+# Add project root's cmake folder to module path
+if (IS_DIRECTORY "${FPRIME_PROJECT_ROOT}/cmake")
```
thomas-bc, line 16: Could this cause an issue if projects using the new build structure have a folder called `cmake/` at the top level of their project? e.g. to store platform files or something

LeStarch: This is to support exactly that use case

LeStarch: When a project stores platform files, autocoders, and targets, these things are intended to be on the CMAKE_MODULE_PATH....which is why it is being added

thomas-bc (approved):

PR #2451, opened 2023-12-22
Switching to absolute paths in CMake. Fixes #2449

### LINKED ISSUE #2449: CMake System Uses `REALPATH` in `normalize_paths` Incorrectly
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

When `REALPATH` is used below, but no where else....then building only functions when building within a realpath.

https://github.com/nasa/fprime/blob/6641547e7ea8ec7593f7055cd2d5ff1e4acea990/cmake/utilities.cmake#L204

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Switches to using `ABSOLUTE` instead of `REALPATH` within CMake to fix #2449  and step towards correct builds in sym-linked directories.

## Future Work

### DIFF
--- a/cmake/utilities.cmake
+++ b/cmake/utilities.cmake
@@ -201,7 +201,7 @@ function(normalize_paths OUTPUT_NAME)
     # Loop over the list and check
     foreach (PATH_LIST IN LISTS ARGN)
         foreach(PATH IN LISTS PATH_LIST)
-            get_filename_component(PATH "${PATH}" REALPATH)
+            get_filename_component(PATH "${PATH}" ABSOLUTE)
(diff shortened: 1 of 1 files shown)


LeStarch:
Note: while this corrects parts of the issue, FPP tools (implemented in Java) still deal in real-paths.  This is not something that can be resolved as it is built into the Java world.  Thus when passing paths to FPP we must convert to `REALPATH` within the FPP autocoder to properly allow symlink builds.

thomas-bc (approved):

PR #2472, opened 2024-01-08
Generalize UT of RateGroupDriver

SMorettini:
| | |
|:---|:---|

---
## Change Description

The UT of `RateGroupDriver` was failing if `RateGroupDriverRateGroupPorts` in `AcConstant.fpp` was different from 3. The new change makes the test work for any value.


thomas-bc (approved):

LeStarch:
I switched the use of `int` (signed) to the used of `FwIndexType` that is the expected type to represent such things (i.e. port indices).  Hope you don't mind the assistance, figured it was easier than the back-and-forth for such a small delta.  Thanks for this correction, it will help substantially!

LeStarch (approved):

SMorettini:
@LeStarch thanks for the fix. It still doesn't pass the CI, I wanted to solve it but I don't know if you prefer casting the `DIVIDER_SIZE` or change the type of `i`

LeStarch:
@SMorettini in this case, I'd cast it.

LeStarch (approved):

PR #2482, opened 2024-01-13
Fix shadow variables

### LINKED ISSUE #2491: Resolve Warnings for `-Wshadow`
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Many bits of code in F´ produce warnings when compiling with `-Wshadow`.  We should fix these warnings.  The style guidelines inform us that recommended practice is to use `m_` prefixes for members to avoid this collision.

## Rationale

Code quality.

JohanBertrand:
| | |
|:---|:---|

---
## Change Description

Add support of `-wShadow` in the code base.

Forseen changes are:
- Add a prefix `m_` to `private` and `protected` class members
- Add a prefix `a_` to arguments shadowing a `public` class member
- Change variable names to avoid shadowing class member
- Remove redundant variable definition with same name

The `a_` prefix has been chosen to avoid breaking changes in the code. Changing the name of public variables can have a bigger impact on other projects. However, I think some of those variables should become `protected` instead of `public`, and the `_m` prefix should be used on those class members in the future. I think it's an open point of discussion.

A full compatibility with `-wShadow` will require some changes also in fpp. See https://github.com/fprime-community/fpp/pull/381
I am not sure what would be the release schedule. We might want to keep the `-Wshadow` flag out until the next fpp release happens.

## Rationale

Shadow variables can make the code more difficult to read and debug.

TBD


CMakeLists.txt
```diff
@@ -25,6 +25,14 @@ add_compile_options(
     -Wno-unused-parameter
 )
 
+# Add -Wshadow only for framework code, not the unit tests
+# Next lines to be uncommented once FPP is not generating files with shadow variables
```
LeStarch: If I find any other needed changes, I'd recommend adding a `TODO` or a `FIXME` so that it is obvious that a future change is necessary.

If no other changes are needed, this is non sufficient to hold up the merge.

Fw/FilePacket/DataPacket.cpp
```diff
@@ -18,15 +18,15 @@ namespace Fw {
   void FilePacket::DataPacket ::
     initialize(
         const U32 sequenceIndex,
-        const U32 byteOffset,
-        const U16 dataSize,
-        const U8 *const data
+        const U32 a_byteOffset,
```
LeStarch, line 21: We should make two issues for this PR:
1. Create an issue to track the removal of shadows, so this work ties to an issue
2. Create an issue to track the future work (e.g. reviewing and removing `a_` shadow collisions with `public` members.

Ref/CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ target_compile_options("${PROJECT_NAME}" PUBLIC -Wall)
 target_compile_options("${PROJECT_NAME}" PUBLIC -Wextra)
 target_compile_options("${PROJECT_NAME}" PUBLIC -Werror)
 target_compile_options("${PROJECT_NAME}" PUBLIC -pedantic)
-#target_compile_options("${PROJECT_NAME}" PUBLIC -Wshadow)
+target_compile_options("${PROJECT_NAME}" PUBLIC -Wshadow)
```
LeStarch, line 60: This should remain commented out until we fix the autocoder.  Otherwise `Ref` won't build warning-free.

STest/STest/Rule/Rule.hpp
```diff
@@ -27,9 +27,9 @@ namespace STest {
 
       //! Construct object Rule
       Rule(
-          const char *const name //!< The name of the rule
+          const char *const a_name //!< The name of the rule
```
LeStarch, line 30: @bocchino we will need to look into this w.r.t. the STest repository.  If and when we move to a full sub-repo, we'll need to ensure shadows are weeded-out.

I'll leave that to you to track before we make STest a subrepo.

Svc/FileUplink/FileUplink.hpp
```diff
@@ -171,17 +171,17 @@ namespace Svc {
...
-            ++this->n;
-            this->fileUplink->tlmWrite_Warnings(n);
+            ++this->m_n;
+            this->m_fileUplink->tlmWrite_Warnings(m_n);
           }
 
         PRIVATE:
 
           //! The total number of warnings
-          U32 n;
```
LeStarch, line 181: Another minor nit, not blocking for this PR....but given this PR is changing much, could we rename this to `m_warning_count` or something more descriptive?

LeStarch (changes requested):
I think there is only one required change I see: comment out the `-Wshadow` in the 
Ref package until the PR on FPP makes its way over.

If you are so-motivated, there are a few other minor comments on other cleanup items....however, none of these are required in this PR.  I'll take the action to add tickets as we need something to track this work.

@bocchino or @thomas-bc this is a big PR but entirely formulaic, would one of you take a look to ensure I did not miss anything?

LeStarch:
The CI failures are likely due to unit testing.  Much of our unit testing relies on white-box testing and as such does access these private members.  We'll need to fix those too.  @JohanBertrand are you up for this fix?

Fw/FilePacket/DataPacket.cpp
LeStarch: I made both issues.  Hope I am not stepping on toes.

LeStarch:
In case this isn't clear, this is solid work.  Thanks for the contribution.  Feel free to communicate us if you have reached the limit of what you are able to work on and we can move it through the process, otherwise we'll let you move it forward!

JohanBertrand:
> I think there is only one required change I see: comment out the `-Wshadow` in the Ref package until the PR on FPP makes its way over.

From what I've seen, it works well because the warning flags apply only to the user code, and not to the auto generated code.  The unit tests and the build on Ref were working fine for me.

> If you are so-motivated, there are a few other minor comments on other cleanup items....however, none of these are required in this PR. I'll take the action to add tickets as we need something to track this work.

I'm happy keeping the work going, all good. I would definitely do the fix of the `a_` prefix in another PR to make it easier to review, but I'll go through the rest.

> The CI failures are likely due to unit testing. Much of our unit testing relies on white-box testing and as such does access these private members. We'll need to fix those too. @JohanBertrand are you up for this fix?

It was compiling well for me. I already did changes in the test files, but just a few were needed. I did not run the `check` though. I'll double check, thanks.

Does this relate to the failure in `CI / Framework (pull_request)`?

> In case this isn't clear, this is solid work. Thanks for the contribution. Feel free to communicate us if you have reached the limit of what you are able to work on and we can move it through the process, otherwise we'll let you move it forward!

Thanks for the kind words, I'll come back with the changes in a few hours/days.

Ref/CMakeLists.txt
JohanBertrand: Correct me if I am wrong, but it looks like the warning here only affects the code written in the Ref project, but does not affect the code generated by fpp

Svc/FileUplink/FileUplink.hpp
JohanBertrand: I changed the `m_n` variable names to something more descriptive, and updated the comments of the variables. Let me know if there is a mistake in the comments

JohanBertrand:
@LeStarch it should be good to go, let me know if something else is need

I'll work on https://github.com/nasa/fprime/issues/2492 once this PR is closed

CMakeLists.txt
JohanBertrand: Added a TODO

LeStarch:
I fixed a comment.  We'll set what CI does, if it fails I can render more advice!

JohanBertrand:
Is there anything else needed before merging?

Ref/CMakeLists.txt
LeStarch: True, this only affects `Ref`

LeStarch (approved):

LeStarch:
I have approved, was hoping for one more set of eyes on this because of the extensive length of the PR.  Once we get that it will be merged!

thomas-bc (approved):

PR #2485, opened 2024-01-16
Data product framework support

bocchino:
This PR adds the following support for data products to the F Prime framework:

* A link to v2.1.0a3 of FPP, which supports data products
* Design documentation for the data product support
* New built-in types to support data products
* A new Com packet type to support data products
* New FPP code generation tests for data products

The next step is to add the data product support components in Svc.


Fw/Com/ComPacket.hpp
```diff
@@ -24,6 +24,7 @@ namespace Fw {
                 FW_PACKET_LOG, // !< Log type - outgoing
                 FW_PACKET_FILE, // !< File type - incoming and outgoing
                 FW_PACKET_PACKETIZED_TLM, // !< Packetized telemetry packet type
+                FW_PACKET_DP, //!< Data product packet
```
Joshua-Anderson, line 27: I'm coming into this fairly ignorant to the detailed F' data product architecture, but if data products are stored as files couldn't we just downlink them like standard files? Is there a benefit to giving DPs their own specific packet type?

bocchino: Yes, the plan is to downlink data products as standard files. The idea is that a data product is stored on disk as a header plus data. The header has a data product packet ID inside it, similarly to other F Prime data units such as com buffers. When a data product file is downlinked, it works like downlinking any file: the file is divided into packets, and each packet has a header with a file packet ID.

bocchino: In a review we discussed the fact that a data product packet is not intended to be used as a "com packet," i.e., as a packet in a transmission protocol. However, in current F Prime practice, "com packet" descriptors are not just for transmission; they also appear in data payloads, e.g., telemetry buffers and event buffers stored on disk. @LeStarch suggested that we move the ComPacket enumeration into a DataPacket enumeration. We can do this in a future PR.

Fw/Dp/Dp.fpp
```diff
@@ -0,0 +1,62 @@
+module Fw {
+
+  # ----------------------------------------------------------------------
+  # Types
+  # ----------------------------------------------------------------------
+
+  enum DpState: U8 {
+    @ The untransmitted state
+    UNTRANSMITTED
+    @ The transmitted state
```
timcanham, line 10: I was thinking about this RE: `DpCatalog`. It would make sense to add a `PARTIAL` to this so it can be set between the start of download and completion. `DpCatalog` would write this until it was done downloading so it would survive a reboot and would persist if the downlink vanished during a downlink.

Fw/Dp/DpContainer.cpp
```diff
@@ -0,0 +1,174 @@
...
+    // Deserialize the processing types
+    if (status == Fw::FW_SERIALIZE_OK) {
+        status = serializeRepr.deserialize(this->m_procTypes);
+    }
+    // Deserialize the user data
+    if (status == Fw::FW_SERIALIZE_OK) {
+        const bool omitLength = true;
+        const FwSizeType requestedSize = sizeof this->m_userData;
+        FwSizeType receivedSize = requestedSize;
+        status = serializeRepr.deserialize(this->m_userData, receivedSize, omitLength);
```
timcanham, line 76: Why not put `omitLength` in as a direct value in the argument? Readability?

Fw/Dp/DpContainer.cpp
```diff
@@ -0,0 +1,174 @@
...
+    }
+    // Deserialize the data product state
+    if (status == Fw::FW_SERIALIZE_OK) {
+        status = serializeRepr.deserialize(this->m_dpState);
+    }
+    // Deserialize the data size
+    if (status == Fw::FW_SERIALIZE_OK) {
+        status = serializeRepr.deserialize(this->m_dataSize);
+    }
+    return status;
```
timcanham, line 89: Should we check the header hash?

FppTest/dp/DpTest.cpp
```diff
@@ -0,0 +1,188 @@
...
+    this->dpRequest_Container4(CONTAINER_4_DATA_SIZE);
+    // Request a buffer for Container 5
+    this->dpRequest_Container5(CONTAINER_5_DATA_SIZE);
+    // Get a buffer for Container 1
+    {
+        DpContainer container;
+        Fw::Success status = this->dpGet_Container1(CONTAINER_1_DATA_SIZE, container);
+        FW_ASSERT(status == Fw::Success::SUCCESS, status);
+        // Check the container
+        this->checkContainer(container, ContainerId::Container1, CONTAINER_1_PACKET_SIZE);
```
timcanham, line 59: No `dpSend()?`

FppTest/dp/DpTest.cpp
```diff
@@ -0,0 +1,188 @@
...
+    if (status == Fw::Success::SUCCESS) {
+        auto serializeStatus = Fw::FW_SERIALIZE_OK;
+        for (FwSizeType i = 0; i < CONTAINER_1_DATA_SIZE; ++i) {
+            serializeStatus = container.serializeRecord_U32Record(this->u32RecordData);
+            if (serializeStatus == Fw::FW_SERIALIZE_NO_ROOM_LEFT) {
+                break;
+            }
+            FW_ASSERT(serializeStatus == Fw::FW_SERIALIZE_OK, status);
+        }
+        // Use the time stamp from the time get port
```
timcanham, line 108: No `checkContainer()?"

timcanham (changes requested):
Minor change and some questions

timcanham (approved):
Changing to approved, since the change I'm requesting is a small change that can be deferred.

LeStarch:
@bocchino I would like you to answer to @timcanham's questions and tell me if we should defer and merge, or update and merge

LeStarch:
Will merge for now, comments can be addressed in a follow-up.

Fw/Dp/DpContainer.cpp
bocchino: Yes, it's for readability, so you know what option is being selected.

Fw/Dp/DpContainer.cpp
bocchino: After discussion, I'll add separate helper functions for checking the hashes.

FppTest/dp/DpTest.cpp
bocchino: There is some redundancy in the test cases, because there are two ways to request buffers (sync and async). To simplify the testing logic, I skipped the `dpSend` for the sync case and had the component check the container directly. The async case performs the `dpSend` and checks the container at the other end. This way all the code under test is checked, but we avoid some duplicate test code.

FppTest/dp/DpTest.cpp
bocchino: See my previous response.

Fw/Dp/Dp.fpp
bocchino: I'll add this in the next PR.

bocchino: This is done on the data-products branch. We'll merge it in when we close #2530.

Fw/Dp/DpContainer.cpp
bocchino: Done on the data-products branch. I will merge in the changes when closing #2530.

PR #2489, opened 2024-01-17
Fix `Baremetal/FileSystem` `readDirectory()` impl

kbotteon:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

This PR fixes the stub implementation of `readDirectory()` under `Baremetal/Filesystem`.

## Rationale

The implementation does not match the signature in [FileSystem.hpp](https://github.com/nasa/fprime/blob/devel/Os/FileSystem.hpp#L30), which will cause an application using this function to fail to link.

CI should be testing that all implementations of these multi-target utilities build and link.

## Future Work

N/A

### DIFF
--- a/Os/Baremetal/FileSystem.cpp
+++ b/Os/Baremetal/FileSystem.cpp
@@ -13,9 +13,11 @@ Status removeDirectory(const char* path) {
     return INVALID_PATH;
 }  // end removeDirectory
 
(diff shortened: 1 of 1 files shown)


LeStarch (approved):

LeStarch:
Thanks @kbotteon!

PR #2494, opened 2024-01-19
Fix issues in FppTest

bocchino:
This PR fixes some issues in FppTest.
1. I noticed that the sanitizer was complaining about an array size of zero. I fixed it.
2. I noticed that a test was assuming a non-empty test string, and it failed in the case where the random test string was empty. I added the constraint that the string must have min length 1.
3. I cleaned up the logic and added some assertions in the function for generating random strings.
4. I removed trailing spaces from the files that I touched.

### DIFF
--- a/FppTest/struct/PrimitiveStructTest.cpp
+++ b/FppTest/struct/PrimitiveStructTest.cpp
@@ -39,8 +39,9 @@ protected:
     }
 
     void assertUnsuccessfulSerialization(Primitive& s, U32 bufSize) {
-        U8 data[bufSize];
-        Fw::SerialBuffer buf(data, sizeof(data));
+        // Avoid creating an array of size zero
+        U8 data[bufSize + 1];
+        Fw::SerialBuffer buf(data, bufSize);
         Fw::SerializeStatus status;
 
         // Serialize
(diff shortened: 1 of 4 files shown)


thomas-bc (approved):

FppTest/struct/PrimitiveStructTest.cpp
```diff
@@ -39,8 +39,9 @@ class PrimitiveStructTest : public ::testing::Test {
     }
 
     void assertUnsuccessfulSerialization(Primitive& s, U32 bufSize) {
-        U8 data[bufSize];
-        Fw::SerialBuffer buf(data, sizeof(data));
+        // Avoid creating an array of size zero
+        U8 data[bufSize + 1];
+        Fw::SerialBuffer buf(data, bufSize);
```
LeStarch, line 44: Was it intentional to not send `bufSize +1`?  Does `Fw::SerialBuffer` handle 0 length buffers in the constructor?

LeStarch (commented):
One minor question.

FppTest/struct/PrimitiveStructTest.cpp
bocchino: I wanted to make the minimal change to avoid the compiler warning. Using `bufSize` as the size of the serial buffer (1) is the original behavior of the test code and (2) does not cause any warnings. A physical array size of zero causes a warning, so I added one to the array size.

Issue #1893, opened 2023-02-15
AcConstants.ini exists but is not used


bocchino:
## Issue

* AcConstants.ini exists in the repo but isn't used for anything: https://github.com/nasa/fprime/blob/devel/config/AcConstants.ini. It was used in the hand-coded XML, which no longer exists. It has been superseded by AcConstants.fpp. This is confusing because users may update AcConstants.ini by mistake, when they should update AcConstants.fpp.
* The docs still refer to AcConstants.ini. See https://github.com/nasa/fprime/blob/devel/docs/UsersGuide/dev/configuring-fprime.md#acconstantsini.

## Suggested Solution

1. Remove AcConstants.ini from the repo.
2. Update the docs.

bocchino:
@gjwatney FYI.

LeStarch:
@bocchino is this something you would fix?  It seems fairly straight-forward.

bocchino:
This also needs to be fixed:

https://github.com/nasa/fprime/blob/devel/docs/UsersGuide/user/settings.md

Also, it looks like something in the F Prime tools is using settings.ini to find AcConstants.ini.

bocchino:
There are also references to AcConstants.ini here and there in the docs, e.g.:

https://github.com/nasa/fprime/blob/devel/Svc/StaticMemory/docs/sdd.md#configuration

We should update these to refer to AcConstants.fpp.

bocchino:
> Also, it looks like something in the F Prime tools is using settings.ini to find AcConstants.ini. This should be scrubbed.

Actually I think this is the first order of business. Otherwise it is not clear what will happen if the AcConstants.ini file is not present, and the tools are expecting it to be there.

LeStarch:
@bocchino doing this change will break our backward support of the XML modeling language.  Have we officially drop that support?

bocchino:
> doing this change will break our backward support of the XML modeling language

Can you elaborate the use case you have in mind that will break? AcConstants.ini defines some symbols that were used by the old handwritten XML and are not used by the FPP-generated XML. So at a minimum we should delete all those unused symbols and document that AcConstants.ini is there for backwards compatibility. Are you thinking that removing AcConstants.ini entirely will break something that users are doing with their own handwritten XML? That seems unlikely to me, since the purpose of AcConstants.ini was to configure the framework, and we are not using it for that (or for any other purpose) any more.

bocchino:
[...]

PR #2496, opened 2024-01-23
Remove AcConstants.ini

### LINKED ISSUE #1579: Reassess process for adding/editing constants defined in config folder
| | |
|:---|:---|
|**_F´ Version_**|3.0+|
|**_Affected Component_**| N/A  |
---
## Problem Description

For repo's using fprime as a submodule, users will need to copy the config folder into their deployment folder and point the `config_directory` item in the `settings.ini` file to the new config folder. Users should then be able to add constants to the AcConstants.ini or AcConstants.fpp file. However, this process does not work(see discussion here: #1555).

Some things that can be improved
1. Remove the need for BOTH the AcConstants.ini and AcConstants.fpp files, as they seem redundant. 
2. Allow users to add custom constants to the `AcConstants` file so that they can be referenced in their custom components. 
3. Allow users to edit the values of the constants that already exist in the `AcConstants` file. 

## How to Reproduce

1. Add fprime as a submodule to your repository
2. Copy the config folder within the fprime repo into your deployment directory
3. Edit the `config_directory` item in the `settings.ini` file to the path of the new config folder
4. Add a custom constant to the `AcConstants.fpp` or `AcConstants.ini` files/or change the value of a constant in those file
5. Generate and build deployment
6. Check build directory and check the `FppConstantsAc.hpp` file, the changes made in Step 4 will not be there. 

## Expected Behavior

The expected behavior should be for users to add or edit constants in the AcConstants.fpp/.ini file and have those changes be applied during the build of their deployment. 

An even better solution would be to avoid having users copy the fprime config folder into their deployment at all. In this case, users would still have a config folder in their deployment directory, but it would only contain any new .hpp files they need for their new components and an AcConstants.fpp file that ONLY contains new constants or existing constants with edited values. The fprime build system would then have to check both config folders and use that for the build. Any constants that are defined twice would default to the user configured value.

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| #1579 , #1893 |

---
## Change Description

Fix #1579 and #1893 

Remove `AcConstants.ini` file as it has been replaced by `AcConstants.fpp`. Updates relevant documentation entries.

May break some of the autocoding capabilities of `codegen.py` (?), but we are only using the dictionary autocoder which doesn't need it.

## Rationale

File is not being used and may cause confusion for users

CI

## Other

Associated fprime-tools PR: https://github.com/fprime-community/fprime-tools/pull/192


cmake/autocoder/ai_xml.cmake
```diff
@@ -76,7 +76,8 @@ macro(__ai_info XML_PATH MODULE_NAME)
         # Next parse the output matching one line at a time
         read_from_lines("${AI_OUTPUT}" XML_TYPE_FROM_FILE MODULE_DEPENDENCIES FILE_DEPENDENCIES)
     endif()
-    list(APPEND FILE_DEPENDENCIES "${FPRIME_AC_CONSTANTS_FILE}")
+    # Post v3.4.1, support for the AcConstants.ini file has been dropped in favor of AcConstants.fpp,
+    # list(APPEND FILE_DEPENDENCIES "${FPRIME_AC_CONSTANTS_FILE}")
```
LeStarch: The Ai.xml is still used for dictionary generation.  By dropping this here, do we not end up in  a situation where the `Ai.xml` is wrong?

LeStarch (commented):
In general I question if we can do this yet. We need the `Ai.xml` for making dictionaries, and we need `Ac.ini` in some form to make the `Ai.xml`.  Thus by deleting this, we theoretically could get corrupted dictionaries.

cmake/autocoder/ai_xml.cmake
thomas-bc: In which case would dropping this corrupt the dictionary ? 
Agreed that a more conservative approach could be to wait for the JSON dictionary to come in and drop the entire XML/Autocoder stuff at once... I went for it because we planned this ticket for this sprint haha

LeStarch: Delete!

LeStarch (changes requested):
Delete the commented out line.

LeStarch (approved):
For those following along at home, the realization is that `AcConstants.ini` is ignored in XML that was generated from FPP.  Thus it is only needed for the `2.x` style of writing XML by hand. Since this has been de-scoped, we can merge this!

Issue #2457, opened 2023-12-22
Refactor Os::File, Directory, and FileSystem into a CMake Selection


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Os::File -> CMake selection

1. Break into front-end and back-end.
2. Register as selection
3. Select as selection in platform

LeStarch:
Done and delivered! Wooo!

PR #2502, opened 2024-01-25
Issue 2457

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Issue #2457


LeStarch:
@bocchino this should be ready for review.  It appears the PI CI is offline again...but those have been passing.

Os/File.cpp
```diff
@@ -0,0 +1,237 @@
+// ======================================================================
+// \title Os/File.cpp
+// \brief common function implementation for Os::File
+// ======================================================================
+#include <Os/File.hpp>
+#include <Fw/Types/Assert.hpp>
+
+#ifdef __cplusplus
```
bocchino: Do we want this ifdef cplusplus block? The file is already including hpp files, so how can it be compiled as a C file?

Fw/SerializableFile/CMakeLists.txt
```diff
@@ -12,13 +12,10 @@ set(SOURCE_FILES
   "${CMAKE_CURRENT_LIST_DIR}/SerializableFile.cpp"
 )
 set(MOD_DEPS
-  Fw/Cfg
-  Fw/Types
+    Fw/Cfg
+    Fw/Types
+    Os
 )
-# For shared libraries, Os must exist in the module list
```
bocchino, line 18: This looks fine, but what changed to cause the deletion?

Os/File.hpp
```diff
@@ -1,66 +1,492 @@
...
+            //! \brief provide a pointer to a file delegate object
+            //!
+            //! This function must return a pointer to a `FileInterface` object that contains the real implementation of the file
+            //! functions as defined by the implementor.  This function must do several things to be considered correctly
+            //! implemented:
+            //!
+            //! 1. Assert that the supplied memory is non-null. e.g `FW_ASSERT(aligned_placement_new_memory != NULL);`
+            //! 2. Assert that their implementation fits within FW_HANDLE_MAX_SIZE.
+            //!    e.g. `static_assert(sizeof(PosixFileImplementation) <= sizeof Os::File::m_handle_storage,
+            //!        "FW_HANDLE_MAX_SIZE to small");`
```
bocchino: to  --> too

Os/Linux/FileSystem.cpp
```diff
@@ -317,19 +317,19 @@ Status initAndCheckFileStats(const char* filePath, struct stat* fileInfo = nullp
  * @param destination File to copy data to
  * @param size The number of bytes to copy
  */
-Status copyFileData(File source, File destination, FwSizeType size) {
```
bocchino, line 320: What is the reason for using a signed size here?

Os/Models/File.fpp
```diff
@@ -0,0 +1,30 @@
...
+@ FPP shadow-enum representing Os::File::Mode
+enum FileMode {
+    OPEN_NO_MODE, @<  File mode not yet selected
+    OPEN_READ, @<  Open file for reading
+    OPEN_CREATE, @< Open file for writing and truncates file if it exists, ie same flags as creat()
+    OPEN_WRITE, @<  Open file for writing
+    OPEN_SYNC_WRITE, @<  Open file for writing; writes don't return until data is on disk
+    OPEN_APPEND, @< Open file for appending
+}
+}
```
bocchino: Suggest adding a newline at the end of the file.

Os/Posix/File.cpp
```diff
@@ -0,0 +1,279 @@
...
+#include <Fw/Types/Assert.hpp>
+#include <Os/Posix/errno.hpp>
+
+namespace Os {
+namespace Posix {
+namespace File {
+
+// Os specific direct writing support
+#ifdef __linux__
+#define DIRECT_FLAGS O_DIRECT
```
bocchino: What is this for? It doesn't look like DIRECT_FLAGS is used anywhere.

Os/Posix/File.cpp
```diff
@@ -0,0 +1,279 @@
...
+#else
+#define DIRECT_FLAGS 0
+#endif
+
+#if defined(S_IREAD) && defined(S_IWRITE)
```
bocchino, line 23: What is going on here? Can we have a comment?

Os/Posix/File.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+PosixFile::Status PosixFile::write(const U8* buffer, FwSignedSizeType &size, PosixFile::WaitType wait) {
+    Status status = OP_OK;
+    FwSignedSizeType accumulated = 0;
+    // Loop up to 2 times for each by, bounded to prevent overflow
+    const FwSignedSizeType maximum = (size > (std::numeric_limits<FwSignedSizeType>::max()/2)) ? std::numeric_limits<FwSignedSizeType>::max() : size * 2;
+
+    for (FwSignedSizeType i = 0; i < maximum && accumulated < size; i++) {
+        // char* for some posix implementations
+        ssize_t write_size = ::write(this->m_handle.m_file_descriptor, reinterpret_cast<const CHAR*>(buffer + accumulated), size - accumulated);
```
bocchino: I recommend that we avoid pointer arithmetic using +. Can we say &buffer[accumulated] instead?

Os/Posix/test/ut/PosixFileTests.cpp
```diff
@@ -0,0 +1,144 @@
...
+    //! \return true
+    //!
+    bool functional() const override{
+        return true;
+    }
+
+};
+
+std::unique_ptr<Os::Test::File::Tester> get_tester_implementation() {
+    return std::unique_ptr<Os::Test::File::Tester>(new Os::Test::File::PosixTester());
```
bocchino, line 133: I think this should be std::make_unique in modern C++ style.

Os/Stub/test/ut/StubFileTests.cpp
```diff
@@ -0,0 +1,215 @@
...
+    //! \return true
+    //!
+    bool functional() const override {
+        return false;
+    }
+
+};
+
+std::unique_ptr<Os::Test::File::Tester> get_tester_implementation() {
+    return std::unique_ptr<Os::Test::File::Tester>(new Os::Test::File::StubsTester());
```
bocchino, line 63: See previous comment about std::unique_ptr.

Os/ValidateFileCommon.cpp
```diff
@@ -18,10 +18,10 @@ namespace Os {
 
         // Get the file size:
         FileSystem::Status fs_status;
-        FwSizeType fileSize = 0;
+        FwSignedSizeType fileSize = 0;
```
bocchino, line 21: Why are we switching to signed size?

Os/test/ut/OsTestMain.cpp
```diff
@@ -49,7 +49,7 @@ TEST(Nominal, DISABLED_IntervalTimerTest) {
    intervalTimerTest();
 }
 TEST(Nominal, FileSystemTest) {
-   fileSystemTest();
+//   fileSystemTest();
```
bocchino: Uncomment code, or remove.

Os/test/ut/file/MyRules.cpp
```diff
@@ -0,0 +1,1150 @@
+// ======================================================================
+// \title Os/test/ut/file/MyRules.cpp
+// \brief rule implementations for common testing
+// ======================================================================
+
+#include "RulesHeaders.hpp"
+#include "STest/Pick/Pick.hpp"
+#ifdef __cplusplus
```
bocchino: See earlier comment. ifdef cplusplus doesn't make sense if we are including hpp files. This is C++ code.

bocchino: Also I see C++ syntax in the file, so a C compiler won't compile it.

Os/test/ut/file/MyRules.cpp
```diff
@@ -0,0 +1,1150 @@
...
+          m_random(randomize_filename) {}
+
+bool Os::Test::File::Tester::OpenBaseRule::precondition(const Os::Test::File::Tester &state  //!< The test state
+) {
+    return state.m_mode == Os::File::Mode::OPEN_NO_MODE;
+}
+
+void Os::Test::File::Tester::OpenBaseRule::action(Os::Test::File::Tester &state  //!< The test state
+) {
+    printf("--> Rule: %s \n", this->name);
```
bocchino: Why are we printing out rules manually instead of using STest to do this? Is the issue that we want the rule printed every time it is applied, instead of just in a random scenario? If so, should we update STest? The manual printing seems awkward.

Os/test/ut/file/MyRules.cpp
```diff
@@ -0,0 +1,1150 @@
+// ======================================================================
```
bocchino: Why is this file called MyRules.cpp? It doesn't seem like a good name for production code developed by a team. (To whom does "my" refer to in this context?) Maybe just Rules.cpp?

Svc/CmdSequencer/CMakeLists.txt
```diff
@@ -59,5 +59,9 @@ set(UT_SOURCE_FILES
...
   "${CMAKE_CURRENT_LIST_DIR}/test/ut/CmdSequencerTester.cpp"
 )
-set(UT_MOD_DEPS "Os_Stubs")
-register_fprime_ut()
+
+set(UT_MOD_DEPS
+    Os_File_Posix
+)
+choose_fprime_implementation(Os/File Os_File_None)
+register_fprime_ut()
```
bocchino: Missing newline.

Svc/CmdSequencer/test/ut/CmdSequencerTester.hpp
```diff
@@ -174,45 +121,31 @@ namespace Svc {
...
-            // ----------------------------------------------------------------------
-
-            //! Intercept an open request
-            //! \return Success or failure
-            bool intercept(
-                Os::File::Status &fileStatus,
-                void *buffer,
-                NATIVE_INT_TYPE &size
-            );
+            class Override : public Os::Posix::File::PosixFile {
```
bocchino: Why is this class called Override? It seems better to give it a descriptive name based on what it does. You could call it PosixFile, if it's a specific implementation of Os::Posix::File::PosixFile.

Svc/PrmDb/test/ut/PrmDbImplTester.cpp
```diff
@@ -346,234 +351,160 @@ namespace Svc {
         PrmDbGTestBase::init();
     }
 
+    PrmDbImplTester* PrmDbImplTester::PrmDbTestFile::s_tester = nullptr;
 
-void PrmDbImplTester::runFileReadError() {
```
bocchino, line 350: What if anything is replacing what this deleted code does?

Svc/PrmDb/test/ut/PrmDbImplTester.cpp
```diff
@@ -346,234 +351,160 @@ namespace Svc {
...
+        this->m_errorType = FILE_DATA_ERROR;
+        for (FwSizeType i = 0; i < 2; i++) {
+            clearEvents();
+            this->m_waits = i;
+            this->m_impl.readParamFile();
+            ASSERT_EVENTS_SIZE(1);
+            switch (i) {
+                case 0:
+                    ASSERT_EVENTS_PrmFileReadError_SIZE(1);
+                    ASSERT_EVENTS_PrmFileReadError(0, PrmReadError::DELIMITER_VALUE, 0, 0xA6);
```
bocchino: What is this magic number? Should we use a defined constant?

Svc/PrmDb/test/ut/PrmDbImplTester.cpp
```diff
@@ -346,234 +351,160 @@ namespace Svc {
...
+                    break;
+                case 1:
+                    ASSERT_EVENTS_PrmFileReadError_SIZE(1);
+                    ASSERT_EVENTS_PrmFileReadError(0, PrmReadError::RECORD_SIZE_VALUE, 0, 16777224);
```
bocchino: What is this magic number?

Svc/PrmDb/test/ut/PrmDbImplTester.cpp
```diff
@@ -588,290 +519,100 @@ void PrmDbImplTester::runFileReadError() {
         ASSERT_CMD_RESPONSE_SIZE(1);
         ASSERT_CMD_RESPONSE(0,PrmDbImpl::OPCODE_PRM_SAVE_FILE,12,Fw::CmdResponse::EXECUTION_ERROR);
 
-        Os::clearOpenInterceptor();
```
bocchino, line 591: Where has the function of all this deleted code gone?

cmake/platform/types/PlatformTypes.h
```diff
@@ -47,8 +47,11 @@ typedef unsigned int PlatformUIntType;
 typedef PlatformIntType PlatformIndexType;
 #define PRI_PlatformIndexType PRI_PlatformIntType
 
-typedef PlatformUIntType PlatformSizeType;
-#define PRI_PlatformSizeType PRI_PlatformUIntType
+typedef int64_t PlatformSignedSizeType;
```
bocchino, line 50: Shouldn't this be ssize_t?

cmake/platform/types/PlatformTypes.h
```diff
@@ -47,8 +47,11 @@ typedef unsigned int PlatformUIntType;
...
+#define PRI_PlatformSignedSizeType PRId64
+
+typedef uint64_t PlatformSizeType;
+#define PRI_PlatformSizeType PRIu64
```
bocchino, line 54: Shouldn't this be size_t?

Os/Linux/FileSystem.cpp
bocchino: It does look more consistent with the uses of signed sizes elsewhere. Is that the reason?

bocchino:
Looks great! An improvement on many fronts. Please take a look at the comments.

Fw/SerializableFile/CMakeLists.txt
LeStarch: Os needed to be track as an actual link dependency.  Before it worked as a quirk of other ordering, but ordering changes revealed that it needed to be explicit....not just used for shared libs.

Os/File.cpp
LeStarch: Agreed.  Will commit a fix.

Os/File.hpp
LeStarch: Agreed.

Os/Models/File.fpp
LeStarch: Agreed.

Os/Posix/File.cpp
LeStarch: Direct flags were used when we supported the direct sync mode, which we have removed.

Os/Posix/File.cpp
LeStarch: Agreed.

Os/Posix/File.cpp
LeStarch: Fixed here and in read too.

Os/Posix/test/ut/PosixFileTests.cpp
LeStarch: std::make_unique is a C++14 feature, while we are at C++11.  I found this information here:

https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

Os/Stub/test/ut/StubFileTests.cpp
LeStarch: Same as above.  I wish this was supported as it is cleaner.

Os/test/ut/OsTestMain.cpp
LeStarch: Restored the UT for now.  Once Directory and FileSystem are refactored then this test will be no longer needed.

Os/test/ut/file/MyRules.cpp
LeStarch: This pattern is from @gjwatney's rule generator.  It starts each rule with this.  With the new update to STest changing the name of rules to `.getName()` this breaks.  I will scrub them.

Would be nice to have for debug output from STest though.

Os/test/ut/file/MyRules.cpp
LeStarch: FileRules as it is rules for the file test!

Svc/CmdSequencer/test/ut/CmdSequencerTester.hpp
LeStarch: It is really an interceptor working with PosixFile so I will call it `PosixFileInterceptor`

Svc/PrmDb/test/ut/PrmDbImplTester.cpp
LeStarch: I think this is the magic record delimiter.  I will see if I can make it a constant.

cmake/platform/types/PlatformTypes.h
LeStarch: I chose to use `int64_t` instead of `ssize_t` as this is a fixed-size integer and would remove the warnings about "use of un-sized integers".   This file is only for supporting the core systems Linux, Arm/Linux, and Darwin.  To the best of my knowledge each system in that list support 64bit.

cmake/platform/types/PlatformTypes.h
LeStarch: Same as for the signed equivalent.

Svc/PrmDb/test/ut/PrmDbImplTester.cpp
LeStarch: The open interceptor wasn't doing anything special.  Just returning known specific statuses.  `Os::Stub::File::Test::TestFile` already supports that built-in, so the code was switched to use that feature rather than reimplementing it.

Svc/PrmDb/test/ut/PrmDbImplTester.cpp
LeStarch: `runFileReadError` still exists.  Instead of doing copy-and-paste of 4 large blocks of code, I reduced the 4 read steps into a `for` loop as all but the record type was the same.  Makes it easier to maintain when you need to update the order.

Os/Linux/FileSystem.cpp
LeStarch: I believe it was done to be consistent with the use of signed-sizes inherent to the Os::File class, which itself mirrors the use of `ssize_t` in the posix API.  I will come back to this in a moment once I have addressed the other comments.

Svc/PrmDb/test/ut/PrmDbImplTester.cpp
LeStarch: Not exactly accurate, it is reduced to several for loops whose purpose is to setup a failed read after `i` successful reads. The process is the same for each so a for loop is appropriate.

An example is here: https://github.com/nasa/fprime/blob/b2fd0791b536489895fa6f434776be50a092c9e4/Svc/PrmDb/test/ut/PrmDbImplTester.cpp#L413-L438

Svc/PrmDb/test/ut/PrmDbImplTester.cpp
LeStarch: Done.

Svc/PrmDb/test/ut/PrmDbImplTester.cpp
LeStarch: I explained this magic number as best I could in a comment....but there is some magic inherited from the original test.

Os/Linux/FileSystem.cpp
LeStarch: Yes, my belief is correct.  The File API deals in signed sizes to match the underlying calls (which reserve the right to set size to -1 on error).  This change has flowed out to all uses of Os::File.

Os/ValidateFileCommon.cpp
LeStarch: The File API uses signed size, and these values are passed in and out of that API.

LeStarch:
@bocchino I think I fixed or addressed each of your comments.  Let me know if there is more you'd like to see!

Fw/SerializableFile/CMakeLists.txt
bocchino: Thanks!

Os/Linux/FileSystem.cpp
bocchino: OK, sounds good.

Os/Posix/test/ut/PosixFileTests.cpp
bocchino: OK, sounds good.

Svc/CmdSequencer/test/ut/CmdSequencerTester.hpp
bocchino: Sounds good!

Svc/PrmDb/test/ut/PrmDbImplTester.cpp
bocchino: Thanks for the explanation. I agree with the changes.

Svc/PrmDb/test/ut/PrmDbImplTester.cpp
bocchino: OK.

bocchino (approved):
Approved, based on the updates and comments. Good work!

bateller (approved):

cmake/implementation.cmake
```diff
@@ -59,10 +59,15 @@ function(setup_executable_implementations MODULE)
     # Check and setup implementors
     get_property(LOCAL_REQUIRED GLOBAL PROPERTY "REQUIRED_IMPLEMENTATIONS")
     if (LOCAL_REQUIRED)
+        init_variables(IMPLEMENTATION_SOURCES IMPLEMENTATION_DEPENDENCIES)
         foreach (IMPLEMENTATION IN LISTS LOCAL_REQUIRED)
             setup_executable_implementation("${IMPLEMENTATION}" "${MODULE}")
+            list(APPEND IMPLEMENTATION_SOURCES ${NEW_IMPLEMENTATION_SOURCES})
+            list(APPEND IMPLEMENTATION_DEPENDENCIES ${NEW_IMPLEMENTATION_DEPENDENCIES})
```
thomas-bc, line 66: I don't understand where these `NEW_*` values are coming from? I also can't find where these lists are being used, in the parent scope or anywhere else. Am I missing something? Or is this legacy code?

thomas-bc (approved):
LGTM! Just one question.

cmake/implementation.cmake
LeStarch: It was an older implementation, replaced by the `target_sources` command.  I removed it.

Good catch!

thomas-bc (approved):

Fw/Types/Serializable.cpp
```diff
@@ -500,10 +504,17 @@ namespace Fw {
     }
 
     SerializeStatus SerializeBufferBase::deserialize(U8* buff, NATIVE_UINT_TYPE& length, bool noLength) {
+        FwSizeType length_in_out = length;
+        SerializeStatus status = this->deserialize(buff, length_in_out, noLength ? Serialization::OMIT_LENGTH : Serialization::INCLUDE_LENGTH);
+        length = length_in_out;
```
timcanham: Add static cast for bigger to smaller type conversion.

Fw/Types/Serializable.cpp
```diff
@@ -500,10 +504,17 @@ namespace Fw {
...
+        return status;
+    }
+
+    SerializeStatus SerializeBufferBase::deserialize(U8* buff, FwSizeType& length, Serialization::t mode) {
 
         FW_ASSERT(this->getBuffAddr());
 
-        if (not noLength) {
+        if (mode == Serialization::INCLUDE_LENGTH) {
             FwBuffSizeType storedLength;
```
timcanham, line 529: Make overflow checks with `FwSizeType` with a serialization status if doesn't fit.

Fw/Types/Serializable.hpp
```diff
@@ -72,6 +80,17 @@ namespace Fw {
 
             SerializeStatus serialize(const U8* buff, NATIVE_UINT_TYPE length, bool noLength = false); //!< serialize data buffer
```
timcanham: Make old calls have DEPRECATED tag.

timcanham (changes requested):

Fw/Types/Serializable.hpp
LeStarch: Not going to do this.  To much would be forced to change....as deprecated is an error under `-Werror`.

bocchino (approved):
Looks good! Just one comment.

timcanham (approved):
Looks good, modulo Rob's most recent email concerning type names, but that can be done in another PR.

bocchino:
I think the type names agree with what we want (FwSizeStoreType). This looks good to me!

LeStarch:
CI passed, hanging on artifact-uploade.  I will merge now.

PR #2509, opened 2024-01-31
Update `ActiveTextLogger` naming to match build system naming expectations

kbotteon:
| | |
|:---|:---|
|**_Related Issue(s)_**|  
|**_Has Unit Tests (y/n)_**| Yes
|**_Documentation Included (y/n)_**|  Yes

---
## Change Description

This PR updates file names and `CMakeLists` for `ActiveTextLogger` to conform with what I believe the build system requires. See _Rationale_ below. 

**I could also be completely overlooking a pattern or small detail, so please let me know if this change is unnecessary!** I can't imagine I'm the first person to run into this problem with such a common Component, so I suspect another issue in my configuration might be involved.

## Rationale

Including `ActiveTextLogger` in a topology leads to a compiler error in an auto-coded file:

```
TopologyAc.hpp:16:10: fatal error: Svc/ActiveTextLogger/ActiveTextLogger.hpp: No such file or directory
 #include "Svc/ActiveTextLogger/ActiveTextLogger.hpp"
```

If I change the fpp to use `ActiveTextLoggerImpl`, now the build fails at `fpp-to-cpp`:

```
Declarations.fpp:79.23
instance logText: Svc.ActiveTextLoggerImpl \
                      ^
error: undefined symbol ActiveTextLoggerImpl
```

I think the latest build system expects naming like `Svc/<ComponentName>/<ComponentName>` without the `ComponentImpl` class or `Impl` file suffixes.

My topology including `ActiveTextLogger` now compiles. `fprime-util check –all` runs with `100% tests passed, 0 tests failed out of 54`.

### DIFF
--- a/Svc/ActiveTextLogger/ActiveTextLogger.cpp
+++ b/Svc/ActiveTextLogger/ActiveTextLogger.cpp
@@ -3,7 +3,7 @@
 // ALL RIGHTS RESERVED.  United States Government Sponsorship
 // acknowledged.
 
-#include <Svc/ActiveTextLogger/ActiveTextLoggerImpl.hpp>
+#include <Svc/ActiveTextLogger/ActiveTextLogger.hpp>
 #include <Fw/Types/Assert.hpp>
 #include <Fw/Logger/Logger.hpp>
 #include <ctime>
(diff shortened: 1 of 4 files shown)


thomas-bc (approved):

LeStarch (approved):

LeStarch:
@kbotteon we previously added headers that `typedef`ed the `Impl` version of the class to the non-impl version.  These headers were named with the new preferred name.

However, renaming is a *much* better fix and since you are doing a single component....I wholeheartedly approve!

Issue #2492, opened 2024-01-18
Resolve Shadowed `public` Members


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

With the merge of #2482, shadowed public members were renamed with the `a_` prefix for "argument".  We should review these uses and decide how to proceed.  

## Rationale

Public members are discouraged but have a purpose. `a_` is not a long-term solution either, but effectively resolves the issue and makes it easy to find these problems for discussion.

JohanBertrand:
For reference, we had some discussions about this point in a FPP issue: https://github.com/fprime-community/fpp/issues/326

PR #2513, opened 2024-02-01
Removed a_ prefix on arguments

JohanBertrand:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2492 |
|**_Documentation Included (y/n)_**| yes |

---
## Change Description

Removed the prefix `_a` previously added in https://github.com/nasa/fprime/pull/2482

Changes:
- Changed public variables of `Fw/FilePacket` and `Svc/FileDownlink` to `PRIVATE` variables

- Changed public variable `name` to private variable `m_name` in `STest/STest/Rule/Rule.hpp`

- Remove object initializion using public variable list in `Svc/FileDownlink/FileDownlink.cpp` and `Fw/FilePacket/StartPacket.cpp`

- Added getters for variables new private variables
The getters have been implemented in the header. Sometimes using references, sometimes copying values.
If it is not standard, I am happy to update it.

- Fixed UT

## Rationale

The `a_` prefix should not be used for arguments

I am not sure if there are some breaking changes for other projects

### DIFF
--- a/STest/STest/Rule/Rule.hpp
+++ b/STest/STest/Rule/Rule.hpp
@@ -61,7 +61,7 @@ namespace STest {
       ) = 0;
 
       //! Get rule name
-      char const * getRuleNamePointer() const {
+      char const * getRuleName() const {
         return this->m_name;
       }
 
--- a/STest/STest/Rule/Rule.hpp
+++ b/STest/STest/Rule/Rule.hpp
@@ -27,9 +27,9 @@ namespace STest {
 
       //! Construct object Rule
       Rule(
-          const char *const a_name //!< The name of the rule
+          const char *const name //!< The name of the rule
       ) :
-        name(a_name)
+        m_name(name)
       {
 
       }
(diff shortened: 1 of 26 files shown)


STest/STest/Rule/Rule.hpp
```diff
@@ -60,6 +60,11 @@ namespace STest {
           const State& state //!< The system state
       ) = 0;
 
+      //! Get rule name
+      char const * getRuleNamePointer() const {
```
bocchino: This seems OK, but why not call the function `getName`?

JohanBertrand: That makes more sense indeed, thanks!

I also removed "Pointer" from the other getters I created

LeStarch (commented):
One question: @bocchino many functions where implemented in the header as opposed to the CPP.  Does this match the C++ style we use?

bocchino:
Function implementations in the header file are fine if they are short functions or they are part of a template.

LeStarch (approved):
I am happy given @bocchino's response.  @bocchino are you ready to merge this?

thomas-bc (approved):

bocchino (approved):
Looks good!

PR #2542, opened 2024-02-22
Address comments from #2485

### LINKED ISSUE #2530: Address Follow-Up to #2485
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

Address @timcanham's comments on: #2485.

bocchino:
Closes #2530.


LeStarch (approved):
I see the requested changes and tests there-in.  Code looks good.  QL checks all seem to be false-positives.

Fw/Dp/DpContainer.cpp
```diff
@@ -127,42 +127,110 @@ void DpContainer::serializeHeader() {
...
+    const FwSizeType bufferSize = this->m_buffer.getSize();
+    FW_ASSERT(bufferSize >= Header::SIZE, static_cast<FwAssertArgType>(bufferSize),
+              static_cast<FwAssertArgType>(Header::SIZE));
+    U8* const buffAddr = this->m_buffer.getData();
+    Utils::HashBuffer computedHash;
+    Utils::Hash::hash(buffAddr, Header::SIZE, computedHash);
+    return computedHash;
+}
+
+void DpContainer::setHeaderHash(Utils::HashBuffer hash) {
```
timcanham: Why not pass by reference?

Fw/Dp/DpContainer.cpp
```diff
@@ -127,42 +127,110 @@ void DpContainer::serializeHeader() {
...
+    const FwSizeType bufferSize = this->m_buffer.getSize();
+    const FwSizeType minBufferSize = HEADER_HASH_OFFSET + HASH_DIGEST_LENGTH;
+    FW_ASSERT(bufferSize >= minBufferSize, static_cast<FwAssertArgType>(bufferSize),
+              static_cast<FwAssertArgType>(minBufferSize));
     U8* const buffAddr = this->m_buffer.getData();
-    Utils::Hash::hash(buffAddr, Header::SIZE, hashBuffer);
     ExternalSerializeBuffer serialBuffer(&buffAddr[HEADER_HASH_OFFSET], HASH_DIGEST_LENGTH);
```
timcanham: Seems a little heavyweight. Could we just `memcpy()` it instead of creating an `ExternalSerializeBuffer` instance?

timcanham (changes requested):
Just a couple more comments...

Fw/Dp/DpContainer.cpp
bocchino: We could pass this one by const reference, but it's small, and I think it's a good idea to make the copy, since reading a serializable object updates the object.

Fw/Dp/DpContainer.cpp
bocchino: Yes, good idea.

Fw/Dp/DpContainer.cpp
bocchino: But if I make the change in the next comment, then we don't need the copy, because there's no destructive read.

bocchino: Fixed

Fw/Dp/DpContainer.cpp
bocchino: Fixed

timcanham (approved):
Good changes!

LeStarch:
@bocchino I think this is ready.  Would you triage the CI annotations?

bocchino:
Done.

bocchino:
There's one more fix I'd like to do on this branch before merging it. That's to merge in devel and replace `FwSizeType` with `FwSizeStoreType` in the data product header.

bocchino:
The last commit is for CI checking. It needs one more change before it's ready for merge.

bocchino:
I reverted the temporary change.

bocchino:
My changes are done, once CI passes.

bocchino:
With the change in FwSizeType to I64, I noticed some new compiler warnings in the FPP compilation checks, which have `-Wconversion` enabled. It looks like we are not consistent in representing size values. Many times we use U32 as the size type, e.g., in the size of a hashed value or an Fw::Buffer. We should improve the consistency. In the mean time, I added casts to suppress the warnings.

LeStarch (approved):

PR #2546, opened 2024-02-27
Add DpManager

bocchino:
This pull request adds the `DpManager` component to support handling of data products. `DpManager` is one of the trio of core data products components. The other two are `DpWriter` and `DpCatalog`.

**Predecessor:** #2485
**Next up:** Add `DpWriter`

For the overall data products design, see https://github.com/nasa/fprime/blob/devel/docs/Design/data-products.md.


timcanham (approved):
LGTM! Great work!

PR #2548, opened 2024-02-28
Remove FPP dependencies on native int types

bocchino:
## Change Description

This PR makes changes necessary to remove the dependencies of the FPP model and generated C++ on the native int types `NATIVE_INT_TYPE` and `NATIVE_UINT_TYPE`.

Incidental changes:

* I reformatted a few files via clang-format. We want to do this reformatting anyway on a larger scale. Reformatting is not the primary purpose of this PR, but the reformatting helped me understand the code. Since reformatting obscures change/delete info, I reformatted only hpp files, so you can still see the change/delete info in the corresponding cpp files.

* I added some comments in `FpConfig.h` explaining what the configurations do.

* I renamed `FwBuffSizeType` to `FwSizeStoreType` as discussed. I added a type alias for `FwBuffSizeType` for backwards compatibility.

## Rationale

This is a first step in removing `NATIVE_INT_TYPE` and `NATIVE_UINT_TYPE` from the code base.

## Next Steps

1. Make an alpha release of FPP that eliminates references to `NATIVE_INT_TYPE` and `NATIVE_UINT_TYPE`. At this time we can replace `NATIVE_INT_TYPE` with `FwIndexType` in the `portNum` arguments of the port handler implementations.
2. Scrub `NATIVE_INT_TYPE` and `NATIVE_UINT_TYPE` from the code in this repo.
3. In `FpConfig.h`, replace configurable types except for PlatformInt/Uint with fixed-width types, to eliminate static analysis warnings.

## Related Issues

https://github.com/fprime-community/fpp/issues/391


bocchino:
Note: There are many false-positive warnings due to (1) problems in F Prime that this PR is addressing or (2) problems with the static analysis. Dismissing these warnings one-by-one in the GUI is too painful. I gave up. (You have to click dismiss, wait a few seconds until it resets to the top of the changed file list, and scroll back manually to where you were.)

bocchino:
> There are many false-positive warnings due to (1) problems in F Prime that this PR is addressing or (2) problems with the static analysis. 

It turns out that the static analysis does not work as expected with regard to fixed-width types. It seems that when you declare a variable
```
Foo bar;
```
It wants `Foo` to be a literal fixed-width numeric type like `int32_t` or maybe `I32`. It won't accept a general type alias, even if the aliased type has a fixed width. This seems like a bug in the static analysis.

LeStarch (changes requested):
1. Review need for pointer cast in model.  If needed, switch to PlatformPointerCase
2. Review type event counter derived types

LeStarch:
Make them all enum store type, but keep queue size as it is.

bocchino:
The event counters are for counting events emitted at runtime, for purposes of throttling. So I think they should be `FwIndexType`.

bocchino:
With the change of `FwSizeType` to `I64`, there are a lot of implicit casts from `I64` to `NATIVE_UINT_TYPE` and/or `U32`. I for this reason, I had to disable `-Wconversion` in the FPP compilation checks (too many warnings). This issue will be fixed in due course as we replace `NATIVE_UINT_TYPE` and `U32` with `FwSizeType`, e.g., in the `Fw::Buffer` interface.

bocchino:
@LeStarch I made the requested changes. I also had to introduce a type alias into `Fw::Serializable`, to resolve the discrepancy between the declared types `NATIVE_UINT_TYPE` and the eventual types `FwSizeType`. We'll have FPP refer to the alias and change the alias later. That way we decouple the FPP changes from the F Prime changes.

LeStarch (approved):
I see the expected changes fixed!

PR #2565, opened 2024-03-05
Force pip to upgrade during setup action

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

I don't expect it to actually solve anything... but you never know 🤷


LeStarch (approved):
But why?

PR #2574, opened 2024-03-07
Update dictionary JSON spec field names and examples

jwest115:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

- move arrays, structs, and enums into `typeDefinitions`
- rename `formatSpecifier` to `format`
- rename `identifier` to `name`
- shorten `identifier` to `id`
- add bool size

### DIFF
--- a/docs/Design/fpp-json-dict.md
+++ b/docs/Design/fpp-json-dict.md
@@ -561,7 +561,7 @@ telemetry Channel1: F64 \
         "name": "c1.Channel1",
         "description": "Telemetry channel 1",
         "type": {
-            "type": "F64",
+            "name": "F64",
             "kind": "float",
             "size": 64
         },
@@ -952,7 +952,7 @@ product container Container2 default priority 10
             "name": "c1.Channel1",
             "description": "Telemetry channel 1",
             "type": {
-                "type": "F64",
+                "name": "F64",
                 "kind": "float",
                 "size": 64
             },
--- a/docs/Design/fpp-json-dict.md
+++ b/docs/Design/fpp-json-dict.md
@@ -38,8 +38,8 @@ This document describes the format of FPP JSON dictionaries.
   - [Telemetry Channels](#telemetry-channels)
   - [Events](#events)
 - [Data Products](#data-products)
-  - [Record](#record)
-  - [Container](#container)
+  - [Records](#records)
+  - [Containers](#containers)
 - [Dictionaries](#dictionaries)
   - [Dictionary Metadata](#dictionary-metadata)
   - [Dictionary Content](#dictionary-content)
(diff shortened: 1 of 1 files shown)


LeStarch:
@bocchino you'll need to review this one!

docs/Design/fpp-json-dict.md
```diff
@@ -793,40 +751,42 @@ module MyModule {
...
 | `libraryVersions` | **Array of Strings** corresponding to the version (semantic versioning) of libraries used by the F´ project | **Array of Strings** | true
-
+| `dictionarySpecVersion` | **String** representing the JSON dictionary specification version | **String** | true |
 ```json
 {
     "deploymentName": "MyDeployment",
     "frameworkVersion": "3.3.2",
     "projectVersion": "1.0.0",
-    "libraryVersions": []
+    "libraryVersions": [],
```
thomas-bc, line 760: As I'm wiring fpp-to-dict in to the CMake system, figuring out how to specify `libraryVersions` is giving me some troubles. I think providing an example would be helpful. Here are my questions:

1. What is the format we want to use? `libraryName=1.0.0`,  `libraryName@1.0.0` , `libraryName==1.0.0` etc... ? We probably need to add this to the spec.
2. Say the library name is very generic e.g. `arduino-i2c`, are we ok with it being impossible to figure out where the library came from unless we can look at the source project? What I'm hinting at here is we could have an optional field e.g. `remoteUrl` or something. I don't actually think it would be useful/beneficial, but I just wanted to double check with everyone.
3. How do we deal with unspecified versions? e.g. a user has created a library locally but it is not a git submodule and/or there is no version to be retrieved.

jwest115: This is a good point @thomas-bc. The format of entries in `libraryVersions` is not specified in the spec. Also, there is no check done in the fpp-to-dict tool to verify that the libraryVersions input conforms to a certain format.

docs/Design/fpp-json-dict.md
```diff
@@ -552,41 +537,35 @@ module MyComponents {
...
 ```
 
 ```json
 [
     {
-        "identifier": "MyModule.FirstComponent.Channel1",
+        "name": "c1.Channel1",
         "description": "Telemetry channel 1",
         "type": {
             "type": "F64",
```
thomas-bc: Is this `type` sub-dictionary here out of date? Seems that F64 should now be `name` and not `type`

thomas-bc (commented):
Just marking things as I notice them

docs/Design/fpp-json-dict.md
jwest115: Good catch, should be resolved now

bocchino (approved):
This PR looks good to me! It makes the changes we discussed.

LeStarch (approved):

PR #2593, opened 2024-03-13
Add DpWriter

bocchino:
This PR adds the `Svc/DpWriter` component and supporting unit tests. It also does the following:
* Add data product-specific configuration files to `config`.
* Add a `Fw::FileNameString` object for storing file names. The buffer size is configurable in FPP.
* Add more fine-grained control of the error status return values in the unit test Os::File stub.
* Add a method to `Utils::HashBuffer` for converting the first four bytes of the hash to a `U32` value. This is useful for error reporting.
* Do incidental reformatting of `Utils/HashBuffer.{hpp,cpp}`.
* Add data product-specific ports to `Svc/DpPorts`.
* Make a few incidental changes to `Svc/DpManager`.
* Fix spelling of header files included from `config`, as discussed in #2640.


bocchino:
Something is wrong with the dp-writer branch on GitHub. It's not picking up my changes.

Utils/Hash/HashBufferCommon.cpp
```diff
@@ -1,55 +1,70 @@
...
 
-    HashBuffer::~HashBuffer() {
-    }
+HashBuffer::HashBuffer() {}
 
-    HashBuffer::HashBuffer(const HashBuffer& other) : Fw::SerializeBufferBase() {
-        Fw::SerializeStatus stat = Fw::SerializeBufferBase::setBuff(other.m_bufferData,other.getBuffLength());
-        FW_ASSERT(Fw::FW_SERIALIZE_OK == stat,static_cast<NATIVE_INT_TYPE>(stat));
-    }
+HashBuffer::HashBuffer(const U8* args, NATIVE_UINT_TYPE size) {
```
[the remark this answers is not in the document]
bocchino: Fixed

Fw/Types/FileNameString.hpp
```diff
@@ -0,0 +1,36 @@
...
+    FileNameString& operator=(const FileNameString& other);  //!< assignment operator
+    FileNameString& operator=(const StringBase& other);      //!< other string assignment operator
+    FileNameString& operator=(const char* other);            //!< char* assignment operator
+    ~FileNameString();                                       //!< destructor
+
+    const char* toChar() const;            //!< gets char buffer
+    NATIVE_UINT_TYPE getCapacity() const;  //!< return buffer size
+
+  private:
+    char m_buf[STRING_SIZE];  //!< storage for string data
```
timcanham: Could we make `STRING_SIZE` a separate config variable like `FILE_STRING_SIZE`?

Fw/Types/FileNameString.cpp
```diff
@@ -0,0 +1,54 @@
...
+    (void)Fw::StringUtils::string_copy(this->m_buf, other.toChar(), sizeof(this->m_buf));
+    return *this;
+}
+
+FileNameString& FileNameString::operator=(const StringBase& other) {
+    if (this == &other) {
+        return *this;
+    }
+
+    (void)Fw::StringUtils::string_copy(this->m_buf, other.toChar(), sizeof(this->m_buf));
```
timcanham, line 36: Not sure if this is a general approach for strings, but should we use [length](https://github.com/nasa/fprime/blob/devel/Fw/Types/StringType.hpp#L28) of source string to avoid overcopying?

Svc/DpPorts/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+####
```
timcanham, line 1: Why wouldn't these go in `Fw/Dp`?

timcanham (changes requested):
Just a few comments...

Fw/Types/FileNameString.hpp
bocchino: `FileNameString::STRING_SIZE` is an alias for `FileNameStringSize`, which is defined in `config/AcConstants.fpp`.

Fw/Types/FileNameString.cpp
bocchino: We want to use the size of the destination buffer, to avoid overrunning the buffer. For example, a string of size 100 that's copied into a buffer of size 40 will be truncated.

Svc/DpPorts/CMakeLists.txt
bocchino: I followed the pattern of ports like `Svc::Cycle` that are used only in the Svc layer.

Fw/Types/FileNameString.cpp
timcanham: You're right on that part that we want to limit the max to be the local buffer size. How about a minimum of the source actual string size and the local storage size?

Svc/DpPorts/CMakeLists.txt
timcanham: The difference is that `Svc::Cycle` isn't in the FP spec, where DP things are. I think we want a pattern of FPP supported types being in `Fw`. Thoughts @LeStarch?

Fw/Types/FileNameString.hpp
timcanham: If you are making any changes, perhaps `FileNameString::STRING_SIZE` could be the array size to make the scope explicit?

Svc/DpPorts/CMakeLists.txt
bocchino: These ports aren't in the FPP spec either. They're in the F Prime data products design. I think Cycle is similar. It supports the Svc F Prime rate group components and their uses; these ports support the Svc data product components and their uses.

Fw/Types/FileNameString.cpp
bocchino: The string copy utility does do a minimum. It copies the actual length of the source string unless that length is greater than the buffer size, in which case it truncates the string to fit in the buffer size. This is a standard size-bounded copy operation. It's just like strncpy, except that it's safer.

Svc/DpPorts/CMakeLists.txt
bocchino: The only port that's currently in Svc/DpPorts is DpWritten, which is the port that DpWriter uses to communicate to DpCatalog. I think it belongs in Svc. You're right that the FPP special ports (DpGet, etc.) belong in Fw/Dp. That is where they are.

Fw/Types/FileNameString.hpp
bocchino: Sure, I can add that qualifier.

bocchino: Done.

timcanham (approved):

LeStarch (approved):

LeStarch:
@bocchino @timcanham ready for merge?

timcanham:
image

bocchino:
Ready! I made some changes to config include paths as discussed.

LeStarch (approved):

timcanham (approved):

LeStarch:
Wooooooooooooo...... :shipit:

Issue #2591, opened 2024-03-13
Integrate JSON Dictionary autocoding in the CMake system


thomas-bc:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

We have fpp-to-dict to generate a JSON dictionary. This issue is to hook it in the CMake system to generate the JSON dictionary on every build.

https://github.com/nasa/fprime/issues/2263

thomas-bc:
Few things are needed for this:
- https://github.com/jwest115/fpp/pull/1 adding Dictionary.json to the list of generated files yielded by the fpp utility
- https://github.com/nasa/fprime/pull/2598 using the changes in FPP from above

PR #2598, opened 2024-03-18
Add JSON dictionary generation to CMake system

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2591 |

---
## Change Description

Address https://github.com/nasa/fprime/issues/2591

Opening as draft to start get some feedback to improve upon


thomas-bc:
This branch requires these changes to FPP: https://github.com/jwest115/fpp/pull/1

thomas-bc:
Rebased to catch latest devel, including fpp v2.1.0a7 with fpp-to-dict

thomas-bc:
@LeStarch I finally managed to fix the bug on the RPI build by explicitly calling out the global `version` target as a dependency to the add_custom_command() for the dictionary... I have no clue why though, especially why it would only fail on that specific deployment and not others... 🤷‍♂️ 

This should be ready for review

cmake/target/version/generate_version_info.py
```diff
@@ -0,0 +1,116 @@
...
+        fid.write("#ifndef _VERSION_HPP_\n")
+        fid.write("#define _VERSION_HPP_\n")
+        fid.write("\n")
+        fid.write(f'static const char* FRAMEWORK_VERSION = "{framework_version}";\n')
+        fid.write(f'static const char* PROJECT_VERSION = "{project_version}";\n')
+        if len(get_library_versions()) > 0:
+            fid.write("static const char* LIBRARY_VERSIONS[] = {\n")
+            for lib, version in get_library_versions().items():
+                fid.write(f'    "{lib}@{version}",\n')
+            fid.write("};\n")
```
thomas-bc: I am not very happy with this being there only when `len(libraries) > 0` but I could not find a way to appease the compiler when LIBRARY_VERSIONS was unused and empty... let me know if you can think of something better to do

thomas-bc: Here's the error I would get
```
versions/version.hpp:10:20: error: zero-size array ‘LIBRARY_VERSIONS’
   10 | static const char* LIBRARY_VERSIONS[] = {
      |                    ^~~~~~~~~~~~~~~~
versions/version.hpp:10:20: error: ‘LIBRARY_VERSIONS’ defined but not used [-Werror=unused-variable]
```

cmake/target/version/generate_version_header.py
Shivaly-Reddy, line 1: What is the reason to replace generate_version_header.py with generate_version_info.py?

thomas-bc: I'm really just renaming to stay in sync with what the script is doing. It used to just generate a version header (hpp). It's now generating more than that, with a JSON file.

thomas-bc: Ohh I realize the diff shows it as removed, which is not ideal... What I did was rename the file.

cmake/target/version/generate_version_info.py
```diff
@@ -0,0 +1,116 @@
...
+            f"    This file has been autogenerated using [{os.path.basename(__file__)}].\n"
+        )
+        fid.write("    This file may be overwritten.\n")
+        fid.write("*/\n")
+        fid.write("#ifndef _VERSION_HPP_\n")
+        fid.write("#define _VERSION_HPP_\n")
+        fid.write("\n")
+        fid.write(f'static const char* FRAMEWORK_VERSION = "{framework_version}";\n')
+        fid.write(f'static const char* PROJECT_VERSION = "{project_version}";\n')
+        if len(get_library_versions()) > 0:
```
Shivaly-Reddy: If there are no libraries, perhaps add an else with dummy string for lib name and version that says no lib versions available?

thomas-bc: If you are ok parsing that in your Version component (maybe empty string is the easiest to parse?), that works for me!

Shivaly-Reddy: I think I will take the strings as-is and put them on an EVR. It's not like we use the values to determine anything. It could be something like {"LIB_NA"}@{"Version_NA"}. Unless @LeStarch has other ideas?

thomas-bc: The current structure has them being single strings, separated with `@`
```
LIBRARY_VERSIONS[] = {
      "lib1@v1",
      "lib2@v2"
}
```
I pushed the following version for when there is no libraries
```
LIBRARY_VERSIONS[] = { "" }
```
but am happy to switch if we think there is a better approach.

cmake/target/version/generate_version_info.py
```diff
@@ -0,0 +1,121 @@
...
+        fid.write("    This file may be overwritten.\n")
+        fid.write("*/\n")
+        fid.write("#ifndef _VERSION_HPP_\n")
+        fid.write("#define _VERSION_HPP_\n")
+        fid.write("\n")
+        fid.write(f'static const char* FRAMEWORK_VERSION = "{framework_version}";\n')
+        fid.write(f'static const char* PROJECT_VERSION = "{project_version}";\n')
+        lib_versions = get_library_versions()
+        # TODO: LIBRARY_VERSIONS is unused for now - remove the unused attribute when used in Version component
+        fid.write("static const char* LIBRARY_VERSIONS[] __attribute__((unused)) = {")
```
thomas-bc: `__attribute__((unused))` is kind of icky, but I don't know what of a better option. This will need to be removed once the Version component uses it.

cmake/target/version/generate_version_info.py
LeStarch: Since we cannot have a 0-length array, I think this is fine.

LeStarch: Scratch that.  Can we do:

LIBRARY_VERSIONS[] = { nullptr }

instead?  Make the check easier.  See below.

cmake/target/version/generate_version_info.py
LeStarch: We should not use `__attribute__((unused)` as these kinds of things tend to be compiler specific.  We can "use" it by putting in the following:

if (LIBRARY_VERSIONS[0] != nullptr ) {
   // Telemeter library versions.

}

cmake/target/version/generate_version_info.py
thomas-bc: Done

cmake/target/version/generate_version_info.py
thomas-bc: Done

LeStarch:
@thomas-bc I bet this is because RPI does not use SystemResources and as such there is no dependence on `version` for the deployment.....except in the install step.

Svc/SystemResources/SystemResources.cpp
```diff
@@ -152,10 +152,10 @@
 }
 
 void SystemResources::Version() {
-    Fw::TlmString version_string(FRAMEWORK_VERSION);
+    Fw::TlmString version_string(Project::Version::FRAMEWORK_VERSION);
     this->tlmWrite_FRAMEWORK_VERSION(version_string);
 
-    version_string= PROJECT_VERSION;
+    version_string= Project::Version::PROJECT_VERSION;
```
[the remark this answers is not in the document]
thomas-bc: huh?

cmake/required.cmake
```diff
@@ -35,4 +35,5 @@ message(STATUS "[python3] python3 found at: ${PYTHON}")
 message(STATUS "[fpp-tools] fpp-depend found at: ${FPP_DEPEND}")
```
LeStarch, line 35: We should probably at some point make these debug messages, and just print the FPP tools path .

Autocoders/Python/src/fprime_ac/utils/version.py
```diff
@@ -1,4 +1,5 @@
 """ fprime version handling and reporting """
```
LeStarch, line 1: Should we take this opportunity to change this path?  We will (hopefully) blow away the AC package soon, so this file will need to find somewhere else to live.

LeStarch: Later.

LeStarch (approved):
There are several things of note:

1. We need a plan for supporting sub-topologies.  Not needed now, but someday.
2. We should switch the CACHE variable used to hold the current dictionary to be a property of the module that creates it.  I do not recommend doing this now, but when we tackle point 1.
3. We need to move version.py at some point.

Autocoders/Python/src/fprime_ac/utils/version.py
thomas-bc: I considered it but also noticed that some other part of the `Autocoders/` package use that module. Didn't want to break those dependencies or duplicate the code so I left it there. I'm ok breaking things if that's what we want to do.

PR #2619, opened 2024-03-26
Split CMake tests and run with minimum cmake

### LINKED ISSUE #2618: Run CMake UT CI On Minimum Required CMake Version
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

When running. the CMake UTs, we should run them on the minimum CMake version to ensure that version isn't inadvertently broken.

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Runs CMake UTs in a separate workflow with a specific minimum CMake version.

## Rationale

F´ must work with the minimum supported version.

Let CI run!

## Future Work

None

### DIFF
--- a/.github/workflows/cmake-test.yml
+++ b/.github/workflows/cmake-test.yml
@@ -34,7 +34,7 @@ jobs:
           export CMAKE_TAR_FILE="https://cmake.org/files/v3.16/cmake-3.16.0-Linux-x86_64.tar.gz"
           export CMAKE_INSTALL_DIRECTORY="${GITHUB_WORKSPACE}/tools-override"
           mkdir -p "${GITHUB_WORKSPACE}/tools-override"
-          curl -Ls "${CMAKE_TAR_FILE}" | tar -JC "${CMAKE_INSTALL_DIRECTORY}"  --strip-components=1 -x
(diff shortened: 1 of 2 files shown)


thomas-bc (approved):

.github/workflows/cmake-test.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          fetch-depth: 0
+          submodules: true
+      - uses: ./.github/actions/setup
+      # Specifically install CMake minimum version
+      - name: Minimum CMake Install
+        shell: bash
+        run: |
+          export CMAKE_TAR_FILE="https://cmake.org/files/v3.16/cmake-3.16.0-Linux-x86_64.tar.gz"
+          export CMAKE_INSTALL_DIRECTORY="${GITHUB_WORKSPACE}/tools-override"
```
thomas-bc, line 35: FYI it's possible to set those env var that you want to share across steps at the job / workflow level https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow
Can make things easier in larger workflows. But this should do too, unless you want to change rn.

LeStarch: I am not going to stress on this now.

PR #2620, opened 2024-03-26
Disambiguate fpp error messages. Fixes: #2618

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|  |
|**_Documentation Included (y/n)_**|  |

---
## Change Description

Cleans-up bad FPP installation errors:
1. Missing Java now reports "missing java"
2. Corrupt now recommends reinstall
3. Missing recommends install

## Rationale

Conflated error messages deconflated.

## Testing/Review Recommendations

Tested the following:

1. Removed `fpp-depend`: missing message reported
2. Invalid binary `fpp-fepend`: invalid install reported, recommend reinstall
3. Installed java wheel w/o Java: requires java reported

## Future Work

Update install guide.

PR #2621, opened 2024-03-26
Add Java requirement to install guide

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix install guide to include Java for non-x86_64/aarch64 systems.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -27,9 +27,10 @@ Requirements:
 3. [CMake 3.16](https://cmake.org/download/) or newer. CLI tool must be available on the system path.
 4. CLang or GNU C and C++ compilers (e.g. gcc and g++)
 5. [Python 3.8+](https://www.python.org/downloads/), virtual environments, and PIP
-6. Java for build machine architectures other than x86_64 and aarch64
 
-> Latest versions of PIP are strongly recommended, as this will install native packages with improved performance and no dependency on Java. See [Recommended PIP Versions](#recommended-pip-versions)
+> Latest versions of PIP are strongly recommended. See [Recommended PIP Versions](#recommended-pip-versions)
+
+> For build host architectures other than x86_64 or aarch64, and systems with older PIP versions, Java is required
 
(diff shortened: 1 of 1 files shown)


docs/INSTALL.md
```diff
@@ -27,6 +27,7 @@ Requirements:
 3. [CMake 3.16](https://cmake.org/download/) or newer. CLI tool must be available on the system path.
 4. CLang or GNU C and C++ compilers (e.g. gcc and g++)
 5. [Python 3.8+](https://www.python.org/downloads/), virtual environments, and PIP
+6. Java for build machine architectures other than x86_64 and aarch64
```
thomas-bc: Any way we can format this to be less scary at first glance?
e.g. 

6. For build host architectures other than x86_64 aarch64, Java is required to run the FPP tools

Or (better in my opinion) make it a note 

> Note: For build host architectures other than x86_64 aarch64, Java is required to run the FPP tools

Or anything really. Just trying not to make a 10second-read of this page mislead people in believing Java is a requirement for F´

Joshua-Anderson: Practically speaking, is anybody actually trying to build fprime on something other than x86_64/aarch64 systems?
Maybe an old 32-bit raspberry pi OS? Can't imagine too many people are using PPC or RISC.

I might just say that fprime only supports development on  x86_64/aarch64 systems.

thomas-bc: @Joshua-Anderson yeah I agree with you, that's why I wouldn't really want to have a big bold "JAVA" requirement in the install guide. 
But we did have one customer that was trying to build on something else, and not having Java installed caused an issue which wasn't as quickly identifiable as it should have been (corrected in #2620). Since we do have a better error handling now, I'm hesitant to add it as an entry in the system requirements as well.

LeStarch: Java is required for RISC-V.  The current recommendations for cross-compiling RISC-V is to compile in an emulated RISC-V machine.  I like @thomas-bc's suggestion of a note.  I will make that change.

thomas-bc (approved):

PR #2626, opened 2024-03-27
feat(ci): migrate to uv for faster python deps installation

SauravMaheshkar:
## Change Description

This PR aims to migrate the python dependency installation to [`uv`](https://github.com/astral-sh/uv) from `pip`. [uv](https://github.com/astral-sh/uv) backed by [Astral](https://astral.sh/), the creators of [Ruff](https://github.com/astral-sh/ruff), is an extremely fast Python package installer and resolver, written in Rust. According to their open-source benchmarks they are [10-100x faster](https://github.com/astral-sh/uv/blob/main/BENCHMARKS.md) than pip and pip-tools (pip-compile and pip-sync).

## Rationale

This change will enable faster CI runs and therefore faster feedback to streamline the contribution workflow.

uv has been adopted by popular repositories such as [diffusers](https://github.com/huggingface/diffusers) with [observed performance boost](https://x.com/RisingSayak/status/1763037626372632884?s=20).

## Future Work

There are some other workflow files which also use pip such as `python-format` but I avoided making those changes in this PR. If the maintainers find it useful other workflows using `pip` can be updated to use `pip`

### DIFF
--- a/.github/workflows/codeql-jpl-standard.yml
+++ b/.github/workflows/codeql-jpl-standard.yml
@@ -44,10 +44,9 @@ jobs:
 
     - name: Build
       run: |
-          python3 -m venv ./fprime-venv
-          . ./fprime-venv/bin/activate
-          pip install -U setuptools setuptools_scm wheel pip
-          pip install -r ./requirements.txt
+          python -m venv /opt/venv && export PATH="/opt/venv/bin:$PATH"
+          python -m uv pip install -U setuptools setuptools_scm wheel pip uv
+          python -m uv pip install -r ./requirements.txt
           fprime-util generate
           fprime-util build --all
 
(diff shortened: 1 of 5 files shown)


thomas-bc:
This is interesting, I didn't know about `uv`. We'll have to look into it, thanks for bringing it up and opening the PR!

LeStarch:
While we deeply appreciate this contribution, we are not ready for such a change.  Most of our users run `pip` to install F´  packages and our CI system reflect this.    This traces to a "test as you fly, fly as you test" methodology.  This means that our tests (in CI) must install using the same tools as our users.

Individuals and projects may of course choose to use `uv` to speed up their individual projects.

SauravMaheshkar:
> While we deeply appreciate this contribution, we are not ready for such a change. Most of our users run `pip` to install F´  packages and our CI system reflect this. This traces to a "test as you fly, fly as you test" methodology. This means that our tests (in CI) must install using the same tools as our users.
> 
> Individuals and projects may of course choose to use `uv` to speed up their individual projects.

Understandable, I appreciate the transparency.

PR #2655, opened 2024-04-01
Fixing Documentation Generation fo macOS

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes some issues with the generate_docs script.


thomas-bc (approved):
LGTM - I'm good merging as is, but since we're on the subject of CI.... 
Should we add this file (and maybe others) to the paths that trigger https://github.com/nasa/fprime/blob/devel/.github/workflows/autodocs.yml on PRs... thoughts?

```
  pull_request:
    paths:
      - 'docs/**.*sh'
```
or something of that nature

LeStarch:
I am not sure.  Autodocs runs frequently enough.

Issue #1713, opened 2022-10-10
Data Products: DpCatalog Prototype


LeStarch:
`DpCatalog` will do the following:

1\) Upon initialization, read the DP directory and build a catalog of products
2\) When commanded, start downloading based on priority and other TBD criterion (date? size?)
3\) Manage deletion of data products (commands or file list)
4\) Generate DP listings for download (file? self-defined DP?)

This was derived from:

_Originally posted by @timcanham in https://github.com/nasa/fprime/discussions/1154_

matt392code:
[dp-catalog-impl.py.txt](https://github.com/user-attachments/files/18304572/dp-catalog-impl.py.txt)
Proposed implementation:
1. Core Data Structures:
   - DpMetadata: Serializable structure storing essential file information
   - Internal catalog dictionary mapping filenames to metadata
   - Priority-based download queue
2. Main Component Operations:
   - Directory scanning and catalog building
   - Prioritized download management
   - File deletion handling
   - Status reporting and listing generation
3. F Prime Integration:
   - Uses F Prime's serializable types for command/telemetry interfaces
   - Implements command handlers for ground system control
   - Provides telemetry points for monitoring
4. Design Considerations:
   - Thread-safe operations for concurrent access
   - Efficient directory scanning using periodic updates
   - Flexible criteria-based download prioritization
   - Multiple output format support for listings
The implementation leaves certain aspects configurable:
- Priority calculation algorithms
- Directory scanning frequency
- Download concurrency limits
- Storage management policies

LeStarch:
I believe this was delivered.

PR #2667, opened 2024-04-08
Data Product Catalog prototype

timcanham:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/1713 |

---
## Change Description

This is a prototype demonstration of `Svc/DpCatalog` It was used to demonstrate an early release of data products in conjunction with [FPP updates](https://github.com/fprime-community/fpp/issues?q=is%3Aissue+is%3Aopen++%5BData+Products%5D) and [`DpManager/DpWriter`](https://github.com/nasa/fprime/pulls?q=is%3Apr+is%3Aclosed+data+product)

`Svc/DpCatalog` is not complete yet - this version is to capture the demo and allow users to play with it.

## Rationale

Get a consolidated set of data product features int he baseline.

The demo can be run and experimented with.

## Future Work

`Svc/DpCatalog` work will be continued with further releases.


timcanham:
Requested @bocchino for changes made to `Svc/DpWriter`.

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -0,0 +1,444 @@
...
+            FW_ASSERT(filesRead <= this->m_numDpSlots - totalFiles, filesRead, this->m_numDpSlots - totalFiles);
+
+            // extract metadata for each file
+            for (FwNativeUIntType file = 0; file < filesRead; file++) {
+                Fw::String fullFile;
+                fullFile.format("%s/%s",
+                    this->m_directories[dir].toChar(),
+                    this->m_fileList[file].toChar()
+                );
+                this->log_ACTIVITY_LO_ProcessingFile(fullFile);
```
Joshua-Anderson, line 161: Might be worth lowering the severity of this EVR to diagnostic, sending a event per data product file processed could add up.

Joshua-Anderson:
Is storing transmit state (UNTRANSMITTED/PARTIAL/TRANSMITTED) in the data product header the long term term plan?

It seems like this would force us to re-write data products once we start transmitting them, which could be a bit of a circular dependency. It also could get a bit confusing because the onboard DP file checksum would change after you start transmitting it.

Did we consider having DpCatalog own data product transmit state instead? (I.E. have a catalog file with a list of data products and their transmit states)

This could also provide a bit more flexibility because different projects may want to track data product transmit state differently (I.E. track not just that a data product is partial but how much of the file was sent, or which portions of the file need to be retransmitted).

timcanham:
@Joshua-Anderson The plan is to modify the header as we process the data products, i.e. `DpWriter` writes it as `UNTRANSMITTED`, then as `DpCatalog` starts downlinking it will change to `PARTIAL`, then `TRANSMITTED` when done. I'm not sure I see the circular dependency part, so elaborate on that for me.


You are correct that having a separate DP state file (or other variants on the idea) is an option and @bocchino and I discussed it, but I've always preferred files to be self-contained in their state so that the `DpCatalog` can be simpler in that it discovers state in the files and doesn't have to keep track of things in two places. Just a design preference.

config/DpCfg.hpp
```diff
@@ -17,6 +17,6 @@
 
 // The format string for a file name
 // The format arguments are container ID, time seconds, and time microseconds
```
bocchino: Let's add the new string field to the comment in line 19.

Svc/DpWriter/DpWriter.hpp
```diff
@@ -32,6 +32,11 @@ class DpWriter : public DpWriterComponentBase {
     //!
     ~DpWriter();
 
+    //! configure writer
+    void configure(
+            const Fw::String& dir //!< directory for file writing
```
bocchino: Maybe this should be Fw::FileNameString?

bocchino:
The changes to DpWriter look good to me! I just had a couple of comments.

config/DpCfg.hpp
timcanham: Added comment.

Svc/DpWriter/DpWriter.hpp
timcanham: Changed them to `Fw::FileNameString`

bocchino: Looks good. In my edits I changed the type in the configure prototype to `const Fw::StringBase&`. I think we should use this pattern going forward whenever the concrete string type is not needed.

config/DpCfg.hpp
bocchino: Looks good!

Ref/SignalGen/SignalGen.fpp
```diff
@@ -35,6 +35,18 @@ module Ref {
     # Special ports
     # ----------------------------------------------------------------------
 
+    @ Data product get port
+    product get port productGetOut
+
+    @ Data product request port
+    product request port productRequestOut
```
LeStarch, line 42: Why is there bothj get and request on this component and one is left in TODO state?

Ref/SignalGen/SignalGen.cpp
```diff
@@ -38,153 +38,222 @@ namespace Ref {
...
+    // Handler implementations for data products
+    // ----------------------------------------------------------------------
+
+    void SignalGen ::
+        dpRecv_DataContainer_handler(
+            DpContainer& container,
+            Fw::Success::T status
+        )
+    {
+        // TODO
```
LeStarch, line 255: This should be removed or demonstrated.

LeStarch (commented):
@bocchino would you mark a final review on this demo component?  Just clarify your comments were resolved.

Ref/SignalGen/SignalGen.cpp
timcanham: I was going to add that update to `SignalGen` after the demo was merged.

bocchino (approved):
My comments have been addressed.

Ref/SignalGen/SignalGen.cpp
timcanham: Hold off for a day or so - decided to implement the missing function.

LeStarch: @timcanham this was merged yesterday.

timcanham: Ah... Okay, I'll do another PR for the update.

PR #2676, opened 2024-04-12
FPP v2.1.0a7

bocchino:
This PR updates FPP to v2.1.0a7. This FPP version addresses issues fprime-community/fpp#397 and fprime-community/fpp#398 and adds JSON dictionary generation.

This PR also updates the F Prime unit tests to match the new FPP version.
* In `FppTest/dp`, check that the default priority is correctly set when getting a DP container. The test now checks that the code gen for default priorities is working.
* In `Svc/DpWriter/test/ut`, remove the default from-port handlers from the handwritten tester class. These handlers are now implemented in the auto-generated base class, so the handwritten code is redundant. Removing the handlers here proves that the base-class code is working.


requirements.txt
```diff
@@ -18,17 +18,17 @@ fprime-fpl-convert-xml==1.0.3
...
+fprime-fpp-depend==2.1.0a7
+fprime-fpp-filenames==2.1.0a7
+fprime-fpp-format==2.1.0a7
+fprime-fpp-from-xml==2.1.0a7
+fprime-fpp-locate-defs==2.1.0a7
+fprime-fpp-locate-uses==2.1.0a7
+fprime-fpp-syntax==2.1.0a7
+fprime-fpp-to-cpp==2.1.0a7
+fprime-fpp-to-json==2.1.0a7
+fprime-fpp-to-xml==2.1.0a7
```
thomas-bc, line 32: We should add a line for 
```
fprime-fpp-to-dict==2.1.0a7
```

It has been successfully released: https://pypi.org/project/fprime-fpp-to-dict/

thomas-bc (changes requested):

requirements.txt
LeStarch: Yes please do!

bocchino: Done!

thomas-bc (approved):

LeStarch (approved):

Issue #1984, opened 2023-04-25
Format C++ codebase


thomas-bc:
| | |
|:---|:---|
|**_F´ Version_**| `devel` |
|**_Affected Component_**| N/A |
---
## Feature Description

We are going to start formatting the entire C++ codebase and enforce the formatting through a CI check.

## Plan

- Create CI workflow to check that C++ is formatted correctly, with an optional exclude list. Start by excluding everything.
- Iteratively format the C++ codebase, removing the exclusion parameter as we go.

## Subtasks

- [x] Create clang-format check workflow https://github.com/nasa/fprime/pull/3778
- [x] Format Fw / Drv / Os / Svc
- [ ] Format other modules

JohanBertrand:
Is anyone working on that right now? I'm happy having a look at it

thomas-bc:
@JohanBertrand if you don't mind I'll take the CI part of this ticket, because we have an idea of how we want to do it. I'm working on this right now, I'll tag you in the PR if you'd like to review it

Once the check is there, feel free to help out formatting the code base!

We will format by small chunks and activate the CI per-module as we go.

thomas-bc:
One issue that is arising, and is loosely related to https://github.com/nasa/fprime-tools/pull/199

`clang-format`, for some obscure reason, has some slight variations in the formatting it outputs across different versions of the tool (even when using the exact same style). I have seen it happen with indentation of comments, and spacing like this one: https://github.com/thomas-bc/fprime/commit/a4585cde58301c0e201e797ea7a3ad9a7a090125

Some conversations about it [here](https://stackoverflow.com/questions/59395507/how-to-maintain-a-clang-format-file-for-different-clang-format-versions) and [here](https://stackoverflow.com/questions/58450914/unify-output-with-different-clang-format-versions)

We are likely going to need to pick a version of the tool we want to use in CI, version that users will need to use if they need to pass CI. It is very inconvenient... but there doesn't seem to be a way around it.

Let me know if you can think of anything else

cc @LeStarch @JohanBertrand

JohanBertrand:
I think it's reasonable to fix the version of the formatter, like it is also done with the static analyser (`clang-tidy-12`). Worst case scenario, the minor differences between two formatted versions would be flagged by the CI, and it could be resolved in one single commit.

It might be good to select a version of `clang-tidy` which is the same as `clang-format` to make it easier for the developers.

[...]

PR #2677, opened 2024-04-12
Format cpp and hpp files in Fw/Types

bocchino:
Reformatting improves the readability of the code. If we do it module by module, the impact should be minimal. `Fw/Types` seems like a logical place to start. 

If someone is developing `Fw/Types` with the old formatting, that person just needs to reformat the code being developed before diffing or making a PR against the new formatting. The catch is that because of the idiosyncratic use of the preprocessor symbols `PRIVATE` and `PROTECTED` in F Prime, you can't just run clang-format. Here is the script that I use to work around this issue:
```bash
#!/bin/sh -e

# fprime-format-cpp
# Use clang-format to format C++ files for F Prime

format()
{
  sed -e 's;^[ \t]*PROTECTED[ \t]*:;protected: // fprime-format-cpp;' -e 's;^[ \t]*PRIVATE[ \t]*:;private: // fprime-format-cpp;' | \
  clang-format --style=file | \
  sed -e 's;protected: *// fprime-format-cpp;PROTECTED:;' -e 's;private: *// fprime-format-cpp;PRIVATE:;'
}

if test $# -eq 0
then
  format
else
  for file in "$@"
  do
    cp "$file" "$file.bak"
    format < "$file.bak" > "$file"
  done
fi
```
I call this tool `fprime-format-cpp`. When run anywhere in the F Prime repo, it picks up the formatting configuration stored in `[fprime-root]/.clang-format`, so the formatting should be consistent for all users. It's designed so that you can run it inside a text editor (e.g., vi or VS Code) or in batch mode on the command line.

Here is how you can reformat all the files in a build module:
```bash
% fprime-format-cpp `find . -name '*.hpp' -or -name '*.cpp'`
```

I recommend that we put `fprime-format-cpp`, or something like it, into the repo so users can use it to format their code.

### DIFF
--- a/Fw/Types/Assert.cpp
+++ b/Fw/Types/Assert.cpp
@@ -13,272 +13,181 @@
 
 namespace Fw {
 
-    void defaultPrintAssert(const CHAR* msg) {
-        (void)fprintf(stderr,"%s\n", msg);
-    }
-
-    void defaultReportAssert
-            (
-            FILE_NAME_ARG file,
-            NATIVE_UINT_TYPE lineNo,
-            NATIVE_UINT_TYPE numArgs,
-            FwAssertArgType arg1,
-            FwAssertArgType arg2,
-            FwAssertArgType arg3,
-            FwAssertArgType arg4,
-            FwAssertArgType arg5,
-            FwAssertArgType arg6,
-            CHAR* destBuffer,
-            NATIVE_INT_TYPE buffSize
-            ) {
-
-        switch (numArgs) {
-            case 0:
-                (void) snprintf(destBuffer, buffSize, fileIdFs, file, lineNo);
-                break;
-            case 1:
-                (void) snprintf(
-                    destBuffer,
-                    buffSize,
-                    fileIdFs " %" PRI_FwAssertArgType,
-                    file,
-                    lineNo,
-                    arg1
-                );
-                break;
-            case 2:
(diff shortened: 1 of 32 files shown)


Joshua-Anderson:
This is awesome!

I wish we could find an alternative to our idiosyncratic use of PRIVATE/PUBLIC, but I think solving that is a longer term issue and shouldn't hold up code formatting. This script is a great workaround.

Only feedback is I'd highly recommend setting up CI to check formatting of `Fw/Types` to prevent regressions going forward. Otherwise by the time you finish applying formatting everywhere other code changes will probably have regressed formatting on previously formatted code.

thomas-bc:
Adding CI to check formatting as we start formatting is the plan, tracked by https://github.com/nasa/fprime/issues/1984

Also, note that we have `fprime-util format` which does the formatting with clang-format and the same mechanism you used in your bash script. It can take in a file list through 
```
fprime-util format -f Main.cpp Main.hpp
``` 
or with stdin 
```
find . | fprime-util format --stdin
```

bocchino:
> Also, note that we have fprime-util format which does the formatting with clang-format and the same mechanism you used in your bash script.

Great!

LeStarch (approved):

thomas-bc (approved):

LeStarch:
Merge it! Merge it!

PR #2679, opened 2024-04-13
Add Fw::ExternalString and revise string implementations

bocchino:
This PR does the following:
1. Add a new class `Fw::ExternalString`. This class is derived from `Fw::StringBase`. It uses an externally supplied character buffer to implement a string. I added tests for `Fw::ExternalString` to the `Fw/Types` unit tests.
2. Refactor the string implementations in the framework to reduce code duplication.

Rationale:
1. `Fw::ExternalString` is needed to improve the C++ code generation strategy for strings. See fprime-community/fpp#406.
2. The refactored implementation eliminates a significant amount of copy-paste code when creating a new concrete string type. The new version should be functionally identical to the old version.

Notes:

1. Review and merge #2677 first. A lot of changes in this PR are formatting changes that will go away once #2677 is merged.
2. Static analysis is telling us that single-argument constructors should be marked `explicit`. When I tried this, I saw compilation errors in the generated C++. I think this issue will have to be addressed separately.


bocchino:
Converting this PR to draft because I need to do some refactoring.

Fw/Cmd/CmdString.hpp
```diff
@@ -1,37 +1,59 @@
...
-#include <Fw/Cfg/SerIds.hpp>
+
+#include "Fw/Cfg/SerIds.hpp"
+#include "Fw/Types/StringBase.hpp"
 
 namespace Fw {
 
-    class CmdStringArg : public Fw::StringBase {
-        public:
+class CmdStringArg final : public StringBase {
```
LeStarch, line 17: Is this PR done?  I expected to review a bunch of changes to string types where they implement from `ExternalString` instead it seems that the code was just moved from the `.cpp` files to the headers....without inheriting implementation from external string

Fw/Cmd/CmdString.hpp
```diff
@@ -1,37 +1,59 @@
...
+        return *this;
+    }
+
+    const char* toChar() const { return this->m_buf; }
+
+    StringBase::SizeType getCapacity() const { return sizeof this->m_buf; }
 
-}
+  private:
+    char m_buf[CmdStringArg::STRING_SIZE];
```
LeStarch, line 55: While we are at it we should change `char` -> `CHAR`.

LeStarch (commented):
I did not see what I expected from this PR.  Awaiting more input.

Fw/Cmd/CmdString.hpp
bocchino: Yes, it's done. I decided not to have the implementations inherit from ExternalString. I realized nothing was gained by doing that, because in C++ operator= is not inherited. Instead I removed the C++ operator= implementations that were copy-pasting code from the superclass operator= and replaced them with direct calls to StringBase::operator=. The resulting code is much simpler and less error-prone and is functionally identical to what we had before -- in particular there's no overhead due to inheriting from ExternalString. If this isn't clear, let's discuss offline.

Fw/Cmd/CmdString.hpp
bocchino: What is the reason for using CHAR instead of char?

bocchino: In Fw/BasicTypes.h, I see this:
```
typedef char CHAR;
```
This isn't a configuration file, so CHAR is not a configurable type. Is it just for style, to make CHAR look like U32, etc.?

LeStarch (approved):

Fw/Cmd/CmdString.hpp
LeStarch: Let's file the above as a separate discussion topic.

thomas-bc (approved):

Issue #2553, opened 2024-02-29
Improve TcpClient, TcpServer, and Udp for Reconnections


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**| TcpClient, TcpServer |
---
## Feature Description

Update the TcpClient, TcpServer, Udp to allow reconnection on `send` as it does already on the recv loop.  Add in a `close` port to these components such that other components may close the connection.

- [ ] Add a mutex as member to components
- [ ] Expose mutex (or lock/unlock) through the `getHandler()` inteface
- [ ] Lock/unlock mutex on `recv` thread reconnect
- [ ] Add lock, reconnect, unlock to the send call
- [ ] Add a close port to each component that will call close

## Rationale

Allows better interaction with sockets.

Ahmad-Bamba:
I can look at this, if the enhancement is still wanted.

PR #2683, opened 2024-04-16
Add Open Port Request on Sending Side of IpSocket

csmith608:
| | |
|:---|:---|
|**_Related Issue(s)_**| #2553  |
|**_Documentation Included (y/n)_**|  y |

---
## Change Description

Update IpSocket to check socket state on sending side and open socket if necessary

## Rationale

This fixes a bug encountered while using IpSocket. The server would close the TCP connection and then the client would try to send and fail because previously IpSocket would only re-open a TCP connection when it received over the socket. 

In my project's version of fprime we diverged at fprime 3.0. I added implementations for the close() port in the TcpClient, TcpServer, and Udp as well as unit tests. This change also fixed the issue we were seeing with sending data, but the socket not being there. 
This does not include the unit tests or implementations for close() in the ByteStreamDriverModel because I wanted to at least get this part of the change in. I'll work on adding them. 
I also ended up commenting out line 85 of Drv/Ip/SocketReadTask.cpp because when I was running other testing without the TCP server, I was being spammed with that logger message. 

## Future Work

Adding implementations for close() to TcpClient, TcpServer, and Udp as well as unit tests. 

(Pipeline failed on "spammed" but I stand behind my spelling, but perhaps not that comment)


LeStarch:
Looks like it passed CI.  We'll need to correct spelling.  Your spelling is right, so we need to add the word to the expect list.

LeStarch:
The list is at: .github/actions/spelling/expect.txt. Just place "spammed" in there is mostly alphabetical order.   I can help this afternoon, but the above shows you how to should you get to it sooner.

LeStarch:
Fixed the spelling issue.  Will review now.

Drv/Ip/SocketReadTask.cpp
```diff
@@ -70,7 +70,22 @@ void SocketReadTask::readTask(void* pointer) {
     FW_ASSERT(pointer);
     SocketIpStatus status = SOCK_SUCCESS;
     SocketReadTask* self = reinterpret_cast<SocketReadTask*>(pointer);
+    bool disconnected;
     do {
+        // Open a network connection if it has not already been open
+        disconnected = false;
+        // Lock mutex to avoid competing opens from other threads
+        self->getSocketHandler().lockSocketMutex();
```
LeStarch: While this code block looks right, I believe it needs to be below (after start-up).  I think a start-up needs to be added to to send too (following the same pattern).

LeStarch: Line 98 of your file contains the original open that we should replace with this code

LeStarch (changes requested):
Well I feel dumb, these comments were left as pending for several weeks.

csmith608:
@LeStarch I added a unit test to TcpClient and TcpServer to show the socket auto opening, so I think this should be okay to merge (if it passes the checks).

TheSkyWalkerGo (approved):

TheSkyWalkerGo (approved):

csmith608:
Better way of doing this:

- Have an active component with one thread that processes events
- Events are send events that come from outside 
- Another thread is blocking on input that sends the receive events 
- If a send event is processed and the socket isn't open, it opens
- If on a receive the socket isn't open it opens 
- All mediated by a queue
- One thread is from the active TCP*Impl.cpp and the receive thread

csmith608:
@LeStarch other than the RPi known issue, I think this PR is ready for review

Drv/Ip/IpSocket.cpp
```diff
@@ -105,88 +103,45 @@ SocketIpStatus IpSocket::addressToIp4(const char* address, void* ip4) {
...
+SocketIpStatus IpSocket::open(NATIVE_INT_TYPE& fd) {
     SocketIpStatus status = SOCK_SUCCESS;
-    this->m_lock.lock();
-    FW_ASSERT(m_fd == -1 and not m_open); // Ensure we are not opening an opened socket
-    this->m_lock.unlock();
     // Open a TCP socket for incoming commands, and outgoing data if not using UDP
     status = this->openProtocol(fd);
     if (status != SOCK_SUCCESS) {
-        FW_ASSERT(m_fd == -1); // Ensure we properly kept closed on error
+        // TODO: move this to SocketComponentHelper?
```
LeStarch: Remove TODO?

Drv/Ip/IpSocket.hpp
```diff
@@ -198,27 +187,25 @@ class IpSocket {
...
      * \brief Protocol specific implementation of recv.  Called directly with error handling from recv.
+     * \param fd: file descriptor to recv from
      * \param data: data pointer to fill
      * \param size: size of data buffer
      * \return: size of data received, or -1 on error.
      */
-    virtual I32 recvProtocol( U8* const data, const U32 size) = 0;
+    virtual I32 recvProtocol(NATIVE_INT_TYPE fd, U8* const data, const U32 size) = 0;
 
-    Os::Mutex m_lock;
```
LeStarch, line 215: Wooooooo!  Success!

Drv/Ip/SocketComponentHelper.cpp
```diff
@@ -0,0 +1,205 @@
...
+    return is_open;
+}
+
+SocketIpStatus SocketComponentHelper::reconnect() {
+    SocketIpStatus status = SOCK_SUCCESS;
+
+    // Handle opening
+
+    // Open a network connection if it has not already been open
+    // TODO: improve this message
```
LeStarch: TODO....probably need to delete.

Drv/Ip/SocketComponentHelper.cpp
```diff
@@ -0,0 +1,205 @@
...
+}
+
+SocketIpStatus SocketComponentHelper::startup() {
+    this->m_started = true;
+    return this->getSocketHandler().startup();
+}
+
+bool SocketComponentHelper::isStarted() {
+    bool is_started = false;
+    this->m_lock.lock();
```
LeStarch: Why is this locked if the startup call above is not?

Drv/Ip/SocketComponentHelper.cpp
```diff
@@ -0,0 +1,205 @@
...
+    this->getSocketHandler().shutdown(this->m_fd);
+    this->m_started = false;
+    this->m_fd = -1;
+    this->m_lock.unLock();
+}
+
+void SocketComponentHelper::close() {
+    this->m_lock.lock();
+    if (this->m_fd != -1) {
```
LeStarch, line 124: Probably safe to leave this check out.....as it is allowed to close a `-1` socket.

Drv/Ip/SocketComponentHelper.cpp
```diff
@@ -0,0 +1,205 @@
...
+    }
+    status = this->getSocketHandler().recv(fd, data, size);
+    if (status == SOCK_DISCONNECTED) {
+        this->close();
+    }
+    return status;
+}
+
+void SocketComponentHelper::readTask(void* pointer) {
+    FW_ASSERT(pointer);
```
LeStarch, line 160: Given the change in behavior of this, we should either:

1. Start before the loop
2. Assert is_started before looping

LeStarch: This function is safe in that it will not break, but it won't notify the user if they failed to call start.

Drv/Ip/TcpServerSocket.cpp
```diff
@@ -13,6 +13,7 @@
 #include <Fw/Logger/Logger.hpp>
 #include <Fw/Types/Assert.hpp>
 #include <FpConfig.hpp>
+#include <iostream>
```
LeStarch: Remove this header.

Drv/Ip/UdpSocket.cpp
```diff
@@ -111,13 +105,9 @@ SocketIpStatus UdpSocket::bind(NATIVE_INT_TYPE fd) {
     if (::getsockname(fd, reinterpret_cast<struct sockaddr *>(&address), &size) == -1) {
         return SOCK_FAILED_TO_READ_BACK_PORT;
     }
-    U16 port = ntohs(address.sin_port);
```
LeStarch, line 114: Was this unused?

Drv/Ip/UdpSocket.cpp
```diff
@@ -111,13 +105,9 @@ SocketIpStatus UdpSocket::bind(NATIVE_INT_TYPE fd) {
...
 
-    this->m_lock.lock();
     FW_ASSERT(sizeof(this->m_state->m_addr_recv) == sizeof(address), sizeof(this->m_state->m_addr_recv), sizeof(address));
     memcpy(&this->m_state->m_addr_recv, &address, sizeof(this->m_state->m_addr_recv));
-    this->m_recv_port = port;
```
LeStarch, line 119: Why scrub this line?

Drv/Ip/test/ut/SocketTestHelper.cpp
```diff
@@ -40,27 +40,29 @@ void fill_random_data(U8 *data, U32 size) {
 
 void validate_random_buffer(Fw::Buffer &buffer, U8 *data) {
     validate_random_data(buffer.getData(), data, buffer.getSize());
-    buffer.setSize(0);
+    // buffer.setSize(0);
```
LeStarch: Delete or restore.

Drv/Ip/test/ut/SocketTestHelper.cpp
```diff
@@ -40,27 +40,29 @@ void fill_random_data(U8 *data, U32 size) {
...
-    EXPECT_EQ(sender.send(buffer_out, MAX_DRV_TEST_MESSAGE_SIZE), Drv::SOCK_SUCCESS);
-    EXPECT_EQ(receiver.recv(buffer_in, size), Drv::SOCK_SUCCESS);
+    EXPECT_EQ(sender.send(sender_fd, buffer_out, MAX_DRV_TEST_MESSAGE_SIZE), Drv::SOCK_SUCCESS);
+    EXPECT_EQ(receiver.recv(receiver_fd, buffer_in, size), Drv::SOCK_SUCCESS);
     EXPECT_EQ(size, static_cast<U32>(MAX_DRV_TEST_MESSAGE_SIZE));
     Drv::Test::validate_random_data(buffer_out, buffer_in, MAX_DRV_TEST_MESSAGE_SIZE);
 }
 
+/*
+This doesn't belong here anymore
```
LeStarch: Delete all "This doesn't belong here..."

Drv/Ip/test/ut/SocketTestHelper.cpp
```diff
@@ -70,7 +72,10 @@ bool wait_on_change(Drv::IpSocket &socket, bool open, U32 iterations) {
     }
     return false;
 }
+*/
 
+/*
+This doesn't belong here anymore
```
LeStarch: Here too.

Drv/TcpClient/TcpClientComponentImpl.cpp
```diff
@@ -35,6 +35,7 @@ SocketIpStatus TcpClientComponentImpl::configure(const char* hostname,
     // Check that ensures the configured buffer size fits within the limits fixed-width type, U32                                                
     FW_ASSERT(buffer_size <= std::numeric_limits<U32>::max(), static_cast<FwAssertArgType>(buffer_size));                                                   
     m_allocation_size = buffer_size; // Store the buffer size
+    (void)startup();
```
LeStarch, line 38: Is this needed in Tcp client?

Drv/TcpClient/test/ut/TcpClientTester.cpp
```diff
@@ -142,9 +157,12 @@ void TcpClientTester ::test_advanced_reconnect() {
   {
     this->pushFromPortEntry_recv(recvBuffer, recvStatus);
     // Make sure we can get to unblocking the spinner
-    EXPECT_EQ(m_data_buffer.getSize(), recvBuffer.getSize()) << "Invalid transmission size";
-    Drv::Test::validate_random_buffer(m_data_buffer, recvBuffer.getData());
-    m_spinner = true;
+    if(recvStatus == Drv::RecvStatus::RECV_OK){
```
LeStarch: Was this added to fix an issue, or patch one of the threading failures we had?

Drv/TcpServer/TcpServerComponentImpl.cpp
```diff
@@ -22,13 +22,14 @@ namespace Drv {
...
-      SocketReadTask() {}
+      SocketComponentHelper() {}
 
 SocketIpStatus TcpServerComponentImpl::configure(const char* hostname,
                                                  const U16 port,
                                                  const U32 send_timeout_seconds,
                                                  const U32 send_timeout_microseconds) {
-    return m_socket.configure(hostname, port, send_timeout_seconds, send_timeout_microseconds);
+    (void)m_socket.configure(hostname, port, send_timeout_seconds, send_timeout_microseconds);
+    return startup();
```
LeStarch, line 32: Auto-start!  Disregard my comment about the recv loop earlier.

Drv/TcpServer/test/ut/TcpServerTester.cpp
```diff
@@ -30,52 +30,50 @@ void TcpServerTester ::test_with_loop(U32 iterations, bool recv_thread) {
...
 
         // If all the opens worked, then run this
         if ((Drv::SOCK_SUCCESS == status1) && (Drv::SOCK_SUCCESS == status2) &&
-            (this->component.getSocketHandler().isOpened())) {
+            (this->component.isOpened())) {
             // Force the sockets not to hang, if at all possible
-            Drv::Test::force_recv_timeout(this->component.getSocketHandler());
-            Drv::Test::force_recv_timeout(client);
+
+            // TODO: is server.m_base_fd right?
```
LeStarch: TODO.

Drv/TcpServer/test/ut/TcpServerTester.cpp
```diff
@@ -84,23 +82,50 @@ void TcpServerTester ::test_with_loop(U32 iterations, bool recv_thread) {
...
+                //from_deallocate_handler(0, m_data_buffer);
                 while (not m_spinner) {}
             }
         }
-        client.close(); // Client must be closed first or the server risks binding to an existing address
+
         // Properly stop the client on the last iteration
         if ((1 + i) == iterations && recv_thread) {
-            this->component.shutdown();
+            // TODO: for some reason you have to stop the server receive thread before closing the client or the server will erroneously receive
```
LeStarch: TODO?  Is the server erroring, or is releasing the last empty packet?

Drv/TcpServer/test/ut/TcpServerTester.cpp
```diff
@@ -138,10 +163,12 @@ void TcpServerTester ::test_advanced_reconnect() {
 
 void TcpServerTester ::from_recv_handler(const NATIVE_INT_TYPE portNum, Fw::Buffer& recvBuffer, const RecvStatus& recvStatus) {
     this->pushFromPortEntry_recv(recvBuffer, recvStatus);
-    // Make sure we can get to unblocking the spinner
-    EXPECT_EQ(m_data_buffer.getSize(), recvBuffer.getSize()) << "Invalid transmission size";
-    Drv::Test::validate_random_buffer(m_data_buffer, recvBuffer.getData());
-    m_spinner = true;
+    if(recvStatus == Drv::RecvStatus::RECV_OK){
```
LeStarch: Same question.

LeStarch (changes requested):
I have lots of questions, but my guess is this is ready to merge once the dead code and TODOs are deleted!

Fantastic work!

Drv/Ip/SocketComponentHelper.cpp
LeStarch: Fixed at component level w/ autostart!

Drv/Ip/test/ut/SocketTestHelper.cpp
LeStarch: Probably restore....doesn't hurt/

Drv/Ip/SocketComponentHelper.cpp
csmith608: Good point (it's because I removed it in startup and then forgot to here :))

Drv/Ip/SocketComponentHelper.cpp
csmith608: When this close call was in IpSocket it had this protection so I brought it up to SocketComponentHelper. Still think we should remove?

Drv/TcpClient/test/ut/TcpClientTester.cpp
csmith608: Ah you're right, I should have realized the thread issue here. Removed those now. Ty ty

Drv/TcpServer/test/ut/TcpServerTester.cpp
csmith608: Thread issue :(

Drv/Ip/UdpSocket.cpp
csmith608: This and the previous question is a really good question. I think I was trying to get rid of the local and messed up. Thank you!

csmith608: I'm kind of not sure why this worked with this issue. Seems like m_recv_port wouldn't have been assigned

EDIT: oh because in configureRecv m_recv_port gets assigned to port

csmith608: So actually, why does m_recv_port need to get assigned in `configureRecv` and `bind`?

Because in `configureRecv` `m_recv_port = port` in theory in host byte order? Then we have a different variable `address.sin_port` in `bind` that stores the port it in network byte order. Then we convert `address.sin_port`(back?) to host byte order and reassign that to `m_recv_port`?

Drv/Ip/SocketComponentHelper.cpp
LeStarch: It simply does not matter.  We can leave as-is it to cut down on churn.

Drv/Ip/UdpSocket.cpp
LeStarch: You are right, it is assigned at the top, and these two statements are superfluous.

Drv/Ip/UdpSocket.cpp
LeStarch: Yes.  Removal is correct.

LeStarch (approved):

PR #2697, opened 2024-04-18
Fix string records for data products

bocchino:
This PR is a companion to nasa/fpp#408. It incorporates the fix for https://github.com/nasa/fpp/issues/392.

Note: Review and merge #2679 first.


requirements.txt
```diff
@@ -18,18 +18,17 @@ fprime-fpl-convert-xml==1.0.3
...
+fprime-fpp-depend==2.1.0a9
+fprime-fpp-filenames==2.1.0a9
+fprime-fpp-format==2.1.0a9
+fprime-fpp-from-xml==2.1.0a9
+fprime-fpp-locate-defs==2.1.0a9
+fprime-fpp-locate-uses==2.1.0a9
+fprime-fpp-syntax==2.1.0a9
+fprime-fpp-to-cpp==2.1.0a9
+fprime-fpp-to-json==2.1.0a9
+fprime-fpp-to-xml==2.1.0a9
```
thomas-bc, line 32: `fprime-fpp-to-dict` should be added back, it probably got removed because this branch is older than the one that introduced it

thomas-bc (changes requested):

requirements.txt
bocchino: Good catch! Fixed.

LeStarch (approved):

thomas-bc (approved):

PR #2700, opened 2024-04-24
code clean up for POSIX.

kevin-f-ortega:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Decoupling POSIX and Linux/Darwin for components to be used by other OS that is POSIX compliant.

## Rationale

Other POSIX OS could use these components

## Future Work


cmake/API.cmake
```diff
@@ -31,7 +31,10 @@ set(FPRIME_AUTOCODER_TARGET_LIST "" CACHE INTERNAL "FPRIME_AUTOCODER_TARGET_LIST
 #####
 macro(restrict_platforms)
     set(__CHECKER ${ARGN})
-    if (NOT FPRIME_TOOLCHAIN_NAME IN_LIST __CHECKER AND NOT FPRIME_PLATFORM IN_LIST __CHECKER)
+    if (FPRIME_TOOLCHAIN_NAME IN_LIST __CHECKER)
```
LeStarch, line 41: We might add a comment to explain all these empty if conditions.

Each of these empty if blocks are the valid-case, that is, the platform is supported.  However, the reason why this is necessary is that this function is a `macro` and not a `function`.  Macros copy-paste the code into the calling context.  Thus, all these valid cases want to avoid the `return` call on line 41.  The return call on line 41 returns from the calling context (i.e. a restricted `CMakeList.txt` will return and not process the component setup).  We do not want this return when the platform is allowed.

This pattern is great for writing a guard statement, but not so easy to understand.

LeStarch (changes requested):

cmake/API.cmake
kevin-f-ortega: Good point. I added a comment.

thomas-bc (approved):

LeStarch (approved):

PR #2704, opened 2024-04-28
Add contributing tips

JohanBertrand:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

I personally spent quite some time trying to figure out how to run tests properly on my computer, specially when fpp needed to be updated for the PR to pass.

I have added here my current workflow.

## Rationale

It is currently hard to figure out how to run all tests for new contributors.

My workflow with modified versions of FPP might not be the best currently. Feel free to update it with the recommended workflow.


CONTRIBUTING.md
```diff
@@ -150,3 +150,49 @@ tools this can increase the effort required to review a submission. Be careful w
...
+
+In case FPP needs to be locally changed, FPP first needs to be installed:
+
+```bash
+# Go into your FPP directory
+cd MY_FPP_DIRECTORY
+# Install FPP. Make sure to have sbt installed.
+./compiler/install
+# Copy
+cp ./compiler/bin/* /usr/local/bin/ -r
```
thomas-bc: I personally do
```
export PATH=<path/to/fpp>/compiler/bin:$PATH
```
Do you think it might be better to recommend that in the guide? Forgetting to uninstall could lead to very hard to debug issues

thomas-bc (commented):
Just one comment, but thank you for compiling this info!

CONTRIBUTING.md
LeStarch: You can also give the destination in the install line....and place it into your virtual environment.

JohanBertrand: I definitely think that there is a better way than mine.

I feel like, ideally, there should be a `setup.py` that we could use to install the fpp packages with `pip install .` after the command `compile/install` has been called. I know the `setup.py` file existed in the past (https://github.com/nasa/fpp/blob/v2.0.1/setup.py), but I'm not sure what was the reason why it was removed.

As you suggested, the `export PATH` way is already better than what I am currently doing.

I have to say that I am not using any virtual environment in my development routine right now, since I am having a dedicated docker image for fprime. However, what would be the command to add it to the virtual environment? 
Maybe both approaches can be suggested.

LeStarch: Do the instructions here work: https://github.com/nasa/fpp/blob/main/compiler/README.adoc?

JohanBertrand: Yes it works. I updated the readme with the link, and added some clang-tidy commands we discussed in other PR.

Let me know if there is something else to change

JohanBertrand:
@LeStarch @thomas-bc is there anything else to be done in this MR?

LeStarch (approved):

thomas-bc (approved):

LeStarch:
PI tests are passing, but failing to upload artifacts.  Will pass this!

PR #2705, opened 2024-04-28
Add log on TCP/UDP UT failure and fix UDP UT

JohanBertrand:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

- Add log on stdout when an error happens in the TCP and UDP UT.
- Fix the following case, where the UDP UT would select twice the same port and fail binding later:
```
[ RUN      ] Nominal.BasicReceiveThread
Setup to send and receive udp to 127.0.0.1:60296
UDP socket open error: Address already in use
Port1: 60296
Port2: 60296
/home/fprime/Drv/Udp/test/ut/UdpTester.cpp:77: Failure
Expected equality of these values:
  status2
    Which is: -9
  Drv::SOCK_SUCCESS
    Which is: 0
[  FAILED  ] Nominal.BasicReceiveThread (10 ms)
```
```
[ RUN      ] Reconnect.MultiMessaging
Setup to send and receive udp to 127.0.0.1:6335
UDP socket open error: Address already in use
Port1: 6335
Port2: 6335
/home/fprime/Drv/Udp/test/ut/UdpTester.cpp:77: Failure
Expected equality of these values:
  status2
    Which is: -9
  Drv::SOCK_SUCCESS
    Which is: 0
[  FAILED  ] Reconnect.MultiMessaging (0 ms)
```

## Rationale

TCP and UDP UT are often failing, and this will give more understanding of the reason of those failures.

### DIFF
--- a/Drv/TcpClient/test/ut/TcpClientTester.cpp
+++ b/Drv/TcpClient/test/ut/TcpClientTester.cpp
@@ -36,12 +36,10 @@ void TcpClientTester ::test_with_loop(U32 iterations, bool recv_thread) {
     server.configure("127.0.0.1", port, 0, 100);
     this->component.configure("127.0.0.1", port, 0, 100);
     serverStat = server.startup();
-    if (serverStat != SOCK_SUCCESS)
-    {
-        printf("TCP server startup error: %s\n", strerror(errno));
-        printf("Port: %u\n", port);
-    }
-    ASSERT_EQ(serverStat, SOCK_SUCCESS);
+
+    ASSERT_EQ(serverStat, SOCK_SUCCESS)
+        << "TCP server startup error: " << strerror(errno) << std::endl
+        << "Port: " << port << std::endl;
 
     // Start up a receive thread
     if (recv_thread) {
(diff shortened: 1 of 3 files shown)


Drv/TcpClient/test/ut/TcpClientTester.cpp
```diff
@@ -36,6 +36,11 @@ void TcpClientTester ::test_with_loop(U32 iterations, bool recv_thread) {
     server.configure("127.0.0.1", port, 0, 100);
     this->component.configure("127.0.0.1", port, 0, 100);
     serverStat = server.startup();
+    if (serverStat != SOCK_SUCCESS)
+    {
+        printf("TCP server startup error: %s\n", strerror(errno));
+        printf("Port: %u\n", port);
+    }
```
LeStarch: As a note, this could be done with the following on line 44:

```
ASSERT_EQ(serverStat, SOCK_SUCCESS) <<  "TCP server startup error:" << strerror(errno) << " Port:" << port;
```

LeStarch (changes requested):
I think it is cleaner to use the messaging built into GTest.

```
ASSERT...() << "Custom error message here"
```

As this allows GTest to colocate the error with the assertion failure.

Thoughts?

Drv/TcpClient/test/ut/TcpClientTester.cpp
JohanBertrand: I agree, I made the changes.

I can see that it is currently done both ways in the unit tests. I might try to make it uniform in the future.

LeStarch (approved):

PR #2703, opened 2024-04-27
Update to FPP v2.1.0a10

bocchino:
* Addresses issue nasa/fpp#407
* Cleans up string code and corrects off-by-one error in buffer sizes
* Refactors code in `Fw::StringBase` that was terse and hard to understand (it was not clear what was going on with serializing and deserializing string lengths; because so much was implicit, it looked like the code was serializing and deserializing null-terminated strings).

**Note:** This PR will require some minor changes in user test code, as discussed in nasa/fpp#409. The changes bring the code more into line with approved C++ standards, by eliminating reliance on implicit conversions. I've made the required changes for the F Prime unit tests in this PR.

PR #2707, opened 2024-04-29
Make single-argument string constructors explicit

bocchino:
This PR makes single-argument string constructors explicit, where possible with the current code generation. Explicit single-argument constructors are recommended (including by the static analysis in this repo) because they avoid unwanted implicit conversions.

Note the following:

* This PR incorporates the changes from #2703. Review and merge that one first.
* I revised the FSW and test code as necessary to eliminate implicit string conversions. In some cases they were doing implicit string copies that were not necessary.
* For parameter, event, and telemetry strings, and for the `char *` constructor of `Fw::String`, we can't make the single-argument constructors explicit until we close this issue: nasa/fpp#421.


Fw/Types/StringBase.cpp
```diff
@@ -130,16 +139,34 @@ SerializeStatus StringBase::serialize(SerializeBufferBase& buffer) const {
...
+    SerializeStatus stat = buffer.deserialize(reinterpret_cast<U8*>(raw), actualSize, Serialization::INCLUDE_LENGTH);
+    if (stat == FW_SERIALIZE_OK) {
+        // Deserialization succeeded: null-terminate string at actual size
+        FW_ASSERT(actualSize <= maxSize, static_cast<FwAssertArgType>(actualSize),
+                  static_cast<FwAssertArgType>(maxSize));
+        raw[actualSize] = 0;
+    } else {
+        // Deserialization failed: leave string unmodified, but ensure that it
+        // is null-terminated
+        raw[0] = maxSize;
```
LeStarch: Minor bug here.

LeStarch (approved):

Fw/Types/StringBase.cpp
LeStarch: Fixed inline.

Issue #2706, opened 2024-04-28
Failures of TCP/UDP unit tests


JohanBertrand:
| | |
|:---|:---|
|**_F´ Version_**| 6bb780d |
|**_Affected Component_**|   |
---
## Problem Description

Similar to https://github.com/nasa/fprime/issues/1290, https://github.com/nasa/fprime/issues/881

Maybe also related to https://github.com/nasa/fprime/issues/2244

I am not sure if it is having an impact outside of the UT.

The tests of `Drv/TcpServer` and  `Drv/TcpClient` are often failing, about 25% of the time (excluding silent failures).

It comes from several different errors:
- [Fixed in #2739] For TcpServer and TcpClient, most of the time, an error comes from the Address already being in use (Port already in use ?), based on `errno`. Also happens on the Github CI: https://github.com/nasa/fprime/actions/runs/8839296509/job/24272334136
```
[ RUN      ] Nominal.BasicMessaging
TCP server startup error: Address already in use
Port: 48514
/home/fprime/Drv/TcpServer/test/ut/TcpServerTester.cpp:49: Failure
Expected equality of these values:
  serverStat
    Which is: -9
  SOCK_SUCCESS
    Which is: 0
[  FAILED  ] Nominal.BasicMessaging (0 ms)
```
```
[ RUN      ] Reconnect.ReceiveThreadReconnect
TCP server startup error:Address already in use
Port: 33260
/home/fprime/Drv/TcpClient/test/ut/TcpClientTester.cpp:44: Failure
Expected equality of these values:
  serverStat
    Which is: -9
  SOCK_SUCCESS
    Which is: 0
[  FAILED  ] Reconnect.ReceiveThreadReconnect (0 ms)
```
- [Fixed in #2739] Sometimes TcpServer UT fail and can not start:
```

[ RUN      ] Nominal.BasicReceiveThread
[WARNING] Failed to open port with status -9 and errno 98
[WARNING] Failed to open port with status -9 and errno 98
/home/fprime/Drv/TcpServer/test/ut/TcpServerTester.cpp:41: Failure
Value of: Drv::Test::wait_on_started(this->component.getSocketHandler(), true, SOCKET_RETRY_INTERVAL_MS/10 + 1)
  Actual: false
Expected: true
/home/fprime/Drv/TcpServer/test/ut/TcpServerTester.cpp:51: Failure
Value of: component.getSocketHandler().isStarted()
  Actual: false
Expected: true
[...]

PR #2709, opened 2024-04-29
Fix failures of tcp UT

JohanBertrand:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2706 |

---
## Change Description

Fix issue where the TCP client/server would fail to connect to a socket, generating an error and making the UT to fail.

Partially solves https://github.com/nasa/fprime/issues/2706, as this does not solve the warnings/silent failures

## Rationale

The TCP unit tests should pass in a reliable manner.
This also can affect the flight code if the "bind to 0" strategy is used to select a port.


Drv/Ip/TcpClientSocket.cpp
```diff
@@ -57,6 +57,11 @@ SocketIpStatus TcpClientSocket::openProtocol(NATIVE_INT_TYPE& fd) {
 #if defined TGT_OS_TYPE_VXWORKS || TGT_OS_TYPE_DARWIN
     address.sin_len = static_cast<U8>(sizeof(struct sockaddr_in));
 #endif
+    // Enable Address reuse to avoid binding to the socket that still might be in TIME_WAIT state.
+    // Can happen if a function like get_free_port() is called to determine an available port by binding to port 0.
+    const int enable = 1;
+    if (setsockopt(socketFd, SOL_SOCKET, SO_REUSEADDR, &enable, sizeof(enable)) < 0)
```
LeStarch: Minor coding-standard note:  all `if` statements regardless of line count should contain explicit block curly braces. i.e.

```c++
if (something) {

}
```

This is done to prevent bug that might creep-in if ever a step is added to this if block on a future revision.

LeStarch (changes requested):
We'd like to see this as a defaulted argument to the `openProtocol` call because this code affects UTs (where turning this option on is **needed**) and flight code where this option might be a vulnerability depending on the remote side of these TCP connections.

```c++
openProtocol(NATIVE_INT_TYPE& fd, bool reuse_address=false)
```

Since there is a potential for some plumbing through components, and changes to the UTs, please let me know if you'd like some assistance performing this work.

Thanks for you contributions!

Drv/Ip/TcpClientSocket.cpp
JohanBertrand: Totally agree. I thought it was already identified by clang-tidy, but it's currently not.

Would you be in to have clang-tidy verifying that there is always a brace after statements? I can prepare a PR for that.

LeStarch: yes, absolutely!

Drv/Ip/TcpClientSocket.cpp
```diff
@@ -57,12 +57,22 @@ SocketIpStatus TcpClientSocket::openProtocol(NATIVE_INT_TYPE& fd) {
 #if defined TGT_OS_TYPE_VXWORKS || TGT_OS_TYPE_DARWIN
     address.sin_len = static_cast<U8>(sizeof(struct sockaddr_in));
 #endif
+    if (reuse_address)
```
LeStarch, line 60: Based on the comments, I didn't expect this code in the client....only in the server.  If that is true, then perhaps it is only needed in the `startup` function and not the `open` functions?

JohanBertrand: I just tried to run the unit tests with the 4 different configurations, and it seems like both sockets are needing the REUSEADDR option to avoid errors. You can double check from your side using the procedure mentioned in the issue https://github.com/nasa/fprime/issues/2706 on the TcpClient UT.

To be honest, I am not an expert in TCP sockets, and I can not really explain the real reason behind this behavior, sadly.

Drv/Ip/UdpSocket.hpp
```diff
@@ -88,7 +88,7 @@ class UdpSocket : public IpSocket {
      * \param fd: (output) file descriptor opened. Only valid on SOCK_SUCCESS. Otherwise will be invalid
      * \return status of open
      */
-    SocketIpStatus openProtocol(NATIVE_INT_TYPE& fd);
+    SocketIpStatus openProtocol(NATIVE_INT_TYPE& fd, const bool reuse_address=false) override;
```
JohanBertrand: Should I not add override here ?

Drv/Ip/TcpClientSocket.cpp
LeStarch: I can attempt to explain why these errors occur.  TCP has two endpoints (a sever and a client).  When the connection is terminated there are two termination possibilities:

1. Uncoordinated termination (typically caused by calling `close` before calling `shutdown`)
2. Coordinated termination (caused by calling `shutdown` and then `close`)

In the uncoordinated case, a port will end-up in `TIME_WAIT*` state, where it waits for a specified timeout (typically 60s) to let any remote that was not properly informed to shutdown notice the connection is down before allowing the port to be reused.  Otherwise, the client might send data to a reused port accidentally leaking data.

In the coordinated case, a message is sent across TCP informing the remote to shutdown and an acknowledgement is sent back.  Then the port can be closed, and immediately reused.

These errors occur when the port cannot be reused immediately due to an uncoordinated shutdown, or an unacknowledged coordinated shutdown.  Thus the port is unable to be reused, but the UTs expect it can.

Given the random nature of these errors, it is hard to reproduce with enough insight to determine which of these cases occur and exactly why.  In an ideal world, the UTs would conduct a proper `shutdown` always, and thus not need the ability to "reuse" waiting ports.  However, your proposed solution is deeply practical as it quickly resolves the issue without trying to reproduce a random bug.

JohanBertrand: @LeStarch Thanks for the explanations.

What would be the path forward for this MR? 
Do you want me to investigate more on the proper shutdown of the sockets for the UT instead of having the reuse option?

LeStarch: I like the reuse option as a fallback. Especially given the time required to figure out the cause of the other.

JohanBertrand: Do you want any other changes in the current MR then?

I will anyways investigate on the proper shutdown of the UT in the future, and come back to you if I have any findings.

LeStarch:
I studied up on TCP and this is what I was able to deduce:

1. TIME_WAITs happen on the initiator of the close/shutdown call. Thus clients should always initiate the shutdown
2. close and/or shutdown is asynchronous unless the SO_LINGER option is set.

I think that means the following should be done to the UTs:

1. Ensure the client always closes first and 
2. Make sure to use SO_LINGER on UT sockets.  This will ensure that things are closed fully before reuse.

Since the error originates from the find port helper, it might be sufficient to just apply number 2 to the port bound there.

This work won't hurt, but may be masking the underlying issue.  Thoughts?

Thanks for your patience while I study up!

JohanBertrand:
Thanks for the details. I just gave it a try. I set the `reuse_address` parameters to `false` (their default values), and set `SO_LINGER` on the port helper function, but I still get the same errors:

```
/home/fprime/Drv/TcpClient/test/ut/TcpClientTester.cpp:40: Failure
Expected equality of these values:
  serverStat
    Which is: -9
  SOCK_SUCCESS
    Which is: 0
TCP server startup error: Address already in use
```

I pushed the changes so that you can give it a try on your side. I agree with your analysis, but I don't understand why we're having such issues even with `SO_LINGER` activated and set to `0`.

I think that we should go ahead with the `reuse_address` option, either as a function parameter, or as something fixed in the code. 
I'm not sure if there is a real drawback of having it activated by default.

LeStarch:
The issue I see is that setting `SO_LINGER` to `0` has almost no effect. This is because that value is the timeout, and a timeout of zero will exit immediately yielding the same problem. I have pushed a fix that sets it to `30`.

JohanBertrand:
I have exactly the same error messages on my side with the timeout of 30

LeStarch:
@JohanBertrand can you remind me again what your testing environment is?  Is it a docker container?  Any special networking?  Host machine type?

I have yet to be able to reproduce the issue to have the failures appear reliably....but I want to get something to reproduce it.

LeStarch:
@JohanBertrand  As a reminder, this is an open forum.  So keep the information succinct.

1. native, VM, docker
2. Host OS (mac/darwin/win), Guest OS (linux/other)
3. Any special networking config

JohanBertrand:
I am currently running an `ubuntu:20.04` docker instance on WSL2, with the script mentioned in https://github.com/nasa/fprime/issues/2706

I don't have any special network settings on the docker

Usually, if you run the test about 20 times, you should get at least one test failing

LeStarch:
First of all, thank you.  That was key-information and I can now reliably reproduce the problem.  I had a thought on how to fix the issue altogether.

In short:
1. Allow TcpServer to accept port "0" (choose a port)
2. Allow TcpServer to return the port hosted at
3. Update the test to skip port-selector altogether, and just choose any port and reuse that in the clients.

I'll take a crack at this tomorrow.  In the meantime, we should determine if we want to keep reuse around (assuming this pans out).

LeStarch:
@JohanBertrand I think I finally figured out some answers!  Why doesn't the original code work?  Why do the things that "should" fix the problem not fix the problem?  Why does logic seem to fail when discussing theses issues?  Why did port-selector need to recurse to avoid "root only" ports in a non-root program?

Here is the answer: notice the asymmetry 
https://github.com/nasa/fprime/blob/0fe467c6e21aad321a0f1c737576733aac582443/Drv/Ip/test/ut/PortSelector.cpp#L23

when compared against:
https://github.com/nasa/fprime/blob/0fe467c6e21aad321a0f1c737576733aac582443/Drv/Ip/test/ut/PortSelector.cpp#L41

There is no network byte swap when reading back the port.  This means that we might select a "good" port, and then byte-swap it into a bad port, root port, reserved port, etc.

1. Why doesn't the original code work? The detected good port is corrupted before use, loosing its guarantee of availability.
2. Why do things that should fix the issue fail?  We were using SO_LINGER and other mechanisms on a port that was the pre-corrruption.
3. Why does logic fail? Because logic was based on a bad understanding of which port was really used.
4. Why could we get root only ports? Because post-swap could enter reserved space

I have addressed this, and added the capability for TcpServer and UdpRecv to use "0" (Os chooses a port) in this PR: https://github.com/nasa/fprime/pull/2739.

Would you mind reviewing?

JohanBertrand:
Awesome! Thanks you for looking into that and for providing those details.

I'm going to look into the fix in the PR.

FYI, we might still need to have a recursive call to the `get_free_port` function to be able to find two ports which are different for the UDP test (See the changes in this PR)

LeStarch:
@JohanBertrand since we found the root cause, I am going to close this PR.  If you feel there is still need for reconnect behavior, we can track that as a dedicated issue.

PR #2710, opened 2024-04-30
Update of clang tidy checks

JohanBertrand:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

- Add clang-tidy check on braces after statements and related flagged errors.
- Merge `release.clang-tidy` into `.clang-tidy` and update CI. Unless it is used for somewhere externally, I do not see the file `.clang-tidy` used anywhere otherwise.
- Also fix an issue in StringBase flagged on my docker image, but not on the CI?

## Rationale

Enforce consistent coding standard

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,6 +1,7 @@
+# Clang-tidy configuration used on the whole code base, including tests and flight code
+
 Checks: >
     bugprone-unhandled-self-assignment,
-    misc-no-recursion,
     modernize-deprecated-headers,
     modernize-redundant-void-arg,
     modernize-use-bool-literals,
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,8 +1,10 @@
 Checks: >
     bugprone-unhandled-self-assignment,
+    misc-no-recursion,
     modernize-deprecated-headers,
     modernize-redundant-void-arg,
     modernize-use-bool-literals,
     modernize-use-nullptr,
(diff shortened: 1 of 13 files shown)


LeStarch (approved):
I am happy with these changes but I will want @Joshua-Anderson and @thomas-bc to weigh in.

Joshua-Anderson:
The logic for separating out `.clang-tidy` and `release.clang-tidy` was because we had some clang tidy checks we wanted to enforce for flight code, but not project unit tests.

See [this github workflow](https://github.com/nasa/fprime/blob/devel/.github/workflows/build-test.yml#L102) where `release.clang-tidy` is used

At this moment, we could remove `release.clang-tidy` if we were okay with disallowing recursion on unit tests, but there may be future checks we only want to enforce on flight code.

JohanBertrand:
@Joshua-Anderson Thank you for the explanation, I am happy keeping two different files. I did not realize that the first call of `clang-tidy` on the CI was also running on the unit tests.

LeStarch:
@JohanBertrand I think we will need to go with @Joshua-Anderson's suggestion.  The failure we see is recursion used in a UT.

JohanBertrand:
I reverted the clang-tidy configuration file changes and added some comments. It should be good to go!

LeStarch (approved):
Looks good.

PR #2711, opened 2024-05-01
Fix copying of data product containers

bocchino:
This PR fixes an issue with copying data product containers. A `DpContainer` contains an `Fw::ExternalSerializeBuffer` as a member that provides a view into the `Fw::Buffer` that it owns. When a `DpContainer`was copied via copy construction or copy assignment, the `ExternalSerializeBuffer` member was not being copied correctly.

I did the following:
* Delete the copy assignment operator from `ExternalSerializeBuffer`.
* Add two new derived classes of `ExternalSerializeBuffer`: `ExternalSerializeBufferWithDataCopy` and `ExternalSerializeBufferWithMemberCopy`. Each derived class adds just that kind of copy semantics to the base class.
* Revise `DpContainer` so that the `ExternalSerializeBuffer` has type `ExternalSerializeBufferWithMemberCopy`. The slice is owned by the `DpContainer`, so this is the behavior we want.
* Revise the tests in `FppTest/dp` to verify that the container assignment is working.
* Make the copy assignment operator of `SerializeBufferBase` protected, so that assigning into a `SerializeBufferBase` reference is not allowed.
* Revise a unit test that was assigning into a `SerializeBufferBase` reference, in a way that was not necessary.

@timcanham This PR pulls out the changes that I submitted to your DpCatalog branch. I thought it would be good to make a separate PR to mainline with just these changes.

### DIFF
--- a/Fw/Buffer/test/ut/TestBuffer.cpp
+++ b/Fw/Buffer/test/ut/TestBuffer.cpp
@@ -83,21 +83,19 @@ void test_representations() {
     }
     Fw::SerializeStatus stat = sbb.serialize(100);
     ASSERT_NE(stat, Fw::FW_SERIALIZE_OK);
+
     // And that another call to repr resets it
-    sbb = buffer.getSerializeRepr();
     sbb.resetSer();
     ASSERT_EQ(sbb.serialize(0), Fw::FW_SERIALIZE_OK);
 
     // Now deserialize all the things
     U32 out;
-    sbb = buffer.getSerializeRepr();
     sbb.setBuffLen(buffer.getSize());
     for (U32 i = 0; i < sizeof(data)/4; i++) {
         ASSERT_EQ(sbb.deserialize(out), Fw::FW_SERIALIZE_OK);
         ASSERT_EQ(i, out);
     }
     ASSERT_NE(sbb.deserialize(out), Fw::FW_SERIALIZE_OK);
-    sbb = buffer.getSerializeRepr();
     sbb.setBuffLen(buffer.getSize());
     ASSERT_EQ(sbb.deserialize(out), Fw::FW_SERIALIZE_OK);
     ASSERT_EQ(0, out);
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -268,13 +268,14 @@ doxyindexer
 doxyrules
 doxysearch
 Doxywizard
-DPCFG
-dpi
+dpc
 DPCAT
 DPCATALOG
+DPCFG
+dpi
 DPMANAGER
-DPWRITER
 dps
+DPWRITER
 DRAINBUFFERS
 drv
 dsdl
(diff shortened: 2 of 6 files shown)


timcanham (approved):
LGTM!

Fw/Buffer/test/ut/TestBuffer.cpp
```diff
@@ -83,21 +83,19 @@ void test_representations() {
     }
     Fw::SerializeStatus stat = sbb.serialize(100);
     ASSERT_NE(stat, Fw::FW_SERIALIZE_OK);
+
     // And that another call to repr resets it
-    sbb = buffer.getSerializeRepr();
```
LeStarch, line 87: I think that we will want to support this.  We should probably change the return type from `getSerializeRepr` to return the `ExternalBufferWithMemberCopy` for the same reason.  `Fw::Buffer` owns the "slice" and thus we want member copy in these instances.

LeStarch (approved):
One minor comment.  Could be fixed on follow-up.

Fw/Buffer/test/ut/TestBuffer.cpp
bocchino: What is the intended purpose of the deleted line here? I suspect that what it is actually doing is not intended:

1. The right-hand side gets a reference to the serialize representation object, expressed as an Fw::SerializeBufferBase.
1. The left-hand side sbb is a reference to the same object, from a previous call to getSerializeRepr, also expressed as an Fw::SerializeBufferBase.
1. According to the copy semantics prior to this PR, assignment of an Fw::SerializeBufferBase copies the data. So I believe the deleted line copies the same data on top of itself, i.e., it is a no-op.
1. In the revised copy semantics, the assignment is not allowed. I think this is the correct behavior, since assigning to a base-class reference is contrary to C++ coding recommendations.

With regard to the internal implementation of the serialize representation, we could use ExternalBufferWithMemberCopy, and that would provide member copy semantics, as in the new version of DpContainer. However, I think this is not necessary. Fw::Buffer already defines its own copy semantics that does the same thing (member copy) in handwritten code.

bocchino: Also, I don't think we want to return an ExternalBufferWithMemberCopy. This would be dangerous, because then users could assign into the buffer and update the external buffer pointer. If we returned ExternalBufferWithDataCopy then we could preserve what was happening in the old code (assignment that copies data) but I don't see a need to assign into the serial buffer here. It seems to me that all access through the serial buffer should be through the serialization interface, not through assignment.

bocchino: So I think the current return type Fw::SerializeBufferBase is correct.

LeStarch: I see.  Ok, I am fine with this as-is.

bocchino:
Note: If we had `getSerializeRepr` return an object, instead of a reference, then we could return `Fw::ExternalSerializeBufferWithMemberCopy` as suggested by @LeStarch. I wrote up an issue on it here: https://github.com/nasa/fprime/issues/2714.

PR #2719, opened 2024-05-06
Add `-Wno-format-truncation` to root project

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

- Attempt to fix the Autodocs CI failure by disabling the format-truncation warnings
- Run autodocs as part of regular CI checks. It is a fairly quick workflow (<5mins) and it's not the first time that it not being ran on PRs is playing tricks on us. 

## Rationale

Format truncation are an expected behavior

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
         $<$<COMPILE_LANGUAGE:CXX>:-Wno-vla-extension>
         $<$<COMPILE_LANGUAGE:CXX>:-Wno-zero-length-array>
     )
-else()
+elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
     add_compile_options(
         -Wno-format-truncation
     )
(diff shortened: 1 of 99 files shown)


CMakeLists.txt
```diff
@@ -56,6 +56,10 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
         $<$<COMPILE_LANGUAGE:CXX>:-Wno-vla-extension>
         $<$<COMPILE_LANGUAGE:CXX>:-Wno-zero-length-array>
     )
+else()
```
Joshua-Anderson: Is there something special about autodocs that triggers truncation warnings? I would have expected that truncation would be failing on normal gcc builds as part of CI.

I'm a little leary of conditionally disabling format truncation only if the compiler isn't clang. Maybe consider always disable the format truncation warning if we're not ready to resolve it yet?

LeStarch: The issue is that clang does not support this flag.  Thus we cannot disable it on clang.

Joshua-Anderson: Gotcha. Maybe just add the flag if the compiler is GCC so we don't break in weird ways on unexpected compilers? That being said I think it's unlikely enough that somebody would build the root cmake project on an odd compiler, so maybe not a significant concern.

LeStarch (approved):

PR #2720, opened 2024-05-08
Work/string search

timcanham:
| | |
|:---|:---|
|**_Related Issue(s)_**| None  |
|**_Documentation Included (y/n)_**| y, headers |

---
## Change Description

Added a function to `string_utils` to search for a substring in a string

## Rationale

Centralizes a feature that could be useful to many people.

## Future Work

Calling this function will be added to `StringBase` as a helper.

### DIFF
--- a/Fw/Types/StringUtils.cpp
+++ b/Fw/Types/StringUtils.cpp
@@ -48,7 +48,7 @@ FwSignedSizeType Fw::StringUtils::substring_find(const CHAR* source_string, FwSi
     FW_ASSERT((source_size - sub_size) <= std::numeric_limits<FwSignedSizeType>::max());
 
     // Loop from zero to source_size - sub_size (inclusive)
-    for (FwSizeType source_index = 0; source_index < (source_size - sub_size + 1); source_index++) {
+    for (FwSizeType source_index = 0; source_index < (source_size - sub_size + 1) && source_index < static_cast<FwSizeType>(std::numeric_limits<FwSignedSizeType>::max()); source_index++) {
         // if the current character matches
         for (FwSizeType sub_index = 0; sub_index < sub_size; sub_index++) {
             // Prevent read overrun
--- a/Fw/Types/StringUtils.cpp
+++ b/Fw/Types/StringUtils.cpp
@@ -29,3 +29,32 @@ U32 Fw::StringUtils::string_length(const CHAR* source, U32 max_len) {
     }
     return length;
 }
+
+I32 Fw::StringUtils::substring_find(const CHAR* source_string, U32 source_size, const CHAR* sub_string, U32 sub_size) {
+
+    FW_ASSERT(source_string != nullptr);
+    FW_ASSERT(sub_string != nullptr);
+
+    for (U32 source_index = 0; source_index < source_size; source_index++) {
+        // if the current character matches
+        U32 dest_index = 0;
+        if (source_string[source_index] == sub_string[dest_index]) {
+            for (U32 sub_index = 0; sub_index < sub_size; sub_index++) {
+                // if there is a mismatch, go to next character
+                if (source_string[source_index + sub_index] != sub_string[sub_index]) {
+                    // if we have reached the end of the source buffer, return match not found
+                    if (source_index + sub_index == source_size-1) {
+                        return -1;
+                    }
+                    // otherwise move to next character in source
+                    continue;
+                }
+            }
+            // if we got here, that means we matched all the way to the end of the substring
+            return source_index;
+        }
+    }
+    
+    // if we make it here, no matches were found
+    return -1;
+}
--- a/Fw/Types/StringUtils.cpp
+++ b/Fw/Types/StringUtils.cpp
@@ -37,21 +37,16 @@ I32 Fw::StringUtils::substring_find(const CHAR* source_string, U32 source_size, 
 
     for (U32 source_index = 0; source_index < source_size; source_index++) {
         // if the current character matches
-        U32 dest_index = 0;
-        if (source_string[source_index] == sub_string[dest_index]) {
+        if (source_string[source_index] == sub_string[0]) {
             for (U32 sub_index = 0; sub_index < sub_size; sub_index++) {
                 // if there is a mismatch, go to next character
                 if (source_string[source_index + sub_index] != sub_string[sub_index]) {
-                    // if we have reached the end of the source buffer, return match not found
-                    if (source_index + sub_index == source_size-1) {
-                        return -1;
-                    }
-                    // otherwise move to next character in source
-                    continue;
+                    break;
+                } else if (sub_index == sub_size - 1) {
+                    // if we matched all the way to the end of the substring
+                    return source_index;
                 }
             }
-            // if we got here, that means we matched all the way to the end of the substring
-            return source_index;
         }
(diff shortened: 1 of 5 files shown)


Fw/Types/StringUtils.cpp
```diff
@@ -29,3 +29,27 @@ U32 Fw::StringUtils::string_length(const CHAR* source, U32 max_len) {
     }
     return length;
 }
+
+I32 Fw::StringUtils::substring_find(const CHAR* source_string, U32 source_size, const CHAR* sub_string, U32 sub_size) {
```
bocchino: We should use` FwSignedSizeType` instead of I32 and U32 here.

Fw/Types/StringUtils.cpp
```diff
@@ -29,3 +29,27 @@ U32 Fw::StringUtils::string_length(const CHAR* source, U32 max_len) {
...
+
+    FW_ASSERT(source_string != nullptr);
+    FW_ASSERT(sub_string != nullptr);
+
+    for (U32 source_index = 0; source_index < source_size; source_index++) {
+        // if the current character matches
+        if (source_string[source_index] == sub_string[0]) {
```
bocchino: I think this code will do an out-of-bounds array access if the sub_size is zero. What is the expected behavior in that case? No match?

Also, what is the reason for the outer if statement? Optimizing the fast path?

bocchino (commented):
Looks good! Just a couple of comments.

Should we add a wrapper function in `Fw::StringBase` to call this function?

Fw/Types/StringUtils.cpp
```diff
@@ -29,3 +29,27 @@ U32 Fw::StringUtils::string_length(const CHAR* source, U32 max_len) {
     }
     return length;
 }
+
+I32 Fw::StringUtils::substring_find(const CHAR* source_string, U32 source_size, const CHAR* sub_string, U32 sub_size) {
+
+    FW_ASSERT(source_string != nullptr);
+    FW_ASSERT(sub_string != nullptr);
+
+    for (U32 source_index = 0; source_index < source_size; source_index++) {
```
bocchino: Use `FwSignedSizeType` instead of `U32` for loop indices.

Fw/Types/StringUtils.cpp
```diff
@@ -29,3 +29,27 @@
     }
     return length;
 }
+
+I32 Fw::StringUtils::substring_find(const CHAR* source_string, U32 source_size, const CHAR* sub_string, U32 sub_size) {
```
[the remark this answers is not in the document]
timcanham: Fixed

Fw/Types/StringUtils.cpp
timcanham: Fixed

Fw/Types/StringUtils.cpp
timcanham: Fixed

Joshua-Anderson:
Did you consider using the libc [`strstr`](https://www.man7.org/linux/man-pages/man3/strstr.3.html) function? It does have a prereq on the strings being null terminated (rather than providing a length argument) but if the plan is to use it with StringBase I think that may be reasonable assumption.

timcanham:
@Joshua-Anderson I thought about it, but we prefer to use the "n" versions of string functions so they don't run off into space if the null terminator is corrupted. This is a home-grown version with an "n" adaptation, and the code is pretty small too.

bocchino:
> I thought about it, but we prefer to use the "n" versions of string functions so they don't run off into space if the null terminator is corrupted. 

I second this. We should avoid using any string library functions that lack a buffer-bound argument. It looks like there is a `strnstr` function, but it is BSD-specific, so it won't work on general POSIX.

Fw/Types/StringUtils.cpp
```diff
@@ -29,3 +29,30 @@ U32 Fw::StringUtils::string_length(const CHAR* source, U32 max_len) {
     }
     return length;
 }
+
+FwSignedSizeType Fw::StringUtils::substring_find(const CHAR* source_string, FwSizeType source_size, const CHAR* sub_string, FwSizeType sub_size) {
+
+    FW_ASSERT(source_string != nullptr);
+    FW_ASSERT(sub_string != nullptr);
+
```
LeStarch, line 40: Probably should assert that:

```
FW_ASSERT(source_size - sub_size <= std::numeric_limits<FwSignedSizeType>::max()); // FwSignedSizeType will overflow if the source size is too-large
```

Fw/Types/StringUtils.cpp
```diff
@@ -29,3 +29,30 @@ U32 Fw::StringUtils::string_length(const CHAR* source, U32 max_len) {
...
+    // zero size sub-strings should always match
+    if ((source_size > 0) && (0 == sub_size)) {
+        return 0;
+    }
+
+    for (FwSizeType source_index = 0; source_index < source_size; source_index++) {
```
LeStarch: Can't you get away with looping until `source_size - sub_size`.  It is impossible to match a substring that would go past the end of your source.

Fw/Types/StringUtils.cpp
```diff
@@ -29,3 +29,30 @@ U32 Fw::StringUtils::string_length(const CHAR* source, U32 max_len) {
...
+        // if the current character matches
+        for (FwSizeType sub_index = 0; sub_index < sub_size; sub_index++) {
+            // if there is a mismatch, go to next character
+            if (source_string[source_index + sub_index] != sub_string[sub_index]) {
```
LeStarch, line 63: Isn't this an overrun without the above bounds fix? Probably want:

```
FW_ASSERT(source_index + sub_index < source_size);
```

Fw/Types/StringUtils.cpp
```diff
@@ -29,3 +29,30 @@ U32 Fw::StringUtils::string_length(const CHAR* source, U32 max_len) {
...
+                break;
+            } else if (sub_index == sub_size - 1) {
```
LeStarch: I feel like my inner @timcanham says "use `(sub_index == (sub_size - 1))` for clarity".

Fw/Types/StringUtils.cpp
```diff
@@ -29,3 +29,30 @@ U32 Fw::StringUtils::string_length(const CHAR* source, U32 max_len) {
...
+FwSignedSizeType Fw::StringUtils::substring_find(const CHAR* source_string, FwSizeType source_size, const CHAR* sub_string, FwSizeType sub_size) {
+
+    FW_ASSERT(source_string != nullptr);
+    FW_ASSERT(sub_string != nullptr);
+
+    // zero size sub-strings should always match
+    if ((source_size > 0) && (0 == sub_size)) {
+        return 0;
+    }
+
```
LeStarch, line 45: Add:

```
if (source_size < sub_size) {
    return -1;
}
```

LeStarch (changes requested):
A few minor changes

Fw/Types/StringUtils.cpp
LeStarch: source_size - sub_size + 1....to catch the equality case

LeStarch (approved):

thomas-bc (approved):

PR #2731, opened 2024-05-13
Add sequence dispatcher component

### LINKED ISSUE #2729: Distributing command sequences throughout CmdSequencers
### Discussed in https://github.com/nasa/fprime/discussions/2608

<div type='discussions-op-text'>

<sup>Originally posted by **zimri-leisher** March 20, 2024</sup>
Hello,
We'd like to run multiple command sequences at the same time. As I understand, that means we need more than one command sequencer. However, it would be nice not to have to decide which command sequencer we want to run a command on. Instead, we want to just say "run this sequence on an available cmd sequencer".
Any suggestions for this? It seems like it would be a common problem. Or is this a bad idea and we should have to explicitly declare which command sequencer we want to use?</div>

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #2608 #2729  |

---
## Change Description

This PR adds a SeqDispatcher component to Svc which allows command sequences to be automatically distributed among CmdSequencers.

## Rationale

CmdSequencers only support running a single command sequence at a time. For complex missions, this necessitates having multiple CmdSequencers. However, without SeqDispatcher, ground operators would have to manually keep track of which CmdSequencers are currently being used. SeqDispatcher automates this process.

This is ported from an internal fork of FPrime which is a few versions out of date. Special care should be taken to make sure that the code seen here is up-to-date with the rest of FPrime. Also, probably should make much more in-depth unit tests.

## Future Work

Closes #2729


zimri-leisher:
Hello,
This is a very WIP PR for the SeqDispatcher. Eager to iterate on it with the maintainers and other members of the community!
At the moment all functionality works and some basic testing has been implemented. I had to make some very small changes to CmdSequencer to notify this component when a sequence starts, so that this component can keep track of which sequencers are busy.

Svc/SeqDispatcher/SeqDispatcher.fpp
```diff
@@ -0,0 +1,48 @@
...
+    @ Dispatches command sequences to available command sequencers
+    active component SeqDispatcher {
+
+        include "SeqDispatcherCommands.fppi"
+        include "SeqDispatcherTelemetry.fppi"
+        include "SeqDispatcherEvents.fppi"
+
+        async input port seqRunIn: Svc.CmdSeqIn
+
+        output port seqRunOut: [types.CMD_SEQUENCERS_COUNT] Svc.CmdSeqIn
```
Joshua-Anderson: It might be worth adding sequence dispatcher custom constant (SeqDispatcherSequencerCount?) for tracking the number of command sequencers, since some projects may want to separately reserve some command sequencers for specific uses. I.e. command sequencer 0 might be used for fault protection sequence dispatch.

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+bool SeqDispatcher::runSequence(NATIVE_INT_TYPE sequencerIdx,
+                                const Fw::String& fileName,
+                                Svc::CmdSequencer_BlockState block) {
+  FW_ASSERT(sequencerIdx >= 0 && sequencerIdx < types::CMD_SEQUENCERS_COUNT,
+            sequencerIdx);
+
+  // if the seqRunOut port at this idx isn't connected
+  if (!this->isConnected_seqRunOut_OutputPort(sequencerIdx)) {
+    this->log_WARNING_HI_INVALID_SEQUENCER(static_cast<U16>(sequencerIdx));
```
Joshua-Anderson: Instead of checking if the port is connected on use, it might make sense to check if the sequence engine is connected on initialization and asserting if not - letting users know if the topology is misconfigured as quickly as possible.

Joshua-Anderson (commented):
This is fantastic @zimri-leisher! Looks like the exact architecture I would have recommended. I threw in a few minor comments for consideration

Svc/CmdSequencer/CmdSequencerImpl.hpp
```diff
@@ -235,6 +235,10 @@ namespace Svc {
           //! \return The log file name
           Fw::LogStringArg& getLogFileName();
 
+          //! Get the normal string file name
+          //! \return The normal string file name
+          Fw::String& getStringFileName();
```
LeStarch, line 240: I see this new function here, and the member variable...but not the implementation of the function in the `.cpp` file changes.  Is that missing?

Svc/SeqDispatcher/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+set(SOURCE_FILES
+  "${CMAKE_CURRENT_LIST_DIR}/SeqDispatcher.fpp"
+  "${CMAKE_CURRENT_LIST_DIR}/SeqDispatcher.cpp"
+)
+
+# Uncomment and add any modules that this component depends on, else
+# they might not be available when cmake tries to build this component.
+
+set(MOD_DEPS
+    common/types
```
LeStarch: I don't recall `common/types` in F´, nor do I see that in the PR.

Svc/SeqDispatcher/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+    Svc/CmdSequencer
+)
+
+register_fprime_module()
+
+# Register the unit test build
+set(UT_SOURCE_FILES
+  "${CMAKE_CURRENT_LIST_DIR}/SeqDispatcherCommands.fppi"
```
LeStarch: Typically you don't need to add `.fppi` files to source lists.  Is this a workaround for the version I have?

Svc/SeqDispatcher/SeqDispatcher.fpp
zimri-leisher: Sorry yeah this types.CMD_SEQUENCERS_COUNT was a leftover from our internal fork. Will do

Svc/SeqDispatcher/SeqDispatcher.cpp
zimri-leisher: Great call! Will do

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,204 @@
+// ======================================================================
+// \title  SeqDispatcher.cpp
+// \author root
+// \brief  cpp file for SeqDispatcher component implementation class
+// ======================================================================
+
+#include <FpConfig.hpp>
+#include <components/fsw/SeqDispatcher/SeqDispatcher.hpp>
```
LeStarch: Probably need to update include paths.

Svc/SeqDispatcher/SeqDispatcherCommands.fppi
```diff
@@ -0,0 +1,9 @@
+async command RUN(
+                      fileName: string size 240 @< The name of the sequence file
+                      $block: Svc.CmdSequencer.BlockState @< Return command status when complete or not
+                    ) \
+    opcode 0
+
+async command LOG_STATUS() opcode 1
+
+async command PING() opcode 2
```
LeStarch: What is PING for?  Typically, we do ping ports as those can be tracked by health.

LeStarch (commented):
In general, I am impressed!  I'll ask @timcanham to look over it too.

LeStarch:
@timcanham want to take a look, it is a sequence dispatcher!

Svc/SeqDispatcher/SeqDispatcherCommands.fppi
zimri-leisher: Hmm, internally we add them so that we can ping the component's thread. How do you hook it up to health?

Svc/SeqDispatcher/SeqDispatcher.cpp
zimri-leisher: Will do

Svc/SeqDispatcher/CMakeLists.txt
zimri-leisher: Not sure if we've tried without including the fppi files. Will try removing if possible

Svc/SeqDispatcher/CMakeLists.txt
zimri-leisher: Artifact from internal code layout

Svc/CmdSequencer/CmdSequencerImpl.hpp
zimri-leisher: That was a mistake, adding back

zimri-leisher:
Thanks for all the comments. Sorry that there are a lot of things that are idiomatic to our code that I didn't remove from here, but honestly part of the reason why I wanted to put this in front of you guys was to get exactly this feedback.

I have a question: where would you recommend storing the `CmdSequencerState` enum and `CMD_SEQUENCERS_COUNT` const? Should I make a new .fpp file inside of `Svc/SeqDispatcher`? What should I call it, if so?

Joshua-Anderson:
Since CMD_SEQUENCERS_COUNT/SeqDispatcherSequencerCount is a per deployment config option, I'd add a `SeqDispatcherCfg.fpp` file with this constant to `config/`, similar to [DpCfg.fpp](https://github.com/nasa/fprime/blob/devel/config/DpCfg.fpp). ~~If you wanted to share `CmdSequencerState` between CmdSequencer and SeqDispatcher, I might add the enum type to the same file.~~

Joshua-Anderson:
Err, since we don't want folks to modify CmdSequencerState, `config/` probably isn't the right home for it. [Fw/Types/Types.fpp](https://github.com/nasa/fprime/blob/devel/Fw/Types/Types.fpp) is where we store common types but this is a bit too specific for that. What about using `FW::Wait` instead of a custom command sequencer enum?

zimri-leisher:
Alright to the `SeqDispatcherCfg.fpp` file, will do

With regards to using Fw::Wait, I'd prefer not to, as CmdSequencerState has three values instead of two: `AVAILABLE`, `RUNNING_BLOCK`, and `RUNNING_NO_BLOCK`. I also see a relatively low cost to creating a new enum for this. What I'm realizing right now as I write this is that it should just be an enum internal to the .hpp/.cpp files, because it's not used in the fpp at all.

Joshua-Anderson:
@zimri-leisher my bad, I got CmdSequencerState and `Svc.CmdSequencer.BlockState` confnused.

For component internal enums, you can either add it in the hpp as you suggested, or add it as a standalone type to the `SeqDispatcher.fpp` file.

The concern we were discussing does arise for the `Svc.CmdSequencer.BlockState` enum. I'm not sure we want SeqDispatcher to depend on CmdSequencer, since we may want to leave open the option for folks to replace CmdSequencer with their own component.

zimri-leisher:
Good point, maybe I should change the argument of the RUN command to just take a boolean: blocking or non-blocking. No need for a two-valued enum to exist when we have bools I guess :P

Joshua-Anderson:
I'd recommend using an enum like FW:Wait (WAIT and NO_WAIT) instead of a bool, because that can make the sequence files more readable `SeqDispatch.Run "/myseq.bin", WAIT` vs `SeqDispatch.Run "/myseq.bin", TRUE`

I'd be curious to get folks opinion on what is the clearest.... WAIT/NO_WAIT, BLOCK/NO_BLOCK, TRUE/FALSE

timcanham:
IMHO not TRUE/FALSE for the reasons @Joshua-Anderson says, WAIT/NO_WAIT seems a little more understandable in plain English.

zimri-leisher:
Ok, will make these changes. However just want to mention that I'm out of office for 2 weeks starting this Friday so they might not happen until then.
I have another question which is related to this PR and also our implementation of complex logic in sequences. Not sure where best to ask it so I'll ask it here:

The idea of this PR is that you shouldn't have to know which sequencer you're running a sequence on ahead of time. However, this means that commands inside of the sequence file don't know either. This is a problem if these commands want to modify something about the sequencer--in our case, we want to change which line the sequencer is on based on some runtime criteria inside of a command.

Is there some way that a command handler can tell from which sequencer a command came from? This would solve my problem. I couldn't find a way and so came up with a really gross solution. Hoping one of you has some input on this.

Joshua-Anderson:
> This is a problem if these commands want to modify something about the sequencer--in our case, we want to change which line the sequencer is on based on some runtime criteria inside of a command.

The way I've seen this approached on past missions is a concept of sequence directives. Effectively these are special commands/opcodes that are only valid within a command sequence, and they dispatch a command to the command sequencer executing the sequence.

Ex: `SEQ_WAIT 10` sequence diretive within a command sequence would execute a command on the command sequencer executing the sequence to delay 10 seconds until the next command.

Fprime doesn't have any concept of sequence directives yet, but it's something on our roadmap that we want to add. It probably requires a change to the sequence binary format to add a flag indicating if the "command" opcode is a standard fprime command or if it should be dispatched as a sequence directive instead.

What do you think about a sequence directive approach? Would that solve your use cases?

Have a fantastic vacation and no rush at all on this pull request!

zimri-leisher:
That is kind of what I'm doing, only my solution is hackier. I made a component called `GenericSequencer` which only has empty commands inside of it. Whenever a real sequencer sees a command from a GenericSequencer (it checks this by deserializing the command, and checking if the opcode is equal to one of the known commands in a generic sequencer), it edits the command packet opcode to instead be the opcode from its own command.
Any pitfalls you can see with this approach? I'm kind of afraid of doing it because messing around with binary packets is not where I want to be, but as far as I can tell it's the most seamless way of doing it.

Joshua-Anderson:
@zimri-leisher Yep, very similar approach. Only pitfall is that is it adds commands in your dictionary that you never want to send from the ground. Sequence directives are a slightly cleaner way of accomplishing the same thing.

I think your approach will work well and as a future improvement we can work on the design and implementation of sequence directives for fprime. I'll make sure you get tagged in the github discussions on the topic once we get started.

zimri-leisher:
Well, I took my time, but I'm back and gonna get this merged. Just pushed some changes to allow the code to build, and incorporated all of your feedback so far.

What's the best way to add my name to the allowed words list? It's failing on spell check lol

Svc/SeqDispatcher/SeqDispatcherCommands.fppi
zimri-leisher: Removed ping command

Svc/SeqDispatcher/SeqDispatcher.cpp
zimri-leisher: Updated include paths

Svc/SeqDispatcher/CMakeLists.txt
zimri-leisher: Removed redundant fppi in source list

Svc/SeqDispatcher/CMakeLists.txt
zimri-leisher: Removed

Svc/CmdSequencer/CmdSequencerImpl.hpp
zimri-leisher: Added back

Svc/SeqDispatcher/SeqDispatcher.cpp
zimri-leisher: Code written, needs further testing

Svc/SeqDispatcher/SeqDispatcher.fpp
zimri-leisher: Done

thomas-bc:
@zimri-leisher you can add words that aren't recognized to the https://github.com/nasa/fprime/blob/devel/.github/actions/spelling/expect.txt

zimri-leisher:
Any help on this error?
image
`/workspaces/fprime/fprime/build-fprime-automatic-native-ut/F-Prime/Svc/SeqDispatcher/SeqDispatcherTesterBase.cpp:1170:5: error: no matching function for call to ‘Fw::String::String(Fw::String&)’`

Looks like I can fix it by removing `explicit` from this line in String.hpp:
image

thomas-bc:
@zimri-leisher could using the `FileNameString` type help?

zimri-leisher:
It seems the problem was that the (previously unused) CmdSeqIn port took in a string like `ref filename: Fw.String` instead of `filename: string size X`
Once I fixed it, the unit test compilation errors went away.

zimri-leisher:
Sorry about the compilation errors. All fixed now, should be good to rerun build jobs.

zimri-leisher:
Ah, sorry, looks like there were some changes to upstream FPrime that I didn't have locally that broke the build in the CI action but not on my computer. Fixed this now.

zimri-leisher:
Most of the CodeQL errors I don't really understand--it doesn't seem to be detecting my assertions in functions with more than 10 lines, for instance. Also, for the markdown-link-check, Format Python, Tutorial: MathComponent, and macOS-CI jobs, I don't see any output from them other than "This job failed":
image

LeStarch:
Here is the error from macOS-CI:

```
Users/runner/work/fprime/fprime/Svc/SeqDispatcher/test/ut/SeqDispatcherTester.hpp:54:8: error: 'Svc::Tester::textLogIn' hides overloaded virtual function [-Werror,-Woverloaded-virtual]
  void textLogIn(const FwEventIdType id,         /*!< The event ID*/
       ^
/Users/runner/work/fprime/fprime/build-fprime-automatic-native-ut/F-Prime/Svc/SeqDispatcher/SeqDispatcherTesterBase.hpp:491:20: note: hidden overloaded virtual function 'Svc::SeqDispatcherTesterBase::textLogIn' declared here: type mismatch at 2nd parameter ('const Fw::Time &' vs 'Fw::Time &')
      virtual void textLogIn(
                   ^
1 error generated.
```

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,210 @@
...
+// ----------------------------------------------------------------------
+// Construction, initialization, and destruction
+// ----------------------------------------------------------------------
+
+SeqDispatcher ::SeqDispatcher(const char* const compName)
+    : SeqDispatcherComponentBase(compName) {}
+
+SeqDispatcher ::~SeqDispatcher() {}
+
+void SeqDispatcher::init(NATIVE_INT_TYPE queueDepth,
```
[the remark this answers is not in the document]
LeStarch: As a note here, we've been removing the `init` function altogether.  It isn't needed, as it calls the parent class function.

Custom setup is best done in a dedicated `configure` or `setup` function anyway, as FPP expects `init` to have only default arguments.

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,210 @@
...
+                         NATIVE_INT_TYPE instance) {
```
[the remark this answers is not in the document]
LeStarch: See comment above.

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,210 @@
...
+    // assert if any of the cmd sequencer connections aren't
+    // connected. otherwise we might think we sent out
+    // a sequence to be run but it doesn't get run
+    if (!this->isConnected_seqRunOut_OutputPort(i)) {
+      FW_ASSERT(0, i);
+    }
+  }
+}
+
+NATIVE_INT_TYPE SeqDispatcher::getNextAvailableSequencerIdx() {
```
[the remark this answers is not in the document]
LeStarch: For port indices, use `FwIndexType` in place of `NATIVE_INT_TYPE`

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,210 @@
...
+  for (int i = 0; i < SeqDispatcherSequencerPorts; i++) {
```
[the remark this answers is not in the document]
LeStarch: Same as above

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,210 @@
...
+    if (this->m_entryTable[i].state == SeqDispatcher_CmdSequencerState::AVAILABLE) {
+      return i;
+    }
+  }
+  return -1;
+}
+
+bool SeqDispatcher::runSequence(NATIVE_INT_TYPE sequencerIdx,
```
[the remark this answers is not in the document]
LeStarch: FwIndexType

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,210 @@
...
+  this->m_dispatchedCount++;
+  this->tlmWrite_dispatchedCount(this->m_dispatchedCount);
+  this->seqRunOut_out(sequencerIdx,
+                      this->m_entryTable[sequencerIdx].sequenceRunning);
+
+  return true;
+}
+
+void SeqDispatcher::seqStartIn_handler(
+    NATIVE_INT_TYPE portNum,                                //!< The port number
```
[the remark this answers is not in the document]
LeStarch: We will eventually move this to `FwIndexType` but for now we the autocoder uses `NATIVE_INT_TYPE`

LeStarch:
I restarted the code-ql job.  Sometimes cloud hosts go down and jobs fail w/o output.  A rerun should work.

Svc/Seq/Seq.fpp
```diff
@@ -2,7 +2,7 @@ module Svc {
 
   @ Port to request a sequence be run
   port CmdSeqIn(
-                 ref filename: Fw.String @< The sequence file
+                 filename: string size 240 @< The sequence file
```
LeStarch, line 5: I'd like to understand why this change was made.

Svc/SeqDispatcher/SeqDispatcher.cpp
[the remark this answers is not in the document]
zimri-leisher: I added this to the `init` function based on @Joshua-Anderson's suggestion:
"Instead of checking if the port is connected on use, it might make sense to check if the sequence engine is connected on initialization and asserting if not - letting users know if the topology is misconfigured as quickly as possible."
I can move this to a different function but I think it would be weird to require the user to do extra work and call it if all it does is check if the ports are connected. Is there another place I could put this checking code so that it will run at bootup?

Svc/SeqDispatcher/SeqDispatcher.cpp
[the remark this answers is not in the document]
zimri-leisher: Will do

zimri-leisher:
Okay, I made the changes you suggested, pending further discussion on the init function.

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+SeqDispatcher ::SeqDispatcher(const char* const compName)
+    : SeqDispatcherComponentBase(compName) {}
+
+SeqDispatcher ::~SeqDispatcher() {}
+
+void SeqDispatcher::init(NATIVE_INT_TYPE queueDepth,
+                         NATIVE_INT_TYPE instance) {
+  SeqDispatcherComponentBase::init(queueDepth, instance);
+  for (FwIndexType i = 0; i < SeqDispatcherSequencerPorts; i++) {
```
LeStarch: I mentioned earlier that this function might be removed. However, this is a fine use of the function: no additional inputs.

I do wonder why this asserts.  Could it be better to just build up a list of sequencers that are available and use those as the pool?

This architecture Implies that every project must attach all command sequencers possible to this component.  You couldn't say configure a sequencer pool max size of 6, but only attach 3.

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,210 @@
...
+    const Fw::StringBase& fileName //!< The sequence file name
+) {
+  FW_ASSERT(portNum >= 0 && portNum < SeqDispatcherSequencerPorts, portNum);
+  if (this->m_entryTable[portNum].state ==
+          SeqDispatcher_CmdSequencerState::RUNNING_SEQUENCE_BLOCK ||
+      this->m_entryTable[portNum].state ==
+          SeqDispatcher_CmdSequencerState::RUNNING_SEQUENCE_NO_BLOCK) {
+    // we were aware of this sequencer running a sequence
+    if (this->m_entryTable[portNum].sequenceRunning != fileName) {
+      // uh oh. m_sequencesRunning is wrong
```
LeStarch: I think this needs to produce a WARNING_HI event.  This is because:

1. You've said to run a sequence
2. Something else in the system ran a different sequence

I agree there is not automatic resolution, but the requested sequence is likely being rejected by the sequence engine.

A more complicated behavior might be to retry dispatch to another member of the pool....but we'd need to think through the implications.

Svc/SeqDispatcher/test/ut/SeqDispatcherTesterHelpers.cpp
```diff
@@ -0,0 +1,58 @@
+// ======================================================================
```
LeStarch: If you did not make changes to this file, I'd recommend adding:

`set(UT_AUTO_HELPERS ON)` to your CMakeList.txt and deleting this file.

Svc/SeqDispatcher/SeqDispatcher.cpp
LeStarch: Looking through this more, I think the right answer is add an isConnected call to the `getNextAvailableSequenceIdx`.

That way users could connect up as many or as few sequencers as they want in their pool.  The "catch" for users who did not connect-up sequencers would be to look for unconnected ports in the topology.

That way a project could use two sequencer pools of different sizes if they wanted, without needing to connect every port.

Svc/SeqDispatcher/SeqDispatcher.cpp
LeStarch: Looking through the whole code, I recommend this check and the whole function be deleted.  See reasons below.

Please feel to comment back if you disagree with my rationale as we should try to reach a consensus.

Svc/SeqDispatcher/SeqDispatcherEvents.fppi
```diff
@@ -0,0 +1,23 @@
+event INVALID_SEQUENCER(
```
LeStarch: Typically in F´  core, we use UPPER_SNAKE_CASE command and parameter names for historical consistency with other missions.  For events and channels we use PascalCase.

LeStarch (changes requested):
This is coming together!  A few more changes.

Also note, we typically resolve codeQL errors on our end to help y'all out because that check is a bit overzealous.

LeStarch:
@zimri-leisher sorry, I was half-way through reviewing when I go side-tracked.  Added more comments.  @ me if you need more insight.

LeStarch:
I'll try to keep an eye on CI and report any other buried errors!

LeStarch:
Framework failure caused by: #2835.  Will rerun.

Svc/Seq/Seq.fpp
zimri-leisher: See my previous comment:
> Any help on this error? image `/workspaces/fprime/fprime/build-fprime-automatic-native-ut/F-Prime/Svc/SeqDispatcher/SeqDispatcherTesterBase.cpp:1170:5: error: no matching function for call to ‘Fw::String::String(Fw::String&)’`
> 
> Looks like I can fix it by removing `explicit` from this line in String.hpp: image


> It seems the problem was that the (previously unused) CmdSeqIn port took in a string like `ref filename: Fw.String` instead of `filename: string size X` Once I fixed it, the unit test compilation errors went away.

Any uses of the previous version of the port seemed to break the autocoder. Would be interested to know if you can replicate this behavior by undoing this change and seeing if the unit tests for SeqDispatcher compile.

Svc/SeqDispatcher/SeqDispatcher.cpp
zimri-leisher: Agreed, I will add a WARNING_HI. I don't think an auto dispatch to another one is a good idea just on the principle that we don't want to be too eager to run a sequence if there is a problem, which there clearly is in this case.

Svc/SeqDispatcher/test/ut/SeqDispatcherTesterHelpers.cpp
zimri-leisher: Ah, I actually never knew what that UT_AUTO_HELPERS was doing. Now I understand!! So many questions about where the init and connectPorts methods were coming from are answered :)

Svc/SeqDispatcher/SeqDispatcherEvents.fppi
zimri-leisher: Okay, will make this change.

zimri-leisher:
Okay, ready for review again.

Svc/Seq/Seq.fpp
LeStarch: Looks like Fw::String lacks a correct copy constructor.

Svc/SeqDispatcher/SeqDispatcherEvents.fppi
```diff
@@ -0,0 +1,31 @@
+event InvalidSequencer(
+    idx: U16
+) \
+    severity warning high \
+    format "Invalid sequence index {}"
+
+event NoAvailableSequencers() \
+    severity warning low \
```
LeStarch: This should probably be a `high` as the user was unable to run sequence.

Svc/SeqDispatcher/docs/sdd.md
```diff
@@ -0,0 +1,66 @@
+# components::SeqDispatcher
```
LeStarch, line 1: If this is to remain empty, then it should be deleted.  Otherwise fill in your design.

LeStarch (approved):
Looks very good!  An EVR could be made `high` instead of `low` because it is a problem with a direct user command.

Tests could be more complete, and SDD is blank.

Other (very minor style points):
1. Typically in compound `if` clauses we put `()` around each sub-phrase.  e.g. `if ((A) || (B))`. for clarity.  Your code is perfectly clear, so no real issue here.
2. Typically we don't return `bool` but rather opt for an enumeration that describes the issue.  This is truly a "did work" or "did not work", so it is ok.

None of this is a blocker.  Good work!  I'd like another set of eyes....let me see if @timcanham will review. In the mean-time, consider that EVR to be made `high`.

zimri-leisher:
I made the changes to the SDD file and event.
I agree, the unit tests are a bit lacking, I would like to add a couple more.
One question: would we like to retroactively come up with requirements for this component?

LeStarch:
Known file CI failure.  I captured the failing random seed, and re-ran.  We have an issue to look into these failures.

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,199 @@
...
+    NATIVE_INT_TYPE portNum,         //!< The port number
+    FwOpcodeType opCode,             //!< Command Op Code
+    U32 cmdSeq,                      //!< Command Sequence
+    const Fw::CmdResponse& response  //!< The command response argument
+) {
+  FW_ASSERT(portNum >= 0 && portNum < SeqDispatcherSequencerPorts, portNum);
+  if (this->m_entryTable[portNum].state !=
+          SeqDispatcher_CmdSequencerState::RUNNING_SEQUENCE_BLOCK &&
+      this->m_entryTable[portNum].state !=
+          SeqDispatcher_CmdSequencerState::RUNNING_SEQUENCE_NO_BLOCK) {
```
garthwatney, line 103: if state is not BLOCK and state is not NO_BLOCK then state should be Available.  I think this code should be simplified to:  if state is Available

timcanham:
@zimri-leisher Do you plan to release `./docs/sdd.md` in a future release?

Svc/SeqDispatcher/SeqDispatcherEvents.fppi
```diff
@@ -0,0 +1,31 @@
...
+
+event NoAvailableSequencers() \
+    severity warning high \
+    format "No available cmd sequencers to dispatch a sequence to"
+
+event UnknownSequenceFinished(
+    idx: U16
+) \
+    severity warning low \
+    format "Sequencer {} said it finished a sequence but we didn't even know it had started one"
```
timcanham: This is funny, but a little too informal. Maybe "Sequencer {} completed with no matching dispatch"

Svc/SeqDispatcher/SeqDispatcher.hpp
```diff
@@ -0,0 +1,95 @@
...
+        // store the state of each sequencer
+        SeqDispatcher_CmdSequencerState state;
+        // store the sequence currently running for each sequencer
+        Fw::String sequenceRunning = "<no seq>";
+    } m_entryTable[SeqDispatcherSequencerPorts];  //!< table of dispatch
+                                                  //!< entries
+
+    NATIVE_INT_TYPE getNextAvailableSequencerIdx();
+
+    bool runSequence(NATIVE_INT_TYPE sequencerIdx,
```
timcanham: @LeStarch: Are we punting on the `NATIVE_INT_TYPE` for now? Should it be `FwIndexType`?

Svc/SeqDispatcher/SeqDispatcherEvents.fppi
```diff
@@ -0,0 +1,31 @@
...
+
+event UnexpectedSequenceStarted(
+    idx: U16,
+    newSequence: string size 240,
+    sequenceInInternalState: string size 240
+) \
+    severity warning high \
+    format "Sequencer {} said it started a sequence {} but we thought it was still running {}"
```
timcanham: Again, a little informal. Maybe "Sequencer {} indicated a started sequence {} while still running {}"

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,199 @@
...
+  for (FwIndexType i = 0; i < SeqDispatcherSequencerPorts; i++) {
+    if (this->isConnected_seqRunOut_OutputPort(i) && 
+        this->m_entryTable[i].state == SeqDispatcher_CmdSequencerState::AVAILABLE) {
+      return i;
+    }
+  }
+  return -1;
+}
+
+bool SeqDispatcher::runSequence(FwIndexType sequencerIdx,
```
timcanham: This doesn't match header prototype. Here it is `FwIndexType`, in header it is `NATIVE_INT_TYPE`

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,199 @@
...
+                                const Fw::StringBase& fileName,
+                                Fw::Wait block) {
+  FW_ASSERT(sequencerIdx >= 0 && sequencerIdx < SeqDispatcherSequencerPorts,
+            sequencerIdx);
+
+  // if the seqRunOut port at this idx isn't connected
+  if (!this->isConnected_seqRunOut_OutputPort(sequencerIdx)) {
```
timcanham: Wondering if these should be `FW_ASSERT()`, since they were checked earlier and these conditions shouldn't have changed.

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,199 @@
...
+  // ok, good to use this sequencer
+  if (block == Fw::Wait::NO_WAIT) {
+    this->m_entryTable[sequencerIdx].state =
+        SeqDispatcher_CmdSequencerState::RUNNING_SEQUENCE_NO_BLOCK;
+  } else {
+    this->m_entryTable[sequencerIdx].state =
+        SeqDispatcher_CmdSequencerState::RUNNING_SEQUENCE_BLOCK;
+  }
+
+  this->m_sequencersAvailable--;
```
timcanham, line 50: Minor efficiency nit, but you could do `--this->m_sequencersAvailable` in `tlmWrite()` call. same with `m_dispatchedCount`.

Svc/SeqDispatcher/test/ut/SeqDispatcherTester.cpp
```diff
@@ -0,0 +1,92 @@
...
+  this->initComponents();
+}
+
+SeqDispatcherTester ::~SeqDispatcherTester() {}
+
+// ----------------------------------------------------------------------
+// Tests
+// ----------------------------------------------------------------------
+
+void SeqDispatcherTester ::testDispatch() {
```
timcanham, line 28: These are pretty basic. It would be nice to really nuke this with the random rules-based tester.

timcanham (commented):
@zimri-leisher This looks really great! Minor comments only. If you don't wish to make any changes, I'll approve.

Svc/SeqDispatcher/SeqDispatcherEvents.fppi
zimri-leisher: It's a personal style thing, I don't shy away from being informal because it tends to be direct and clear. That said, I think using pronouns here is a bit much, I can change it.

Svc/SeqDispatcher/SeqDispatcherEvents.fppi
zimri-leisher: Agreed, will change.

Svc/SeqDispatcher/SeqDispatcher.cpp
zimri-leisher: That was a mistake, thanks for catching it. Will fix.

Svc/SeqDispatcher/SeqDispatcher.cpp
zimri-leisher: If you don't mind, I'd prefer to keep it this way. There are two operations (increment/decrement, and write to tlm) and so they should happen on two different lines.

Svc/SeqDispatcher/test/ut/SeqDispatcherTester.cpp
zimri-leisher: Would love to do this! Is there an example of that in the code I can look at?

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,199 @@
...
+    // we were not aware that this sequencer was running. ground must have
+    // directly commanded that specific sequencer
+
+    // update the state
+    this->m_entryTable[portNum].state =
+        SeqDispatcher_CmdSequencerState::RUNNING_SEQUENCE_NO_BLOCK;
+    this->m_entryTable[portNum].sequenceRunning = fileName;
+    this->m_sequencersAvailable--;
+    this->tlmWrite_sequencersAvailable(this->m_sequencersAvailable);
+  }
```
garthwatney: If a Sequence is in Available state and you receive a Start, that seems to be an anomaly.  I would suggest sending an EVR Warning instead of just silently transitioning to a NonBlock state.

Svc/SeqDispatcher/SeqDispatcher.cpp
timcanham: No problem!

Svc/SeqDispatcher/SeqDispatcher.hpp
timcanham: Resolved with fix below.

Svc/SeqDispatcher/SeqDispatcher.hpp
```diff
@@ -0,0 +1,95 @@
...
+    // number of errors from dispatched sequences (CmdResponse::EXECUTION_ERROR)
+    U32 m_errorCount = 0;
+    // number of sequencers in state AVAILABLE
+    U32 m_sequencersAvailable = SeqDispatcherSequencerPorts;
+
+    struct DispatchEntry {
+        FwOpcodeType opCode;  //!< opcode of entry
+        U32 cmdSeq;
+        // store the state of each sequencer
+        SeqDispatcher_CmdSequencerState state;
```
garthwatney, line 66: Next month (September 2024) Fprime and FPP will be released to directly support state machines within a component.
The Phase 1 delivery will allow developers to instantiate an externally defined state machine in FPP.  The externally defined state machine can be specified in the open source QM modeling tool or in PlantUML text and the STARS (https://github.com/JPLOpenSource/STARS) state machine autocoder will generate the code.

My suggestion is to specify the state of the Sequencer as a state machine model that is then integrated into the component.  Here is a cut at the state machine specified using the QM modeling tool.  The developer would then write the code snippets to implement the state machine functions.
image

In addition the state machine can just be specified in PlantUML and then run through the same STARS autocoder. 

```
@startuml

[*] --> Available

state Available {
    Available:Internal: Done / UnknownSeqFinished(); incrSeqAvailable(e) 
}

state Running {
    state c1 <<choice>>
    [*] --> NonBlock
    Running:Internal: Start [seqRunningNotFile(e)] / unexpectedSeqStart(e)
    Running:Internal: Run / invalidSequencer(e); sendExecutionError(e)

    state NonBlock {
    }

    state Block {
    }
}

Available --> c1 : Run /seqRunOut(e); decSeqAvailable(e)
c1 --> Block
c1 --> NonBlock : [NoWait(e)]
NonBlock --> Available : Done / incrSeqAvailable(e)
Block --> Available : Done / sendCmdResponse(e); incrSeqAvailable(e)
Available --> Running : Start / decSeqAvailable(e)



@enduml

```

The graphics of the PlantUML can be automatically rendered by python -plantuml or in Visual Studio as:

image

In Phase 2 of the state machine delivery, the state machine model will be specified directly in the component FPP language eliminating the need to use STARS to define the external model or generate the code.

If you do not want to go down this route then at a minimum include the state machine diagrams in the SDD.

garthwatney (approved):
Overall looks very good.  I have included my review comments.

Svc/SeqDispatcher/test/ut/SeqDispatcherTester.cpp
thomas-bc: @zimri-leisher I'll respond since I see your comment and I also have been learning about the rule-based tester recently.

Here is the source / README : https://github.com/nasa/fprime/tree/devel/STest

If you search/grep for "STest" across the codebase you'll find a bunch of modules and components that use it. `STest::RandomScenario` is the random one.

This makes me realize, we really should write a HowTo guide on how to use this...

timcanham: Here is an example: https://github.com/nasa/fprime/tree/devel/Svc/DpManager/test/ut

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -0,0 +1,210 @@
...
+}
+
+void SeqDispatcher::seqDoneIn_handler(
+    NATIVE_INT_TYPE portNum,         //!< The port number
```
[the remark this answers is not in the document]
LeStarch: As @timcanham pointed out, this needs to be `FwIndexType`.

Svc/SeqDispatcher/SeqDispatcher.hpp
LeStarch: Since this component is in its final review (and needed by customers).  I recommend we track this as a issue to work in the future.

Svc/SeqDispatcher/SeqDispatcher.cpp
LeStarch: I agree with this, can we fix?

Svc/SeqDispatcher/SeqDispatcher.cpp
LeStarch: This catches the case where the state enum is in an invalid state.  If we do this simplification, we probably should also do:
```
FW_ASSERT(this->m_entryTable[portNum].state.isValid());
```

LeStarch: Note, that goes before the if.

Svc/SeqDispatcher/SeqDispatcher.cpp
LeStarch: This makes sense, but I don't think this is a show stopper.

Svc/SeqDispatcher/SeqDispatcher.cpp
LeStarch: Great!

Svc/SeqDispatcher/test/ut/SeqDispatcherTester.cpp
LeStarch: This could also be a future work issue.

LeStarch:
@zimri-leisher I noted some things that could be future work, some things that would be nice to do now.  Would you please note those items you intend to close-out for this PR, and which you'd prefer to see as a future work item?

That way I know what to confirm is done before merging and what I should track as future work.

Thanks for all the hard work!

Svc/SeqDispatcher/SeqDispatcher.cpp
garthwatney: I'm not sure I understand how this catches an invalid state.  The condition checks if the state is Available.  It seems awkward to check if the state is not in every other state that is not Available.

Svc/SeqDispatcher/SeqDispatcher.hpp
garthwatney: I recommend that the model diagram be placed in the SDD.

Svc/SeqDispatcher/SeqDispatcher.cpp
zimri-leisher: Yes, will fix.

Svc/SeqDispatcher/SeqDispatcher.cpp
zimri-leisher: @garthwatney , I think @LeStarch is saying that the original way I did the check catches the case in which the state is invalid (i.e. none of the given options). That said, I do see how it would be more direct and readable to just check if it's available. Happy to make this change and also add the isValid check.

Svc/SeqDispatcher/test/ut/SeqDispatcherTester.cpp
zimri-leisher: Yeah, in the interest of getting this merged, I think I will just increase the coverage of the normal unit tests.

Svc/SeqDispatcher/SeqDispatcher.cpp
zimri-leisher: I guess I don't see why the software should prefer to die rather than just warning that something failed. Is this a philosophical FPrime thing?

Svc/SeqDispatcher/SeqDispatcher.hpp
zimri-leisher: I will add the model diagrams you produced to the SDD, if that's okay with you.

Svc/SeqDispatcher/SeqDispatcher.cpp
garthwatney: The asserts are intended to catch software bugs.  If some internal inconsistency reflects a software bug, then yes it is preferable for the software to die (and in flight gets reset) that try to forge ahead in an unknown state.

zimri-leisher: Okay, that makes sense. Fixed.

zimri-leisher:
Okay, I fixed most of the new comments. Any pointers for the unit tests?

LeStarch (approved):

LeStarch:
I am going to merge and take action for adding rules testing!

Meganbrentshandacrain (approved):

PR #2742, opened 2024-05-23
Fix CI bash script for macOS

JohanBertrand:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Small change of the bash script to display errors happening on the CI. The bash version used on the Mac OS CI does not support negative values (See https://github.com/nasa/fprime/actions/runs/8902005968/job/24448082623 for the error, and https://github.com/nasa/fprime/actions/runs/8903107367/job/24450396620 for the fix)

### DIFF
--- a/ci/helpers.bash
+++ b/ci/helpers.bash
@@ -38,7 +38,10 @@ function fail_and_stop()
     if [ -f "${LASTLOG_ERR}" ]
     then
         # Check if a related stdout log exist
-        LASTLOG_OUT="${LASTLOG_ERR::-7}out.log"
+        # Mac OS does not support negative length
+        # so LASTLOG_ERR::-7 cannot be used here
+        LASTLOG_ERR_LENGTH=${#LASTLOG_ERR}
+        LASTLOG_OUT="${LASTLOG_ERR:0:LASTLOG_ERR_LENGTH-7}out.log"
 
         if [ -f "${LASTLOG_OUT}" ]
         then


LeStarch (approved):

LeStarch:
I am glad you remembered this!

JohanBertrand:
@LeStarch Is there anything else needed in this PR?
The CI did not pass, but I don't think it's because of the current changes

thomas-bc (approved):
Yeah that link died and was removed in another PR. Thanks for making this change!

thomas-bc:
The RPI tests seem to be failing consistently though... 🤔

JohanBertrand:
Interesting. Is there any link between this bash script and the RPI integration tests?

Otherwise, are there any artefacts from previous jobs that could have been impacted?

It used to pass without any problem in a previous PR: https://github.com/nasa/fprime/pull/2709/commits/c98a07056ed0cd497089c5cf341a4c3ebce21b61

JohanBertrand:
It seems like the file descriptor stays at -1 during the test. I don't expect it to come from this bash script, and I don't think that's expected in the integration test?

```
[GDS] Received EVR: 2024-05-28 13:18:20: fatalAdapter.AF_ASSERT_0 (16896) (2(0)-1716927500:896756) EventSeverity.FATAL : Assert in file /home/runner/work/fprime/fprime/fprime/Drv/LinuxGpioDriver/LinuxGpioDriverCompon, line 276
```

From `Drv/LinuxGpioDriver/LinuxGpioDriverComponentImpl.cpp`:
```
void LinuxGpioDriverComponentImpl ::
    gpioWrite_handler(
        const NATIVE_INT_TYPE portNum,
        const Fw::Logic& state
    )
  {
      FW_ASSERT(this->m_fd != -1);

      NATIVE_INT_TYPE stat;

      stat = gpio_set_value(this->m_fd,(state == Fw::Logic::HIGH) ? 1 : 0);

      if (0 != stat) {
          this->log_WARNING_HI_GP_WriteError(this->m_gpio,stat);
          return;
      }
  }
```

LeStarch:
This error means the RPIs need to be rebooted.....some orphaned process is holding onto the GPIO pins.  They are supposed to reboot after each test, but this seems to be breaking.

Regardless this PR can be merged.

PR #2745, opened 2024-05-23
Update FPP to v2.1.0a11

bocchino:
* Update FPP to v2.1.0a11
* Make corresponding changes to client code

## FPP Changes

* Document dictionary generation in User's Guide by @bocchino in https://github.com/nasa/fpp/pull/422
* Update example in spec by @bocchino in https://github.com/nasa/fpp/pull/428
* Revise string code generation by @bocchino in https://github.com/nasa/fpp/pull/426
* Invalidate the buffer when sending a container by @bocchino in https://github.com/nasa/fpp/pull/427
* Add constant expressions for record sizes by @bocchino in https://github.com/nasa/fpp/pull/429
* Fixing removing the nullptr from join calls by @LeStarch in https://github.com/nasa/fpp/pull/425


bocchino:
CI is complaining because a link in a Markdown file in the cmake docs is unavailable. The issue is unrelated to this PR.

docs/Design/data-products.md
```diff
@@ -48,7 +51,7 @@ components:
    These components produce data products and are typically mission-specific.
    For example, they may produce science data.
 
-1. Standard F Prime components for managing data products.
+1. Standard F' components for managing data products.
```
LeStarch: This should say `F´`  not `F'`

Fw/Types/test/ut/TypesTest.cpp
```diff
@@ -607,6 +608,14 @@ TEST(SerializationTest, Serialization1) {
...
     str2.deserialize(buff);
     ASSERT_EQ(str1, str2);
+
+    // serialize string template
+    Fw::StringTemplate<80> strTmpl1;
+    Fw::StringTemplate<40> strTmpl2;
+    buff.resetSer();
+    strTmpl1.serialize(buff);
+    strTmpl2.deserialize(buff);
+    ASSERT_EQ(strTmpl1, strTmpl2);
```
LeStarch, line 618: Shouldn't this test include a strTmpl1 is too large for the 40 test?
Why not use real strings rather than uninitialized strings.

LeStarch (changes requested):
Two minor comments, otherwise looks great!

Fw/Types/test/ut/TypesTest.cpp
bocchino: Good points. I switched both strings to size 80 and initialized them. This corresponds to what we are doing for non-templated strings, and it seems good enough for a basic test that the template works.

docs/Design/data-products.md
bocchino: OK, I switched all instances of F' to F Prime, which I think is also acceptable. I have some questions about this, which we can resolve offline:
1. What is that forward tick character, and how do I type it?
2. Since forward tick seems to be a Unicode and not an ASCII character, is there a way to represent it in ASCII? Otherwise a Unicode text editor is required.
3. Why are we using forward tick instead of prime ' when we write F'?

bocchino: Note that if I were writing F' in Markdown as math I would write `_F'_` which resolves to _F'_.

thomas-bc: Rob, I forwarded you an email from Jeff regarding guidance on spelling F Prime, or F´ for short with the [prime symbol](https://en.wikipedia.org/wiki/Prime_(symbol)). This is what we are trying to adhere to. Both F´ or F Prime would work here.

On MacOS, the ´ character can be typed with Option+E.

bocchino: Thanks. Note that option-e generates an acute accent, not a prime symbol. There is a Unicode prime symbol, but it looks like it has no shortcut.

Single quote: F'
Prime symbol: Fʹ
Acute accent: F´

thomas-bc:
CI failure (markdown link not found) was fixed in https://github.com/nasa/fprime/pull/2746 and can safely be ignored

thomas-bc (approved):
LGTM! Ready to merge when we are

LeStarch (approved):

Issue #2584, opened 2024-03-12
Create Events Interface


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Make an interface out of the standard Event ports. See #2255 for an example interface change set.

The following three ports should be put into an interface at `Fw/Interfaces/EventsInterface.fppi` such that component implementers can quickly use them.

https://github.com/nasa/fprime/blob/a73bff28617b96736c0e3e222e4c6e1dea1c6f5e/Ref/SignalGen/SignalGen.fpp#L50-L54

menaman123:
Hi, I was wondering if this issue still needs to be worked on? This would be my first contribution.

Issue #2578, opened 2024-03-12
Create F´ Interfaces From Existing Concepts


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Many F´ concepts can be reworked as F´ interfaces. These are specialized `.fppi` files that define a set of ports.  An example would be: https://github.com/nasa/fprime/blob/devel/Drv/Interfaces/ByteStreamDriverInterface.fppi.

Interfaces are expected to live in a folder `Interfaces` under a given package (e.g. `Drv/Interfaces` or `Svc/Sinterfaces`).

> New interfaces should use the filename pattern `*.interface.fppi` and not just `.fppi`. This is a change since the example was created.

Here is a list of interfaces that should be created:
- [x] Event Ports: https://github.com/nasa/fprime/issues/2584
- [x] Telemetry Ports: https://github.com/nasa/fprime/issues/2585
- [x] Command Ports: https://github.com/nasa/fprime/issues/2586
- [x] `Svc::ComInterface`: https://github.com/nasa/fprime/issues/2257
- [x] `Drv::I2cDriver`: https://github.com/nasa/fprime/issues/2580
- [x] `Drv::GpioDriver`: https://github.com/nasa/fprime/issues/2581
- [x] `Drv::SpiDriver`: https://github.com/nasa/fprime/issues/2582
- [x] `Drv::TickDriver`: https://github.com/nasa/fprime/issues/2583
- [x] UART and TCP Driver: https://github.com/nasa/fprime/issues/2255
- [x] Time: https://github.com/nasa/fprime/issues/2579, https://github.com/nasa/fprime/issues/2256

> I2c, Gpio, Spi, and ByteStreamDriver items should also have a `*UserInterface.fppi` for callers/managers using these inferaces.  These issues have not been created.  These are the same ports with the directions reversed.

LeStarch:
@kevin-f-ortega is this done?  The two outstanding bits were already implemented.

kevin-f-ortega:
@LeStarch, yes this is done now

Issue #2255, opened 2023-08-24
Establish FPP Interfaces for ByteSteamDriverModel


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

ByteSteamDriverModel is used as a key feature in F´ in that it drives the communication layer.  Instead of using the model and different implementers, it should be made into an interface.

This has the primary advantage that this interface can be augmented while still fitting correctly into the topology.

PR #2749, opened 2024-05-27
Add Events FPP interface

menaman123:
…e set

| | |
|:---|:---|
|**Create Events Interface #2584**|  |
|**No**|  |
|**No_**|  |

---
## Change Description

Created an Event Interface out of the standard event ports using #2255 as an example. There were only 2 event ports compared to 3 stated in the issue description. 

## Rationale

Part of #2578.

None

## Future Work

Must work on the other interfaces.


Ref/SignalGen/SignalGen.fpp
```diff
@@ -92,6 +86,8 @@ module Ref {
     # ----------------------------------------------------------------------
 
     include "Events.fppi"
+    include "../Interfaces/EventsInterface.fppi"
```
thomas-bc: This is the events section where events are to be defined. Interfaces should likely be their own section, or another type of grouping.
@LeStarch do we have a convention for this? Should we?

thomas-bc: After we have an agreement, we probably want to do the same change on all components across `Ref/` and `RPI/`

LeStarch: I agree. We should create an interfaces section and place this there.

Also, `"../Interfaces/EventsInterface.fppi"` should read `"../../Fw/Interfaces/EventsInterface.fppi"`

menaman123:
I added an interfaces section

LeStarch (approved):

LeStarch:
I corrected the merge conflict with the other interface PR we merged.

thomas-bc (approved):

menaman123:
Thank you so much for helping me out in my first adventure in open source development. Is there anything else I should do?

LeStarch:
Nope! Welcome to the Open Source world, and thanks for completing this!

Issue #2585, opened 2024-03-12
Create Telemetry/Channel Interface


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Make an interface out of the standard Channel ports. See #2255 for an example interface change set.

The following three ports should be put into an interface at `Fw/Interfaces/ChannelInterface.fppi` such that component implementers can quickly use them.

https://github.com/nasa/fprime/blob/a73bff28617b96736c0e3e222e4c6e1dea1c6f5e/Ref/SignalGen/SignalGen.fpp#L56-L57

PR #2750, opened 2024-05-27
created an interface for the standard Channel ports using #2255 as an…

menaman123:
… example interface change set

| | |
|:---|:---|
|** #2585**|  |
|**No**|  |
|**No**|  |

---
## Change Description

Create Telemetry/Channel Interface #2585

## Rationale

Part of Create F´ Interfaces From Existing Concepts #2578

None

## Future Work

Continue #2578


LeStarch:
The file is empty.  Can you fill it in?

Fw/Interfaces/ChannelInterfaces.fppi
```diff
@@ -0,0 +1,2 @@
+@ Telemetry port
+telemetry port tlmOut
```
LeStarch, line 2: Last step is to delete line 68-69 in `SignalGen.fpp` as your interface adds that port!

Ref/SignalGen/SignalGen.fpp
```diff
@@ -93,6 +93,12 @@ module Ref {
 
     include "Events.fppi"
 
+    # ----------------------------------------------------------------------
+    # Interfaces
+    # ----------------------------------------------------------------------
+
+    include "../../Fw\Interfaces\ChannelInterfaces.fppi"
```
LeStarch: You should use `/` instead of `\` as well.

thomas-bc: Should the channel interface be used for all components in Ref and RPI as well? Or are we happy to keep it hybrid?

LeStarch: Not just yet.  Let's get it in and proven in SignalGen, then we can slowly refactor.

LeStarch (approved):

thomas-bc:
Closing in favor of https://github.com/nasa/fprime/pull/2757/files

PR #2757, opened 2024-06-04
Added Channel Interface for Telemetry

### LINKED ISSUE #2585: Create Telemetry/Channel Interface

menaman123:
| | |
|:---|:---|
|**#2585**|  |
|**No**|  |
|**No**|  |

---
## Change Description

Added ChannelInterface.fppi and included the telemetry within it.

## Rationale

Part of #2578.

## Future Work

Work on other interfaces


thomas-bc:
Some tests are failing on `devel` branch: https://github.com/nasa/fprime/actions/runs/9357605666/job/25757702996
An oversight on our end.

Would you please add `CommandInterface.fppi` `ChannelInterface.fppi` and `EventInterface.fppi` to 
- https://github.com/nasa/fprime/blob/7d10786a10339d22bacb315f733296fb4b3465c3/cmake/test/src/test_unittests.py#L133
- https://github.com/nasa/fprime/blob/7d10786a10339d22bacb315f733296fb4b3465c3/cmake/test/src/test_ref_shared.py#L112

thomas-bc (commented):
Some tests are failing on `devel` branch: https://github.com/nasa/fprime/actions/runs/9357605666/job/25757702996
An oversight on our end.

Would you please add `CommandInterface.fppi` `ChannelInterface.fppi` and `EventInterface.fppi` to 
- https://github.com/nasa/fprime/blob/7d10786a10339d22bacb315f733296fb4b3465c3/cmake/test/src/test_unittests.py#L133
- https://github.com/nasa/fprime/blob/7d10786a10339d22bacb315f733296fb4b3465c3/cmake/test/src/test_ref_shared.py#L112

thomas-bc:
@LeStarch we probably should extend triggers on https://github.com/nasa/fprime/blob/devel/.github/workflows/cmake-test.yml. The tests are pretty quick so I would be in favor of making it run on every PR

Drv/LinuxSpiDriver/LinuxSpiDriver.fpp
```diff
@@ -6,8 +6,8 @@ module Drv {
     # General ports
     # ----------------------------------------------------------------------
 
-    sync input port SpiReadWrite: Drv.SpiReadWrite
-
+    sync input port SpiReadWrite: Drv.SpiReadWrite 
+
```
menaman123: This is from a different commit that i did not want to be added to this PR

cmake/test/src/test_ref_shared.py
```diff
@@ -109,7 +109,15 @@ def test_ref_module_info(REF_BUILD):
...
     ], "Did not find expected sources"
-    expected_ac = ["SignalGen.fpp", "Commands.fppi", "Events.fppi", "Telemetry.fppi"]
+    expected_ac = [
+        "SignalGen.fpp", 
+        "Commands.fppi", 
+        "Events.fppi", 
+        "Telemetry.fppi",
+        "CommandInterface.fppi", 
+        "ChannelInterface.fppi", 
+        "EventsInterface.fppi",
```
thomas-bc, line 119: We should probably choose between singular or plural for the 3 interfaces here and adapt filenames accordingly

menaman123:
@thomas-bc do you happen to know why the format keeps failing?

LeStarch:
@menaman123 format is failing because you have spaces ` ` after the commas in that modified python file.

@thomas-bc yes, we can run cmake on every commit.  It is faster now.

menaman123:
@LeStarch I think i fixed the spaces. I hope it doesnt fail again

cmake/test/src/test_ref_shared.py
LeStarch: Yes, we probably should rename `Events` interface to `Event` interface...but that would be another PR, no?

LeStarch (approved):

LeStarch:
I went ahead and touched up the formatting and reverted the file you did not want.  Hopefully it will build cleanly now.

Issue #2074, opened 2023-06-08
Hardcoded buffer size in TcpClient/TcpDriver/Udp


thomas-bc:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

See https://github.com/nasa/fprime/blob/da4b1daae77f46a940733fbb20a4be8efd3a89b8/Drv/TcpClient/TcpClientComponentImpl.cpp#L50-L51

We should do like UARTDriver (see [here](https://github.com/nasa/fprime/blob/da4b1daae77f46a940733fbb20a4be8efd3a89b8/Drv/LinuxUartDriver/LinuxUartDriver.cpp#L46-L47)) and configure the size at configure() time so that it can be configured for each instance in the topology. Set a default size of 1024 so that it doesn't break any existing code.

LeStarch:
Also in UDP and possibly UART.  Should be a configuration parameter (on startup).

menaman123:
So would a fix be having an allocation parameter in the getBuffer method?

LeStarch:
@menaman123 the fix would be to add an "allocation size" to the `configure` method:

https://github.com/nasa/fprime/blob/da4b1daae77f46a940733fbb20a4be8efd3a89b8/Drv/TcpClient/TcpClientComponentImpl.cpp#L32-L36

bdshenker:
This issue appears to have been fixed in commit 7FF002C by menamen123:

```
TcpClientComponentImpl.hpp
  PROTECTED: FwSizeType m_allocation_size;
  SocketIpStatus configure(const char* hostname,
                             const U16 port,
                             const U32 send_timeout_seconds = SOCKET_SEND_TIMEOUT_SECONDS,
                             const U32 send_timeout_microseconds = SOCKET_SEND_TIMEOUT_MICROSECONDS,
                             FwSizeType buffer_size = 1024);
TcpClientComponentImpl.cpp
  Fw::Buffer TcpClientComponentImpl::getBuffer() {
      return allocate_out(0, static_cast<U32>(m_allocation_size));
  }
```
If I am correct then this ticket can now be closed per GIT ticket # 2759.

LeStarch:
@bdshenker the above ticket did resolve the problem in TcpClient, but not TcpServer nor Udp.

[...]

PR #2759, opened 2024-06-06
	modified:   Drv/TcpClient/TcpClientComponentImpl.cpp

menaman123:
| | |
|:---|:---|
|**_Related Issue(s)_**| #2074  |
|**_Has Unit Tests (y/n)_**|  no |
|**_Documentation Included (y/n)_**| no |

---
## Change Description

Changed the getBuffer function to have a tunable size. There is a default size of 1024 so that it does not break any existing code.

## Rationale

This removes the need of having to fork off of the fprime repo or autopatch everytime.

There is no unit tests to see if it'll break anything. Unsure where to begin to test to see.

### DIFF
--- a/Drv/TcpClient/TcpClientComponentImpl.cpp
+++ b/Drv/TcpClient/TcpClientComponentImpl.cpp
@@ -42,8 +42,13 @@ IpSocket& TcpClientComponentImpl::getSocketHandler() {
     return m_socket;
 }
 
-Fw::Buffer TcpClientComponentImpl::getBuffer() {
-    return allocate_out(0, 1024);
+// Define a constant for the default buffer size
+static const size_t DEFAULT_BUFFER_SIZE = 1024;
+
+// Modify the function signature to accept the buffer size as an argument
+Fw::Buffer TcpClientComponentImpl::getBuffer(size_t bufferSize = DEFAULT_BUFFER_SIZE) {
+    // Allocate the buffer with the specified size
+    return allocate_out(0, bufferSize);
 }
 
 void TcpClientComponentImpl::sendBuffer(Fw::Buffer buffer, SocketIpStatus status) {
--- a/Drv/TcpClient/TcpClientComponentImpl.cpp
+++ b/Drv/TcpClient/TcpClientComponentImpl.cpp
@@ -10,6 +10,7 @@
 //
 // ======================================================================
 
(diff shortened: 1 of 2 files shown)


LeStarch:
@menaman123 sorry, I didn't catch your question.  I do like that you have defaulted this, but I'd move the allocation size to the `configure` method.  Store the values a a member (e.g. m_allocation_size) and then pass that member to the call.

Note: We should do separate PRs for TcpServer and UDP too (once this one is approved)!

menaman123:
@LeStarch Should i modify the getBuffer such that it returns the member allocate(0, m_allocation_size)?

menaman123:
@LeStarch I changed the header file, I was wondering should the default buffer size be hardcoded as 1024 or a variable called DEFAULT_BUFFER_SIZE should be created instead?

LeStarch:
One last change: use `FwSizeType` instead of `size_t` for these sizes. In F´ we use `FwSizeType` as the type of sizes.

I am ok with the hardcoded 1024.  It is a default after all, and easily changed.  Thanks for this change, we've needed it for some time!

LeStarch:
You missed the member variable change to `FwSizeType` too.  I made that edit for you so you don't have to keep chasing nits.

Your code looks good.  Normally, we'd prefix all member variables with `this->m_allocation_type` .... but this file does not do that anywhere and you were correct to maintain style within this file.

Once CI passes, we'll merge this!

menaman123:
Sweet thank you so much for the insight and teaching me the way!! @LeStarch

LeStarch:
The issue with CI is this: F´ is in the process of transitioning from fixed-width types (e.g. U32) to project configurable types (e.g. FwSizeType).  In this case `Fw::Buffer` uses an older style (`U32`) and your changes now update to the new style.

Here is how to fix this. Add a cast to the allocation call to convert types:

```
    return allocate_out(0, static_cast<U32>(m_allocation_size));
```

Add this check to the `configure` method:

```
    FW_ASSERT(buffer_size <= std::numeric_limits<U32>::max(), buffer_size);
```
This check ensures that the configured buffer size fits within the limit of the old structure.  We do this in `configure` so that the assert trips as early as possible.

You may need to include these headers (if not already done so):

```
#include <limits>
```

LeStarch (approved):

Issue #2582, opened 2024-03-12
Create Interface from LinuxSpiDriver


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Make an interface out of Drv/LinuxSpiDriver protocol. See #2255 for an example interface change set.

The following three ports should be pulled out of `Drv::LinuxSpiDriver` into `Svc/Interfaces/SpiInterface.fppi` such that other Spi implementations can use it.

https://github.com/nasa/fprime/blob/a73bff28617b96736c0e3e222e4c6e1dea1c6f5e/Drv/LinuxSpiDriver/LinuxSpiDriver.fpp#L9

Note: annotations should be added to explain what each port does.

menaman123:
Hello, I was wondering if there is documentation on what this port does. Im assuming it reads and writes?

menaman123:
Would the description be like this:
@ Port to perform synchronous read and write operations over the SPI bus

LeStarch:
I would say "Port to perform a synchronous read/write operation over the SPI bus"...but you basically have it.  The nuance I am trying to capture here is that SPI calls are both read and write at the same time.

menaman123:
Follow up question, do i need to add anything to the cmake file so that there wont be any sort of failures?

PR #2762, opened 2024-06-07
Interface from LinuxSpiDriver

menaman123:
| | |
|:---|:---|
|**_Related Issue(s)_**|  #2582|

---
## Change Description

SpiReadWrite port pulled out into a Spi Interface.

## Rationale

Part of https://github.com/nasa/fprime/issues/2578.


Drv/LinuxSpiDriver/LinuxSpiDriver.fpp
```diff
@@ -6,8 +6,6 @@ module Drv {
     # General ports
     # ----------------------------------------------------------------------
```
thomas-bc, line 10: You will need to include SpiInterface.fppi here

thomas-bc (changes requested):

Svc/Interfaces/SpiInterface.fppi
```diff
@@ -0,0 +1,2 @@
+@ Port to perform a synchronous read/write operation over the SPI bus
```
LeStarch: This interface should go in `Drv/Interfaces`.

If an interface is used in service it goes in `Svc/Interfaces`, if it is used in drivers `Drv/Interfaces`, if it is for the whole framework `Fw/Interfaces`.

LeStarch (changes requested):

LeStarch (approved):
Looks good to me.  I changed the section banner to be called Interfaces because I like that style and there are no more ports in that section.

thomas-bc (approved):

PR #2777, opened 2024-06-12
feat: Add prepend flag to autocode register macro

mosa11aei:
| | |
|:---|:---|
|**_Related Issue(s)_**| n |

---
## Change Description

This pull request implements a flag in the API.cmake file macro `register_fprime_list_helper` to be able to prepend target file paths to a target list. This is in the form of a boolean argument in that macro.

## Rationale

The rationale for this addition is to be able to pre-register autocoders before other autocoders, in the event that one would like to be able to register an autocoder to execute before any other autocoder.


cmake/FPrime.cmake
```diff
@@ -141,7 +141,7 @@ macro(fprime_setup_override_targets)
 endmacro(fprime_setup_override_targets)
 
 macro(fprime_initialize_build_system)
-    cmake_minimum_required(VERSION 3.16)
+    cmake_minimum_required(VERSION 3.13)
```
thomas-bc: Why is this changed?

mosa11aei: Good catch, thank you! I am unsure why it changed; it shouldn't have. 3.16 is what is in devel right now.

LeStarch (approved):

Issue #2722, opened 2024-05-10
Refactor Os::Mutex into CMake Selection


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

- [x] Refactor Os::Mutex into the Interface, Os::Mutex, and Implementation classes (following pattern of Os::File)
- [x] Implement Posix implementation
- [x] Implement Stub implementation
- [x] Implement Test stub implementation
- [x] Build common rules-based tests
- [x] Build specific posix tests
- [x] Build stub interface tests
- [x] Setup default posix implementation for Posix platforms
- [ ] Document

PR #2790, opened 2024-06-27
Refactor OS::Mutex in CMake selection

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2722 |
|**_Documentation Included (y/n)_**| OSAL |

---
## Change Description

Refactor the Mutex part of the OSAL layer to use the new system!

## Rationale

See https://github.com/nasa/fprime/issues/2722


cmake/platform/Linux.cmake
```diff
@@ -11,6 +11,13 @@ if (NOT DEFINED FPRIME_USE_BAREMETAL_SCHEDULER)
 endif()
 choose_fprime_implementation(Os/File Os/File/Posix)
 choose_fprime_implementation(Os/Task Os/Task/Posix)
+choose_fprime_implementation(Os/Mutex Os/Mutex/Posix)
```
thomas-bc, line 14: `/` kind of remind of a file path, but in this case it should maybe correspond to the actual path, which is e.g. Os/Posix/Task and not Os/Task/Posix
We may want to either (a) rename all those targets to match the path in which they are defined e.g. Os_Mutex_Posix becomes Os_Posix_Mutex, or (b) use the actual name Os_Mutex_Posix (with `_` instead of `/`) to avoid confusion
Thoughts?

config/FpConfig.h
```diff
@@ -394,11 +394,11 @@ typedef FwIndexType FwQueueSizeType;
 // OS configuration
 
 #ifndef FW_HANDLE_MAX_SIZE
-#define FW_HANDLE_MAX_SIZE 24  //!< Maximum size of a handle for OS resources (files, queues, locks, etc.)
+#define FW_HANDLE_MAX_SIZE 72  //!< Maximum size of a handle for OS resources (files, queues, locks, etc.)
```
thomas-bc, line 397: Should there be a handle size per type of resource? Posix mutexes need 72, where it seems files and tasks only needed 24

thomas-bc:
@LeStarch ready for review when you are - no rush

Os/CMakeLists.txt
```diff
@@ -96,7 +95,11 @@ if (FPRIME_USE_BAREMETAL_SCHEDULER)
     list(APPEND SOURCE_FILES "${CMAKE_CURRENT_LIST_DIR}/Baremetal/SystemResources.cpp")
 endif()
 register_fprime_module()
+
 require_fprime_implementation(Os/File)
+# require_fprime_implementation(Os/Task) # should be added in
```
thomas-bc, line 100: Uncommenting this was causing errors, I think because of the `Os` MOD_DEPS in Os/Tasks, but haven't been able to fix so I kept it commented out here for further exploration

Os/Posix/CMakeLists.txt
```diff
@@ -65,4 +65,29 @@ set(UT_MOD_DEPS
...
 )
 choose_fprime_implementation(Os/Task Os_Task_Posix)
-register_fprime_ut(PosixTaskTest)
\ No newline at end of file
+register_fprime_ut(PosixTaskTest)
+
+#### Os/Mutex/Posix Section ####
+set(SOURCE_FILES
+    "${CMAKE_CURRENT_LIST_DIR}/Mutex.cpp"
+    # "${CMAKE_CURRENT_LIST_DIR}/DefaultMutex.cpp"
```
thomas-bc: correct - delete and make sure it's not in other modules as well

Os/Mutex.cpp
```diff
@@ -0,0 +1,43 @@
...
+    this->m_delegate.lock();
+}
+
+void Mutex::unLock() {
+    FW_ASSERT(&this->m_delegate == reinterpret_cast<MutexInterface*>(&this->m_handle_storage[0]));
+    this->m_delegate.unLock();
+}
+
+Mutex::Status Mutex::take() {
+    FW_ASSERT(&this->m_delegate == reinterpret_cast<MutexInterface*>(&this->m_handle_storage[0]));
```
thomas-bc, line 35: implement take/release calling the others here

Os/Posix/Mutex.cpp
```diff
@@ -1,51 +1,60 @@
...
+namespace Posix {
+namespace Mutex {
 
-    Mutex::Mutex() {
-        pthread_mutex_t* handle = new(std::nothrow) pthread_mutex_t;
-        FW_ASSERT(handle != nullptr);
+PosixMutex::PosixMutex() : Os::MutexInterface(), m_handle() {
+    // set attributes
+    pthread_mutexattr_t attribute;
+    pthread_mutexattr_init(&attribute);
```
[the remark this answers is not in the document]
thomas-bc: fix this static analysis alert, do an assert

Os/Posix/test/ut/PosixMutexTests.cpp
```diff
@@ -0,0 +1,66 @@
...
+// ----------------------------------------------------------------------
+// Posix Test Cases
+// ----------------------------------------------------------------------
+
+// Attempt to delete a locked mutex - expect an assertion
+TEST_F(FunctionalityTester, PosixDeleteLockedMutex) {
+    Os::Test::Mutex::Tester::LockMutex lock_rule;
+    Os::Test::Mutex::Tester::UnlockMutex unlock_rule;
+    lock_rule.apply(*tester);
+    ASSERT_DEATH_IF_SUPPORTED(delete tester.release(), Os::Test::Mutex::Tester::ASSERT_IN_MUTEX_CPP);
```
thomas-bc: Why `tester.release()` and not just tester ?

Os/test/ut/mutex/MutexRules.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+bool Os::Test::Mutex::Tester::ProtectDataCheck::precondition(const Os::Test::Mutex::Tester &state) {
+    return true;
+}
+
+void Os::Test::Mutex::Tester::ProtectDataCheck::action(Os::Test::Mutex::Tester &state) {
+    state.m_mutex.lock();
+    state.m_state = Os::Test::Mutex::Tester::MutexState::LOCKED;
+
+    state.m_value = 42;
```
thomas-bc: Use STest random number here instead

thomas-bc (commented):
Peer review comments

Os/Posix/test/ut/PosixMutexTests.cpp
thomas-bc: `tester` is a `std::unique_ptr`. The `.release()` call here is that of the unique_ptr, **not** Mutex::release(). I will add a comment.

thomas-bc: But actually, since the deletion is not expected to succeed... maybe we should `get()` instead so that the destructor can be called again and the unique_ptr and underlying Mutex does get destroyed successfully?

Os/CMakeLists.txt
thomas-bc: @LeStarch I believe I was able to fix by removing the Os/Task includes and adding some Fw/Time link dependencies, see https://github.com/nasa/fprime/pull/2790/commits/7a27ec9238631e39a5b71688bc531322631c9e3a
Please confirm this is ok

thomas-bc: Ugh - lots of new errors popping in CI haha

cmake/platform/Linux.cmake
LeStarch: _ and / are the same in our system.  We probably should use _, but it is not blocking.

config/FpConfig.h
LeStarch: Not yet, no.

Os/CMakeLists.txt
LeStarch: Reverting is fine, we can tracks as a cleanup item

LeStarch (approved):

PR #2743, opened 2024-05-23
docs: Add documentation for F' subtopologies

mosa11aei:
| | |
|:---|:---|
|**_Related Issue(s)_**| n/a |
|**_Has Unit Tests (y/n)_**| n/a |
|**_Documentation Included (y/n)_**| y |

---
## Change Description

F' subtopologies are a very powerful tool to allow for breaking out one large topology into smaller, more manageable and reusable ones. This PR provides documentation on this tool.

PR #2794, opened 2024-07-05
Create github workflow for the subtopology autocoder

mosa11aei:
| | |
|:---|:---|
|**_Related Issue(s)_**| n |

---
## Change Description

This change implements a Github CI workflow for testing the [Subtopology Autocoder](https://github.com/mosa11aei/fprime-subtopology-tool). This will ensure that changes made to F Prime do not break the tool itself.


.github/workflows/ext-build-subtopology-ac.yml
```diff
@@ -0,0 +1,87 @@
...
+          submodules: false
+          path: ./fprime-subtopology-tool
+          repository: mosa11aei/fprime-subtopology-tool
+          ref: main
+      - name: "Go to Deployment Directory"
+        run: |
+          cd ./MainDeployment
+      - name: "Generate Subtopology Example Build Cache"
+        run: |
+          fprime-util generate -DFPRIME_SKIP_TOOLS_VERSION_CHECK=ON -DCMAKE_DEBUG_OUTPUT=ON
```
thomas-bc: This will not work as you expect it to because each step is run in its own shell. 
The recommended way is to use `working-directory` directive in your generate step (https://stackoverflow.com/a/58142276)
e.g.

```
     - name: "Generate Subtopology Example Build Cache"
        working-directory: ./MainDeployment
        run: |
          fprime-util generate -DFPRIME_SKIP_TOOLS_VERSION_CHECK=ON -DCMAKE_DEBUG_OUTPUT=ON
```

.github/workflows/ext-build-subtopology-ac.yml
```diff
@@ -0,0 +1,87 @@
...
+          path: ./fpp
+          ref: main
+      - name: "[fpp] Install sbt"
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y apt-transport-https curl gnupg -yqq
+          echo "deb https://repo.scala-sbt.org/scalasbt/debian all main" | sudo tee /etc/apt/sources.list.d/sbt.list
+          echo "deb https://repo.scala-sbt.org/scalasbt/debian /" | sudo tee /etc/apt/sources.list.d/sbt_old.list
+          curl -sL "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x2EE0EA64E40A89B84B2DF73499E82A75642AC823" | sudo -H gpg --no-default-keyring --keyring gnupg-ring:/etc/apt/trusted.gpg.d/scalasbt-release.gpg --import
+          sudo chmod 644 /etc/apt/trusted.gpg.d/scalasbt-release.gpg
```
thomas-bc: Does it not work to leave all that out and run `apt-get install sbt` directly?

.github/workflows/ext-build-subtopology-ac.yml
```diff
@@ -0,0 +1,87 @@
...
+          ref: main
+      - name: "Go to Deployment Directory"
+        run: |
+          cd ./MainDeployment
+      - name: "Generate Subtopology Example Build Cache"
+        run: |
+          fprime-util generate -DFPRIME_SKIP_TOOLS_VERSION_CHECK=ON -DCMAKE_DEBUG_OUTPUT=ON
+      - name: "Build Example"
+        run: |
+          fprime-util build -v
```
thomas-bc: Could be good to add a `-j4` here to speed up

thomas-bc (changes requested):

.github/workflows/ext-build-subtopology-ac.yml
mosa11aei: Seems like it worked. But I will implement the better practice.

.github/workflows/ext-build-subtopology-ac.yml
mosa11aei: Install instructions came from the sbt website:

https://www.scala-sbt.org/1.x/docs/Installing-sbt-on-Linux.html#Ubuntu+and+other+Debian-based+distributions

.github/workflows/ext-build-subtopology-ac.yml
thomas-bc: Yeah it probably built the top-level project and not within the deployment directory, which probably doesn't matter for this

.github/workflows/ext-build-subtopology-ac.yml
thomas-bc: It actually looks like SBT might be installed on the runners already
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md

Not sure of the version requirements with FPP though

mosa11aei: It should be sufficient!

mosa11aei:
@thomas-bc can you run CI again? It should work now.

The rest of the spelling mistakes will be fixed when the docs PR (https://github.com/nasa/fprime/pull/2743) is merged in.

.github/workflows/ext-build-subtopology-ac.yml
```diff
@@ -0,0 +1,87 @@
...
+          sudo apt-get update
+          sudo apt-get install -y cmake
+      - name: "Get FPP"
+        uses: actions/checkout@v4
+        with:
+          submodules: false
+          repository: nasa/fpp
+          path: ./fpp
+          ref: main
+      - name: "[fpp] Install sbt"
```
mosa11aei: Get rid of non native fpp install

thomas-bc: @mosa11aei if it were me I would get rid of all the installs, since all those tools come pre-installed on the GitHub runners. This may prevent breakages in the future. Your call.

mosa11aei: This specifically relates to something @LeStarch  and I talked about today. See https://github.com/nasa/fprime/pull/2816 😉

LeStarch: This should be better/fixed.

mosa11aei: This relates to the issue we were seeing with incomplete bitmap. Was this resolved on FPP side? I think the next step should be idea of running fpp-to-json on a wider range of tests to ensure a complete bitmap. 

I had floated the idea that of running fpp-to-json on the Ref deployment, because it's my opinion that the Ref deployment should be a model complete data set that the bitmap can be run on. I believe this was a point of disagreement with Rob.

mosa11aei:
Actually it seems like fpp-to-json now runs on Ref! Cool! We can make the change to now use installed fpp tools.

LeStarch:
Closing this PR as it is overcome by events.  These concepts demonstrated herein are in the process of being pulled into core FPP!

PR #2795, opened 2024-07-08
FPP v2.1.0a12

bocchino:
This PR updates the FPP version to v2.1.0a12.  It also makes conforming changes to F Prime, including the following:
* Remove configuration symbols that are no longer used in the generated code.
* Update framework code so that it doesn't use these symbols (i.e., use the same implementation that FPP uses).
* Replace an `Ai.xml` file that was lingering in a test with the corresponding FPP file.
* Replace `== 1` in a configuration check with `!= 0` -- this seems more robust.


thomas-bc (approved):

LeStarch (approved):

LeStarch:
@bocchino ready to merge?

bocchino:
Ready!

PR #2796, opened 2024-07-09
Revise file name string sizes in AssertFatalAdapter events

bocchino:
## Change Description

This PR updates the sizes of the file name strings used in the AssertFatalAdapter events:

1. Make the event string size configurable, instead of hard-coded to 80.
2. In the default configuration, set each of the three kinds of truncation that an event string passes through to 200. (Note: 256 is too large for the AMPCS compatibility code which we don't use, but is still in the generated C++.)
3. Add comments to the config files identifying the three kinds of truncation applied to file names in assertion failure events.
4. Fix issues in the framework uncovered by these changes.
5. Incidentally fix an off-by-one error in the CmdSplitter unit tests.

## Rationale

On workstation development systems, e.g., Linux and Mac, an 80-character string is not long enough to report the location where an assertion has failed. For example, if an assertion failure occurs in an autocoded C++ file, the file name can have the following form:
```
/several/file/path/to/fprime/project/build-fprime-automatic-native/F-Prime/Fw/Buffer/BufferGetPortAc.cpp
```
That's 104 characters; with an 80-character string, the GDS displays the unimportant part (the path prefix that is common to all FSW files) and chops off the part you care about, which is at the end:
```
% printf /several/file/path/to/fprime/project/build-fprime-automatic-native/F-Prime/Fw/Buffer/BufferGetPortAc.cpp | head -c 80
/several/file/path/to/fprime/project/build-fprime-automatic-native/F-Prime/Fw/Bu
```
This behavior makes assertion failure event reports essentially unusable on these systems.

The assertion file strings that users get out of the box should satisfy two criteria:
1. They should be configurable, so that users can use different sizes on different systems.
2. In the basic configuration, which should run on Linux and Mac, the strings should be long enough to display the information needed to debug the FSW. On smaller systems with less preamble in the file names (or on systems that use numerical hashes instead of file name strings), the configured size can be made smaller.
3. It should be straightforward to change the configuration to enable longer or shorter assertion failure strings. 

The framework unit tests should also compile and run correctly under different configurations. I fixed the issues that I saw when making the configuration changes here. I'm sure there are more issues, which we won't see until we have a general way to test other configurations. This is a CI issue.


config/AcConstants.fpp
```diff
@@ -48,6 +48,12 @@ constant DpWriterNumProcPorts = 5
 @ The size of a file name string
 constant FileNameStringSize = 256
 
+@ The size of a file name in an AssertFatalAdapter event
+@ Note: File names in assertion failures are also truncated by
+@ the constants FW_ASSERT_TEXT_SIZE and FW_LOG_STRING_MAX_SIZE, set
+@ in FpConfig.h.
```
Joshua-Anderson, line 54: This is fantastic - I really appreciate properly documenting this. It'll save a bunch of pain for fprime users that need to raise the assert message size

config/FpConfig.h
```diff
@@ -272,7 +273,7 @@ typedef FwIndexType FwQueueSizeType;
 
 // Specifies the size of the buffer that contains a communications packet.
 #ifndef FW_COM_BUFFER_MAX_SIZE
-#define FW_COM_BUFFER_MAX_SIZE 128  //!< Max size of Fw::Com buffer
+#define FW_COM_BUFFER_MAX_SIZE (FW_LOG_STRING_MAX_SIZE + sizeof(FwEventIdType) + sizeof(FwPacketDescriptorType))
```
Joshua-Anderson: Since the max log string length tends to be the largest buffers fprime needs to send this makes logical sense, but conceptually it feels a bit odd to tie the size of this core construct to our logging string length.

Practically, it's also nice to have the com buffer size defined to be a constant because it makes it easy to check and see if you're going over the max combuffer size without needing to do any napkin math.

I might consider keeping this a constant with a note that this should be larger than the log max string size.

Joshua-Anderson (approved):
Thanks for fixing this Rob, this looks great!

I added one comment about the sizing of the combuffer you may want to consider. I suspect we'll also want @LeStarch's signoff on this.

config/FpConfig.h
bocchino: Thanks, Josh, I changed it to 512. The logic of what needs to be greater than what is somewhat complicated -- I decided to leave it to the comments in the static checks.

config/AcConstants.fpp
bocchino: Thanks, Josh!

Joshua-Anderson (approved):

thomas-bc (approved):

Svc/CmdSplitter/test/ut/CmdSplitterTester.cpp
```diff
@@ -116,7 +116,7 @@ void CmdSplitterTester ::test_response_forwarding() {
     Fw::CmdResponse response;
     response.e = static_cast<Fw::CmdResponse::T>(STest::Pick::lowerUpper(0, Fw::CmdResponse::NUM_CONSTANTS));
     U32 cmdSeq = static_cast<U32>(STest::Pick::any());
-    this->active_command_source = static_cast<NATIVE_INT_TYPE>(STest::Pick::lowerUpper(
```
LeStarch, line 119: Seems an unrelated change.  Is this recommended instead of lowerUpper?

LeStarch (approved):
Looks good, ready to merge?

Svc/CmdSplitter/test/ut/CmdSplitterTester.cpp
bocchino: Yes, this fixes an off-by-one error that caused an intermittent out-of-bounds array access. If your array has 10 elements and you call `lowerUpper(0,10)` you can get an index of 10, which is out of bounds. If you call `startLength(0, 10)` you will get an element in the range [0,9].

bocchino:
Yes, ready!

PR #2816, opened 2024-07-25
workflow: Test Ref deployment (topology and instances.fpp) with fpp-to-json

mosa11aei:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/discussions/2803 |

---
## Change Description

This PR implements a workflow that tests fpp-to-json on the Ref deployment. This ensures that the F Prime `Ref` deployment (and thus its dependencies) are all model complete. It will also notify if there are potentially any issues with `fpp-to-json` itself.

### DIFF
--- a/.github/workflows/fpp-json-model-test.yml
+++ b/.github/workflows/fpp-json-model-test.yml
@@ -1,0 +1,63 @@
+# Runs the subtopology autocoder tool to test the subtopology tool (https://github.com/mosa11aei/fprime-subtopology-tool)
+
+name: "fpp-to-json Analysis Test"
+
+on:
+  push:
+    branches: [ devel, release/** ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


mosa11aei:
Spelling can be fixed when one of the other open PRs re: subtopologies gets committed. We can merge back from fprime.

mosa11aei:
FYI @bocchino - This implements the workflow that uses fpp-to-json to check the Ref deployment to make sure everything is model complete. 

It fails with a bitmap error, which means that the bitmap is still not updated enough to fully analyze the Ref deployment. 

cc @LeStarch

thomas-bc:
Implemented in https://github.com/nasa/fprime/pull/3430 - will close this PR

PR #2818, opened 2024-07-28
Fix double-promotion in fprime codebase

JohanBertrand:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fixed a warning of `float` being promoted to `double`.

Fixed `%lg` parameter being used for `double`, instead of `%g`.

Fixed warning on `FW_ASSERT`.

## Rationale

When a `float` is used as a `printf` parameter, it is promoted to `double`, triggering a warning when `-Wdouble-promotion` is used.

## Future Work

FFP is currently generating code for in `Ac.cpp` files which are using `sb.format` and `printf` in the background. 

I had an attempt to fix it in the branch here: https://github.com/nasa/fpp/commit/f495ce1579c275107f2522c859ee4b572dc6f2d1

However, this patch is currently not compatible with all platform. A platform not supporting `double`/`F64` would not work with this change.

I personally don't have a platform that does not support `F64`/`double` types, so I could not check the behavior.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,12 +24,14 @@ add_compile_options(
     -Wno-unused-parameter
 )
 
-# Add -Wshadow and -pedantic only for framework code, not the unit tests
+# Add -Wshadow, -Wconversion and -pedantic only for framework code, not the unit tests
+# TODO: uncomment -Wdouble-promotion once fpp is fixed
 if (NOT BUILD_TESTING)
     add_compile_options(
         -Wshadow
         -pedantic
         -Wconversion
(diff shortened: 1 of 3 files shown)


LeStarch (approved):
This looks like the right fix.   Are there more uses of this to fix?  If not, can we uncomment -Wdouble-promotion?

LeStarch:
I see the note on the FPP changes required.  Will look into that.

LeStarch:
I merged this as it.  Sorry for the delay, I went down a rabbit hole on the idea of `HAS_F64`.

PR #2823, opened 2024-07-29
Update Ref and RPI for compatibility with FPP #349 (fully qualified component instance names)

jwest115:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fpp/issues/349 |

---
## Change Description

As part of https://github.com/nasa/fpp/issues/349, component instance names are now fully qualified in topology c++ auto generated files. Updated Ref and RPI deployment fpp files and integration tests to match this change.

## Rationale

Needed for compatibility with https://github.com/nasa/fpp/issues/349

Built both deployments successfully. Ran integration tests for Ref deployment successfully.

- [x] Create the FPP alpha release - waiting on reviewers
- [x] Update FPP version in requirements.txt
- [x] Check that CI passes 
- [x] Convert draft pull request to regular pull request
- [ ] Review and merge


thomas-bc:
Will probably need to wait for both the FPP and GDS releases for that one

RPI/Top/RPITopologyDefs.hpp
```diff
@@ -46,15 +46,15 @@ namespace RPI {
...
-    namespace cmdDisp { enum { WARN = 3, FATAL = 5 }; }
-    namespace cmdSeq { enum { WARN = 3, FATAL = 5 }; }
-    namespace chanTlm { enum { WARN = 3, FATAL = 5 }; }
-    namespace eventLogger { enum { WARN = 3, FATAL = 5 }; }
-    namespace prmDb { enum { WARN = 3, FATAL = 5 }; }
-    namespace fileDownlink { enum { WARN = 3, FATAL = 5 }; }
-    namespace fileUplink { enum { WARN = 3, FATAL = 5 }; }
+    namespace RPI_rateGroup10HzComp { enum { WARN = 3, FATAL = 5 }; }
+    namespace RPI_rateGroup1HzComp { enum { WARN = 3, FATAL = 5 }; }
+    namespace RPI_cmdDisp { enum { WARN = 3, FATAL = 5 }; }
```
thomas-bc, line 51: I know this was likely discussed already but I want to capture in writing and refresh my mind:

Why are we namespacing with `_` here? Could FPP not handle `::` ?

And do we need to namespace at all since they are used-defined?

These patterns should probably be documented somewhere if they are required

bocchino: We need one C++ name for each qualified FPP name. The easiest way seems to be to use the underscores as shown. We could use :: but this seems awkward. You would have to write each of the names inside a namespace.

bocchino: Also we can't use namespaces here because RPI::cmdDisp is already a variable name. The idea is that RPI_cmdDisp::WARN is the WARN value that corresponds to RPI::cmdDisp.

bocchino: This should be documented in the FPP User's Guide.

jwest115: I can add that to the docs!

bocchino (approved):
Looks good!

RPI/Top/RPITopologyDefs.hpp
bocchino: Sounds good! I opened an issue on it. See https://github.com/nasa/fpp/issues/510.

thomas-bc: It is imo a little confusing from the user perspective that some stuff (e.g. components) would be namespaced `RPI::rateGroup10HzComp` but its configuration has to be `RPI_rateGroup10HzComp`

From a user perspective it is not clear what makes the distinction, and a little confusing. But maybe with the User Guide it will make more sense.

bocchino: > It is imo a little confusing from the user perspective that some stuff (e.g. components) would be namespaced RPI::rateGroup10HzComp but its configuration has to be RPI_rateGroup10HzComp

The rationale is that `RPI::rateGroup10HzComp` is a fully qualified name in C++, declared at the top level, whereas `RPI_rateGroup10HzComp` is a name declared inside the topology namespace. Note the following:

1. I agree the use of underscores is not ideal, but I don't know how to do it better. I don't think we should nest the instance namespaces inside the topology namespaces.
2. We could be consistent and use underscores everywhere, but we decided to move the instance names out of the topology namespace and use namespaces for them.

We will explain this in the FPP User's Guide.

LeStarch (approved):

PR #2831, opened 2024-08-13
Implements #2723: Adds Os::Console to replace Os::Log

### LINKED ISSUE #2723: Refactor Os::Log into CMake Selection
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

- [x] Consider if `Os::Log` should be renamed `Os::Console`. 
- [x] Refactor Os::Log into the Interface, Os::Log, and Implementation classes (following pattern of Os::File)
- [x] Implement Posix implementation
- [x] Implement Stub implementation
- [x] Implement Test stub implementation
- [x] Build common rules-based tests
- [x] Build specific posix tests
- [x] Build stub interface tests
- [x] Setup default posix implementation for Posix platforms
- [x] Document

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Refactors Os::Log into a cmake implementation called Os::Console.


Fw/Types/StringBase.cpp
```diff
@@ -54,14 +54,22 @@
...
-    FW_ASSERT(us);
     va_list args;
     va_start(args, formatString);
-    (void)vsnprintf(us, cap, formatString, args);
+    this->vformat(formatString, args);
     va_end(args);
-    // null terminate
+}
+
+void StringBase::vformat(const CHAR* formatString, va_list args) {
```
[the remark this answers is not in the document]
LeStarch: Not detecting `FW_ASSERT` as assert.

Fw/Logger/Logger.hpp
```diff
@@ -4,87 +4,59 @@
...
-            );
+  public:
+    //! \brief log a formated string with supplied arguments
+    //!
+    //! Logs a format string with the arguments filled-in. This delegates to StringBase.format, which in-turn
+    //! delegates to snprintf. This implies that the caller is fully responsible for handling the type safety of
+    //! the supplied format string. The format string uses C-style (printf function family) formatting.
+    //! \param format: format string
+    //! \param ...: var-args list of arguments to inject into format string.
+    static void log(const char* format, ...);
```
thomas-bc, line 32: Could we be interested in aliasing this to `logMsg(format, ...)` as well for backwards compatibility? Otherwise we should flag this PR as a breaking change for release notes

thomas-bc (approved):

Fw/Logger/Logger.hpp
thomas-bc: nevermind ... just saw it has the label already

LeStarch: The old format was very dangerous, so we want to force users to update with consideration.  Typically, this is a better approach except when the old way is dangerous.

Issue #2837, opened 2024-08-16
fprime-venv uses hardcoded full path to project folder instead of building a full path at time of call


mprather:
| | |
|:---|:---|
|**_F´ Version_**| latest|
|**_Affected Component_**| virtual env |
---
## Feature Description
Let's say you initially create a project at the following location: /folder1/category1/MyProject
Then you decide to move the project to a different location: /someOtherFolder/MyProject

The call to `fprime-venv/bin/activate` runs successfully in the new location but it uses old information from when the project was first created (i.e. it has a full path to the original /folder1/ location).  The `activate` and `fprime-util` scripts contain full path references. The result is that the user is presented with an error message that states `fprime-util: command not found`.

It would be nice to have the pathing information use relative paths or dynamically created full paths. This way, it doesn't matter where the project is placed - the project continues to work.  

Alternatively, the Troubleshooting FAQ should be updated to include the relocation scenario under the [fprime-util: Command not found](https://nasa.github.io/fprime/troubleshooting.html#fprime-util-command-not-found) section.

## Rationale

Folder reorganization is a common task. Scripts need handle this more effectively. 

If this information is already available, then a link to info could be added to the new user tutorial in the place where the user is [first told to use `activate`](https://fprime-community.github.io/fprime-tutorial-hello-world/docs/NewProject.html) (in addition also updating the FAQ).

thomas-bc:
Hey @mprather, thanks for opening an issue.

Unfortunately, this is a limitation of Python venv's, and not something we can control. There's a lot of discussion about this online (stackoverflow etc...). 
The feeling I have from reading on this is that re-creating a virtual environment is rather easy, so it's not worth adding complexity to `venv`s to handle those edge cases.

The solution is to `rm -rf fprime-venv` and recreate a new one in the new location `python -m venv fprime-venv`

mprather:
Thanks for the quick reply, @thomas-bc. 

I don't quite agree that folder renames is an edge case. New users to the framework are bound to reorg folders as they determine how to best work with the tools. Literally in the span of a half day, we've already tried 2 different layouts on how best to integrate our work with the default F' setup.

Nonetheless, I understand a workable fix is out of scope. For newbie users of python (and python's virtual environments) that are following the tutorials, the line between what is python's fault and F''s, could be lead to confusion. A little guidance would be extremely helpful. The solution outlined above should go into the troubleshooting FAQ. I'm happy to submit a PR with a doc update if that helps.

[...]

PR #2843, opened 2024-08-20
Updated troubleshooting md

mprather:
## Change Description

Improved 'fprime-util: command not found' section. Related to #2837
Also added a link to the navbar to make it easier to find the FAQ.

## Rationale

Better documentation makes it easier to use the framework.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -23,11 +23,26 @@ Some of the F´ Python packages are built in a way that it is recommended to ins
 
 ## fprime-util: command not found
 
-If the user is using a virtual environment and receives the ‘command not found’, the problem is likely caused by the environment not being sourced in a new terminal. Make sure to source the environment before running:
-```
-. <path/to/project>/fprime-venv/bin/activate
-```
-If installing without a virtual environment, PIP occasionally uses `$HOME/.local/bin` as a place to install user tools. Users running without virtual environments should add this directory to the path.
+There are at least three common reasons for this error.
+
+1. If the user is using a virtual environment and receives the ‘command not found’, the problem is likely caused by the environment not being sourced in a new terminal. Make sure to source the environment before running:
+    ```sh
+    # In MyProject
+    . fprime-venv/bin/activate
+    ```
+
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -23,11 +23,26 @@ Some of the F´ Python packages are built in a way that it is recommended to ins
...
-If installing without a virtual environment, PIP occasionally uses `$HOME/.local/bin` as a place to install user tools. Users running without virtual environments should add this directory to the path.
+There are at least three common reasons for this error.
+
+1. If the user is using a virtual environment and receives the ‘command not found’, the problem is likely caused by the environment not being sourced in a new terminal. Make sure to source the environment before running:
+    ```sh
+    # In MyProject
+    . fprime-venv/bin/activate
+    ```
+
+1. If the project folder was moved after the virtual environment was provisioned (e.g. project was originally setup in `/folderA/MyProject` and moved to `/someOtherFolder/MyProject`), various files in the original `fprime-venv` folder will contain references to the original path. The easist resolution is to remove and reprovision the virtual environment with the following:
```
thomas-bc: typo: `easist` -> `easiest`

docs/troubleshooting.md
```diff
@@ -23,11 +23,26 @@ Some of the F´ Python packages are built in a way that it is recommended to ins
...
+    # In MyProject
+    rm -rf fprime-venv
+    python -m venv fprime-venv
+    . fprime-venv/bin/activate
+    pip install -r fprime/requirements.txt
+    ```
+
+    > Some projects ship their own `requirements.txt`.  Install using that file if it exists.
+
+1. If installing without a virtual environment, PIP occasionally uses `$HOME/.local/bin` as a place to install user tools. Users running without virtual environments should add this directory to the path.
```
thomas-bc, line 45: Numbering is off - should be 1 - 2 - 3 and not 1 - 1 - 1

thomas-bc (changes requested):
Typo and numbering issue - otherwise looks good to me!

docs/troubleshooting.md
mprather: Did you know that list numbering is automatically handled by the md engine? If you use 1-1-1, then future reordering doesn't seem as messy in diff views.  See https://www.markdownguide.org/basic-syntax/#ordered-lists

I'll submit the fixes shortly.

thomas-bc: Oh wow I did not know haha. Today I learned! No need to fix then - thanks!

thomas-bc (approved):

LeStarch (approved):

Issue #2066, opened 2023-06-01
ComQueue Does Not Deallocate Buffers on Overflow


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

If a set Fw::Buffer overflows ComQueue it is potentially droped (internally and on queue) which would leak the memory.

thomas-bc:
https://github.com/nasa/fprime/issues/2571

thomas-bc:
Need an FPP alpha release to get the hook feature in

thomas-bc:
See `hook` behavior on https://nasa.github.io/fpp/fpp-users-guide.html#Defining-Components_Port-Instances_Queue-Full-Behavior -> deallocate on overflow

PR #2853, opened 2024-08-30
Fix ComQueue Does Not Deallocate Buffers on Overflow

DJKessler:
| | |
|:---|:---|
|**_Related Issue(s)_**|  #2066  |

---
## Change Description

Fix potential memory leak when a sent `Fw::Buffer` overflows `ComQueue`.

* Added `Fw.BufferSend` `deallocate` port to `ComQueue`
* Updated FPP to v2.2.0a2
* Change `buffQueueIn` port from `drop` to `hook`
* Added queue overflow hook method implementation
* Added return status to `ComQueue::enqueue()`
* `Fw::Buffer` is now deallocated on queue overflow
* Enabled `UT_AUTO_HELPERS` in `ComQueue` UT build
* Updated `ComQueue` UTs


Svc/ComQueue/ComQueue.cpp
```diff
@@ -139,7 +139,11 @@ void ComQueue::buffQueueIn_handler(const NATIVE_INT_TYPE portNum, Fw::Buffer& fw
     // Ensure that the port number of buffQueueIn is consistent with the expectation
     FW_ASSERT(portNum >= 0 && portNum < BUFFER_PORT_COUNT, portNum);
     FW_ASSERT(queueNum < TOTAL_PORT_COUNT);
-    this->enqueue(queueNum, QueueType::BUFFER_QUEUE, reinterpret_cast<const U8*>(&fwBuffer), sizeof(Fw::Buffer));
+    bool status =
+        this->enqueue(queueNum, QueueType::BUFFER_QUEUE, reinterpret_cast<const U8*>(&fwBuffer), sizeof(Fw::Buffer));
+    if (!status) {
+        this->buffQueueIn_overflowHook(portNum, fwBuffer);
```
LeStarch: Perhaps we can just call this->deallocate_out?

Svc/ComQueue/ComQueue.hpp
```diff
@@ -151,13 +151,23 @@ class ComQueue : public ComQueueComponentBase {
...
     );
 
+    // ----------------------------------------------------------------------
+    // Hook implementations for typed async input ports
+    // ----------------------------------------------------------------------
+
+    //! Queue overflow hook method that deallocates the fwBuffer
+    //!
+    void buffQueueIn_overflowHook(FwIndexType portNum, /*!< The port number*/
+                                  Fw::Buffer& fwBuffer /*!< The buffer*/
```
LeStarch: The new style is:

```
//!< The port number
```

LeStarch (changes requested):
There are several things that need to be fixed:

1. In the com queue function, we should explicitly discard the new enqueue status with `(void) this->enqueue...`
2. I don't see any tests proving that the drop function is called on overflow.  Can we add those.
3. CI needs to pass

Svc/ComQueue/ComQueue.cpp
DJKessler: Fixed in 658d2a43e.

Svc/ComQueue/ComQueue.hpp
DJKessler: Fixed in c2e203c7d.

LeStarch (approved):

LeStarch:
Skipping known ci failure.  Nice work!

Issue #2835, opened 2024-08-16
Random UT Failure: PosixFile


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

```
55/66 Test #55: PosixFileTest ....................***Failed    3.34 sec
[STest::Random] Generated seed 965653 from system time
[==========] Running 32 tests from 3 test suites.
[----------] Global test environment set-up.
[----------] 14 tests from Functionality
[ RUN      ] Functionality.OpenWithCreation
[       OK ] Functionality.OpenWithCreation (7 ms)
[ RUN      ] Functionality.Close
[       OK ] Functionality.Close (0 ms)
[ RUN      ] Functionality.AssignmentOperator
[       OK ] Functionality.AssignmentOperator (0 ms)
[ RUN      ] Functionality.CopyConstructor
[       OK ] Functionality.CopyConstructor (0 ms)
[ RUN      ] Functionality.OpenWithCreationOverwrite
[       OK ] Functionality.OpenWithCreationOverwrite (0 ms)
[ RUN      ] Functionality.OpenInvalidModes
[       OK ] Functionality.OpenInvalidModes (0 ms)
[ RUN      ] Functionality.PreallocateWithoutOpen
[       OK ] Functionality.PreallocateWithoutOpen (0 ms)
[ RUN      ] Functionality.SeekWithoutOpen
[       OK ] Functionality.SeekWithoutOpen (0 ms)
[ RUN      ] Functionality.FlushInvalidModes
[       OK ] Functionality.FlushInvalidModes (0 ms)
[ RUN      ] Functionality.ReadInvalidModes
[       OK ] Functionality.ReadInvalidModes (0 ms)
[ RUN      ] Functionality.WriteInvalidModes
[       OK ] Functionality.WriteInvalidModes (0 ms)
[ RUN      ] Functionality.RandomizedInterfaceTesting
Ran 100 steps.
[       OK ] Functionality.RandomizedInterfaceTesting (1245 ms)
[ RUN      ] Functionality.FullCrcInvalidMode
[       OK ] Functionality.FullCrcInvalidMode (0 ms)
[ RUN      ] Functionality.IncrementalCrcInvalidMode
[       OK ] Functionality.IncrementalCrcInvalidMode (0 ms)
[----------] 14 tests from Functionality (1256 ms total)

[----------] 9 tests from FunctionalIO
[...]

PR #2860, opened 2024-09-04
Fix Random UT Failure: PosixFile (#2835)

rlcheng:
* Fix UT failing FinalizeCrc test when crc calculation miss match. Miss match due to file crc and shadow being out of sync when both are not reinialized at the same time.

 * Update File::open in File.ccp to match shadow operation in FileRules.cpp to only reset crc when status is OP_OK.

 * Update FileRules.cpp to print out rule name during test.

 * Remove changing file mode in shadow_open() in FileRules.cpp to OPEN_NO_MODE when status is not OP_OK.

| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

A description of the changes contained in the PR.

## Rationale

A rationale for this change. e.g. fixes bug, or most projects need XYZ feature.

## Future Work

### DIFF
--- a/Os/test/ut/file/FileRules.cpp
+++ b/Os/test/ut/file/FileRules.cpp
@@ -2,7 +2,7 @@
 // \title Os/test/ut/file/MyRules.cpp
 // \brief rule implementations for common testing
 // ======================================================================
-
+#include <cstdio>
 #include "RulesHeaders.hpp"
 #include "STest/Pick/Pick.hpp"
 extern "C" {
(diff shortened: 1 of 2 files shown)


Os/test/ut/file/FileRules.cpp
```diff
@@ -256,6 +258,7 @@ bool Os::Test::File::Tester::OpenBaseRule::precondition(const Os::Test::File::Te
 
 void Os::Test::File::Tester::OpenBaseRule::action(Os::Test::File::Tester &state  //!< The test state
 ) {
+    printf("--> Rule: %s \n", this->getName());
```
thomas-bc, line 261: Should this be implemented at the STest level? It seems pretty convenient

How does one get access to those logs btw? I am assuming it gets captured by GoogleTest and put in a log file somewhere?

LeStarch (approved):

PR #2862, opened 2024-09-05
Fix random UT failures on PosixFile

rlcheng:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

A description of the changes contained in the PR.

## Rationale

A rationale for this change. e.g. fixes bug, or most projects need XYZ feature.

## Future Work


thomas-bc (approved):

Os/test/ut/file/FileRules.cpp
```diff
@@ -497,7 +497,7 @@ void Os::Test::File::Tester::Preallocate::action(
     state.assert_file_consistent();
     FileState original_file_state = state.current_file_state();
     FwSignedSizeType offset = static_cast<FwSignedSizeType>(STest::Pick::lowerUpper(0, FILE_DATA_MAXIMUM));
-    FwSignedSizeType length = static_cast<FwSignedSizeType>(STest::Pick::lowerUpper(0, FILE_DATA_MAXIMUM));
+    FwSignedSizeType length = static_cast<FwSignedSizeType>(STest::Pick::lowerUpper(1, FILE_DATA_MAXIMUM));
```
LeStarch, line 500: @rlcheng do we have a test that proves that preallocate errors on a "0" size argument?

LeStarch (approved):
Just a quick question, but otherwise this looks good.

LeStarch:
Merging despite known CI failure.

Os/test/ut/file/FileRules.cpp
rlcheng: We dont have one. Currently the SeekInvalidSize test randomly picks using the range original_file_state.position + 1, std::numeric+_limits<U32>. Seems to be always a positive non zero value that gets multiplied by -1. We can change the range to: 0, std::numeric_limits<U32>.

This should guarantee in a randomized test we can get a 0 or negative value.

LeStarch: Let's track as a separate issue.  Rather than randomizing, we should just add a specific test to the interface tests.

Issue #2826, opened 2024-07-31
Remove versions from System Resources and fix Versions component


Shivaly-Reddy:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description
As the title suggests, this ticket is in fact capturing two separate issues

- [ ]  Remove versions from System Resources 
- [ ]  Fix versions component so commanding to get custom versions will not need enable to be left on.See [here](https://github.com/nasa/fprime/pull/2747#discussion_r1671183951) for details.

PR #2866, opened 2024-09-10
Remove versions from SystemResources and updates to Version component

Shivaly-Reddy:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

This pull request is for ticket #2826 changes

## Rationale

1. Removes redundant version events and channels generated in SystemResources
2. Updated Versions component to produce events on custom versions when verbosity is disabled (channel generation still depends on verbosity configuration)

Updated UTs to ensure events are generated regardless of verbosity configuration but channel depend upon it. 

## Future Work

Currently there is no future work identified related to this issue.

### DIFF
--- a/Ref/Top/RefPackets.xml
+++ b/Ref/Top/RefPackets.xml
@@ -72,12 +72,7 @@
         <channel name="systemResources.NON_VOLATILE_FREE"/>
     </packet>
 
-    <packet name="SystemRes2" id="6" level="2">
-        <channel name="systemResources.FRAMEWORK_VERSION"/>
-        <channel name="systemResources.PROJECT_VERSION"/>
-    </packet>
-
-    <packet name="SystemRes3" id="7" level="2">
+    <packet name="SystemRes3" id="6" level="2">
         <channel name="systemResources.CPU"/>
         <channel name="systemResources.CPU_00"/>
         <channel name="systemResources.CPU_01"/>
--- a/Svc/SystemResources/SystemResources.cpp
+++ b/Svc/SystemResources/SystemResources.cpp
@@ -1,4 +1,3 @@
-// ======================================================================
 // \title  SystemResourcesComponentImpl.cpp
 // \author sfregoso
 // \brief  cpp file for SystemResources component implementation class
@@ -12,7 +11,6 @@
 
 #include <cmath>  //isnan()
 #include <Svc/SystemResources/SystemResources.hpp>
-#include <versions/version.hpp>
 #include <FpConfig.hpp>
 
 namespace Svc {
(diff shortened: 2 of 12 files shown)


Svc/SystemResources/SystemResources.cpp
```diff
@@ -1,4 +1,3 @@
-// ======================================================================
```
thomas-bc: Removed by mistake?

Svc/SystemResources/SystemResources.fpp
```diff
@@ -129,12 +109,6 @@ module Svc {
     @ System's CPU Percentage
     telemetry CPU_15: F32 id 20 format "{.2f} percent"
 
-    @ Software framework version
-    telemetry FRAMEWORK_VERSION: string size 40 id 21
-
-    @ Software project version
-    telemetry PROJECT_VERSION: string size 40 id 22
```
thomas-bc, line 136: This requires changes to the tutorial reference repositories because they have the Telemetry spec XML file... that is why CI is breaking in some places.

Which begs the question, should that file be removed from the tutorials? It's really not that useful...
I vote yes 👀 

@kevin-f-ortega thoughts? Since you are reworking the LedBlinker one

Svc/SystemResources/SystemResources.cpp
Shivaly-Reddy: Lol, yes. Thanks!

Svc/SystemResources/SystemResources.fpp
Shivaly-Reddy: Looks like the xml file in Ref/Top needed updating as well. That took care of some of these warnings

Ref/Top/RefPackets.xml
```diff
@@ -72,12 +72,7 @@
         <channel name="systemResources.NON_VOLATILE_FREE"/>
     </packet>
 
-    <packet name="SystemRes2" id="6" level="2">
-        <channel name="systemResources.FRAMEWORK_VERSION"/>
-        <channel name="systemResources.PROJECT_VERSION"/>
-    </packet>
-
-    <packet name="SystemRes3" id="7" level="2">
+    <packet name="SystemRes3" id="6" level="2">
```
Shivaly-Reddy, line 75: Is it okay that I changed id for SystemRes3 from 7 to 6?

thomas-bc: As long as it is unique (which it is) it should not matter. We could also rename `SystemRes3` to `SystemRes2` since there is no `SystemRes2` anymore, but not really needed

thomas-bc (approved):
LGTM!

LeStarch (approved):

PR #2896, opened 2024-09-25
Fix Os::Console singleton to be heapless

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Os::Console singleton -> heap free placement new singleton.

### DIFF
--- a/Os/Console.cpp
+++ b/Os/Console.cpp
@@ -4,6 +4,7 @@
 // ======================================================================
 #include <Os/Console.hpp>
 #include <Fw/Types/Assert.hpp>
+#include <new>
 
 alignas(Os::Console) U8 _singleton_store[sizeof Os::Console];
 
--- a/Os/Console.cpp
+++ b/Os/Console.cpp
@@ -5,6 +5,8 @@
 #include <Os/Console.hpp>
 #include <Fw/Types/Assert.hpp>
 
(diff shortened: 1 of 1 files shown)


Os/Console.cpp
```diff
@@ -4,6 +4,18 @@
...
+// While we desire the singleton to be in global space to avoid placing
+// such object on the heap, the singleton must be constructed at first
+// request due to C++'s undefined ordering of constructors between
+// compilation units.
+//
+// To meet both requirements, we place an aligned and sized block of
+// memory on the in global space as seen here, and then placement-new
+// the singleton into that region. Thus constructing on the fly while
+// avoiding heap usage.
+alignas(Os::Console) U8 _singleton_store[sizeof(Os::Console)];
```
thomas-bc: I was reading about placement-new [here](https://en.cppreference.com/w/cpp/language/new) and saw this note about objects created in such a way:

```
    tptr->~T();           // You must **manually** call the object's destructor
                          // if its side effects is depended by the program.
                          // Leaving this block scope automatically deallocates “buf”.
```

Just want to make sure we're aware of this - not sure if it matters in our case

LeStarch: Yeah, for that we'd need a deinit function.  I am trying something else that might work.

thomas-bc (approved):

Issue #2073, opened 2023-06-08
Remove Polling Buffer in Deframer Code


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

Deframer has poll and `schedIn` functionality.  It does not need this functionality. Drivers that need to support polling can have a schedIn method themselves.

matt392code:
Proposed detailed implementation plan for removing the polling buffer functionality from the Deframer component:
[deframer-changes.txt](https://github.com/user-attachments/files/18342501/deframer-changes.txt)
Here's a breakdown of the proposed changes:

1. **Architectural Change**:
   - Remove polling functionality entirely from Deframer
   - Simplify to only support callback-based operation
   - Move polling responsibility to driver components

2. **Key Benefits**:
   - Cleaner separation of concerns
   - Simplified component interface
   - More flexible driver implementations
   - Reduced complexity in Deframer code

3. **Migration Steps**:
   - Remove `schedIn` and `poll` ports
   - Remove polling-related internal state
   - Update existing drivers to handle polling if needed
   - Update component documentation

4. **Implementation Considerations**:
   - Ensure backward compatibility path for existing systems
   - Update unit tests to remove polling scenarios
   - Verify buffer management remains correct
   - Document migration path for existing implementations

LeStarch:
@thomas-bc this will be fixed by the deframer rework, yes?  If so, would you close?

PR #2900, opened 2024-09-25
Split Deframer into FrameAccumulator and Router

### LINKED ISSUE #2763: Break Apart Deframer
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

The Deframer is performing several peices:

1. It reassembles frames
2. It calls custom deframing via a subclass
3. It routes data to F´ components.

On modern projects, we have seen the following needs that the deframer cannot support. 

1. Custom deframing needs to use custom port calls (e.g. call out to a helper component, database component, etc)
2. Routing to F´ components often needs to be customized
3. Reassembly is optional (e.g. a packet radio is already produces complete frames)

Thus we will break the deframer into 3 F´ items:

1. An optional accumulator that takes in byte streams and outputs complete frames in Fw::Buffer: https://github.com/nasa/fprime/issues/2764
2. A deframer interface that allows custom deframing to a standard component "drop in":
3. A router implementation for basic F´ constructs:

### EPIC Subtasks
- [x] https://github.com/nasa/fprime/issues/2764
- [x] https://github.com/nasa/fprime/issues/2765

### LINKED ISSUE #2764: Build A Frame Reassembler
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Build a component that can be used to reassemble communication frames given protocol information. This will break-off this functionality from the Deframer.

Basic requirements:
1. Accept incoming Fw::Buffers of any size
2. Accumulate buffers in-order
3. Call frame detector helper class
4. On "NO_FRAME" discard first byte and try again
5. On "NEED DATA" continue to accumulate data
6. On "FRAME" allocate buffer, copy-out frame

Helper class requirements:
1. Must implement `Enum detect_frame(const CircularBuffer& buffer, FwSizeType& size_output)` method
2. Cannot alter circular buffer (uses peeks)
3. Must set `size_output` when data is needed and when frame detected

### LINKED ISSUE #2765: Build uplink router
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**| `Router` |
---
## Feature Description

Following https://github.com/nasa/fprime/issues/2763, build a router that sits after the Deframer and routes F´ packets to either FileUplink or CommandDispatcher

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  https://github.com/nasa/fprime/issues/2763  |

---
## Change Description

Implements https://github.com/nasa/fprime/issues/2763

Split up Deframer into 3 components:
- FrameAccumulator
- FprimeDeframer 
- Router (rename `UplinkRouter` ?)

## Rationale

Better reusability and chaining of deframers, preliminary work for CCSDS integrations

## TODO

!!! UPDATE `fprime-util new --deployment` 

## To discuss

- [ ] Should we keep old Deframer in (and what to do with [config/DeframerCfg.hpp](https://github.com/thomas-bc/fprime/blob/76820e63fa820da917fc8ce39b94974261641147/config/DeframerCfg.hpp#L1))


thomas-bc:
Will fix https://github.com/nasa/fprime/issues/2073

ReggieMarr:
@thomas-bc  @LeStarch I made a fork and kinda ran with it in [my repo](https://github.com/ReggieMarr/space-vehicle-system/tree/using_ccsds_framers_deframers) I found what I think is an issue with the frame detection [here](https://github.com/ReggieMarr/fprime/commit/3b9a1a6508d44cb8948fc80e8fc0b4425e395d1b)

Currently I'm workin to fix my TM framer impl but let me know if I'm stepping on anyones toes or going in a bad direction

Svc/FrameAccumulator/FrameAccumulator.cpp
```diff
@@ -0,0 +1,176 @@
...
+                static_cast<FwAssertArgType>(offset),
+                static_cast<FwAssertArgType>(serSize));
+        // Process the data
+        this->processRing();
+        // Update buffer offset and remaining
+        offset += serSize;
+        remaining -= serSize;
+    }
+    // Either all the bytes from the data buffer must be processed, or the ring must be full
+    FW_ASSERT(remaining == 0 || this->m_inRing.get_free_size() == 0, static_cast<FwAssertArgType>(remaining));
```
celskeggs, line 93: Does this mean that data will be silently dropped under some circumstances? Can we have a way to make it not silent (either a FATAL or some other kind of response)?

Svc/FrameAccumulator/FrameAccumulator.cpp
```diff
@@ -0,0 +1,176 @@
...
+                        size_out > remaining,
+                        static_cast<FwAssertArgType>(size_out),
+                        static_cast<FwAssertArgType>(remaining));
+                // Break out of loop: suspend detection until we receive another buffer
+                break;
+            }
+            // No frame was detected or an unknown status was received
+            else {
+                // Discard a single byte of data and start again
+                m_inRing.rotate(1);
```
celskeggs, line 168: What if we _know_ how much data is invalid and should be discarded? Stepping through one byte at a time is likely to be needlessly slow. Yes, some frame detectors won't know how many bytes to discard, but some will.

Svc/FrameAccumulator/FrameAccumulator.fpp
```diff
@@ -0,0 +1,18 @@
...
+
+        @ Port for deallocating buffers received on dataIn.
+        output port dataDeallocate: Fw.BufferSend
+
+
+        @ Port for allocating buffer to hold extracted frame
+        output port frameAllocate: Fw.BufferGet
+
+        @ Port for sending an extracted frame out
+        output port frameOut: Fw.DataWithContext
```
celskeggs, line 15: Why do we send DataWithContext if there's always no context?

Svc/FrameAccumulator/FrameDetector.hpp
```diff
@@ -0,0 +1,50 @@
...
+        //!  3. MORE_DATA_NEEDED status implies that a frame might be possible but more data is needed before a
+        //!     determination is possible. size_out must be set to the total amount of data needed.
+        //!
+        //!     For example, if a frame start word is 4 bytes, and 3 bytes are available in the circular buffer then the
+        //!     return status would be NO_FRAME_DETECTED and size_out must be set to 4 to ensure that at least the start
+        //!     word is available.
+        //!
+        //! \param data: circular buffer with read-only access
+        //! \param size_out: set as output to caller indicating size when appropriate
+        //! \return status of the detection to be pared with size_out
```
celskeggs: pared -> paired?

Svc/FrameAccumulator/FrameDetector/StartLengthCrcDetector.hpp
```diff
@@ -0,0 +1,343 @@
...
+        this->m_value = token;
+        return FrameDetector::FRAME_DETECTED;
+    }
+
+    //! \brief get value of token after read, or 0 before
+    const TokenType& getValue() { return m_value; }
+
+  protected:
+    TokenType m_value;
+};
```
celskeggs, line 96: I don't know if this is my place to comment, since I won't be using the StartLengthCrcDetector, but I do want to say that I think all the templates in this file are needlessly convoluted. I don't see what benefit all of this abstraction adds. If you want to make it easy to implement a custom Start-Length-CRC detector, I'd just provide example code and tell people to copy it. It shouldn't be very complicated.

Svc/Deframer/Deframer.cpp
```diff
@@ -1,312 +1,33 @@
...
-    }
+void Deframer ::framedIn_handler(FwIndexType portNum, Fw::Buffer& data, Fw::Buffer& context) {
 
-    // If more not needed, circular buffer should be empty
-    if (status != DeframingProtocol::DEFRAMING_MORE_NEEDED) {
-        FW_ASSERT(remaining == 0, static_cast<FwAssertArgType>(remaining));
-    }
+    // TODO: add asserts?
+    data.setData(data.getData() + FrameConfig::HEADER_SIZE);
+    data.setSize(data.getSize() - FrameConfig::HEADER_SIZE - FrameConfig::CHECKSUM_SIZE);
```
celskeggs, line 33: I suspect that this isn't going to be something you'll change, but I do want to note that modifying the buffer passed in as a reference seems dangerous. What if the caller didn't expect its buffer pointer and size to change? Yes, the caller should probably realize this could happen and not rely on it (especially since it should expect the data in the buffer to change anyway), but it still makes me nervous.

Svc/FrameAccumulator/FrameAccumulator.cpp
```diff
@@ -0,0 +1,176 @@
...
+                // size_out must be set to the size of the buffer and must fit within the existing data
+                FW_ASSERT(size_out != 0);
+                FW_ASSERT(
+                        size_out <= remaining,
+                        static_cast<FwAssertArgType>(size_out),
+                        static_cast<FwAssertArgType>(remaining));
+                Fw::Buffer buffer = this->frameAllocate_out(0, size_out);
+                if (buffer.isValid()) {
+                    // Copy out data and rotate
+                    FW_ASSERT(this->m_inRing.peek(buffer.getData(), size_out) == Fw::SerializeStatus::FW_SERIALIZE_OK);
```
celskeggs: I'm worried about the performance overhead of this additional copy. In the case that a custom protocol requires _transformation_ of the received data, the decision to split out FrameAccumulator and Deframer means that we'll have to yet another copy in what is already a many-copy architecture.

Here are the copies:

1. The driver loads the received data bytes from the device into a buffer and passes the buffer to FrameAccumulator.
2. FrameAccumulator copies the bytes into its internal ring buffer.
3. FrameAccumulator copies the bytes out of its internal ring buffer into a new allocated frame. (This is the extra copy added by the new model, compared to the existing Deframer.) This frame is passed to the custom deframer.
4. The custom deframer copies the bytes into a new buffer (or reuses the existing buffer) as it translates the received data. The resulting decoded packet is passed onward.
5. The receiving component processes the data and copies it to its final destination, for example saving the data into the filesystem.

I like the idea of splitting out the Router from the Deframer, but I'm not so sure I like the idea of splitting out the FrameAccumulator from the Deframer, at least not in this way.

Also, while I know that F Prime does not discard the condition of an assertion when assertions are disabled, it still unnerves me that we have an assertion condition with a side effect, because that's dangerous in many frameworks.

celskeggs (changes requested):
Thanks for putting together these changes, @thomas-bc. I like the splitting out of the Router from the Deframer, but I'm not so sure I like splitting out the FrameAccumulator from the Deframer; this will actually make it _less_ likely that I can use these stock components in my design. More details in individual comments.

Svc/FrameAccumulator/FrameAccumulator.cpp
LeStarch: The decision to split out the Accumulator was done to provide support for projects that receive a complete packet from the radio.  This allows them to push the buffer from the driver straight through the deframer chain, removing both the copy-in and copy-out of the ring buffer.

I do see your point, that when a rearranging of the incoming data is needed, copying to a buffer only to copy again to rearrange adds overhead.

LeStarch: We discussed this today.  In the near term we will proceed with the current architecture as we'd like to make progress.  There are several ways to mitigate the problem in the near term.

1. (Near term) we will keep the existing deframer in our alternative implementation set. Projects can fall back and use this one if they want the efficiency of  the monolith solution.
2. (Near term) a project might choose to implement a more efficient assembler that assembles a frame into a circular buffer.

Either of these approaches should work, as F Prime is designed to mix and match components to trade efficiency an other project goals.

In the longer term, we have devised an approach that could allow the best of both worlds.  We could provide two interfaces: zero-copy, and fw.buffer.  This could allow deframers to use the circular buffer, or a linear buffer.  The topology engineer would then choose which path to hook-up and as such choose the approach. 

e.g. you have a packet radio that provides an assembled packet.  Use Fw.Buffer side port, and bypass the accumulator. 

e.g. you want to random-access bytes in an assembled packet, use the zero-copy path.

It is a concept for now and it will add some complexity, so solid engineering is needed.  However, it is a viable path forward if the two near-term solutions proved innefficient.

thomas-bc:
After further review, we're going to try to move away from the templated FrameDetectors and instead leverage FPP to specify the header format, and leverage the struct's serializing functions to do the grunt work.

Closing for now so as not to spam CI and yall with notifications - will reopen when we're closer to done

PR #2939, opened 2024-10-10
Random UT Failure: PosixFile

rlcheng:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Update get_test_filename() to exclude certain special characters and change minimum filename length to prevent posix open() from getting an unintentional bad file name.

## Rationale

This fix removes possibility of creating a filename with single character . that causes unintentional error with posix open() in otherwise nominal tests (OpenForRead, OpenFileCreate, and OpenForWrite). Filenames are now 2 character minimum and cannot start with . or /, which can also cause problems.

Tested with seed values known to create the single char . filename: 522402, 588311, and 501569.

## Future Work

None


LeStarch:
@rlcheng do we need to limit to 2+ characters if we are removing specials?

LeStarch (approved):

LeStarch:
I am going to merge the fix anyway.

rlcheng:
We dont have to set it to 2+ but I did find it can be puzzling when you get a file name that is just ; which I did get.

PR #2958, opened 2024-10-16
Farewell, pirate logo!

thomas-bc:
## Change Description

So long, old friend! 
You shall forever be in our git history.


LeStarch (approved):
No!

PR #2995, opened 2024-10-29
Fix posix Task for cases where _SC_PAGESIZE is undefined

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

VxWorks posix implementation does not define `_SC_PAGESIZE` for a call to sysconf.  However, this is only used for an additional check on stack size.  This should not prevent the use of Posix::Tasks should it be undefined.


pcrosemurgy:
Approved. Can confirm that this allows for VxWorks with POSIX support for me.

thomas-bc (approved):

Issue #1378, opened 2022-04-11
FileUplink packet sequence repeat and CRC


timcanham:
| | |
|:---|:---|
|**_F´ Version_**| 3.0 |
|**_Affected Component_**| Svc/FileUplink |
---
## Feature Description

Add duplicate packet detection based on sequence number and don't add to CRC computation

## Rationale

If the radio link is marginal, a radio can do a retry and send the same packet. In that case, you get the following warning:

```
FileUplink_PacketOutOfOrder: Received packet 342 after packet 342
```

The repeated packet still gets added to the CRC computation, so even with a successful retry, you get this warning at the end:

```
FileUplink_BadChecksum: Bad checksum value during receipt of file XXXXX: computed 0x6A4F6972, read 0x92A51AAA
```

The update would to be to not update the CRC only in the case of a repeated packet sequence number. This wouldn't be to try to track all cases of dropped and retried packets.

DJKessler:
@timcanham -- In the case of a repeated packet, should the repeated packet still get written to the file? In that case, I'll simply protect against updating the checksum by wrapping [THIS LINE](https://github.com/nasa/fprime/blob/2c8d0db8411fe955476925fad8e046e8100653ff/Svc/FileUplink/File.cpp#L56) in a condition. 

If we don't want to write the repeated packet to the file, then I'll wrap [THIS BLOCK OF CODE](https://github.com/nasa/fprime/blob/2c8d0db8411fe955476925fad8e046e8100653ff/Svc/FileUplink/FileUplink.cpp#L131-L138) in a condition that uses the same logic as `checkSequenceIndex()`.

timcanham:
@DJKessler I would skip the second write if the first copy was received successfully.

LeStarch:
@DJKessler I think this should be reassigned given the schedule change-up.  Do you agree?

DJKessler:
@LeStarch -- Schedule change-up? I've actually got this fix implemented. I'm about to submit the pull request.

LeStarch:
Closing as complete.

PR #3007, opened 2024-11-04
Fix FileUplink packet sequence repeat and CRC #1378

DJKessler:
| | |
|:---|:---|
|**_Related Issue(s)_**| #1378 |

---
## Change Description

Added duplicate packet detection based on sequence number

- Added duplicated packet event/warning
- Added method to check for duplicated packet
- Added last packet write status `FileUplink` member variable
- Duplicate packets are now detected and skipped
- Added duplicate file packet unit test

### DIFF
--- a/Svc/FileUplink/FileUplink.cpp
+++ b/Svc/FileUplink/FileUplink.cpp
@@ -174,6 +174,18 @@ namespace Svc {
     this->m_lastSequenceIndex = sequenceIndex;
   }
 
+  bool FileUplink ::
+    checkDuplicatedPacket(const U32 sequenceIndex)
+  {
+    // check for duplicate packet
+    if (sequenceIndex == this->m_lastSequenceIndex) {
+      this->m_warnings.packetDuplicate(sequenceIndex);
+      return true;
+    }
+
+    return false;
+  }
+
   void FileUplink ::
     compareChecksums(const Fw::FilePacket::EndPacket& endPacket)
   {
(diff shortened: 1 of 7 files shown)


LeStarch (approved):
This is what I see:  if the last packet was successful *and* this packet is a duplicate, then the rest of the function (writing, CRC) is skipped.  To my estimate this completes the work!

LeStarch:
@timcanham would you review too, given you were more involved with this work?

Svc/FileUplink/FileUplink.cpp
```diff
@@ -174,6 +185,18 @@ namespace Svc {
     this->m_lastSequenceIndex = sequenceIndex;
   }
 
+  bool FileUplink ::
+    checkDuplicatedPacket(const U32 sequenceIndex)
+  {
+    // check for duplicate packet
+    if (sequenceIndex == this->m_lastSequenceIndex) {
+      this->m_warnings.packetDuplicate(sequenceIndex);
```
timcanham, line 193: How do you avoid a duplicate on the first packet if the first packet sequence number is 0?

DJKessler: Fair point. I suppose `m_lastSequenceIndex` should be initialized to some non-zero value that also won't be a valid packet index (eg. `U32MAX`).

timcanham: I think it would actually drop the first packet, since the last status was OK.

DJKessler: You're right; `m_lastPacketWriteStatus` should be initialized to something other than `Os::File::OP_OK`.

timcanham: Whatever solution you come up with, you should try running it against the ground system and uplink a file nominally to make sure something doesn't get through.

timcanham (approved):
Looks good!

LeStarch:
Thanks @DJKessler!

PR #3008, opened 2024-11-04
moved Drv ports to a Ports directory

kevin-f-ortega:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Moving ports in fprime/Drv to fprime/Drv/Ports

## Rationale

Having a `Ports` directory for ports is inline with how F Prime structures its directories

Run through CI

## Future Work


LeStarch (changes requested):
1. Are you going to merge the port directories?
2. Can we move Drv_DataTypes into the Ref App, as it is only used by sample components and has no business outside of that App.

kevin-f-ortega:
@LeStarch,

I merged the port directories.

Drv_DataTypes is being used by Drv.BlockDriver; so, I cannot move DataTypes outside of Drv. I went ahead and added the DataTypes directory into the Ports directory.

LeStarch (approved):
Ok, I understand.  We need a better replacement for BlockDriver.....but it is not available today.

PR #3009, opened 2024-11-06
Os file vxworks

kevin-f-ortega:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Adding macros to include/exclude code when compiling for VxWorks. 

## Rationale

This allows VxWorks to build with the Posix file implementation. 

I ran a simple hello world program in the VxSim. The hello world program opened a file and wrote some bytes to the file.

## Future Work

### DIFF
--- a/Os/Posix/Directory.cpp
+++ b/Os/Posix/Directory.cpp
@@ -2,7 +2,7 @@
 // \title Os/Posix/Directory.cpp
 // \brief Posix implementation for Os::Directory
 // ======================================================================
-#include <sys/errno.h>
+#include <errno.h>
 #include <sys/stat.h>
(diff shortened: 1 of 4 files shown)


kevin-f-ortega:
NOTE: I used F Prime's .clang-format to format the files I modified.

LeStarch (commented):
This is currently a runtime-check that asserts when running.  Would it be better to do a build-time check?  This could be done by conditionally setting the constructors to =delete in the header in the VxWorks build. 

Rather than crash the FSW, this would fail to link if that code was ever used.  Thoughts?

LeStarch (approved):

Issue #3005, opened 2024-11-04
fprime-seqgen passes but FSW crashes on empty seq


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 3.4.3 |
|**_Affected Component_**|   CmdSequencer, fprime-seqgen |
---
## Problem Description

1. Create an empty `.seq` file
2. Run fprime-seqgen on it
```
$ fprime-seqgen empty.seq -d <some dictionary>
Sequence is 0 bytes with timebase ANY
CRC: 4017330436 (0xEF739904)
$ echo $?
0
```
3. Run with `CS_RUN`:
```
Assert: "<path>/CmdSequencer/FPrimeSequence.cpp:97" 3
FATAL 14273 handled.
```
## Context / Environment

Execute `fprime-util version-check` and share the output.
```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-6.5.0-26-generic-x86_64-with-glibc2.35
Python version: 3.10.12
CMake version: 3.27.2
Pip version: 22.0.2
Pip packages:
    fprime-tools==3.4.4
    fprime-gds==3.4.3
    fprime-fpp-*==2.1.0a3
```

## Expected Behavior

1. Either CmdSequencer should not assert, or fprime-seqgen should fail.

LeStarch:
We merged a fix for this.

PR #3025, opened 2024-11-14
Bugfix/empty seq

jsilveira1409:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3005  |

---
## Change Description

Implemented FileEmpty warning for the `Svc::CmdDispatcher` Component, which is called when deserializing the next record returns an empty buffer.

## Rationale
The Flight Software should raise a warning on empty command sequence files instead of crashing.

### DIFF
--- a/Svc/CmdSequencer/Events.fppi
+++ b/Svc/CmdSequencer/Events.fppi
@@ -214,3 +214,11 @@ event CS_JoinWaitingNotComplete() \
   severity warning high \
   id 24 \
   format "Still waiting for sequence file to complete"
+
+event CS_FileEmpty(
+                      #fileName: string size 60 @< The name of the sequence file
+                    ) \
+  severity warning high \
+  id 25 \
+  format "Sequence file is empty. Ignoring."
+  #format "Sequence file {} is empty. Ignoring."
--- a/Svc/CmdSequencer/Events.fppi
+++ b/Svc/CmdSequencer/Events.fppi
@@ -215,10 +215,9 @@ event CS_JoinWaitingNotComplete() \
   id 24 \
   format "Still waiting for sequence file to complete"
 
-event CS_FileEmpty(
-                      #fileName: string size 60 @< The name of the sequence file
+event CS_NoRecords(
+                      fileName: string size 60 @< The name of the sequence file
                     ) \
-  severity warning high \
+  severity warning low \
   id 25 \
-  format "Sequence file is empty. Ignoring."
-  #format "Sequence file {} is empty. Ignoring."
+  format "Sequence file {} has no records. Ignoring."
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -470,6 +470,7 @@ jdperez
 jenkins
 jenkinsci
 jishii
+Joaquim
 jobrestrictions
 joshuaa
 jpl
@@ -594,6 +595,7 @@ nogen
 noncomma
 NONINFRINGEMENT
 noparent
+norecords
 normalwidths
 NOSIZE
 NOSPEC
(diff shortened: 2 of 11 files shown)


Svc/CmdSequencer/Events.fppi
```diff
@@ -214,3 +214,11 @@ event CS_JoinWaitingNotComplete() \
...
   id 24 \
   format "Still waiting for sequence file to complete"
+
+event CS_FileEmpty(
+                      #fileName: string size 60 @< The name of the sequence file
+                    ) \
+  severity warning high \
+  id 25 \
+  format "Sequence file is empty. Ignoring."
+  #format "Sequence file {} is empty. Ignoring."
```
LeStarch: We need to clean-out the commented out code.  Preferably we'd include the sequence name.

What scenario would this hit?  Doesn't the sequence validation catch this (e.g. the CRC)?

Additionally, can this code be hit while a sequence is running? (e.g. the sequence file wasn't fully uplinked).

LeStarch (changes requested):
A few minor changes and questions.

LeStarch:
@timcanham and @zimri-leisher I'd appreciate your thoughts on this PR.

Svc/CmdSequencer/Events.fppi
jsilveira1409: My bad for the comment, forgot about them!

`cmdSeq.CS_VALIDATE` does validate empty command sequences: 

image

So actually the `CS_VALIDATE` would also need to trigger an empty command sequence event.

jsilveira1409: So actually, to encompass both cases, a more elegant solution could be to simply verify the number of records when loading the sequence file at `FPrimeSequence` level, and return false if the number of records is zero, which would trigger the following events: 

image

This would be the minor change: 

```
bool CmdSequencerComponentImpl::FPrimeSequence ::
    validateRecords()
  {
    Fw::SerializeBufferBase& buffer = this->m_buffer;
    const U32 numRecords = this->m_header.m_numRecords;
    Sequence::Record record;

    if (numRecords == 0)
    {
      return false;
    }

( ... )

}
```

zimri-leisher: I think the FileEmpty solution may be technically inaccurate as the file still contains a sequence header and so is not literally empty. If we're going to use an event, maybe we'd better call it `SeqEmpty` or `NoRecords`, and make it clear in the error message that there specifically weren't any commands in the sequence. I like the second solution better, if this is the direction we choose to go.

zimri-leisher:
I would add/modify a unit test to check this behavior.

EDIT--On second thought, instead of EXECUTION_ERRORing, should we just allow the user to run empty sequences? My team intentionally did this and thought it would work (we were using the seq as a placeholder). In general tools should allow as much valid input as possible to make the fewest decisions for users. Maybe we could still raise a warning just in case it was a mistake.

jsilveira1409:
@zimri-leisher @LeStarch how about this event logs:

image

zimri-leisher:
In my opinion this could be a low warning, but otherwise the event looks good! I assume it's the same for CS_VALIDATE.

jsilveira1409:
@zimri-leisher yeah makes sense to make it low warning, here is the full logs with CS_VALIDATE too : 
image

zimri-leisher:
Looks good, just missing a unit test.

zimri-leisher (approved):

timcanham (approved):
LGTM!

LeStarch (approved):

LeStarch:
Looks good to me!  I added a commit to allow the improperly flagged misspellings.  Thanks for working this through!  @zimri-leisher thanks for the review!

PR #3037, opened 2024-11-21
Adding restrict_platform support for FPRIME_HAS_XYZ flags

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

`restrict_platforms` is a nice feature to prevent code from attempting to build on platforms that do not support a given module.  However, ongoing work in the VxWorks area has shown that some platforms support features (e.g. SOCKETS) without the platform (e.g. VxWorks) in the restrict-to list nor the restrict-to pseudo list (e.g. Posix).

This implies that restrict_platforms should also support feature sets.  This PR allows syntax like the following:

```
restrict_plarforms(SOCKETS)
```

and allows platform developers to set the corresponding flag

```
set(FPRIME_HAS_SOCKETS)
```
 
This will allow platforms to pass for code built on a feature set that isn't as clear cut as platform (e.g. Linux), toolchain (e.g. arm-linux), or Posix sudo-os.

Extending to the future, this could be used to allow code that is built on platforms with pre-supplied libraries installed (e.g. the platform and cross-compiler supply opencv). This could be represented as FPRIME_HAS_OPENCV.

Note: OPENCV here is a hypothetical example, and is not implemented in this PR.


LeStarch:
@thomas-bc and @kevin-f-ortega  this PR is needed to build VxWorks w/ sockets when the VxWorks posix support layer is turned off.

Svc/PosixTime/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@
 #
 # Note: using PROJECT_NAME as EXECUTABLE_NAME
 ####
-restrict_platforms(Posix)
+#restrict_platforms(Posix)
```
kevin-f-ortega: Did you mean to comment this line out?

kevin-f-ortega (commented):
Looks good to me, thanks! Just one question

Svc/PosixTime/CMakeLists.txt
LeStarch: No!  Good catch!

kevin-f-ortega (approved):

PR #3040, opened 2024-11-22
Clarify buffer ownership for ByteStreamDriverModel

celskeggs:
| | |
|:---|:---|
|**_Documentation Included (y/n)_**| yes |

---
## Change Description

`TcpClientComponentImpl::send_handler` and `TcpServerComponentImpl::send_handler` deallocate the provided `Fw::Buffer` if (and only if) they return SEND_OK or SEND_ERROR. If they return SEND_RETRY, they do not deallocate the buffer.

## Rationale

To make sure that users of the ByteStreamDriverInterface can implement the correct buffer handling logic, it's important to define what happens with ownership of the buffer.

N/A

## Future Work

`UdpComponentImpl::send_handler` does not implement the interface in the same way: notably, it also deallocates the buffer when it returns SEND_RETRY. I believe this is incorrect and has not been noticed because SEND_RETRY is unlikely (or maybe impossible) for UDP connections. Some revision to that code may be necessary.

### DIFF
--- a/Drv/ByteStreamDriverModel/docs/sdd.md
+++ b/Drv/ByteStreamDriverModel/docs/sdd.md
@@ -11,11 +11,11 @@ The manager component (typically the ground interface) initiates the transfer of
 The caller will provide a `Fw::Buffer` containing the data to send and the port call will return a status of that send.
 These responses are an enumeration whose values are described in the following table:
 
-| Value | Description |
-|---|---|
-| Drv::SEND_OK    | Send functioned normally. |
-| Drv::SEND_RETRY | Send should be retried, but a subsequent send should return SEND_OK. |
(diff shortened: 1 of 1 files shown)


LeStarch (approved):

LeStarch:
@celskeggs thanks for this clarification!  Also, you are right on the UDP issue and we should fix it for conformity with this pattern.

I took a look at the UART driver (also a byte stream) and it has one line we might review:

https://github.com/nasa/fprime/blob/00c5c8e11d11f8aa027e83fc28db5da5b7cbf5c7/Drv/LinuxUartDriver/LinuxUartDriver.cpp#L334

That is, it only takes ownership in the case when the port is attached.  I'd argue that it would be better to require that port to be attached.  There are other patterns to prevent deallocation that would be clearer given the intended constraints of this pattern you documented here.

Issue #3030, opened 2024-11-18
Build fails


garthwatney:
| | |
|:---|:---|
|**_F´ Version_**| |
v3.5.0
|**_Affected Component_**|   |
---
## Problem Description
Build fails:
In member function ‘void Drv::LinuxGpioDriver::pollLoop()’:
fprime/Drv/LinuxGpioDriver/LinuxGpioDriver.cpp:290:37: error: comparison of integer expressions of different signedness: ‘const FwSizeType’ {aka ‘const long unsigned int’} and ‘int’ [-Werror=sign-compare]
  290 |     static_assert(GPIO_POLL_TIMEOUT < std::numeric_limits<int>::max(), "Poll timeout would overflow");
      |                   ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
fprime/Drv/LinuxGpioDriver/LinuxGpioDriver.cpp:292:55: error: comparison of integer expressions of different signedness: ‘long int’ and ‘long unsigned int’ [-Werror=sign-compare]
  292 |     static_assert(std::numeric_limits<ssize_t>::max() <= std::numeric_limits<FwSizeType>::max(), "FwSizeType too small");
      |                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors

A description of the problem with sufficient detail to understand the issue.

## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.29
Python version: 3.8.10
CMake version: 3.23.1
Pip version: 20.0.2
Pip packages:
    fprime-tools==3.5.0
    fprime-gds==3.5.0
    fprime-fpp-*==2.2.0
--INSERT HERE--
```

## How to Reproduce

1. cd to the top level directory of the fprime repo
2. fprime-util generate -f
3. fprime-util build

## Expected Behavior

A description of the expected behavior.

jsilveira1409:
I tried reproducing this bug but was not able to :(

LeStarch:
This was fixed!

PR #3118, opened 2025-01-08
LinuxGpioDriver: fix static_assert's for FwSizeType. Possible fix for #3030

AlesKus:
| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime#3030 |

---
## Change Description

A possible fix to avoid warning for signed/unsigned comparison in LinuxGpioDriver.
The signed max() values are cast to corresponding unsigned types. The cast is safe because max() returns nonnegative value which shall fit to unsigned type.

## Rationale

Possible fix for issue nasa/fprime#3030.

Steps to reproduce and test:
1. cd to the top level directory of the fprime repo
2. fprime-util generate -f
3. fprime-util build

There should be no build errors.

## Future Work

N/A.

### DIFF
--- a/Drv/LinuxGpioDriver/LinuxGpioDriver.cpp
+++ b/Drv/LinuxGpioDriver/LinuxGpioDriver.cpp
@@ -20,6 +20,7 @@
 #include <sys/ioctl.h>
 #include <unistd.h>
 #include <cerrno>
+#include <type_traits>
 
 namespace Drv {
 
@@ -287,9 +288,13 @@ Drv::GpioStatus LinuxGpioDriver ::gpioWrite_handler(const NATIVE_INT_TYPE portNu
 }
 
 void LinuxGpioDriver ::pollLoop() {
-    static_assert(GPIO_POLL_TIMEOUT < std::numeric_limits<int>::max(), "Poll timeout would overflow");
+    // Ensure size of FwSizeType is large enough to fit the necessary ranges
+    // NOTE: casts to unsigned types for int and ssize_t are made to avoid sign-compare warning;
+    //       in both cases the cast is safe because max() returns nonnegative value.
+    static_assert(GPIO_POLL_TIMEOUT < static_cast<unsigned int>(std::numeric_limits<int>::max()), "Poll timeout would overflow");
     static_assert(sizeof(struct gpioevent_data) < std::numeric_limits<FwSizeType>::max(), "FwSizeType too small");
-    static_assert(std::numeric_limits<ssize_t>::max() <= std::numeric_limits<FwSizeType>::max(), "FwSizeType too small");
+    typedef std::make_unsigned<ssize_t>::type unsigned_ssize;
+    static_assert(static_cast<unsigned_ssize>(std::numeric_limits<ssize_t>::max()) <= std::numeric_limits<FwSizeType>::max(), "FwSizeType too small");
     // Setup poll information
     pollfd file_descriptors[1];
     // Loop forever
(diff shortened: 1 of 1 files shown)


Drv/LinuxGpioDriver/LinuxGpioDriver.cpp
```diff
@@ -287,9 +288,13 @@ Drv::GpioStatus LinuxGpioDriver ::gpioWrite_handler(const NATIVE_INT_TYPE portNu
...
 void LinuxGpioDriver ::pollLoop() {
-    static_assert(GPIO_POLL_TIMEOUT < std::numeric_limits<int>::max(), "Poll timeout would overflow");
+    // Ensure size of FwSizeType is large enough to fit the necessary ranges
+    // NOTE: casts to unsigned types for int and ssize_t are made to avoid sign-compare warning;
+    //       in both cases the cast is safe because max() returns nonnegative value.
+    static_assert(GPIO_POLL_TIMEOUT < static_cast<unsigned int>(std::numeric_limits<int>::max()), "Poll timeout would overflow");
     static_assert(sizeof(struct gpioevent_data) < std::numeric_limits<FwSizeType>::max(), "FwSizeType too small");
-    static_assert(std::numeric_limits<ssize_t>::max() <= std::numeric_limits<FwSizeType>::max(), "FwSizeType too small");
+    typedef std::make_unsigned<ssize_t>::type unsigned_ssize_t;
+    static_assert(static_cast<unsigned_ssize_t>(std::numeric_limits<ssize_t>::max()) <= std::numeric_limits<FwSizeType>::max(), "FwSizeType too small");
```
LeStarch, line 297: Should we not just use `size_t` here?  That should be the signed equivalent of `ssize_t`.

I am hesitant to use `std::make_unsigned<ssize_t>::type` as it is unclear if this is C++11 or C++14.   F Prime currently must be c++11 compliant.

LeStarch (changes requested):

Drv/LinuxGpioDriver/LinuxGpioDriver.cpp
AlesKus: At the moment, I didn't find 100% evidence that `ssize_t` is always a signed equivalent of `size_t`.
POSIX [does not define](https://pubs.opengroup.org/onlinepubs/007904975/basedefs/sys/types.h.html) any relations between these types. However, usually they are equivalent ([example](https://awesomekling.github.io/How-SerenityOS-declares-ssize_t/)).
The [template< class T > struct make_unsigned](https://en.cppreference.com/w/cpp/types/make_unsigned) is from C++11.

AlesKus: An alternative approach is to use SSIZE_MAX from <limits.h>.
`static_assert(SSIZE_MAX <= std::numeric_limits<FwSizeType>::max(), "FwSizeType too small");`

LeStarch: After more review, I accept your original approach.  I've asked another colleague to look at it since this is a new C++ism for the project.

bocchino:
Looks good to me! I just have a couple of comments:

1. Shouldn't `std::make_unsigned<ssize_t>::type` be the same as `size_t`?
2. If we go with `std::make_unsigned<ssize_t>::type`, let's write `using` instead of `typedef` per C++11 coding practice.

LeStarch (approved):

LeStarch:
@bocchino that was in the discussion i resolved.  The belief is that they should be the same, but there isn't any documented guarantee and as such this approach was chosen.

@AlesKus I agree with switching to `using` and I went ahead and made the change.  Will merge once CI passes!

LeStarch:
Merging!  Thanks @AlesKus !

PR #3132, opened 2025-01-14
Port patterns

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Documentation of some common port patterns.


docs/documentation/user-manual/framework/patterns/common-port-patterns.md
```diff
@@ -0,0 +1,276 @@
...
+        ... do work ...
+    }
+}
+```
+
+### Conclusion
+
+The synchronous cancel  pattern is used to stop work running inside an `async` handler. It is performed by atomically setting a cancel flag in a `sync` handler and polling that flag during the work running in the `async` handler.
+
+An example of this pattern can be found as part of the [Worker Component](--TODO--) in F Prime examples.
```
thomas-bc: I would call this out at the beginning of this document

docs/documentation/user-manual/framework/patterns/common-port-patterns.md
```diff
@@ -0,0 +1,276 @@
+# Common Port Design Patterns
+
```
thomas-bc: Move this file into the soon-to-be `user-manual/design-pattern` directory

LeStarch:
I think the bad links will be fixed by the other PR as I did not touch those files.

docs/user-manual/design-patterns/common-port-patterns.md
```diff
@@ -0,0 +1,285 @@
...
+```
+
+Finally, the port handler should be minimal often returning just returning the value of a member variable previously set.
+
+```c++
+F32 getTemperature_handler() {
+    return this->m_previously_set_temperature;
+}
+```
+> ![NOTE]
```
thomas-bc: Could you add a newline above here? Between ``` and `> ![NOTE]` 
Seems GitHub is unhappy about rendering this

thomas-bc (approved):

docs/user-manual/design-patterns/common-port-patterns.md
thomas-bc: My bad it's not the newline, it should be `> [!NOTE]` and not `> ![NOTE]`

thomas-bc (approved):

Issue #1708, opened 2022-10-04
Remove vsnprintf from StringType.cpp - pulls in libraries for embedded


timcanham:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

Analyze the usage of of printf family of functions to minimize the code space requirements on deeply embedded systems.

- [x] Ref. Note: Kept `printf` in Main.cpp
- [x] RPI. Note: Kept `printf` in Main.cpp
- [x] Drv
- [ ] Fw
- [x] Os
- [x] Svc
- [x] Utils
- [ ] Everything Else

LeStarch:
Command:

```
grep -R "printf" . | grep -v "./.git"  | grep -v "/test/ut" | grep -v "./googletest" | grep -v "./STest" | grep -v "./Autocoders" | grep -v "tst_crc.c" | grep -v ".md:" | grep -v "UnitTest" | awk -F: '{print "|" $1 "|`" $2 "`| | |"}' 
```

Results

|   | File                                                               | Line                                                                                                                                                                          | Risk | Recommentation |
|---|--------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------|----------------|
| <ul><li>[ ] Done</li></ul> | ./Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp              | `//#define DEBUG_PRINT(...) printf(##__VA_ARGS__); fflush(stdout)`                                                                                                             |      |                |
| <ul><li>[ ] Done</li></ul> | ./Drv/LinuxSpiDriver/LinuxSpiDriverComponentImpl.cpp              | `        snprintf(devName,sizeof(devName),"/dev/spidev%d.%d",device,select);`                                                                                                  |      |                |
| <ul><li>[ ] Done</li></ul> | ./Drv/LinuxGpioDriver/LinuxGpioDriverComponentImpl.cpp            | `//#define DEBUG_PRINT(...) printf(##__VA_ARGS__); fflush(stdout)`                                                                                                             |      |                |
| <ul><li>[ ] Done</li></ul> | ./Drv/LinuxGpioDriver/LinuxGpioDriverComponentImpl.cpp            | `        len = snprintf(buf, sizeof(buf), "%u", gpio);`                                                                                                                       |      |                |
[...]

PR #3140, opened 2025-01-16
Removing DEBUG_PRINT like statements. Part of: #1708

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Removes DEBUG_PRINT and other variants of this pattern from flight code.

## Rationale

This is a step to addressing #1708. 

Additionally, this pattern causes a lot of dead and commented-out code, which is not the best practice.

Please assert that only these changes have been made:

1. DEBUG_PRINT lines were removed
2. DEBUG_PRINT defines were removed
3. Clauses and functions left empty by 1 or 2 were removed.

## Future Work

### DIFF
--- a/Utils/Types/CircularBuffer.hpp
+++ b/Utils/Types/CircularBuffer.hpp
@@ -132,9 +132,6 @@ class CircularBuffer {
          */
         void clear_high_water_mark();
 
-#ifdef CIRCULAR_DEBUG
-        void print();
-#endif
     PRIVATE:
         /**
          * Returns a wrap-advanced index into the store.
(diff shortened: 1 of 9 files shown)


Utils/Types/CircularBuffer.hpp
```diff
@@ -20,8 +20,6 @@
 #include <FpConfig.hpp>
 #include <Fw/Types/Serializable.hpp>
 
-//#define CIRCULAR_DEBUG
-
 namespace Types {
 
 class CircularBuffer {
```
thomas-bc, line 25: We should also remove the declaration
```
#ifdef CIRCULAR_DEBUG
        void print();
#endif
```
from this file

thomas-bc (changes requested):
Confirmed 1), 2) and 3) from Review recommendations

thomas-bc (approved):

PR #3143, opened 2025-01-16
UDP does not bind on send only port. Fixes: #3127

### LINKED ISSUE #3127: UDP driver errno 22: Invalid Argument
| | |
|:---|:---|
|**_F´ Version_**| 3.5.1 |
|**_Affected Component_**|  Udp |
---
## Problem Description

When trying to send to a UDP port with the `Udp` component, we are seeing errno 22 raised (invalid argument)

The driver is being configured like so:
```
instance x: Drv.Udp base id 0x1234 {
    phase Fpp.ToCpp.Phases.configComponents """
    x.configureSend(ip_c_str, port_int, 0, UDP_TIMEOUT_US);
    x.open();
    """

    phase Fpp.ToCpp.Phases.tearDownComponents """
    x.stop();
    x.join();
    """
}
```
From internal testing, we have validated that the data we are passing to the send port is a valid... Unsure what could be causing this.
## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python version: 3.10.12
CMake version: 3.22.1
Pip version: 22.0.2
Pip packages:
    fprime-tools==3.5.1
    fprime-gds==3.5.1
    fprime-fpp-*==2.2.1
Project submodules:
```

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Removes the attempt to bind on UDP ports where only `send` was configured.

I also fixed the reconnect handling now that `send` can also attempt to reconnect.

UTs to check the following would be nice:

- [x] UDP (send only)
- [x]  UDP (recv only)
- [x] Disabled autoconnect properly results in autoconnect disabled.
   - [x] In thread
   - [x] In send
   - [x] Via thread start method
   - [x] Via `setAutoConnect` method 
- [x] Update SDDs for this behavior
  - [x] UDP in send/recv only mode
  - [x] Autoconnect enabled
  - [x] Autoconnect disabled
  - [x] Implication on client recv loop
  - [x] Implication on server recv loop

## Rationale

UDP may send, receive, or both depending.  When a send only UDP calls `bind` it fails because the receive port is invalid.

Fixed: #3127 #3133.

Tested with UDP in the downlink chain of Ref (only downlink) works as expected with the following caveat:

`configureSend` must be called before starting threads.

## Future Work

Decided to do the future work!

### DIFF
--- a/Drv/Ip/SocketComponentHelper.cpp
+++ b/Drv/Ip/SocketComponentHelper.cpp
@@ -28,7 +28,7 @@ void SocketComponentHelper::start(const Fw::StringBase &name,
                                      const Os::Task::ParamType cpuAffinity) {
     FW_ASSERT(m_task.getState() == Os::Task::State::NOT_STARTED);  // It is a coding error to start this task multiple times
     this->m_stop = false;
-    m_reconnect = reconnect;
+    this->setAutoconnect(reconnect);
     // Note: the first step is for the IP socket to open the port
     Os::Task::Arguments arguments(name, SocketComponentHelper::readTask, this, priority, stack, cpuAffinity);
     Os::Task::Status stat = m_task.start(arguments);
@@ -77,13 +77,28 @@ bool SocketComponentHelper::isOpened() {
     return is_open;
 }
 
+void SocketComponentHelper::setAutoConnect(bool auto_connect) {
+    Os::ScopeLock scopedLock(this->m_lock);
+    this->m_reconnect = auto_connect;
+}
+
 SocketIpStatus SocketComponentHelper::reconnect() {
     SocketIpStatus status = SOCK_SUCCESS;
-    // Open a network connection if it has not already been open
     if (not this->isOpened()) {
-        status = this->open();
-        if (status == SocketIpStatus::SOCK_ANOTHER_THREAD_OPENING) {
-            status = SocketIpStatus::SOCK_SUCCESS;
+        // Check for autoconnect before attempting to reconnect
+        bool reconnect = false;
+        {
+            Os::ScopeLock scopedLock(this->m_lock);
+            reconnect = this->m_reconnect;
+        }
+        // Open a network connection if it has not already been open
+        if (not reconnect) {
+            status = SOCK_AUTOCONNECT_DISABLED;
+        } else {
+            status = this->open();
+            if (status == SocketIpStatus::SOCK_ANOTHER_THREAD_OPENING) {
+                status = SocketIpStatus::SOCK_SUCCESS;
+            }
         }
     }
     return status;
--- a/Drv/Ip/SocketComponentHelper.cpp
+++ b/Drv/Ip/SocketComponentHelper.cpp
@@ -85,7 +85,7 @@ void SocketComponentHelper::setAutoConnect(bool auto_connect) {
 SocketIpStatus SocketComponentHelper::reconnect() {
     SocketIpStatus status = SOCK_SUCCESS;
     if (not this->isOpened()) {
-        // Check for autoconnect before attempting to reconnect
+        // Check for auto-connect before attempting to reconnect
         bool reconnect = false;
         {
             Os::ScopeLock scopedLock(this->m_lock);
@@ -213,7 +213,7 @@ void SocketComponentHelper::readLoop() {
             this->sendBuffer(buffer, status);
         }
     }
-    // This will loop until stopped. If autoconnect is disabled, this will break at the moment of reconnect
+    // This will loop until stopped. If auto-connect is disabled, this will break at the moment of reconnect
     while (this->running());
     // Close the socket
     this->close(); // Close the port entirely
--- a/Drv/Ip/SocketComponentHelper.cpp
+++ b/Drv/Ip/SocketComponentHelper.cpp
@@ -93,7 +93,7 @@ SocketIpStatus SocketComponentHelper::reconnect() {
         }
         // Open a network connection if it has not already been open
         if (not reconnect) {
-            status = SOCK_AUTOCONNECT_DISABLED;
+            status = SOCK_AUTO_CONNECT_DISABLED;
         } else {
             status = this->open();
             if (status == SocketIpStatus::SOCK_ANOTHER_THREAD_OPENING) {
@@ -182,7 +182,7 @@ void SocketComponentHelper::readLoop() {
         if ((not this->isOpened()) and this->running()) {
             status = this->reconnect();
             // When reconnect is disabled, just break as this is a exit condition for the loop
-            if (status == SOCK_AUTOCONNECT_DISABLED) {
+            if (status == SOCK_AUTO_CONNECT_DISABLED) {
                 break;
             }
             // If the reconnection failed in any other way, warn, wait, and retry
--- a/Drv/Ip/SocketComponentHelper.cpp
+++ b/Drv/Ip/SocketComponentHelper.cpp
@@ -22,13 +22,11 @@ SocketComponentHelper::SocketComponentHelper() {}
 SocketComponentHelper::~SocketComponentHelper() {}
 
 void SocketComponentHelper::start(const Fw::StringBase &name,
-                                     const bool reconnect,
-                                     const Os::Task::ParamType priority,
-                                     const Os::Task::ParamType stack,
-                                     const Os::Task::ParamType cpuAffinity) {
+                                  const Os::Task::ParamType priority,
+                                  const Os::Task::ParamType stack,
+                                  const Os::Task::ParamType cpuAffinity) {
     FW_ASSERT(m_task.getState() == Os::Task::State::NOT_STARTED);  // It is a coding error to start this task multiple times
     this->m_stop = false;
-    this->setAutoConnect(reconnect);
     // Note: the first step is for the IP socket to open the port
     Os::Task::Arguments arguments(name, SocketComponentHelper::readTask, this, priority, stack, cpuAffinity);
     Os::Task::Status stat = m_task.start(arguments);
@@ -77,22 +75,22 @@ bool SocketComponentHelper::isOpened() {
     return is_open;
 }
 
-void SocketComponentHelper::setAutoConnect(bool auto_connect) {
+void SocketComponentHelper::setAutomaticOpen(bool auto_open) {
     Os::ScopeLock scopedLock(this->m_lock);
-    this->m_reconnect = auto_connect;
+    this->m_reopen = auto_open;
 }
 
-SocketIpStatus SocketComponentHelper::reconnect() {
+SocketIpStatus SocketComponentHelper::reopen() {
     SocketIpStatus status = SOCK_SUCCESS;
     if (not this->isOpened()) {
-        // Check for auto-connect before attempting to reconnect
-        bool reconnect = false;
+        // Check for auto-open before attempting to reopen
+        bool reopen = false;
         {
             Os::ScopeLock scopedLock(this->m_lock);
-            reconnect = this->m_reconnect;
+            reopen = this->m_reopen;
         }
         // Open a network connection if it has not already been open
-        if (not reconnect) {
+        if (not reopen) {
             status = SOCK_AUTO_CONNECT_DISABLED;
         } else {
             status = this->open();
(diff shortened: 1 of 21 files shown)


LeStarch:
@zimri-leisher want to take a look?

Drv/Ip/UdpSocket.cpp
```diff
@@ -151,23 +151,34 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
...
+    } else if ((port > 0) && (recv_port > 0))  {
         Fw::Logger::log("Setup to receive udp at %s:%hu and send to %s:%hu\n",
                            m_recv_hostname,
                            recv_port,
                            m_hostname,
                            port);
     }
+    // Neither configuration method was called
+    else {
+        FW_ASSERT(port > 0 || recv_port > 0, port, recv_port);
```
LeStarch, line 187: I added this assert because it is invalid to attempt connection before configure is called. However,  I think TCP does not assert in this case.  It just fails and tries again until the user calls configure.

@csmith608 what do you think?  Should we assert if the auto-connect happens before configure, or just record an error and try again?

If we go with the assert, then calls to tcp/udp/....configure would be forced into the configure phase (where they belong), but this would be a disruptive change to users.

How it is now is more lenient...but if you do not use autoconnect then your code won't assert on the missordering when there is not change to recover.

Thoughts?

Drv/Ip/SocketComponentHelper.cpp
```diff
@@ -77,13 +77,28 @@ bool SocketComponentHelper::isOpened() {
...
-            status = SocketIpStatus::SOCK_SUCCESS;
+        // Check for auto-connect before attempting to reconnect
+        bool reconnect = false;
+        {
+            Os::ScopeLock scopedLock(this->m_lock);
+            reconnect = this->m_reconnect;
+        }
+        // Open a network connection if it has not already been open
+        if (not reconnect) {
+            status = SOCK_AUTO_CONNECT_DISABLED;
```
zimri-leisher, line 94: reconnect is a public method. not guaranteed that this method is only called when automatically reconnecting--it may be the case that someone is manually calling reconnect. if someone is manually calling reconnect, in my opinion i would trust them. It would be confusing if I had to set auto reconnect to true before I manually reconnected. i'd suggest either making it private or having the check for `m_reconnect` before calling `reconnect`, or even better just inlining this function as it's only used in a couple of places.

Drv/Ip/SocketComponentHelper.cpp
```diff
@@ -77,13 +77,28 @@ bool SocketComponentHelper::isOpened() {
     return is_open;
 }
 
+void SocketComponentHelper::setAutoConnect(bool auto_connect) {
+    Os::ScopeLock scopedLock(this->m_lock);
+    this->m_reconnect = auto_connect;
+}
+
 SocketIpStatus SocketComponentHelper::reconnect() {
```
zimri-leisher: one thing I was initially confused about is that `reconnect`, for `UdpSocket.cpp`, does not do any connecting (of course because UDP is stateless). Maybe a docstring describing its intended behavior or method rename would be useful, or maybe that's scope creep. This method is only used in the context of the receive loop, so maybe it could be `setupReceive`? It's hard to find a nice abstract name because in the case of Udp, it's binding, in the case of TcpClient, it's connecting, in the case of TcpServer it's accepting. All three result in the ability to receive data.

Alternatively, because it calls and checks `open`, maybe it could be called reopen()? Otherwise the naming scheme isn't self-consistent (is it connecting or is it opening?)

Drv/Ip/SocketComponentHelper.cpp
zimri-leisher: as a side note--it would be nice to be able to set a timeout on the connection attempt. this is because on line 99 (pre-changes in this PR), it attempts to reconnect on the sender's thread. This means that even though we may have set the socket up with a 1 ms timeout, the send port call could take longer than this. In our testing we were observing it delay 10-20 ms before giving up, but this is likely OS dependent.

Drv/Ip/SocketComponentHelper.cpp
```diff
@@ -166,7 +181,12 @@ void SocketComponentHelper::readLoop() {
         // Prevent transmission before connection, or after a disconnect
         if ((not this->isOpened()) and this->running()) {
             status = this->reconnect();
-            if (status != SOCK_SUCCESS) {
+            // When reconnect is disabled, just break as this is a exit condition for the loop
```
zimri-leisher: some comments on the surrounding lines:
`SocketComponentHelper::recv` checks if the status is SOCK_DISCONNECTED and closes the socket if so. Then, this read loop checks if the status is not SUCCESS, TRY_AGAIN or NO_DATA, and if none of them, closes it. It's minor but this means we're calling `close` twice.

Drv/Ip/SocketComponentHelper.cpp
```diff
@@ -193,9 +213,8 @@ void SocketComponentHelper::readLoop() {
             this->sendBuffer(buffer, status);
         }
     }
-    // As long as not told to stop, and we are successful interrupted or ordered to retry, keep receiving
-    while (this->running() &&
-           (status == SOCK_SUCCESS || (status == SOCK_NO_DATA_AVAILABLE) || status == SOCK_INTERRUPTED_TRY_AGAIN || this->m_reconnect));
```
zimri-leisher, line 198: can you help me understand why you removed the status checks? I understand the check for m_reconnect is obviated by the `break` statement above if status is SOCK_AUTO_CONNECT_DISABLED. I think the primary difference is that after removing this, if `recv` returns a status that isn't one of these three and isn't DISCONNECTED, and m_reconnect is false, after these changes you will loop back around whereas before you wouldn't. I guess it's just code simplicity?

Drv/Ip/UdpSocket.cpp
zimri-leisher: I think asserting on an invalid configuration would make me trust the framework more. It's clearly a coding error to try to send on a socket that hasn't been opened, and I would prefer the framework to be very loud about it rather than let it get buried in a console printf.

zimri-leisher:
Awesome, thanks for figuring out the issue!

Drv/Ip/UdpSocket.cpp
LeStarch: Agreed.  I will leave the assert in!

Drv/Ip/SocketComponentHelper.cpp
LeStarch: I think the best way to handle that condition is to turn off autoconnect, and reconnect when not on a critical loop.  That way your sends are specifically limited to the time.

A timeout could be applied to other methods `connect` `accept` etc, but there is no guarantee that the overall code duration would remain constant...just that the i/o of the socket would not take more time.  OS operations (Allocating ports, etc) would still take time and would not respect the timeout.

"Timeouts only have effect for system
              calls that perform socket I/O (e.g., [accept(2)](https://man7.org/linux/man-pages/man2/accept.2.html),
              [connect(2)](https://man7.org/linux/man-pages/man2/connect.2.html), [read(2)](https://man7.org/linux/man-pages/man2/read.2.html), [recvmsg(2)](https://man7.org/linux/man-pages/man2/recvmsg.2.html), [send(2)](https://man7.org/linux/man-pages/man2/send.2.html), [sendmsg(2)](https://man7.org/linux/man-pages/man2/sendmsg.2.html));
              timeouts have no effect for [select(2)](https://man7.org/linux/man-pages/man2/select.2.html), [poll(2)](https://man7.org/linux/man-pages/man2/poll.2.html),
              [epoll_wait(2)](https://man7.org/linux/man-pages/man2/epoll_wait.2.html), and so on."

Drv/Ip/SocketComponentHelper.cpp
LeStarch: The thought was code simplicity.

In all explicit status cases above, the socket should still be open and we just need to loop and try reading again.  In the implict cases where the socket is closed, we can let `reconnect` be called and sort out there if it should reconnect or exit.

I am happy to reconsider this change as it did take me some time to think through.  So, if you prefer the other more explicit model, I am happy to revert.

Drv/Ip/SocketComponentHelper.cpp
LeStarch: @zimri-leisher in considering what you wrote, I did make `reconnect` private.  This function is only sensible when it is autoconnecting from our code.  Users are expected to use the `open` call to connect/reconnect as this does not contain reconnect checks.

SDDs were updated to reflect user expectations.

Drv/Ip/SocketComponentHelper.cpp
LeStarch: I made this function private and noted what it should do.  It is effectively a `reopen` call.

Drv/Ip/SocketComponentHelper.cpp
LeStarch: This is true.  That redundant close is for other ways to exit the loop (e.g. the stop method was called).  Closeing twice, although a touch inefficient, is safe.

The first close is needed to ensure it is ready for a reopen.

LeStarch:
@zimri-leisher I think this is now done.  Here is a summary of the decisions and changes I made:

1. reconnect -> reopen as specifically the code is reopening a socket.  This change makes all of the language of the SDDs more understadable.
2. `start` no longer accepts a `reconnect`/`reopen` flag.  Since this property affects the component in both `send` and `recv` I decided it did not belong in the `start` call.  The start call now looks identical to other threads.
3. `configure` on UDP now asserts.  The user **must** call the `configureSend`/`configureRecv` variants explicitly.
4. I am keeping the simplified loop break behavior.

Let me know if you like these changes, 2 and 3 will require some users to update code.

LeStarch:
@zimri-leisher this work should be ready. Would you mind reading the summary and let me know any more comments?

@thomas-bc would you do the official review?

Drv/Ip/SocketComponentHelper.cpp
zimri-leisher: No, I think this makes sense. It was hard to understand why these checks were here in the first place.

zimri-leisher:
Sorry for the late response. 1 and 2 are great! 3 is interesting, it makes sense to me, but generally when things like this happen (a class has a method that should not ever be called), it's pointing towards a larger issue with the class hierarchy. I think this is a fine fix for now, but maybe in the future the socket class hierarchy could be simplified into a more linear, C-style layout (depth-one or -zero inheritance, and helper libraries of static functions). I think most of the problems we're facing, and code that's confusing, is just us fighting abstraction that could be removed altogether. I will keep this in mind for future contributions to FPrime. Finally, I agree with 4.

Drv/TcpServer/docs/sdd.md
```diff
@@ -5,7 +5,7 @@ connects and sends/receives bytes. It implements the callback formation (shown b
 and producing the callback port call. Since it is a server, it must startup and listen for client connections. Designed
 for single client communication, it does not permit a queue of connecting clients.
 
-For more information on the supporting TCP implementation see: Drv::TcpServerSocket.
+For more information on the supporting TCP implementation see: [Drv::TcpServerSocket]().
```
thomas-bc: Did you mean to add a link here?

Drv/Ip/SocketComponentHelper.cpp
```diff
@@ -77,13 +75,28 @@ bool SocketComponentHelper::isOpened() {
...
+            Os::ScopeLock scopedLock(this->m_lock);
+            reopen = this->m_reopen;
+        }
+        // Open a network connection if it has not already been open
+        if (not reopen) {
+            status = SOCK_AUTO_CONNECT_DISABLED;
+        } else {
+            status = this->open();
+            if (status == SocketIpStatus::SOCK_ANOTHER_THREAD_OPENING) {
+                status = SocketIpStatus::SOCK_SUCCESS;
```
thomas-bc, line 98: Why is this considered a success? What are the consequences if for example the other thread fails its `open()` call?

LeStarch: The internal reopen thread considers this a success because some other thread is handling the repoen. Thus, this thread need not be responsible for it.

thomas-bc (approved):

PR #3157, opened 2025-01-23
Stop ignoring real buffers shrunk to 0 size

celskeggs:
| | |
|:---|:---|

---
## Change Description

If BufferManager is asked to allocate a buffer when it has no more buffers, it will return a null, zero-size buffer. Therefore, if BufferManager receives a request to deallocate a zero-sized buffer, it discards the buffer, on the assumption that it resulted from a failed allocation.

However, if a user allocates a buffer from BufferManager, they are allowed to decrease the size before they deallocate it. Therefore, a zero size buffer passed to the deallocation function cannot _necessarily_ be discarded; it could be a real buffer that needs to be freed. Only discard deallocated buffers if they are empty _and null_.

## Rationale

Without this fix, trying to return allocated buffers to a BufferManager with sizes updated to 0 can result in a memory leak.

No specific recommendations.

## Future Work

No future work.

### DIFF
--- a/Svc/BufferManager/BufferManagerComponentImpl.cpp
+++ b/Svc/BufferManager/BufferManagerComponentImpl.cpp
@@ -82,9 +82,11 @@ namespace Svc {
       // make sure component has been set up
       FW_ASSERT(this->m_setup);
       FW_ASSERT(m_buffers);
-      // check for empty buffers - this is just a warning since this component returns
-      // empty buffers if it can't allocate one.
-      if (fwBuffer.getSize() == 0) {
+      // check for null, empty buffers - this is a warning because this component returns
+      // null, empty buffers if it can't allocate one.
+      // however, empty non-null buffers could potentially be previously allocated
+      // buffers with their size reduced. the user is allowed to make buffers smaller.
+      if (fwBuffer.getData() == nullptr && fwBuffer.getSize() == 0) {
           this->log_WARNING_HI_ZeroSizeBuffer();
           this->m_emptyBuffs++;
           return;
--- a/Svc/BufferManager/BufferManagerComponentImpl.cpp
+++ b/Svc/BufferManager/BufferManagerComponentImpl.cpp
@@ -87,7 +87,7 @@ namespace Svc {
       // however, empty non-null buffers could potentially be previously allocated
       // buffers with their size reduced. the user is allowed to make buffers smaller.
       if (fwBuffer.getData() == nullptr && fwBuffer.getSize() == 0) {
-          this->log_WARNING_HI_ZeroSizeBuffer();
+          this->log_WARNING_HI_NullEmptyBuffer();
           this->m_emptyBuffs++;
           return;
       }
--- a/Svc/BufferManager/Events.fppi
+++ b/Svc/BufferManager/Events.fppi
@@ -7,9 +7,9 @@ event NoBuffsAvailable(
   format "No available buffers of size {}" \
   throttle 10
 
-@ The buffer manager received a zero-sized buffer as a return. Probably undetected empty buffer allocation
-event ZeroSizeBuffer \
+@ The buffer manager received a null pointer and zero-sized buffer as a return. Probably undetected failed buffer allocation
+event NullEmptyBuffer \
   severity warning high \
   id 0x01 \
-  format "Received zero size buffer" \
+  format "Received null pointer and zero size buffer" \
   throttle 10


LeStarch:
My first thought was to recommend using `not buffer.isValid();` to check for invalid buffers to skip.  However, to be valid according to the function, a buffer must be not null, and not zero-sized.

https://github.com/nasa/fprime/blob/e09462f283494deffc7a792c14cd92c13d928c2e/Fw/Buffer/Buffer.cpp#L64-L66

I agree with you fix, because it is absolutely legal to shrink a buffer (even to zero) and still have it be deallocated.  My question is, should the `isValid` buffer method be updated to reflect this? 

@bocchino you added this `Fw::Buffer::isValid` method  originally (if my memory holds).  Was there a reason to consider a buffer valid only if it has positive size?  Why not just check to see if the memory pointer is null

LeStarch:
This would be a good issue to put into a Unit-Test to prevent regressions.

LeStarch (approved):
Fix looks solid.

bocchino:
> Was there a reason to consider a buffer valid only if it has positive size?

I think a buffer with a zero size and a non-null pointer is invalid. It can't be used to store any data, any more than a buffer with a positive size and a null pointer.

It seems like we are now using the pointer value to track side-band information of the form "this buffer was valid when the buffer manager gave it out, but it was made invalid by shrinking its size to zero" as opposed to "this buffer was invalid when the buffer manager gave it out." The size-shrinking and consequent valid-to-invalid transformation were not part of the original design.

I've never been entirely comfortable with the size-shrinking pattern, but I think it is now ingrained in F Prime, because there's no field in the buffer class for storing how much of the original allocation is used. In hindsight it might have been better to have two fields: one for the original allocation, and one for how much is used. But given how buffers currently work, the solution proposed in this PR seems good.

Svc/BufferManager/BufferManagerComponentImpl.cpp
```diff
@@ -82,9 +82,11 @@ namespace Svc {
...
       FW_ASSERT(m_buffers);
-      // check for empty buffers - this is just a warning since this component returns
-      // empty buffers if it can't allocate one.
-      if (fwBuffer.getSize() == 0) {
+      // check for null, empty buffers - this is a warning because this component returns
+      // null, empty buffers if it can't allocate one.
+      // however, empty non-null buffers could potentially be previously allocated
+      // buffers with their size reduced. the user is allowed to make buffers smaller.
+      if (fwBuffer.getData() == nullptr && fwBuffer.getSize() == 0) {
           this->log_WARNING_HI_ZeroSizeBuffer();
```
bocchino: Should we rename this event to NullEmptyBuffer instead of ZeroSizeBuffer? Or maybe just NullBuffer?

Svc/BufferManager/BufferManagerComponentImpl.cpp
```diff
@@ -82,9 +82,11 @@ namespace Svc {
...
       FW_ASSERT(this->m_setup);
       FW_ASSERT(m_buffers);
-      // check for empty buffers - this is just a warning since this component returns
-      // empty buffers if it can't allocate one.
-      if (fwBuffer.getSize() == 0) {
+      // check for null, empty buffers - this is a warning because this component returns
+      // null, empty buffers if it can't allocate one.
+      // however, empty non-null buffers could potentially be previously allocated
+      // buffers with their size reduced. the user is allowed to make buffers smaller.
+      if (fwBuffer.getData() == nullptr && fwBuffer.getSize() == 0) {
```
bocchino, line 89: Do we need the size == 0 check here? It seems that if the pointer is null, there is nothing to deallocate, regardless of the size.

bocchino:
It looks like the current BufferManager implementation allows a zero-size buffer allocation. Allocating and returning such a buffer would also cause a memory leak in the old code (though it's an unlikely use case). Checking for a null pointer instead of a zero size fixes this issue as well.

bocchino:
It's not clear to me that the BufferManager should allow allocations of zero-size buffers, which are invalid, but it does.

Svc/BufferManager/BufferManagerComponentImpl.cpp
celskeggs: I include both checks for the sake of defensiveness. If the pointer is null but the size is non-zero, that's likely a serious error and should result in an assertion trip. By excluding the null/non-zero case from this conditional, I make sure it will fall through into the subsequent assertions and cause a FATAL. So I don't think I should change this.

celskeggs:
Allowing zero-sized allocations is a strange edge case, but it can make some tests simpler for us to allow it. (I caught this bug in a test where I randomly generated buffers of length 0 to N, and occasionally it landed on 0.) I don't see any reason to forbid them outright in a core framework component. I agree that this PR also fixes that issue.

Svc/BufferManager/BufferManagerComponentImpl.cpp
bocchino: Thanks, that makes sense!

LeStarch:
@celskeggs will you rename this event? I think this is the only requested change.

celskeggs:
@LeStarch Change complete.

LeStarch (approved):

Issue #2959, opened 2024-10-16
Upgrade fprime-vxworks to latest F Prime


kevin-f-ortega:
| | |
|:---|:---|
|**_F´ Version_**| 3.5.0 |
|**_Affected Component_**| [`fprime-vxworks`](https://github.com/fprime-community/fprime-vxworks/tree/master) |
---
## Feature Description

Upgrade fprime-vxworks to use the latest OSAL refactor features of F Prime. 

## Rationale

Support other projects that want to use VxWorks.

LeStarch:
@kevin-f-ortega is this done?

kevin-f-ortega:
Almost. Currently trying to get CI to pass. After that, then we can close this.

kevin-f-ortega:
This is done with fprime-vxworks release v0.2.0

PR #3158, opened 2025-01-24
Update mutex and condition-variable interface

kevin-f-ortega:
| | |
|:---|:---|
|**_Related Issue(s)_**|  https://github.com/nasa/fprime/issues/2959|
|**_Documentation Included (y/n)_**|  y|

---
## Change Description

Updated ConditionVariable wait API to pend with a return status.

## Rationale

Originally, the F Prime ConditionVariable wait function was locking the mutex. This led to double locking the mutex because the underlying OS' ConditionalVariable wait library call was also locking the mutex. This is an issue because we only do a single unlock.

Run UTs, run CI and nominal FIT tests.

### DIFF
--- a/Os/Stub/ConditionVariable.cpp
+++ b/Os/Stub/ConditionVariable.cpp
@@ -9,10 +9,11 @@ namespace Os {
 namespace Stub {
 namespace Mutex {
 
-void StubConditionVariable::wait(Os::Mutex& mutex) {
+StubConditionVariable::Status StubConditionVariable::pend(Os::Mutex& mutex) {
     // This stub implementation can only be used in deployments that never need to wait on any ConditionVariable.
     // Trigger an assertion if anyone ever tries to wait.
     FW_ASSERT(0);
+    return StubConditionVariable::Status::ERROR_OTHER;
 }
 void StubConditionVariable::notify() {
     // Nobody is waiting, because we assert if anyone tries to wait.
@@ -27,6 +28,6 @@ ConditionVariableHandle* StubConditionVariable::getHandle() {
     return &m_handle;
 }
 
-}
-}
-}
+}  // namespace Mutex
+}  // namespace Stub
+}  // namespace Os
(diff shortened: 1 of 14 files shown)


Os/Stub/test/ConditionVariable.cpp
```diff
@@ -20,28 +20,28 @@ TestConditionVariable::~TestConditionVariable() {
     StaticData::data.lastCalled = StaticData::LastFn::DESTRUCT_FN;
 }
 
-
-void TestConditionVariable::wait(Os::Mutex& mutex) {
+TestConditionVariable::Status TestConditionVariable::pend(Os::Mutex& mutex) {
     StaticData::data.lastCalled = StaticData::LastFn::WAIT_FN;
     StaticData::data.passed = &mutex;
+    // Need to take mutex because we expect conditional var wait API to grab mutex
+    mutex.take();
```
LeStarch: I don't follow this line.

Os/Stub/ConditionVariable.cpp
```diff
@@ -9,10 +9,11 @@ namespace Os {
 namespace Stub {
 namespace Mutex {
 
-void StubConditionVariable::wait(Os::Mutex& mutex) {
+StubConditionVariable::Status StubConditionVariable::pend(Os::Mutex& mutex) {
     // This stub implementation can only be used in deployments that never need to wait on any ConditionVariable.
     // Trigger an assertion if anyone ever tries to wait.
     FW_ASSERT(0);
```
LeStarch: Do we need this ASSERT anymore?

LeStarch (changes requested):

Os/Stub/ConditionVariable.cpp
kevin-f-ortega: I don't think so. We can just return an error and rely on the calling code to decide for itself what to do. Removing it...

LeStarch (approved):

LeStarch:
## Update Instructions Synopsis

We need to write-up instructions for OS implementers that the ConditionVariable API has changed to return a status. 

Users should know that they **must** hold the lock to wait.

PR #3167, opened 2025-01-28
Improve error checking in mutex stub

celskeggs:
| | |
|:---|:---|

---
## Change Description

Replace the stub implementation of Os::Mutex with a more complete nonblocking implementation that still avoids platform dependencies.

## Rationale

The default mutex stub is intended to be usable on any platform, even platforms that do not have threads and cannot block. This stub is inappropriate for applications that need to contend over mutexes. However, if inadvertently used in applications in a way that would result in mutex contention, they would silently allow incorrect and dangerous behavior. (Notably, they allowed multiple threads to enter the same critical section.)

The new implementation still works on all platforms, and never blocks. However, it ensures that only one thread enters each critical section at a time. Attempting to acquire a mutex that is already taken will result in a failure to acquire the mutex, and likely an assertion. Because this should only occur in the case of a coding defect, this is an improvement over the previous implementation.

No specific testing has been performed yet, so I am leaving this PR as a draft for the moment.

## Future Work

It is possible that some existing programs work correctly by luck but happen to rely on the ability to have multiple threads acquire the same mutex. These programs could be broken by this upgrade. Therefore, this change in behavior should be documented in the release notes.

### DIFF
--- a/Os/Mutex.cpp
+++ b/Os/Mutex.cpp
@@ -34,13 +34,13 @@ Mutex::Status Mutex::release() {
 void Mutex::lock() {
     FW_ASSERT(&this->m_delegate == reinterpret_cast<MutexInterface*>(&this->m_handle_storage[0]));
     Mutex::Status status = this->take();
-    FW_ASSERT(status == Mutex::Status::OP_OK, static_cast<U32>(reinterpret_cast<FwSizeType>(this)), status);
+    FW_ASSERT(status == Mutex::Status::OP_OK, static_cast<FwAssertArgType>(reinterpret_cast<PlatformPointerCastType>(this)), status);
 }
 
 void Mutex::unLock() {
     FW_ASSERT(&this->m_delegate == reinterpret_cast<MutexInterface*>(&this->m_handle_storage[0]));
     Mutex::Status status = this->release();
-    FW_ASSERT(status == Mutex::Status::OP_OK, static_cast<U32>(reinterpret_cast<FwSizeType>(this)), status);
+    FW_ASSERT(status == Mutex::Status::OP_OK, static_cast<FwAssertArgType>(reinterpret_cast<PlatformPointerCastType>(this)), status);
 }
 
 ScopeLock::ScopeLock(Mutex& mutex) : m_mutex(mutex) {
--- a/Os/Mutex.cpp
+++ b/Os/Mutex.cpp
@@ -34,13 +34,13 @@ Mutex::Status Mutex::release() {
 void Mutex::lock() {
     FW_ASSERT(&this->m_delegate == reinterpret_cast<MutexInterface*>(&this->m_handle_storage[0]));
     Mutex::Status status = this->take();
-    FW_ASSERT(status == Mutex::Status::OP_OK, status);
+    FW_ASSERT(status == Mutex::Status::OP_OK, static_cast<U32>(reinterpret_cast<FwSizeType>(this)), status);
 }
 
 void Mutex::unLock() {
     FW_ASSERT(&this->m_delegate == reinterpret_cast<MutexInterface*>(&this->m_handle_storage[0]));
     Mutex::Status status = this->release();
-    FW_ASSERT(status == Mutex::Status::OP_OK, status);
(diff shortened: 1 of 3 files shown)


LeStarch:
Is there a reason this is in draft?  It looks right.

celskeggs:
As mentioned above, I have not tested this PR. I won't have the opportunity to test it until next week. Do you want to merge this without testing?

LeStarch:
I can wait for you to run your tests.  I just wanted to let CI have a crack at it too.  That passed!

celskeggs:
Does CI actually test the stub mutex at all? I ask because this code turns out to be completely broken: I omitted a `!` by accident in the second conditional, which makes it impossible to successfully release the mutex! I'm surprised to see that CI didn't catch the issue.

thomas-bc:
CI was not testing the behavior of the StubMutex because there wasn't much behavior to test in the first place (in hindsight, not the best idea).
We should probably throw our common set of mutex tests at this implementation https://github.com/nasa/fprime/blob/devel/Os/test/ut/mutex/CommonTests.cpp ?

celskeggs:
@LeStarch I've performed basic testing on this PR, so it should be ready to review.

@thomas-bc I did not add tests for this mutex, because that appeared to require a bigger change. (Maybe a new platform is needed, since none of the default platforms include the stub mutex?) Do you want me to file a ticket for that?

thomas-bc:
We can take care of adding those tests in this PR here if you'd like

celskeggs:
Sounds good. Please go ahead.

celskeggs:
@thomas-bc I'm hoping we could get this PR merged in the near future. If you're not going to have time to add those tests under this PR this week, maybe we should file a separate ticket so that this can be merged?

thomas-bc:
So after more thought, here are a few things:

- To my knowledge, we currently do **not** have tests set up for the Stub implementations of our OSAL layer. The contents of the `Os/Stub/test/` folder is used to test the delegation mechanisms (for example by creating [Os::Stub::Mutex::Test::TestMutex](https://github.com/nasa/fprime/blob/dd463e1f6d4c64e884ec231e9e3fc7e7c9d313f0/Os/Stub/test/Mutex.hpp#L44) – another Os::Mutex implementation that allows for tracking/testing of the delegation&function calls).
- The above-mentioned testing taking place under `Os/Stub/test` is a little misleading - it's easy to think that the Os/Stub implementations are being tested. To my knowledge, they are not.
- I suggest we should probably move the interface testing under e.g. `Os/test/delegation` or `Os/test/interface/` or something. Once this is done, tests for the OS Stub implementations should be added to `Os/Stub/test` 

@LeStarch would you weigh in here? If you agree I can create a ticket for the third bullet point

LeStarch:
> So after more thought, here are a few things:
> 
>     * To my knowledge, we currently do **not** have tests set up for the Stub implementations of our OSAL layer. The contents of the `Os/Stub/test/` folder is used to test the delegation mechanisms (for example by creating [Os::Stub::Mutex::Test::TestMutex](https://github.com/nasa/fprime/blob/dd463e1f6d4c64e884ec231e9e3fc7e7c9d313f0/Os/Stub/test/Mutex.hpp#L44) – another Os::Mutex implementation that allows for tracking/testing of the delegation&function calls).
> 
>     * The above-mentioned testing taking place under `Os/Stub/test` is a little misleading - it's easy to think that the Os/Stub implementations are being tested. To my knowledge, they are not.
> 
>     * I suggest we should probably move the interface testing under e.g. `Os/test/delegation` or `Os/test/interface/` or something. Once this is done, tests for the OS Stub implementations should be added to `Os/Stub/test`
> 
> 
> @LeStarch would you weigh in here? If you agree I can create a ticket for the third bullet point

I like these suggestions.  I think we should pull the more formal testing into a new Issue so that this work does not code-rot.

LeStarch (approved):

LeStarch:
I follow the logic implemented here.  @thomas-bc will spin the testing requests into a new issue.

LeStarch:
## Update Instructions Snippet

Use of `Os::StubMutex` will now only cause an error in the case that there should have been mutex contention.  In cases where the lock is uncontested, no error will result.

Issue #3072, opened 2024-12-11
Address Issues With DpCatalog


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

We agreed to merge DpCatalog contingent on future fixes for the unaddressed items here: https://github.com/nasa/fprime/pull/2713

timcanham:
Issues are addressed in PR #3191.

PR #3191, opened 2025-02-09
Work/dp catalog review updates

timcanham:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3072 |

---
## Change Description

Minor changes to `Svc::DpCatalog` to address review items.

## Rationale

These fixes were requested prior to the next F Prime review

None, other than verifying unit tests work. Unit tests passed as of this PR,

## Future Work

There will be continuing work on `Svc::DpManager`


timcanham:
@LeStarch The word that the spell check is complaining about is in the `expect.txt` file...

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -147,6 +149,8 @@ namespace Svc {
         this->m_dpTree = nullptr;
         // reset number of records
         this->m_numDpRecords = 0;
+        // reset number of files
+
```
LeStarch, line 152: There is a comment for resetting number of files, but no code to do it.  Is this comment wrong, or was something omitted?

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -308,9 +318,13 @@ namespace Svc {
...
     void DpCatalog::appendFileState(const DpStateEntry& entry) {
         FW_ASSERT(this->m_stateFileData);
+        FW_ASSERT(entry.dir < static_cast<FwIndexType>(this->m_numDirectories),
+            entry.dir,
+            static_cast<FwAssertArgType>(this->m_numDirectories)
+        );
 
         // We will append state to the existing state file
-        // FIXME: Have to handle case where state file has partially transmitted
+        // TODO: Have to handle case where state file has partially transmitted
```
LeStarch, line 326: Remaining TODO.

Svc/DpCatalog/test/ut/DpCatalogTester.cpp
```diff
@@ -93,6 +94,7 @@ namespace Svc {
             } else {
                 ASSERT_TRUE(res != nullptr);
             }
+            //printf("CE: %u\n",entry);
```
LeStarch, line 97: More dead code.

Svc/DpCatalog/test/ut/seq/Makefile
```diff
@@ -9,4 +9,4 @@ install: all
 	cp *.bin ../../../../../Ref/seq
```
LeStarch, line 9: Delete the makefile.

LeStarch (changes requested):
One more thing I did not see fixed from the original PR:

```
In configure, the partitioning of memory into three pieces uses reinterpret cast on the address of an out-of-bounds array element, and it may not respect alignment requirements. This kind of code should do array access on the underlying byte buffer (so it's in bounds), it should use placement new to create the arrays, and it should respect alignment requirements.
```

Svc/DpCatalog/test/ut/DpCatalogTester.cpp
timcanham: This is UT code, so I like to keep them around in case I need to dump it again. It's not flight code, so does it apply?

Svc/DpCatalog/test/ut/seq/Makefile
timcanham: I will if you help me write a CMake one! Otherwise I'll keep it for my testing until I have time to learn CMake.

Svc/DpCatalog/DpCatalog.cpp
timcanham: That's on purpose, since this is still technically a prototype.

Svc/DpCatalog/DpCatalog.cpp
timcanham: That was a stray comment from something I realized I didn't need to do. I'll push a delete of the line.

LeStarch (approved):
Approving as future work will be spun off into new issues.

Issue #3147, opened 2025-01-17
Add telemetry packets to the dictionary specification


bocchino:
As part of the migration of telemetry packets from XML to FPP, we need to include the packet definitions in the JSON dictionary.

bocchino:
Done in https://github.com/nasa/fprime/pull/3221.

PR #3221, opened 2025-02-13
Add telemetry packets to JSON dictionary spec

jwest115:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3147 |

---
## Change Description

Updated the JSON dictionary specification to include telemetry packets and telemetry packet sets. Also cleaned up examples so they are more consistent/accurate.

### DIFF
--- a/docs/reference/fpp-json-dict.md
+++ b/docs/reference/fpp-json-dict.md
@@ -497,7 +497,7 @@ Example JSON of a struct:
 | `kind` | **String** indicating that the kind of value is invalid | invalid | true |
 
 
-## Commands, Telemetry Channels, Telemetry Packet Groups, Events, and Parameters
+## Commands, Telemetry Channels, Events, and Parameters
 
 ### Formal Parameters
 
@@ -1033,7 +1033,7 @@ module M {
     param Parameter1: A \
       id 0x02 \
       set opcode 0x03 \
-      sav opcode 0x04
+      save opcode 0x04
 
     @ Event with one StatusEnum argument
     event Event1(
--- a/docs/reference/fpp-json-dict.md
+++ b/docs/reference/fpp-json-dict.md
@@ -1077,7 +1077,7 @@ module M {
 
     telemetry packets Packets {
       packet P1 id 0 level 0 {
-        M.c.Channel1
+        M.c1.Channel1
       }
     }
   }
--- a/docs/reference/fpp-json-dict.md
+++ b/docs/reference/fpp-json-dict.md
@@ -938,11 +938,11 @@ module M {
 
     telemetry packets Packets {
       packet P1 id 0 level 0 {
-        M.C.Channel0
-        M.C.Channel1
+        M.C1.Channel0
+        M.C1.Channel1
       }
     } omit {
-      M.C.Channel2
+      M.C1.Channel2
     }
   }
 }
@@ -958,13 +958,13 @@ module M {
           "id" : 0,
           "level" : 0,
           "members" : [
-            "M.C.Channel0",
-            "M.C.Channel1"
+            "M.C1.Channel0",
+            "M.C1.Channel1"
           ]
         }
       ],
       "omitted" : [
-        "M.C.Channel2"
+        "M.C1.Channel2"
       ]
     }
 ]
@@ -1318,7 +1318,7 @@ module M {
           "id" : 0,
           "level" : 0,
           "members" : [
-            "M.C.Channel1"
+            "M.C1.Channel1"
           ]
         }
       ],
--- a/docs/reference/fpp-json-dict.md
+++ b/docs/reference/fpp-json-dict.md
@@ -494,7 +494,7 @@ Example JSON of a struct:
 | `kind` | **String** indicating that the kind of value is invalid | invalid | true |
 
 
-## Commands, Telemetry Channels, Events, and Parameters
+## Commands, Telemetry Channels, Telemetry Packet Groups, Events, and Parameters
 
 ### Formal Parameters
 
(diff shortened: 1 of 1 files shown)


docs/user-manual/design/fpp-json-dict.md
```diff
@@ -729,39 +789,130 @@ product record Record1: U32 id 0x102
...
+}
+```
+
+```json
+[
+  {
+      "name" : "Packets",
+      "members" : [
+        {
+          "name" : "P1",
```
LeStarch: I would expect the packet name to be qualified, shouldn't I?  i.e. `M.T.P1`.

LeStarch (commented):
Just the one question.  Otherwise the packet stuff looks good.  The other items look consistent with recent updates.

docs/user-manual/design/fpp-json-dict.md
```diff
@@ -729,39 +789,130 @@ product record Record1: U32 id 0x102
...
+        "id": 265,
         "defaultPriority": 259
     }
 ]
 ```
 
+## Telemetry Packet Sets
+
+### Telemetry Packets
+| Field | Description | Options | Required |
```
thomas-bc: Can we add a newline between the heading and the table here? This used to break the rendering on the old website - not sure about the new one, but might as well be careful.

docs/user-manual/design/fpp-json-dict.md
thomas-bc: Agreed that it would likely make things easier on the GDS to have a qualified name

docs/user-manual/design/fpp-json-dict.md
thomas-bc: And in the below headings too

docs/user-manual/design/fpp-json-dict.md
bocchino: We can make the name qualified if you want, but to me that seems unusual. The packet name is part of the packet set definition, similarly to the way a struct member name is part of a struct definition. In the past the convention we've used is that definitions use simple names and uses use qualified names. The issue with uses is that without the explicit qualification you have to do some name lookup. With definitions there is no such issue -- the name is qualified by the definition in which it appears.

bocchino: I think the name of the packet set in which the packet appears should be explicitly qualified by the (fully qualified) name of the topology in which it appears in the FPP source. Is that what we are currently doing?

bocchino: > In the past the convention we've used is that definitions use simple names and uses use qualified names.

Unless the definition is stripped out of the qualifying context. So for example (I think) we use qualified names for the names of struct types but simple names for the members. We have to use qualified names for the types because the enclosing module context is stripped away when passing from the FPP source to the dictionary.

Here I think we need the module (and topology) context for the packet set. We could also add that to every packet in the set, but to me it seems overly verbose and maybe inconsistent.

bocchino: So in the example I think

* The packet set name should be `M.T.Packets`
* The packet name could be `M.T.Packets.P1` but I would argue against this -- I would argue that `P1` is more like a struct member.

bocchino: However, I think there's also an argument that we don't need the qualifier for the packet set (so the example is correct as is) because everything in the dictionary pertains to that topology. So the topology qualifier is redundant (indeed, everything in the dictionary could be qualified that way).

bocchino: When we generate C++ code, the packet sets will become classes, and their names will be appropriately qualified, because in C++ we don't assume that the code exists in an outer topology context.

bocchino (commented):
Looks good! My only comment is that in the examples the instance names are written in `CapitalCase`. For stylistic consistency we may want to use `camelCase`. For example, instead of `instance C1: C` write `instance c1: C`. When I saw `C1` used as an instance qualifier, my first thought was that `C1` named a component rather than an instance.

docs/user-manual/design/fpp-json-dict.md
jwest115: Resolved in my latest commit!

docs/user-manual/design/fpp-json-dict.md
jwest115: In my current implementation, telemetry packet set names are not fully qualified. This is also the case for the 
 set members/telemetry packet names. @LeStarch @thomas-bc is there a specific use case you have in mind for the GDS that would benefit from making these names fully qualified? After talking to Rob, we were thinking of adding a topology name field to the dictionary metadata (for ex: `topologyName: M.T`) which would make that information available in the dictionary.

bocchino: > we were thinking of adding a topology name field to the dictionary metadata

The latest thinking is that the `deployment name` field should hold the fully-qualified topology name.

bocchino: > When we generate C++ code, the packet sets will become classes

Note: In the current C++ implementation, the packet sets are stored as instances of a generic class called `Svc::TlmPacketizerPacketList`. The names are used for generating local variables that are hidden inside the implementation. There is no name qualification, and the names are not globally visible at all. This is true for both packets and packet sets.

docs/reference/fpp-json-dict.md
```diff
@@ -920,7 +920,7 @@ module M {
 
 | Field | Description | Options | Required |
 | ----- | ----------- | ------- | -------- |
-| `deploymentName` | **String** representing the deployment name | **String** | true |
+| `deploymentName` | Fully qualified name of the topology | **String** | true |
```
bocchino: This looks good, but maybe we should keep the form "**String** representing ..." for consistency with the other fields? Or revise the other fields to match.

jwest115: I'll update it for consistency

thomas-bc:
@bocchino @jwest115 are we ready to merge this?

bocchino (approved):
Looks good!

bocchino:
Yes, I approved it.

docs/user-manual/design/fpp-json-dict.md
LeStarch: Works for me.

docs/reference/fpp-json-dict.md
```diff
@@ -794,11 +947,91 @@ product container Container2 default priority 10
...
+
+  instance c1: Component1 base id 0x100 \
+    queue size 10
+
+  topology T {
+    instance c1
+
+    telemetry packets Packets {
+      packet P1 id 0 level 0 {
+        M.c.Channel1
```
LeStarch: Should say M.c1.Channel1, right?

LeStarch (changes requested):

docs/reference/fpp-json-dict.md
jwest115: Good catch! Fixed in my latest commit.

LeStarch (approved):

PR #3223, opened 2025-02-13
Improve BufferManager assertion context args

celskeggs:
| | |
|:---|:---|

---
## Change Description

Previously, in some cases, no information or context was provided when certain assertions tripped. Include more information in several cases.

## Rationale

Lack of information in an assertion makes it harder to identify the cause.

N/A

## Future Work

N/A

### DIFF
--- a/Svc/BufferManager/BufferManagerComponentImpl.cpp
+++ b/Svc/BufferManager/BufferManagerComponentImpl.cpp
@@ -194,7 +194,7 @@ namespace Svc {
     // make sure the memory returns was non-zero and the size requested
     FW_ASSERT(memory != nullptr && memorySize == allocatedSize,
       mgrId, memId,
-      static_cast<FwAssertArgType>(reinterpret_cast<FwSizeType>(memory)),
+      static_cast<FwAssertArgType>(reinterpret_cast<PlatformPointerCastType>(memory)),
       static_cast<FwAssertArgType>(memorySize),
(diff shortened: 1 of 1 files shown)


Svc/BufferManager/BufferManagerComponentImpl.cpp
```diff
@@ -192,9 +192,9 @@ namespace Svc {
     // allocate memory
     void *memory = allocator.allocate(memId,allocatedSize,recoverable);
     // make sure the memory returns was non-zero and the size requested
-    FW_ASSERT(memory);
-    FW_ASSERT(
-      memorySize == allocatedSize,
+    FW_ASSERT(memory != nullptr && memorySize == allocatedSize,
+      mgrId, memId,
+      static_cast<FwAssertArgType>(reinterpret_cast<FwSizeType>(memory)),
```
thomas-bc: Why use `reinterpret_cast<FwSizeType>` on `memory` here? Are you looking to print the memory address? We have `PlatformPointerCastType` to represent pointers as integers

celskeggs: I somehow didn't realize that type existed! Yes, that's definitely what I want.

celskeggs:
Thanks for the fix. I didn't hit that in my local tests, so I'm glad CI caught it.

LeStarch (approved):

PR #3227, opened 2025-02-13
Adding manager-worker pattern documentation

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Adding a manager worker pattern.


thomas-bc (changes requested):
The file needs to be added to [mkdocs.yml](https://github.com/nasa/fprime/blob/34a4e2c96d4f6153a0f2cc666aa9621ea7b3d61d/docs/mkdocs.yml#L189).

Also I noticed a mistake in the mkdocs.yml [here](https://github.com/nasa/fprime/blob/34a4e2c96d4f6153a0f2cc666aa9621ea7b3d61d/docs/mkdocs.yml#L193) the link `documentation/....` should be `docs/....` instead

Can we fix as part of this PR ?

LeStarch:
Yes, I will fix both

thomas-bc (approved):

PR #3248, opened 2025-02-19
Remove include of internal glibc header

celskeggs:
| | |
|:---|:---|

---
## Change Description

`sys/cdefs.h` does not exist on musl because it is not a standard header. Remove the include, which does not appear to be necessary.

## Rationale

See the musl FAQ: [Q: When compiling something against musl, I get error messages about sys/cdefs.h](https://wiki.musl-libc.org/faq.html#Q:-When-compiling-something-against-musl,-I-get-error-messages-about-%3Ccode%3Esys/cdefs.h%3C/code%3E)

N/A

## Future Work

N/A

### DIFF
--- a/STest/STest/Random/bsd_random.h
+++ b/STest/STest/Random/bsd_random.h
@@ -9,8 +9,6 @@
 // acknowledged.
 // ======================================================================
 
-
-#include <sys/cdefs.h>
 #include <stdint.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


LeStarch (approved):

LeStarch:
@bocchino you will need to pull this change into the core STest repo for when we eventually switch over to it.

PR #3261, opened 2025-02-24
Add new Fw.Ready port for coordinating startup

celskeggs:
| | |
|:---|:---|

---
## Change Description

Add a dedicated port without arguments for use in coordinating software startup. See comments for an explanation of the port's proposed semantics.

## Rationale

We currently have an equivalent to this port in a JPL-internal repository for a microcontroller platform. However, this port is not specific to our platform; it's something that would be generally useful. We may want to include it in the standard F Prime framework.

Note that the details of the proposed semantics are up for discussion. We did not specifically settle on these semantics; they're just my suggestion for what this port should mean.

N/A

## Future Work

This could potentially replace ByteStreamReady.


LeStarch:
@timcanham what do you think of this?

timcanham (approved):
This looks fine to me!

Issue #3023, opened 2024-11-13
Advanced sequencing language C++ sequencer


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**| <new component> |
---
## Feature Description

A new sequencer for the advanced sequencing language bytecode.

## Rationale

This is necessary to implement the advanced sequencing language.

zimri-leisher:
@LeStarch please assign myself to this task when you get the chance

zimri-leisher:
(taking some development logs for myself)

![Image](https://github.com/user-attachments/assets/a3f71d3e-c87f-4cae-9c18-9af1954f43e5)
^ moving away from this cmd handling methodology because we don't know how to return cmd response
problem arises when you're awaiting a cmd response from a previous RUN, and then you send a cmd which is invalid (i.e. RUN again). this system would save the RUN opcode/cmdseq and then return an execution error to that, and the original one would be lost.

if you're already executing a cmd (i.e. if opcode and cmdseq are already set), then you can't run another blocking task i guess.

Ended up moving towards explicitly preventing signal from being raised if in wrong state:

![Image](https://github.com/user-attachments/assets/63461b74-e4f1-45af-b9b3-c0a0136f59c5)

zimri-leisher:
considering splitting into two sms: Sequencer and Runtime. They really are quite independent, Sequencer just exists to get Runtime ready to go... but how complex is runtime? is it always just STEP, AWAIT, CHECK RESULT?

zimri-leisher:
TODO still bugging me having these two separate states that are identical... 

![Image](https://github.com/user-attachments/assets/d99e75d6-7e13-44dc-8e1a-06f0b9c85701)

EDIT--I collapsed them

zimri-leisher:
very annoying: in this situation, forced to make stmt non const?

![Image](https://github.com/user-attachments/assets/87e8248a-0425-427c-bc61-c157a6bd8b21)
even though all i want to do is deserialize data

gonna const cast

![Image](https://github.com/user-attachments/assets/4d4c2fc0-ee2b-4ebf-8ff1-e117698a865b)

zimri-leisher:
incidentally also bothers me that there's no DESERIALIZE_OK

![Image](https://github.com/user-attachments/assets/ddc48a5f-9b15-4a38-bf0c-22eff947a0e9)

zimri-leisher:
Going to use the following binary structure for statements:

[...]

PR #3334, opened 2025-03-05
FpySequencer 0.1

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3023 |
|**_Has Unit Tests (y/n)_**| WIP |
|**_Documentation Included (y/n)_**| WIP |

---
## Change Description

The 0.1 version of the `FpySequencer`, which replicates the functionality of the existing `CmdSequencer` component.
Features:
* Dispatching arbitrary commands
* Absolute and relative delays between commands
* Configurable timeout
* Cancelling sequences
* Validating and running as separate commands
* Blocking/non-blocking running of sequences

It uses FPP state machines extensively.
~1300 lines of CPP
~550 lines of FPP

## Rationale

This is the first part of the roadmap for the implementation of the advanced sequencing language.


Fw/Fpy/StatementArgBuffer.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    const U8* StatementArgBuffer::getBuffAddr() const {
+        return this->m_bufferData;
+    }
+
+    U8* StatementArgBuffer::getBuffAddr() {
+        return this->m_bufferData;
+    }
+
+    bool StatementArgBuffer::operator==(const StatementArgBuffer& other) const {
```
zimri-leisher, line 45: had to make another one of these Bufs cuz none of the other ones were usable in an fpp struct

Ref/Top/RefPackets.xml
```diff
@@ -7,10 +7,8 @@
         <channel name="Ref.rateGroup1Comp.RgMaxTime"/>
         <channel name="Ref.rateGroup2Comp.RgMaxTime"/>
         <channel name="Ref.rateGroup3Comp.RgMaxTime"/>
-        <channel name="Ref.cmdSeq.CS_LoadCommands"/>
-        <channel name="Ref.cmdSeq.CS_CancelCommands"/>
-        <channel name="Ref.cmdSeq.CS_CommandsExecuted"/>
-        <channel name="Ref.cmdSeq.CS_SequencesCompleted"/>
+        <channel name="Ref.fpySeq.SequencesCompleted"/>
```
zimri-leisher: added to ref topo for now, down to discuss

Ref/Top/instances.fpp
```diff
@@ -102,6 +102,10 @@ module Ref {
     stack size Default.STACK_SIZE \
     priority 96
 
+  instance fpySeq: Svc.FpySequencer base id 0x1200 \
```
zimri-leisher: todo delete old seq

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -0,0 +1,204 @@
...
+    // send it to where it needs to go
+    bool result = false;
+
+    if (nextStatement.gettype() == Fpy::StatementType::DIRECTIVE) {
+        result = dispatchDirective(nextStatement);
+    } else {
+        result = dispatchCommand(nextStatement);
+    }
+
+    m_runtime.statementIndex++;
```
zimri-leisher: rename to indicate this is the next statement idx

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -0,0 +1,204 @@
...
+    if (!m_runtime.sleeping) {
+        // we are not sleeping
+        return;
+    }
+
+    // okay, we are sleeping
+
+    Fw::Time currentTime = getTime();
+
+    if (currentTime.getTimeBase() != m_runtime.wakeupTime.getTimeBase()) {
```
zimri-leisher: comment explaining what has happened... tb changed

Autocoders/Python/bin/tlm_packet_gen.py
```diff
@@ -162,6 +162,12 @@ def get_type_size(self, type_name, size):
             return 8
         elif type_name == "bool":
             return 1
+        elif type_name == "FwOpcodeType":
+            return 4
+        elif type_name == "Fw::Time":
+            return 11
+        elif type_name == "Fw::TimeInterval":
+            return 8
```
zimri-leisher, line 170: need a more permanent solution for this.

zimri-leisher:
@LeStarch  can you add everyone who you think should review this PR?

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+#include <Svc/FpySequencer/FpySequencer.hpp>
+
+namespace Svc {
+
+// ----------------------------------------------------------------------
+// Construction, initialization, and destruction
+// ----------------------------------------------------------------------
+
+FpySequencer ::FpySequencer(const char* const compName) : FpySequencerComponentBase(compName) {}
```
zimri-leisher: TODO construct member types here

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -0,0 +1,392 @@
...
+
+        static constexpr U32 CRC_INITIAL_VALUE = 0xFFFFFFFFU;
+
+    // allocated at startup
+    Fw::ExternalSerializeBuffer m_sequenceBuffer;
+    // id of allocator that gave us m_sequenceBuffer
+    FwEnumStoreType m_allocatorId;
+
+    // assigned by the user via cmd
+    Fw::String m_sequenceFilePath;
```
zimri-leisher, line 289: todo consider putting validation state items inside a Validation struct

timcanham:
FYI, I'm preparing for a review early next week and then traveling to a university from 3/26 - 4/1. I will get to the review after that.

zimri-leisher:
* remove priorities from fpp because not supported on some queue backends
* but they're a good way to solve my current problem...
* TODO use cmdOut context with a custom cmd sequence number to track what cmd i sent out
* TODO return sm signal ids instead of raising them in the code. this way you can unit test
* when aliases get released, use them for FwFpyVarIdx etc

zimri-leisher:
question:
should the timeout timer also stop a purposeful sleep?
i.e. if we say "SLEEP 100 seconds" and set a timeout of 50 seconds, should the sleep fail?

Autocoders/Python/bin/tlm_packet_gen.py
LeStarch: Do not worry about this. It will be overcome by events soon.

Fw/Fpy/StatementArgBuffer.cpp
```diff
@@ -0,0 +1,72 @@
+#include <Fw/Fpy/StatementArgBuffer.hpp>
+#include <Fw/Types/Assert.hpp>
+#include <Fw/Types/StringBase.hpp>
+
+namespace Fw {
+
+    StatementArgBuffer::StatementArgBuffer(const U8 *args, NATIVE_UINT_TYPE size) {
```
LeStarch: Use `FwSizeType` instead of `NATIVE_UINT_TYPE`.   The changes are in to allow this in `devel`

Fw/Fpy/StatementArgBuffer.cpp
```diff
@@ -0,0 +1,72 @@
...
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+#if FW_SERIALIZABLE_TO_STRING
+    void StatementArgBuffer::toString(Fw::StringBase& text) const {
+        static const char * formatString = "(data = %p, size = %u)";
```
LeStarch: Should use `%" PRI_FwSizeType "` once you update to FwSizeTyp

Fw/Fpy/StatementArgBuffer.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+    class StatementArgBuffer : public SerializeBufferBase {
+        public:
+
+            enum {
+                SERIALIZED_TYPE_ID = FW_TYPEID_TLM_BUFF,
+                SERIALIZED_SIZE = FW_STATEMENT_ARG_BUFFER_MAX_SIZE + sizeof(FwBuffSizeType)
+            };
+
+            StatementArgBuffer(const U8 *args, NATIVE_UINT_TYPE size);
```
LeStarch: FwSizeType

Ref/Top/RefPackets.xml
```diff
@@ -7,10 +7,11 @@
         <channel name="Ref.rateGroup1Comp.RgMaxTime"/>
         <channel name="Ref.rateGroup2Comp.RgMaxTime"/>
         <channel name="Ref.rateGroup3Comp.RgMaxTime"/>
-        <channel name="Ref.cmdSeq.CS_LoadCommands"/>
-        <channel name="Ref.cmdSeq.CS_CancelCommands"/>
-        <channel name="Ref.cmdSeq.CS_CommandsExecuted"/>
-        <channel name="Ref.cmdSeq.CS_SequencesCompleted"/>
+        <channel name="Ref.fpySeq.SequencesSucceeded"/>
```
LeStarch: Back out all changes in `Ref` for now.

Svc/CMakeLists.txt
```diff
@@ -60,3 +60,4 @@ endif()
 add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/PosixTime/")
 add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/LinuxTimer/")
 add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/Version/")
+add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/FpySequencer/")
```
LeStarch, line 68: Sooooooo AWESOME!

Svc/FpySequencer/FpySequencerCommands.fppi
```diff
@@ -0,0 +1,27 @@
+@ Loads, validates and runs a sequence
+# prio: before sm sig and CANCEL
+async command RUN(
+                      fileName: string size 240 @< The name of the sequence file
```
LeStarch: Should use a constant here for the string length of a file name.

Try this one: https://github.com/nasa/fprime/blob/5287d2f5844e414c8457279b603472adf1184e93/config/AcConstants.fpp#L52

Svc/FpySequencer/FpySequencerCommands.fppi
```diff
@@ -0,0 +1,27 @@
...
+                      $block: BlockState @< Return command status when complete or not
+                    ) \
+    opcode 0 priority 7 assert
+
+@ Loads and validates a sequence
+# prio: before sm sig and CANCEL
+async command VALIDATE(
+                      fileName: string size 240 @< The name of the sequence file
```
LeStarch: Use above constant.

Svc/FpySequencer/FpySequencerTelemetry.fppi
```diff
@@ -0,0 +1,16 @@
...
+@ the number of sequences that failed to validate or execute, or were cancelled
+telemetry SequencesFailed: U64 update always
+
+@ the number of statements successfully dispatched total
+telemetry StatementsDispatched: U64 update always
+
+@ the number of statements that failed to execute
+telemetry StatementsFailed: U64 update always
+
+telemetry SeqPath: string size 240 update always
```
LeStarch: Use: https://github.com/nasa/fprime/blob/5287d2f5844e414c8457279b603472adf1184e93/config/AcConstants.fpp#L52

Svc/FpySequencer/FpySequencerEvents.fppi
```diff
@@ -0,0 +1,148 @@
+event InvalidCommand($state: I32) \
```
LeStarch, line 1: I think we should talk through the severities here.

Svc/FpySequencer/FpySequencerParams.fppi
```diff
@@ -0,0 +1,3 @@
+@ the number of seconds to wait before giving up
+@ on a directive or command. if <= 0, never time out.
+param STATEMENT_TIMEOUT_SECS: F32 default 0
```
LeStarch, line 5: Perhaps this should be a time pair? (sec, usec)?

Svc/FpySequencer/FpySequencerTypes.fpp
```diff
@@ -0,0 +1,58 @@
...
+            patchVersion: U8
+            @ the schema version of this file
+            schemaVersion: U8
+
+            @ the number of input arguments to this sequence
+            @ these will become locals in the sequence
+            argumentCount: U8
+
+            @ the number of statements in the sequence
+            statementCount: U16
```
LeStarch, line 32: Future work: define statement count as an alias type.  

```
FpySequenceStatementCountType = U16
```

Svc/FpySequencer/FpySequencerTypes.fpp
```diff
@@ -0,0 +1,58 @@
...
+            $type: StatementType
+            opCode: FwOpcodeType
+            argBuf: Fw.StatementArgBuffer
+        }
+
+        struct Sequence {
+            header: Header
+            @ an array of size m_header.argumentCount mapping argument position to local
+            @ variable count
+            args: [MAX_SEQUENCE_ARG_COUNT] U8
```
LeStarch, line 53: Are all the args limited to U8s?

Svc/FpySequencer/FpySequencerDirectives.fppi
```diff
@@ -0,0 +1,11 @@
+struct WaitRelDirective {
```
LeStarch, line 1: Are these types meant for types.fppi?

Svc/FpySequencer/docs/sdd.md
```diff
@@ -0,0 +1,26 @@
+# Svc::FpySequencer
```
LeStarch, line 1: We should fill out the SDD, or remove it.

LeStarch: Could be future work.

Svc/FpySequencer/test/ut/FpySequencerTester.cpp
```diff
@@ -0,0 +1,45 @@
...
+    this->initComponents();
+}
+
+FpySequencerTester ::~FpySequencerTester() {}
+
+// ----------------------------------------------------------------------
+// Tests
+// ----------------------------------------------------------------------
+
+void FpySequencerTester::test_waitRel() {
```
LeStarch, line 30: Future work: Add more and better UTs.

config/FpySequencerCfg.fpp
```diff
@@ -0,0 +1,10 @@
+module Svc {
+    module Fpy {
+        @ The maximum number of arguments a sequence can have
+        constant MAX_SEQUENCE_ARG_COUNT = 16
```
LeStarch, line 4: Are these static asserted as fitting in the sizes defined earlier?

config/FpConfig.h
```diff
@@ -324,6 +324,11 @@ typedef FwIndexType FwQueueSizeType;
 #define FW_TLM_BUFFER_MAX_SIZE (FW_COM_BUFFER_MAX_SIZE - sizeof(FwChanIdType) - sizeof(FwPacketDescriptorType))
 #endif
 
+// Specifies the size of the buffer that contains statement args for the FpySequencer
+#ifndef FW_STATEMENT_ARG_BUFFER_MAX_SIZE
+#define FW_STATEMENT_ARG_BUFFER_MAX_SIZE (FW_CMD_ARG_BUFFER_MAX_SIZE)
```
LeStarch, line 234: Using FW_CMD_ARG_BUFFER_MAX_SIZE seems wrong?

config/FpConfig.h
```diff
@@ -196,7 +196,7 @@ typedef FwIndexType FwQueueSizeType;
 
 // This allows tracing calls through ports for debugging
 #ifndef FW_PORT_TRACING
-#define FW_PORT_TRACING 1  //!< Indicates whether port calls are traced (more code, more visibility into execution)
+#define FW_PORT_TRACING 0  //!< Indicates whether port calls are traced (more code, more visibility into execution)
```
LeStarch: Why was this turrned off?

Svc/FpySequencer/FpySequencerStateMachine.fppi
```diff
@@ -0,0 +1,228 @@
+struct SequenceExecutionArgs {
+    filePath: string
```
LeStarch: In events and here, I see string w/o sizes.  What is the default?  Should we specify the size w/ the aforementioned constant?

Svc/FpySequencer/FpySequencerStateMachine.fppi
```diff
@@ -0,0 +1,228 @@
...
+    @ called on CANCEL cmd. raised in all states except IDLE
+    signal cmd_CANCEL
+
+    @ generic failure of an action
+    signal result_failure
+    @ generic success of an action
+    signal result_success
+
+    @ generic entry of a state
+    signal entered
```
LeStarch, line 32: May wish to explain why we need an entered signal.  There was a reason.

Svc/FpySequencer/FpySequencerStateMachine.cpp
```diff
@@ -0,0 +1,313 @@
...
+//! Implementation for action setSequenceFilePath of state machine
+//! Svc_FpySequencer_SequencerStateMachine
+//!
+//! sets the current sequence file path member var
+void FpySequencer::Svc_FpySequencer_SequencerStateMachine_action_setSequenceFilePath(
+    SmId smId,                                              //!< The state machine id
+    Svc_FpySequencer_SequencerStateMachine::Signal signal,  //!< The signal
+    const Svc::FpySequencer_SequenceExecutionArgs& value    //!< The value
+) {
+    m_sequenceFilePath = value.getfilePath();
```
LeStarch: `this->` is preferred.

Svc/FpySequencer/FpySequencerStateMachine.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+//! Implementation for action report_seqSucceeded of state machine
+//! Svc_FpySequencer_SequencerStateMachine
+//!
+//! reports that a sequence was completed
+void FpySequencer::Svc_FpySequencer_SequencerStateMachine_action_report_seqSucceeded(
+    SmId smId,                                             //!< The state machine id
+    Svc_FpySequencer_SequencerStateMachine::Signal signal  //!< The signal
+) {
+    m_tlm.sequencesSucceeded++;
```
LeStarch: Do you need to send out telemetry here?

Svc/FpySequencer/FpySequencerStateMachine.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+//! Implementation for action report_seqCancelled of state machine
+//! Svc_FpySequencer_SequencerStateMachine
+//!
+//! reports that a sequence was cancelled
+void FpySequencer::Svc_FpySequencer_SequencerStateMachine_action_report_seqCancelled(
+    SmId smId,                                             //!< The state machine id
+    Svc_FpySequencer_SequencerStateMachine::Signal signal  //!< The signal
+) {
+    m_tlm.sequencesFailed++;
```
LeStarch: You are counting canceled as failed.

Svc/FpySequencer/FpySequencerStateMachine.cpp
```diff
@@ -0,0 +1,313 @@
...
+//! Svc_FpySequencer_SequencerStateMachine
+//!
+//! responds to the calling command with OK
+void FpySequencer::Svc_FpySequencer_SequencerStateMachine_action_sendCmdResponse_OK(
+    SmId smId,                                             //!< The state machine id
+    Svc_FpySequencer_SequencerStateMachine::Signal signal  //!< The signal
+) {
+    if (this->m_sequenceBlockState == FpySequencer_BlockState::BLOCK) {
+        // respond if we were waiting on a response
+        this->cmdResponse_out(m_savedOpCode, m_savedCmdSeq, Fw::CmdResponse::OK);
```
LeStarch: Note to mstarch: in the NONBLOCK state, is cmdResponse ever sent?

Svc/FpySequencer/FpySequencerStateMachine.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+//! Implementation for action resetRuntime of state machine
+//! Svc_FpySequencer_SequencerStateMachine
+//!
+//! resets the sequence runtime
+void FpySequencer::Svc_FpySequencer_SequencerStateMachine_action_resetRuntime(
+    SmId smId,                                             //!< The state machine id
+    Svc_FpySequencer_SequencerStateMachine::Signal signal  //!< The signal
+) {
+    m_runtime = Runtime();
```
LeStarch: Why not placement new here, and not use a temp + the copy constructor?

Svc/FpySequencer/FpySequencerStateMachine.cpp
```diff
@@ -0,0 +1,313 @@
...
+//! Implementation for action validate of state machine
+//! Svc_FpySequencer_SequencerStateMachine
+//!
+//! performs all steps necessary for sequence validation, and raises a signal
+//! result_success or result_failure
+void FpySequencer::Svc_FpySequencer_SequencerStateMachine_action_validate(
+    SmId smId,                                             //!< The state machine id
+    Svc_FpySequencer_SequencerStateMachine::Signal signal  //!< The signal
+) {
+    bool result = this->validate();
```
LeStarch: Enumerated status?

Svc/FpySequencer/FpySequencerStateMachine.cpp
```diff
@@ -0,0 +1,313 @@
...
+    SmId smId,                                             //!< The state machine id
+    Svc_FpySequencer_SequencerStateMachine::Signal signal  //!< The signal
+) {
+    Fw::Time currentTime = this->getTime();
+
+    if (currentTime.getTimeBase() != m_runtime.wakeupTime.getTimeBase()) {
+        // cannot compare these times.
+        this->log_WARNING_LO_MismatchedTimeBase(currentTime.getTimeBase(), m_runtime.wakeupTime.getTimeBase());
+
+        this->sequencer_sendSignal_result_timeOpFailed();
```
LeStarch: What do we do in this context?  Fail the sequence?

Svc/FpySequencer/FpySequencerStateMachine.cpp
```diff
@@ -0,0 +1,313 @@
...
+        return;
+    }
+
+    if (currentTime.getContext() != m_runtime.wakeupTime.getContext()) {
+        // cannot compare these times.
+        this->log_WARNING_LO_MismatchedTimeContext(currentTime.getContext(), m_runtime.wakeupTime.getContext());
+
+        this->sequencer_sendSignal_result_timeOpFailed();
```
LeStarch: I don't know that context is as-fatal to the sequence as base is.... perhaps warn low and continue?

Svc/FpySequencer/FpySequencerStateMachine.cpp
LeStarch: WARN HI?

Svc/FpySequencer/FpySequencerStateMachine.cpp
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    Fw::Time currentTime = getTime();
+
+    if (currentTime.getTimeBase() != m_runtime.currentStatementDispatchTime.getTimeBase()) {
+        // can't compare time base. must have changed
+        this->log_WARNING_LO_MismatchedTimeBase(currentTime.getTimeBase(),
+                                                m_runtime.currentStatementDispatchTime.getTimeBase());
+        this->sequencer_sendSignal_result_timeOpFailed();
+        return;
```
LeStarch: Same belief as earlier w.r.t  failing on base/vontext.

Svc/FpySequencer/FpySequencerStateMachine.cpp
```diff
@@ -0,0 +1,313 @@
...
+    }
+    if (currentTime.getContext() != m_runtime.currentStatementDispatchTime.getContext()) {
+        // can't compare time ctx. must have changed
+        this->log_WARNING_LO_MismatchedTimeContext(currentTime.getContext(),
+                                                   m_runtime.currentStatementDispatchTime.getContext());
+        this->sequencer_sendSignal_result_timeOpFailed();
+        return;
+    }
+
+    F64 currentTimeSecs =
```
LeStarch: Seems odd to need a F64 for these calculations. Could probably subtract the seconds and sub-seconds independently, then convert to the F32.

This would be nicer on systems with soft double support.

Svc/FpySequencer/FpySequencerEvents.fppi
LeStarch, line 1: This is a confusing name, `InvalidCommand` as it parallels `InvalidSequence` but has an entirely different purpose.  Perhaps `InvalidCommandingState`?

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -0,0 +1,192 @@
...
+//! Handler for command VALIDATE
+//!
+//! Loads and validates a sequence
+void FpySequencer::VALIDATE_cmdHandler(FwOpcodeType opCode,              //!< The opcode
+                                       U32 cmdSeq,                       //!< The command sequence number
+                                       const Fw::CmdStringArg& fileName  //!< The name of the sequence file
+) {
+    // can only validate a seq while in idle
+    if (sequencer_getState() != FpySequencer_SequencerStateMachineStateMachineBase::State::IDLE) {
+        this->log_WARNING_LO_InvalidCommand(static_cast<I32>(sequencer_getState()));
```
LeStarch: Definitely a HI severity.

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -0,0 +1,192 @@
...
+void FpySequencer::cmdResponseIn_handler(FwIndexType portNum,             //!< The port number
+                                         FwOpcodeType opCode,             //!< Command Op Code
+                                         U32 cmdSeq,                      //!< Command Sequence
+                                         const Fw::CmdResponse& response  //!< The command response argument
+) {
+    if (sequencer_getState() !=
+        FpySequencer_SequencerStateMachineStateMachineBase::State::RUNNING_AWAITING_STATEMENT_RESPONSE) {
+        this->log_WARNING_LO_UnexpectedStatementResponseForState(static_cast<I32>(sequencer_getState()), opCode,
+                                                                 response);
+        // ignore it, hopefully that wasn't important :D
```
LeStarch: I feel that this error should cancel/fail the sequence.

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -0,0 +1,192 @@
...
+        return;
+    }
+    if (opCode != m_runtime.currentStatementOpcode) {
+        this->log_WARNING_LO_WrongStatementResponseOpcode(m_runtime.currentStatementOpcode, opCode, response);
```
LeStarch: I feel that this error should cancel/fail the sequence.

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -0,0 +1,192 @@
...
+    } else {
+        this->sequencer_sendSignal_cmdResponse_failure();
+    }
+}
+
+//! Handler for input port tlmWrite
+void FpySequencer::tlmWrite_handler(FwIndexType portNum,  //!< The port number
+                                    U32 context           //!< The call order
+) {
+    this->tlmWrite_StatementsDispatched(m_tlm.statementsDispatched);
```
LeStarch: Since this seems to be polled, I'd argue these should be "on change"

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -0,0 +1,409 @@
...
+    void parametersLoaded() override;
+    void parameterUpdated(FwPrmIdType id) override;
+
+  public:
+    void allocateBuffer(FwEnumStoreType identifier, Fw::MemAllocator& allocator, FwSizeType bytes);
+
+    void deallocateBuffer(Fw::MemAllocator& allocator);
+    PRIVATE :
+
+        static constexpr U32 CRC_INITIAL_VALUE = 0xFFFFFFFFU;
```
LeStarch, line 281: Fix formatting?

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -0,0 +1,409 @@
...
+        Fw::Time currentStatementDispatchTime = Fw::Time();
+
+        // the absolute time we should wait for until returning
+        // a statement response
+        Fw::Time wakeupTime = Fw::Time();
+    };
+
+    Runtime m_runtime;
+
+    struct {
```
LeStarch: Try not to use an anonymous struct.

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -0,0 +1,52 @@
+#include "Svc/FpySequencer/FpySequencer.hpp"
+
+namespace Svc {
+
+using Signal = FpySequencer_SequencerStateMachineStateMachineBase::Signal;
```
LeStarch: Redundant.  Was in header.

Svc/FpySequencer/FpySequencerStateMachine.fppi
```diff
@@ -0,0 +1,228 @@
...
+        # validate does not take as input a block state, only a path
+        on cmd_VALIDATE do { setGoalState_VALID, setSequenceFilePath } enter VALIDATING
+        # run takes both path and block state
+        on cmd_RUN do { setGoalState_RUNNING, setSequenceFilePath, setSequenceBlockState } enter VALIDATING
+    }
+
+    state VALIDATING {
+
+        # do it this way so we are only running validate while we
+        # are in the VALIDATING state... although TODO does this matter?
```
LeStarch: Remove todo.  While not "important" it makes logical since and will be assumed true by users.

Svc/FpySequencer/FpySequencerStateMachine.fppi
```diff
@@ -0,0 +1,228 @@
...
+        on entered do { validate }
+        # -->
+        on result_failure do { report_invalidSeq, sendCmdResponse_EXECUTION_ERROR } enter IDLE
+        on result_success enter VALID
+
+        ####################
+        # commands         #
+        ####################
+        @ cancelled the sequence while it was validating
+        # TODO remove support for this? validate is an atomic action, so we still wait for
```
LeStarch: Remove TODO.  A cancel during validate should cancel the upcoming run.

Svc/FpySequencer/FpySequencerStateMachine.fppi
```diff
@@ -0,0 +1,228 @@
...
+            on result_checkShouldWake_wakeup enter DISPATCH_STATEMENT
+            # this can occur if the time base/ctx changes, or was invalid from the start
+            on result_timeOpFailed do { report_seqFailed, sendCmdResponse_EXECUTION_ERROR } enter IDLE
+            # if neither is raised, just stay asleep
+        }
+
+        ####################
+        # commands         #
+        ####################
+        # TODO is it okay to cancel if we're awaiting a cmd? or sleeping?
```
LeStarch, line 222: Need to resolve and/or remove this TODO.

Svc/FpySequencer/FpySequencerStateMachine.fppi
```diff
@@ -0,0 +1,228 @@
...
+        state AWAITING_STATEMENT_RESPONSE {
+            on checkTimersIn do { checkStatementTimeout }
+            # -->
+            on statementTimeout do { report_seqTimedOut, sendCmdResponse_EXECUTION_ERROR } enter IDLE
+            on result_timeOpFailed do { report_seqFailed, sendCmdResponse_EXECUTION_ERROR } enter IDLE
+
+            on cmdResponse_success enter DISPATCH_STATEMENT
+            on cmdResponse_failure do { report_seqFailed, sendCmdResponse_EXECUTION_ERROR } enter IDLE
+
+            on directiveResponse_success enter DISPATCH_STATEMENT
```
LeStarch: A comment here on the reason for replication between command and directive would help.

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -0,0 +1,181 @@
...
+    return true;
+}
+
+// reads some bytes from the open file into the m_sequenceBuffer.
+// return true if successful
+bool FpySequencer::readBytes(Os::File& file, FwSizeType readLen, bool updateCRC) {
+    FW_ASSERT(file.isOpen());
+    // this has to be fwsignedsizetype cuz that's what file.read takes
+    // it also has to be declared a var because file.read must take a ref
+    FwSignedSizeType expectedReadLen = readLen;
```
LeStarch: Use FwSizeType after upcoming PR.  Much nicer.

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    const FwSizeType capacity = m_sequenceBuffer.getBuffCapacity();
+
+    // if this asserts, then you need to give the sequencer more buffer memory. pass in a bigger number
+    // to fpySeq.allocateBuffer(). This is usually done in topology setup CPP
+    FW_ASSERT(static_cast<FwSignedSizeType>(capacity) >= expectedReadLen, static_cast<FwAssertArgType>(capacity),
```
LeStarch: Always upcast to the larger ranged type.
```
FW_ASSERT(capacity >= static_cast<FwSizeType>(expectedReadLen), 
```
While probably ok, your cast could overflow.  This is easier to contend with.

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -0,0 +1,181 @@
...
+        this->log_WARNING_HI_FileReadError(m_sequenceFilePath, static_cast<I32>(fileStatus));
+        return false;
+    }
+
+    if (static_cast<FwSignedSizeType>(readLen) < expectedReadLen) {
+        this->log_WARNING_HI_EndOfFileError(m_sequenceFilePath);
+        return false;
+    }
+
+    if (static_cast<FwSignedSizeType>(readLen) > expectedReadLen) {
```
LeStarch: This should be an assert.

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -0,0 +1,181 @@
...
+    this->m_allocatorId = identifier;
+    U8* allocatedMemory = static_cast<U8*>(allocator.allocate(identifier, bytes, recoverable));
+    // if this fails, unable to allocate the requested amount of money
+    FW_ASSERT(bytes >= originalBytes, static_cast<FwAssertArgType>(bytes));
+    this->m_sequenceBuffer.setExtBuffer(allocatedMemory, bytes);
+}
+
+void FpySequencer::deallocateBuffer(Fw::MemAllocator& allocator) {
+    allocator.deallocate(this->m_allocatorId, this->m_sequenceBuffer.getBuffAddr());
+    this->m_sequenceBuffer.clear();
```
LeStarch, line 25: Clear before deallocate?

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    // read header bytes
+    bool readStatus = readBytes(sequenceFile, Fpy::Header::SERIALIZED_SIZE);
+    if (!readStatus) {
+        return false;
+    }
+
+    // deser header
+    Fw::SerializeStatus deserStatus = m_sequenceBuffer.deserialize(m_sequenceObj.getheader());
```
LeStarch: Does `get` actually return a reference?  @bocchino valid code?

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -0,0 +1,181 @@
...
+        return false;
+    }
+
+    // check matching schema version
+    if (m_sequenceObj.getheader().getschemaVersion() != Fpy::SCHEMA_VERSION) {
+        this->log_WARNING_LO_WrongSchemaVersion(Fpy::SCHEMA_VERSION, m_sequenceObj.getheader().getschemaVersion());
+        return false;
+    }
+
+    if (m_sequenceObj.getheader().getargumentCount() > Fpy::MAX_SEQUENCE_ARG_COUNT) {
```
LeStarch: Static assert MAX_SEQUENCE_ARG_COUNT is < max U8, otherwise there is a second (implicit) limit.

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -0,0 +1,181 @@
...
+        this->log_WARNING_HI_TooManyArgs(m_sequenceObj.getheader().getargumentCount(), Fpy::MAX_SEQUENCE_ARG_COUNT);
+        return false;
+    }
+
+    if (m_sequenceObj.getheader().getstatementCount() > Fpy::MAX_SEQUENCE_STATEMENT_COUNT) {
```
LeStarch: static assert MAX_SEQUENCE_STATEMENT_COUNT <= max U16 otherwise there is another implicit statement count.

Use aliases when available.

Svc/FpySequencer/FpySequencerValidationState.cpp
LeStarch: This should not be an assert.  This can happen in the middle of the runtime and should be a failure + event

LeStarch: Note: this is most critical when reading the body of the sequence.

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    deserStatus = m_sequenceBuffer.deserialize(m_sequenceObj.getfooter());
+    if (deserStatus != Fw::FW_SERIALIZE_OK) {
+        this->log_WARNING_HI_FileReadDeserializeError(m_sequenceFilePath, static_cast<I32>(deserStatus),
+                                                      m_sequenceBuffer.getBuffLeft(), m_sequenceBuffer.getBuffLength());
+        return false;
+    }
+
+    // need this for some reason to "finalize" the crc TODO get an explanation on this
+    this->m_computedCRC = ~this->m_computedCRC;
```
LeStarch, line 127: You could also call the library finalize method.

LeStarch: Or switch to our CRC32 hash implementation

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    if (this->m_computedCRC != m_sequenceObj.getfooter().getcrc()) {
+        this->log_WARNING_LO_WrongCRC(m_sequenceObj.getfooter().getcrc(), m_computedCRC);
+        return false;
+    }
+
```
LeStarch, line 133: Should validate now at end-of-file. Could be a sequence (valid) overlaid on something else?

LeStarch: Or do we care?

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -0,0 +1,120 @@
...
+    const Fpy::Statement& nextStatement = m_sequenceObj.getstatements()[m_runtime.nextStatementIndex];
+    m_runtime.nextStatementIndex++;
+    m_runtime.currentStatementOpcode = nextStatement.getopCode();
+
+    bool result;
+
+    // based on the statement type (directive or cmd)
+    // send it to where it needs to go
+    if (nextStatement.gettype() == Fpy::StatementType::DIRECTIVE) {
+        // the problem is that this is both parsing and completely executing
```
LeStarch: I thought you fixed this problem?

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -0,0 +1,120 @@
...
+        result = dispatchCommand(nextStatement);
+    }
+
+    if (result) {
+        m_tlm.statementsDispatched++;
+        this->sequencer_sendSignal_result_dispatchStatement_success();
+    } else {
+        this->sequencer_sendSignal_result_dispatchStatement_failure();
+    }
+    // was it dispatched successfully?
```
LeStarch: I don't follow this comment.

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -0,0 +1,120 @@
...
+    }
+    stat = cmdBuf.serialize(stmt.getargBuf().getBuffAddr(), stmt.getargBuf().getBuffLength(), true);
+    if (stat != Fw::SerializeStatus::FW_SERIALIZE_OK) {
+        this->log_WARNING_HI_SerializeError(cmdBuf.getBuffCapacity(), cmdBuf.getBuffLength(),
+                                            stmt.getargBuf().getBuffLength(), stat);
+        return false;
+    }
+
+    // little note--theoretically this could produce a cmdResponse before we send the
+    // dispatchSuccess signal. however b/c of priorities the dispatchSuccess signal will
```
LeStarch, line 72: Need to nix the priorities.

LeStarch (commented):
First, I want to commend you on an **exceptionally** easy to review component.  The structure you laid out was painfully easy to review.  Nice work!  As a good take away, I was able to give this code a really deep introspection.

General Fixes:
1. Change NATIVE_UINT_TUPE to FwSizeType.  This should be in.
2. Remove priorities.  We cannot easily depend on them, and you have them backwards maybe anyway.  We discussed before.
3. `this->` on member variables is preferred. Style

For @timcanham :
- [ ] Telemetry updates "always" and is polled via rate group. This means we will downlink redundant telemetry.  Is this ok?
- [ ] Command response back  when the sequencer is in a wrong state is currently ignored.  Should we assert as this is likely a coding/topology error, or fail the sequence instead?
- [ ] When the wrong opcode returns to the sequencer it is ignored.  Should we fail the sequence instead?
- [ ] Should we fail validate if this sequence is a sub-slice of a larger file (e.g. uplinked on-top-of a larger file)? It will run fine as there is good validation of the sub-slice.
- [ ] Time permitting, a review of the state machine (FPP) is most important.

LeStarch:
@zimri-leisher I have done my review.  Your structure made this review both easy to follow and really easy to rest confidence into.

As such, I had many nits, and a few cross-cutting concerns.  Mostly questions.  Really nice job.  This was a distinctly easy review.

timcanham:
•	[X ] Telemetry updates "always" and is polled via rate group. This means we will downlink redundant telemetry. Is this ok?
   Can we scrub them and make them update on change? 
•	[X ] Command response back when the sequencer is in a wrong state is currently ignored. Should we assert as this is likely a coding/topology error, or fail the sequence instead?
   I would send an event with the command response and cancel the sequence since something odd is going on. If no sequence is running, WARNING_HI should be enough. I wouldn't assert.
•	[X ] When the wrong opcode returns to the sequencer it is ignored. Should we fail the sequence instead?
   Yes, in general if something off-nominal happens, you should cancel the sequence
•	[X ] Should we fail validate if this sequence is a sub-slice of a larger file (e.g. uplinked on-top-of a larger file)? It will run fine as there is good validation of the sub-slice.
   That sounds like an ops error, so I would fail the validate and let the ground recover. Running it silently could cover up an ops mistake.
•	[ ] Time permitting, a review of the state machine (FPP) is most important.
   I will look at it in a week or so.

Svc/FpySequencer/FpySequencerParams.fppi
zimri-leisher: can't be because Fw.Time is not a displayable type

Svc/FpySequencer/FpySequencerTypes.fpp
zimri-leisher: so, these aren't the arg values, they just map input arg position to local var index. this field doesn't really make sense right now because i don't have args or local vars lol

zimri-leisher: should i remove this for now and just update the schema version when I add args?

Svc/FpySequencer/FpySequencerDirectives.fppi
zimri-leisher: hmm I guess they could be, it was just nice to declare them all in one place. I could theoretically delete this file

Svc/FpySequencer/FpySequencerEvents.fppi
zimri-leisher: yes--any guidelines? i kind of guessed for all of them.

config/FpConfig.h
zimri-leisher: Hmm, well often the arguments to a statement are just the args to a cmd. So it seems natural that they should default to the same size.

Svc/FpySequencer/FpySequencerStateMachine.fppi
zimri-leisher: actually I'd rather wait for @bocchino  to review and determine if this is useful

Svc/FpySequencer/FpySequencerStateMachine.cpp
zimri-leisher: would you prefer to count cancelled sequences successfully?

Svc/FpySequencer/FpySequencerStateMachine.cpp
zimri-leisher: yes--see cmdHandler in .cpp

Svc/FpySequencer/FpySequencerStateMachine.cpp
zimri-leisher: yes, in the SM it reports the seq failed and goes to IDLE

Svc/FpySequencer/FpySequencerStateMachine.cpp
zimri-leisher: i don't see why it should or shouldn't be fatal, basically. what is the purpose of context??? we can't know, so is it better to be safe and fail? or to potentially erroneously try?

Svc/FpySequencer/FpySequencerEvents.fppi
zimri-leisher: Hmm, can you explain further? InvalidSequence and InvalidCommand both reflect an inability to execute some action.

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: Okay yeah let's go over all the event severities--why choose HI vs LO?

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: Hmm but it could be caused by a badly implemented external component. I'd rather not let it kill us. Also this can happen normally if we cancel during a long command and it returns later. If we kill the seq, then this could be really annoying, as you'd have to wait for the prev cmd to finish even after cancelling it.

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: Again, could be caused by badly implemented external component. For what it's worth, the current CmdSequencer doesn't die on case 1 (your above comment), and for this case it doesn't even check, it just assumes whatever response it gets is to the cmd it sent out.

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: Okay. I put them to always because that's the default behavior and I like to be explicit about it. In general, I wish I didn't have to choose this for the user of the component.

Svc/FpySequencer/FpySequencer.hpp
zimri-leisher: will do later once I don't have to reformat it any more. my autoformatter does this automatically so I have to manually go back and change it every time I use it

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: Hmm, removing it causes the code to fail to compile. I don't know if the 'using' statement works like that

Svc/FpySequencer/FpySequencerStateMachine.fppi
zimri-leisher: I'd like @bocchino  to take a look at this to determine if it's gross abuse of the SM. Also, if I'm doing it this way here, why shouldn't I do it for all entry actions?

Rob, if you're seeing this, this could be solved by `before entry` and `after entry` actions

Svc/FpySequencer/FpySequencerStateMachine.fppi
zimri-leisher: Would like @timcanham 's or @bocchino 's opinion on this

Svc/FpySequencer/FpySequencerValidationState.cpp
zimri-leisher: image
No, because clear isn't actually modifying any of the underlying buf mem, just setting some member variables. If I clear first then getBuffAddr is null

Svc/FpySequencer/FpySequencerValidationState.cpp
zimri-leisher: image

Does appear to. It's only for primitive types that it is by value

Svc/FpySequencer/FpySequencerValidationState.cpp
zimri-leisher: For what it's worth, the previous CmdSeq asserts on this
image
I'd like to get @bocchino 's opinion

Svc/FpySequencer/FpySequencerValidationState.cpp
zimri-leisher: I think we should validate this. It should not be able to be overlaid on another sequence.

Svc/FpySequencer/FpySequencerRunState.cpp
zimri-leisher: oh yep

Svc/FpySequencer/FpySequencerRunState.cpp
zimri-leisher: Okay, I'm gonna keep em in until Tim sees them so he knows people want to use them lol

zimri-leisher:
> • [X ] Telemetry updates "always" and is polled via rate group. This means we will downlink redundant telemetry. Is this ok?
> Can we scrub them and make them update on change?
> • [X ] Command response back when the sequencer is in a wrong state is currently ignored. Should we assert as this is likely a coding/topology error, or fail the sequence instead?
> I would send an event with the command response and cancel the sequence since something odd is going on. If no sequence is running, WARNING_HI should be enough. I wouldn't assert.
> • [X ] When the wrong opcode returns to the sequencer it is ignored. Should we fail the sequence instead?
> Yes, in general if something off-nominal happens, you should cancel the sequence
> • [X ] Should we fail validate if this sequence is a sub-slice of a larger file (e.g. uplinked on-top-of a larger file)? It will run fine as there is good validation of the sub-slice.
> That sounds like an ops error, so I would fail the validate and let the ground recover. Running it silently could cover up an ops mistake.

1. Done, made them `on change`. Brings up the question if we should just default to on change instead of always? In general, I don't even feel comfortable making this decision for the user.
2. Disagree, this can happen nominally due to cancellation. E.g.: run seq, cancel during a long running command, then cmd returns  while you're still in IDLE. Sequencer should hear it but not freak out.
3. Same as above, can happen nominally with no coding errors, if you cancel, then run again and then previous long-running cmd returns.
4. Done.

zimri-leisher:
TODO should I add a "seqstarted" event?

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -103,18 +92,96 @@ bool FpySequencer::dispatchDirective(const Fpy::Statement& stmt) {
...
+    }
+
+    if (currentTime.getContext() != this->m_runtime.wakeupTime.getContext()) {
+        // cannot compare these times.
+        this->log_WARNING_LO_MismatchedTimeContext(currentTime.getContext(), this->m_runtime.wakeupTime.getContext());
+
+        return Signal::result_timeOpFailed;
+    }
+
+    printf("wakeup %d %d current %d %d\n", m_runtime.wakeupTime.getSeconds(), m_runtime.wakeupTime.getUSeconds(), currentTime.getSeconds(), currentTime.getUSeconds());
```
zimri-leisher: todo remove

zimri-leisher:
@timcanham @bocchino sorry to bother you guys, just getting antsy on this. I would really like to keep on making progress on this component, and I'm really just stuck here waiting for a review. If you'd prefer we can schedule a call and do a live overview if you think that would be a better use of your time.

@LeStarch thanks for your review, if you want to take a look I had some responses to some of your comments.

Svc/FpySequencer/FpySequencerValidationState.cpp
zimri-leisher: Added code to check this, please confirm as expected

Ref/Top/RefPackets.xml
zimri-leisher: willco, have not done this yet because i've been using the ref deployment for testing. but will do this before i merge

bocchino:
@zimri-leisher Thanks for following up. I haven't been attending the design meetings for this component, and it looks like there's no SDD available yet, so it's hard for me to jump in and do a review of the state machine. I think a short remote meeting would be productive.

zimri-leisher:
@bocchino got it, I will email you about this. If you think I should really write the SDD I'm happy to. Originally Michael was okay with forgoing it for now and including it as future work, but maybe it's better that we start off on a good foot on documentation here.

bocchino:
I would recommend adding at least a minimal SDD explaining what the component does and how to use it.

Ref/Top/RefPackets.xml
LeStarch: Excellent!

Svc/FpySequencer/FpySequencerParams.fppi
LeStarch: One could still use a U32, U32 for the same effect.  Consider it, but I will defer to your judgement.

Svc/FpySequencer/FpySequencerTypes.fpp
LeStarch: I'd fix the annotation to use "index" and then it is fine

config/FpConfig.h
LeStarch: Works for me!

Svc/FpySequencer/FpySequencerStateMachine.fppi
LeStarch: Ack!

Svc/FpySequencer/FpySequencerStateMachine.cpp
LeStarch: I'd count them as a separate item.  Canceled seems neither failed nor success.

Svc/FpySequencer/FpySequencerStateMachine.cpp
LeStarch: I did confirm this, thanks!

Svc/FpySequencer/FpySequencerStateMachine.cpp
LeStarch: What is the context used for in this Cmd Sequencer?  Likely nothing.   It might be used by the underlying components, but those should check themselves.

Svc/FpySequencer/FpySequencerStateMachine.cpp
LeStarch: Base may be justified as failing, as it crucially represents the time used by command sequencer.  Context just isn't used.

Svc/FpySequencer/FpySequencer.cpp
LeStarch: HI -> foreground warning.  Something critical has happend in the foreground of the component, or in respons to a direct port call or command.

LO -> Background warning.  Something anomolous happened while a component was running background tasks.

Svc/FpySequencer/FpySequencer.hpp
LeStarch: For the record, we are pushing to remove `PRIVATE` and `PROTECTED` such that they don't keep tripping us up.

Svc/FpySequencer/FpySequencerDirectives.cpp
LeStarch: Should work.  Was the using statement in the header under a different scope (e.g. within the component class?)

Svc/FpySequencer/FpySequencerValidationState.cpp
LeStarch: Gotcha!

Svc/FpySequencer/FpySequencerValidationState.cpp
LeStarch: Makes sense.

Svc/FpySequencer/FpySequencerValidationState.cpp
LeStarch: If the system is beyond initialization, it should not assert because the user tried to run a sequence that is too-large.  It should print a warning HI and go back to idle.

This is effectively asserting on user input, which is something we typically try to avoid.  If the old CmdSequencer had this behavior, I'd argue that would be a bug as it shouldn't assert on bad user input either.

Asserts should be used to ensure software correctness though invariant properties.  e.g. the contract of this function assumes that the caller never pass `nullptr` this we assert `!= nullptr`.  Doing so would be an error in the software.

Asserts should not be on user input, nor hardware produced data, as those don't reflect correctness, but rather anomalous conditions.

Svc/FpySequencer/FpySequencerValidationState.cpp
LeStarch: Will do!

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -382,8 +380,10 @@ class FpySequencer : public FpySequencerComponentBase {
     // Run state
     // ----------------------------------------------------------------------
 
+    using Signal = FpySequencer_SequencerStateMachineStateMachineBase::Signal;
```
LeStarch: Yeah, signal is defined within the `class` so in the C++ it would be `class::Signal`.  This is why it is not resuable

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    // okay, time to wake up
+    m_runtime.sleeping = false;
+    m_runtime.wakeupTime = Fw::Time();
+
+    // say we've finished our sleep
+    this->sequencer_sendSignal_statementResponseIn(
+        FpySequencer_StatementResponse(m_runtime.currentStatementOpcode, Fw::CmdResponse::OK));
+}
+}  // namespace Svc
```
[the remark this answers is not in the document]
LeStarch: @zimri-leisher it would be nice to fix these to appeas the warnings.  Just add a newline at the end.

Svc/FpySequencer/FpySequencer.cpp
LeStarch: Still need to construct members, no?

Svc/FpySequencer/FpySequencer.cpp
LeStarch: I'd also argue that anytime the sequence is canceled it should be a "HI".  Things that happen that we ignore and let the sequence keep running should be a LO.

Svc/FpySequencer/FpySequencer.cpp
LeStarch: One way to resolve this would be to put any "awaiting commands" at moment of cancel on an "lost command" list, and only allow those back.  Cancel the sequence if the returned command is not the active await, nor on the lost commands list.

Might not be worth the effort, but it would be one way to fix the problem.

Svc/FpySequencer/FpySequencerEvents.fppi
LeStarch: I gave some guidelines, but reach-out if we need to go into further depth.

Svc/FpySequencer/docs/sdd.md
LeStarch: @bocchino wants a minimal SDD.

Svc/FpySequencer/FpySequencerEvents.fppi
LeStarch: Invalid command isn't an invalid command, but rather a command could not be issued in the current state.  At least that's how I understood it.

Svc/FpySequencer/FpySequencer.cpp
LeStarch: Something to think through, because I don't like making that choice either.

Svc/FpySequencer/FpySequencerValidationState.cpp
LeStarch: You should use `FwSizeType` in all `Os::File` calls except seek now!

Svc/FpySequencer/FpySequencerValidationState.cpp
LeStarch: It is as I expected.  ....well rather I expected read to see if out of data, but size checking is a better approach.

LeStarch (commented):
I addressed all your comments as best I could.

zimri-leisher:
Awesome, just the feedback I was looking for. Will fix.

Svc/FpySequencer/FpySequencerStateMachine.cpp
zimri-leisher: image
To be consistent with Time::compare, we should check the context. The old CmdSequencer uses Time::compare

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: I would rather not have to track this state. In general:
1. this "error" can happen when everything is functioning perfectly nominally
2. distinguishing nominal instances of this error from off-nominal ones takes tracking an arbitrarily large amount of state (although in practice we'd probably put a cap on the lost command list size)
3. off-nominal instances would only be caused by a coding error external to this component
4. Not failing on this error is consistent with the previous sequencer behavior

Based on all this, I'm still on the side of not doing anything but raise a warning about this.

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: Same as above

Svc/FpySequencer/FpySequencerStateMachine.fppi
zimri-leisher: Rob looked and thought it was vaguely acceptable

Svc/FpySequencer/FpySequencerValidationState.cpp
zimri-leisher: That's fair, I will do this

Svc/FpySequencer/FpySequencerStateMachine.cpp
zimri-leisher: Same response as above

LeStarch:
Finalizing changes:
1. Add sequence file ID.  Add it to context. -- TODO
   1. If sequence ID matches active but the returned opcode is not expected:  WARN then stop sequence (wrong opcode from current sequence).
   2. If sequence ID does not match, WARN but continue (spurious response from a previous sequence).
   3. Capture sequence command id (distinguish between multiple same opcodes)
   4. Capture sequence file ID (dustingush against different runs)
 2. Add SDD w/ statement about being in development.
 5. Remove priorities.
 6. Remove test usage in Ref.
 7. No invalid sequence
 8. Make sequence failed specific to reason why
 9. Statement Index to events where approrpiate

Once done, I will review one more time and then merge as-is.

To make this prime-time we will additionally need to:
1. Author better UTs (close to 100% coverage)
2. Author a full SDD with documented binary format
3. Author users guide for tool support
4. Add to references and templates.
10. Constant for the buffer size
11. Failed on line 
12. Consider: fixing object ids
These steps are a separate PR.

zimri-leisher:
Finalizing changes:

* Add SDD w/ statement about being in development. **DONE**
* Remove priorities. **WONTDO** -- I know I said I would remove priorities, but due to some slight changes in internal architecture it's now hard to remove them. Tim said it was okay, so I'm gonna keep them in.
* Remove test usage in Ref. **WILLCO**
* No invalid sequence **DONE**
* Make sequence failed specific to reason why **DONE**
* Statement Index to events where approrpiate **DONE**

The remainder are about the "what to do if we get an unexpected response" problem.

Here is the behavior that I think we want:
1. If we get a response back from a different sequence, warn but do not cancel current.
   * Rationale: This can happen nominally if we cancelled the previous sequence while a cmd was executing
2. If we get a response back from the same sequence:
     1. If the opcode or cmdIdentifierNumber (a number incrementing from 0 modulo 2^16) do not match, fail the sequence
         * Rationale: it must be a coding error if the current sequence gave us a different response. Do not proceed!
     2. Otherwise, all's well, proceed as normal.

Working on implementing this now.

zimri-leisher:
The implementation of the new, more precise command response behavior is quite complex and needs another pair of eyes. @LeStarch  please review `FpySequencer::cmdResponseIn_handler` and the RUNNING state (specifically, the `on stmtResponse_unexpected` signal listener) in the `FpySequencerStateMachine.fppi`.

EDIT--this is ready to review now

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -0,0 +1,268 @@
...
+                                         const Fw::CmdResponse& response  //!< The command response argument
+) {
+    // TODO ask Rob if there's a better way to check if we're in a superstate. I don't want to have
+    // to update this every time I add a new substate to the RUNNING state.
+
+    // if we aren't in the RUNNING state:
+    if (this->sequencer_getState() != State::RUNNING_AWAITING_STATEMENT_RESPONSE
+        && this->sequencer_getState() != State::RUNNING_DISPATCH_STATEMENT
+        && this->sequencer_getState() != State::RUNNING_SLEEPING) {
+        // must be a coding error.
```
LeStarch: Must be a coding error or a canceled sequence response.

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -0,0 +1,268 @@
...
+        this->log_WARNING_LO_CmdResponseWhileNotRunningSequence(static_cast<I32>(this->sequencer_getState()), opCode,
+                                                                response);
+        // ignore it, hopefully that wasn't important :D
+        return;
+    }
+
+    // okay, we're running a sequence. now let's use the cmdUid to check if the response was for a cmd
+    // from this sequence
+
+    // the cmdSeq arg is confusingly not the cmdSeq in this case, according to the current implementation
```
LeStarch, line 172: Excellent.  Thank you!

Svc/FpySequencer/FpySequencer.cpp
LeStarch: Perhaps this should also be "CmdFromOldSequence"

Svc/FpySequencer/FpySequencerStateMachine.fppi
LeStarch: I think we determined "yes", could remove the TODO.

Svc/FpySequencer/FpySequencer.cpp
LeStarch: We have a plan, see comment lower in PR.

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+    // based on the statement type (directive or cmd)
+    // send it to where it needs to go
+    if (nextStatement.gettype() == Fpy::StatementType::DIRECTIVE) {
+        result = this->dispatchDirective(nextStatement);
+    } else {
+        result = this->dispatchCommand(nextStatement);
+    }
+
+    this->m_statementsDispatched++;
```
LeStarch, line 30: Won't incrementing this here, cause the waiting code to be off-by-one?  The statement `context` gets constructed in `dispatchCommand`  then this upates...then theoretically the wait for response happens.

Am I missing something?

LeStarch (changes requested):
Still need to confirm 2 more things:
1. `priorities` when we discussed, @timcanham heard "priorities aren't needed in this component", and @zimri-leisher heard "Priorities can be used".  Both statements were true, but created different expectations to the path forward.  I am working on a resolution
2.  I see exactly what I expect from the code block for the new command response handling, except it seems to me that the m_statementsDispatched is incremented in the wrong place as it is incremented just after dispatch, resulting in a wait for the next command.  What am I missing?
3. Remove from Ref when the above two are resolved.

Very good work.  We are in the final steps.

Svc/FpySequencer/FpySequencer.cpp
LeStarch: Not blocker, not required. Resolve if and when you decide what to do.

Svc/FpySequencer/FpySequencer.cpp
LeStarch: The comment was exactly what was needed here!

LeStarch:
> Still need to confirm 2 more things:
> 
>     1. `priorities` when we discussed, @timcanham heard "priorities aren't needed in this component", and @zimri-leisher heard "Priorities can be used".  Both statements were true, but created different expectations to the path forward.  I am working on a resolution
> 
>     2. I see exactly what I expect from the code block for the new command response handling, except it seems to me that the m_statementsDispatched is incremented in the wrong place as it is incremented just after dispatch, resulting in a wait for the next command.  What am I missing?
> 
>     3. Remove from Ref when the above two are resolved.
> 
> 
> Very good work. We are in the final steps.

I confirmed Priorities are green-light.  Thus to close out this PR

- [x] Address/Resolve my last round of feedback
- [x] Answer my ordering question
- [x] Remove from Ref
- [x] Pass CI
- [x] Merge
- [x] Celebrate 🎉🎉 🎉 🎉 🎉

Svc/FpySequencer/FpySequencer.cpp
LeStarch: I am going to resolve this.  It is a superset of the event type lacking specific parameters.

Svc/FpySequencer/FpySequencerRunState.cpp
zimri-leisher: Correct! Good catch. I would have caught this had I run my integration tests for this commit. I will run them now.

zimri-leisher:
Okay, I think I have completed all of those items.

bocchino (approved):
This PR looks good  to me!

* I reviewed and approved the state machine design
* I asked for a minimal SDD, which was provided

LeStarch:
@zimri-leisher we'll need to fix the CI errors that came in when you bumped F Prime.  Let me know if you need help!

zimri-leisher:
Okay, believe they're all fixed now.

zimri-leisher:
@LeStarch  let's get this merged!!! So excited.

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+//! Handler for command VALIDATE
+//!
+//! Loads and validates a sequence
+void FpySequencer::VALIDATE_cmdHandler(FwOpcodeType opCode,              //!< The opcode
+                                       U32 cmdSeq,                       //!< The command sequence number
+                                       const Fw::CmdStringArg& fileName  //!< The name of the sequence file
+) {
+    // can only validate a seq while in idle
+    if (sequencer_getState() != State::IDLE) {
```
LeStarch, line 71: While logically this makes sense, is there any technical reason one couldn't pre-validate a sequence while another is running?

LeStarch: Curiosity, not blocking!

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -0,0 +1,272 @@
...
+        return;
+    }
+
+    // validate always blocks until finished, so save opcode/cmdseq
+    // so we can respond once done
+    this->m_savedOpCode = opCode;
+    this->m_savedCmdSeq = cmdSeq;
+
+    this->sequencer_sendSignal_cmd_VALIDATE(
+        FpySequencer_SequenceExecutionArgs(fileName, FpySequencer_BlockState::BLOCK));
```
LeStarch, line 83: Because it runs the state machine, answered!

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -0,0 +1,272 @@
...
+    this->pingOut_out(0, key);
+}
+
+//! Handler for input port cmdResponseIn
+void FpySequencer::cmdResponseIn_handler(FwIndexType portNum,             //!< The port number
+                                         FwOpcodeType opCode,             //!< Command Op Code
+                                         U32 cmdSeq,                      //!< Command Sequence
+                                         const Fw::CmdResponse& response  //!< The command response argument
+) {
+    // TODO ask Rob if there's a better way to check if we're in a superstate. I don't want to have
```
LeStarch, line 154: TODO.  Not blocking.

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -0,0 +1,272 @@
...
+        this->log_WARNING_HI_WrongCmdResponseOpcode(opCode, response, this->m_runtime.currentStatementOpcode);
+        this->sequencer_sendSignal_stmtResponse_unexpected();
+        return;
+    }
+
+    // okay, we were awaiting this opcode. but was it from this exact statement, or a different one with the same opcode
+    // in the same file?
+
+    // pull the cmd index (modulo 2^16) out of cmdUid. this should be equal to the first 16 bits of the 
+    // m_statementsDispatched variable - 1. the -1 is because
```
LeStarch, line 212: Comment trails off.

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -0,0 +1,206 @@
...
+                                                this->m_runtime.currentStatementDispatchTime.getTimeBase());
+        return Signal::result_timeOpFailed;
+    }
+#endif
+#ifdef FW_USE_TIME_CONTEXT
+    if (currentTime.getContext() != this->m_runtime.currentStatementDispatchTime.getContext()) {
+        // can't compare time ctx. must have changed
+        this->log_WARNING_HI_MismatchedTimeContext(currentTime.getContext(),
+                                                   this->m_runtime.currentStatementDispatchTime.getContext());
+        return Signal::result_timeOpFailed;
```
LeStarch, line 172: This still fails.  Need to note this as future work.

LeStarch: Along with Fw::Time fix for the same behavior.

LeStarch (approved):

Svc/FpySequencer/FpySequencer.cpp
LeStarch: Pull into future PR.

LeStarch:
- Versioning

Issue #3335, opened 2025-03-05
Implement a Time provider using OSAL RawTime


kubiak-jpl:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Implement a Time provider component using an OSAL RawTime implementation as the backing time provider

## Rationale

The Time providers currently in F Prime are specific to an underlying implementation. Ie. Posix or c++ Chrono libraries. However, the RawTime abstraction should provide enough information to implement the Time interface directly. This would allow a platform that implements the RawTime abstraction to gain a Time provider without needing to re-implement the abstraction inside a Time component

kubiak-jpl:
I have an initial version here
https://github.com/kubiak-jpl/fprime-kubiak/blob/ostime-rawtime/Svc/OsTime/OsTime.cpp

It pretty much boils down to an initialize `set_epoch` function to get `Fw::Time` and `Os::RawTime` structures that represent the same moment in time, then calls to the Time port will create a new `Os::RawTime` instance, get the elapsed `Fw::TimeInterval` and add that to the epoch `Fw::Time`.

```cpp
void OsTime::set_epoch(const Fw::Time& fw_time, const Os::RawTime& os_time) {
    m_epoch_fw_time = fw_time;
    m_epoch_os_time = os_time;
}
```

```cpp
void OsTime ::timeGetPort_handler(FwIndexType portNum, Fw::Time& time) {
    time = Fw::ZERO_TIME;
    if (m_epoch_fw_time == Fw::ZERO_TIME) {
        return;
    }

    Os::RawTime time_now;
    Os::RawTime::Status stat = time_now.now();
    if (stat != Os::RawTime::OP_OK) {
        return;
    }

    Fw::TimeInterval elapsed;
    stat = time_now.getTimeInterval(m_epoch_os_time, elapsed);
    if (stat != Os::RawTime::OP_OK) {
        return;
    }

    time = m_epoch_fw_time;
    time.add(elapsed.getSeconds(), elapsed.getUSeconds());
}
```

PR #3355, opened 2025-03-10
Add OsTime component

kubiak-jpl:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3335 |

---
## Change Description

Creates a Time provides that uses the underlying OSAL RawTime interfaces to generate timestamps.

The RawTime layer only provides time intervals so the OsTime class needs to be passed a pair of epoch times, an `Fw::Time` and `Os::RawTime` object that correspond to the same moment in time. OsTime calculates the offset from the epoch RawTime as an `Fw::TimeInterval`, then adds this interval to the epoch `Fw::Time` object to calculate the final `Fw::Time` object.

## Rationale

Provide a generic Time provider that can use any RawTime implementation. Original rationale is for microcontroller platforms which cannot use the existing PosixTime or ChronoTime components.

Provided unit tests are fairly barebones and just show the general behavior. It is expected that the underlying RawTime implementations will be more throughly tested.

- How should overflow of the Fw::Time be handled, if at all?

## Future Work

Possible extensions:
- EVR on RawTime errors. These are not implemented because EVRs themselves require a Time call.


Svc/OsTime/docs/sdd.md
```diff
@@ -0,0 +1,33 @@
...
+
+The `Svc::OsTime` is a component that provides system time on any system that implements the RawTime OSAL class. The RawTime interface provides Fw::TimeInterval objects.
+
+## 2. Requirements
+
+| Requirement     | Description                                                                                                        | Verification |
+|-----------------|--------------------------------------------------------------------------------------------------------------------|--------------|
+| SVC-OS-TIME-001 | `Svc::OsTime` shall return current system time as an `Fw::Time` objects in response to the `timeGetPort` port call | Unit Test    |
+| SVC-OS-TIME-002 | `Svc::OsTime` shall return ZERO_TIME as an `Fw::Time` object if the RawTime layer returns an error code            | Unit Test    |
+
```
LeStarch, line 13: Would you add a section for `configuration` or `setup` or `usage`?  I get what you are trying to do, but I a write-up of what I need to call in the topology to set this up would be nice.

Svc/OsTime/OsTime.cpp
```diff
@@ -0,0 +1,58 @@
+// ======================================================================
+// \title  OsTime.cpp
+// \author kubiak
```
LeStarch, line 3: Add `kubiak` to .github/actions/spelling/expect.txt

LeStarch (changes requested):

LeStarch:
@thomas-bc would you be willing to review too?  You are the author of RawTime.

Svc/OsTime/OsTime.cpp
```diff
@@ -0,0 +1,58 @@
...
+    m_epoch_fw_time = fw_time;
+    m_epoch_os_time = os_time;
+    m_epoch_valid = true;
+}
+
+// ----------------------------------------------------------------------
+// Handler implementations for user-defined typed input ports
+// ----------------------------------------------------------------------
+
+void OsTime ::timeGetPort_handler(FwIndexType portNum, Fw::Time& time) {
```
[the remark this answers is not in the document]
kubiak-jpl: I would rather not assert in this code. Error cases are handled by returning a ZERO_TIME to the caller

Svc/OsTime/OsTime.hpp
```diff
@@ -0,0 +1,53 @@
...
+    //! Handler implementation for timeGetPort
+    //!
+    //! Port to retrieve time
+    void timeGetPort_handler(FwIndexType portNum,  //!< The port number
+                             Fw::Time& time        //!< Reference to Time object
+                             ) override;
+
+    Fw::Time m_epoch_fw_time;
+    Os::RawTime m_epoch_os_time;
+    bool m_epoch_valid;
```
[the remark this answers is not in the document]
kubiak-jpl: @LeStarch Does this error suggest that `bool` types aren't allowed? Is there something else that F Prime recommends in these situations?

Svc/OsTime/docs/sdd.md
kubiak-jpl: Sure. Is there an existing component with a documentation structure you'd prefer I'd use here?

kubiak-jpl:
@thomas-bc I noticed the MacOS unit test is failing, probably because the UTs rely on sleeping and time checks. I though 100 ms was generous enough, but I guess not. Is there a recommended way to stub out the OSAL layer for components UTs like this?

thomas-bc:
@kubiak-jpl I would have thought 100ms be more than enough too... Running your tests on an M4 macOS the tests start failing when approaching 15-10ms threshold, but the GitHub VMs are busy bees... also not sure how much more precision we can expect from `std::sleep_for` + wake up time

I'm going to re-run your tests in CI once and see if that was transient or if we need to rethink a little. What are you thinking of stubbing the OSAL layer for here?

Svc/OsTime/OsTime.cpp
```diff
@@ -0,0 +1,58 @@
...
+OsTime ::OsTime(const char* const compName) :
+    OsTimeComponentBase(compName),
+    m_epoch_fw_time(Fw::ZERO_TIME),
+    m_epoch_os_time(),
+    m_epoch_valid(false) {}
+
+OsTime ::~OsTime() {}
+
+void OsTime::set_epoch(const Fw::Time& fw_time, const Os::RawTime& os_time) {
+    m_epoch_fw_time = fw_time;
```
thomas-bc, line 31:
```suggestion
    (void) m_epoch_fw_time = fw_time;
```
To appease CI warnings, and in the other uses of `time_x = time_y` below

Svc/OsTime/test/ut/OsTimeTester.cpp
```diff
@@ -0,0 +1,118 @@
...
+      ASSERT_NEAR(0.4, time_400, 0.1);
+
+      // 600 ms
+      std::this_thread::sleep_for(dur_200ms);
+      Fw::Time time_600ms;
+      invoke_to_timeGetPort(0, time_600ms);
+      ASSERT_NE(time_600ms, Fw::ZERO_TIME);
+
+      const double time_600 = time_600ms.getSeconds() + (time_600ms.getUSeconds() / (1000.*1000.));
+      ASSERT_NEAR(0.6, time_600, 0.1);
```
thomas-bc: For the tests that are causing issues, maybe rather than `ASSERT_NEAR`, you could e.g. 
```
ASSERT_GT(time_600, 0.6) // should be deterministically true
ASSERT_LT(time_600, 0.6 * 1.5) // give some extra time to the slow github runners
```

Svc/OsTime/docs/sdd.md
kubiak-jpl: I updated the sdd in the latest comment

Svc/OsTime/OsTime.cpp
kubiak-jpl: That approach gives me compiler errors on my machine
```
[ 77%] Building CXX object F-Prime/Svc/OsTime/CMakeFiles/Svc_OsTime.dir/OsTime.cpp.o
/home/kubiak/source/fprime/fprime-kubiak/Svc/OsTime/OsTime.cpp: In member function ‘void Svc::OsTime::set_epoch(const Fw::Time&, const Os::RawTime&)’:
/home/kubiak/source/fprime/fprime-kubiak/Svc/OsTime/OsTime.cpp:31:30: error: invalid use of ‘void’
   31 |     (void) m_epoch_fw_time = fw_time;
      |                              ^~~~~~~
/home/kubiak/source/fprime/fprime-kubiak/Svc/OsTime/OsTime.cpp:32:30: error: invalid use of ‘void’
   32 |     (void) m_epoch_os_time = os_time;
      |                              ^~~~~~~
/home/kubiak/source/fprime/fprime-kubiak/Svc/OsTime/OsTime.cpp:33:28: error: invalid use of ‘void’
   33 |     (void) m_epoch_valid = true;
      |                            ^~~~
/home/kubiak/source/fprime/fprime-kubiak/Svc/OsTime/OsTime.cpp: In member function ‘virtual void Svc::OsTime::timeGetPort_handler(FwIndexType, Fw::Time&)’:
/home/kubiak/source/fprime/fprime-kubiak/Svc/OsTime/OsTime.cpp:60:19: error: invalid use of ‘void’
   60 |     (void) time = m_epoch_fw_time;
```

Svc/OsTime/test/ut/OsTimeTester.cpp
kubiak-jpl: I tried something similar to this. Let's see if it passes CI.

I think the ideal solution would be for a set a OSAL implementations that are designed for unit testing. Like a `UtRawTime` implementation that I could pass `Fw::Time` structures to for unit testing, but that would probably be a bigger discussion outside of the PR

Svc/OsTime/OsTime.cpp
thomas-bc: My bad sorry, looks like the syntax should be the following for return values of the assignment operator
```
    (void) (m_epoch_fw_time = fw_time);
```
That's pretty ugly 😮‍💨 

`m_epoch_valid = true;` probably doesn't need a cast

thomas-bc: We can also just dismiss the warning

kubiak-jpl:
@LeStarch @thomas-bc 

I made a few more substantial changes since the initial commit
- I added a `setEpoch` port call so that another component could set the new epoch time. I'm thinking of a time sync command handler here
- Added an `Os::Mutex` and a `ScopeLock` to protect the epoch variables

Svc/OsTime/OsTime.cpp
kubiak-jpl: I'd prefer to dismiss the warnings. I'm not really sure what it even means to check the return type of an assignment outside of using the assigned variable

Svc/OsTime/OsTime.fpp
```diff
@@ -0,0 +1,16 @@
+module Svc {
+
+    @ A pair of timestamps representing an Epoch time in 
+    @ an Fw::Time and Os::RawTime object
+    port OsTimeEpoch(
```
LeStarch: For module independence, ports should typically be defined outside of the component.

LeStarch (changes requested):
Need to fix UTs on mac.  It would be best to define the port outside of the component as that keeps components from having module-link dependencies on one-another.

Svc/OsTime/OsTime.fpp
kubiak-jpl: Sure. Where would you like me to add it?

thomas-bc: There's an `Svc/Ports/` module that this would likely fall under nicely.

kubiak-jpl:
@thomas-bc @LeStarch I reworked the unit tests to remove timing dependencies by adding a `RawTimeTester` module under the OsTime test directory. I suspect this could be more broadly useful though

Svc/OsTime/OsTime.cpp
LeStarch: Sorry, just noticing this.  Yes, dismiss all "operator=" does not check the return type.  This check is from C and not C++ compliant.  We will fix eventually.

LeStarch (approved):
As a note:
1. You could easily set a time with TIME_BASE to something akin to boot time and read the raw time during topology start-up.
2. Then during runtime, use the port to set "coordinated time" once you have an external time sync.

I like it!

LeStarch:
"@thomas-bc @LeStarch I reworked the unit tests to remove timing dependencies by adding a RawTimeTester module under the OsTime test directory. I suspect this could be more broadly useful though"

This was an excellent approach, thank you!

PR #3378, opened 2025-03-13
Removing NATIVE_ and POINTER_CAST from FppTest, Utils, cmake, config,…

### LINKED ISSUE #3377: Remove NATIVE_* from FppTest, CMake Tests, Documentation, and Other small bits

LeStarch:
… and docs

| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Removing NATIVE_ and POINTER_CAST from FppTest, Utils, cmake, config.

Fixes #3377


Utils/Hash/Hash.hpp
```diff
@@ -30,7 +30,7 @@ namespace Utils {
       // ----------------------------------------------------------------------
 
       //! The size type
-      using SizeType = NATIVE_INT_TYPE;
+      using SizeType = FwSizeType;
```
thomas-bc: It appears this `Hash::SizeType` is not actually used in anywhere in `Utils/Hash`. I would recommend either getting rid of it (my preference - not sure why it's there in the first place), or switch uses of `FwSizeType` to `Hash::SizeType`.

config/FpConfig.h
```diff
@@ -284,7 +284,7 @@ typedef FwIndexType FwQueueSizeType;
 
 // Specifies the size of the buffer attached to state machine signals.
 #ifndef FW_SM_SIGNAL_BUFFER_MAX_SIZE
-#define FW_SM_SIGNAL_BUFFER_MAX_SIZE 128  // Not to exceed size of NATIVE_UINT_TYPE
+#define FW_SM_SIGNAL_BUFFER_MAX_SIZE 128  // Not to exceed size of FwSizeType
```
thomas-bc: Should this say `Not to exceed max value of FwSizeType`? I would have understood it as "Must be < sizeof(FwSizeType)" if it were not set to 128 right now

Utils/Hash/Hash.hpp
LeStarch: Removed.

thomas-bc (approved):

Issue #3056, opened 2024-12-04
ROSES - Documentation Request: Integrating External Libraries into F Prime


elliotfayman:
### Integrating External Libraries into F Prime

I’ve been exploring how to integrate external libraries into F Prime by wrapping them in a new passive or active component. However, I haven’t found much documentation on the specifics of this process. I received the following guidance, which outlines two approaches: one for adding .so libraries (to be dynamically linked) and another for adding .a libraries (to be statically linked).

**First Approach: Adding a .so (Shared Object) Library**

1. Build the external library to generate a .so file, if one is not already available. 
2. Copy the generated .so and header files into the F Prime project.
3. Create a new passive or active F Prime component to wrap the semantics of the library.
4. Use mod_deps in the component’s CMake file to establish the necessary link dependencies.
5. Use target_include_directories to configure the header search paths for the compiler.

**Second Approach: Adding a .a (Static Archive) Library**

1. Build the external library to generate a .a archive file if one is not already available.
2. Copy the generated .a and header files into the F Prime project.
3. Use ExternalProject_Add in CMake to manage the library’s build process within F Prime 
4. Update mod_deps and target_include_directories as needed to support static linking and header inclusion.

**Additional Notes:**

1. In addition to the above guidance, I referenced the following CMake file which integrates an external library in the F Prime reference application: [FS/App/PiCamera/CMakeLists.txt at dev · FP-Mission/FS](https://github.com/FP-Mission/FS/blob/dev/App/PiCamera/CMakeLists.txt)
2. You must cross compile the external library if your target is a different architecture or platform than the one with which you’re building the library.
3. Whether you choose to wrap the external library in a passive or active component depends on the semantics of the external library. For example, if your library is similar to NumPy in Python, you should consider wrapping it inside a passive component because (and this is my impression - feel free to correct) that library does not require its own thread to operate, while reducing the number of active components can help conserve CPU resources.

thomas-bc:
@elliotfayman thanks for posting this issue. This is indeed a perfect candidate for https://github.com/nasa/fprime/discussions/2732 and will likely be implemented as such.

ronaldo-monteiro:
I would like to see the use of AI in code improvements for the project, especially in Python.

matt392code:
[library-integration-guide.md](https://github.com/user-attachments/files/18284360/library-integration-guide.md)
This documentation:
-Provides clear step-by-step instructions
-Includes practical examples
-Covers both static and dynamic linking
[...]

PR #3387, opened 2025-03-15
Add minor improvements for handling external libraries

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Incremental step in addressing https://github.com/nasa/fprime/issues/3056

**Proposed change:** If a `MOD_DEPS` entry is an absolute path, it should be recognized as a linker_only flag.

## Rationale 

When building a non-CMake-based library from source as part of the CMake build (e.g. with `ExternalProject_Add`), it can be useful to reference in the `MOD_DEPS` the path to a soon-to-be-built `lib<libName>.a` file.

However the current handling would not recognize this as a library file, since it only checks for the existence of the file. This addresses it by adding a check "is this an absolute path?". If yes, consider it a linker-only dependency.

### DIFF
--- a/Fw/Types/CMakeLists.txt
+++ b/Fw/Types/CMakeLists.txt
@@ -30,7 +30,7 @@ set(UT_SOURCE_FILES
   "${CMAKE_CURRENT_LIST_DIR}/test/ut/TypesTest.cpp"
 )
 set(UT_MOD_DEPS
-  "${FPRIME_FRAMEWORK_PATH}/Os"
+  Os
 )
 register_fprime_ut()
 set (UT_TARGET_NAME "${FPRIME_CURRENT_MODULE}_ut_exe")
--- a/cmake/FPrime.cmake
+++ b/cmake/FPrime.cmake
@@ -62,6 +62,7 @@ function(fprime_setup_global_includes)
     include_directories("${FPRIME_FRAMEWORK_PATH}")
     include_directories("${FPRIME_CONFIG_DIR}")
     include_directories("${FPRIME_PROJECT_ROOT}")
+    include_directories("${FPRIME_PROJECT_ROOT}/lib/include")
 
     # Setup the include directories that exist within the build-cache
     include_directories("${CMAKE_BINARY_DIR}")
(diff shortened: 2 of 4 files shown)


cmake/FPrime.cmake
```diff
@@ -62,6 +62,7 @@ function(fprime_setup_global_includes)
     include_directories("${FPRIME_FRAMEWORK_PATH}")
     include_directories("${FPRIME_CONFIG_DIR}")
     include_directories("${FPRIME_PROJECT_ROOT}")
+    include_directories("${FPRIME_PROJECT_ROOT}/lib/include")
```
thomas-bc, line 65: This is not required, users can always 
```cmake
target_include_directories(${FPRIME_CURRENT_MODULE} PUBLIC "/path/to/lib/include")
```
but it felt like it could be a QOL improvement. Happy to take it out if we feel uneasy about having this at the global level.

thomas-bc: This would only be useful in the case where a user is being provided with pre-built `lib.a` library along with some `include/<libName>/[...]` files (unsure if that's common?). If the library is being built from source, installed includes should go in the build cache.

Fw/Types/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ set(UT_SOURCE_FILES
   "${CMAKE_CURRENT_LIST_DIR}/test/ut/TypesTest.cpp"
 )
 set(UT_MOD_DEPS
-  "${FPRIME_FRAMEWORK_PATH}/Os"
+  Os
```
thomas-bc, line 33: Looks like the proposed change would be a breaking change for projects that may have been defining `MOD_DEPS` using full paths like this.

thomas-bc:
@LeStarch I'm keeping this open to see if you have quick thoughts, but I think we should close (not merge). I didn't realize it would be a breaking change, and it's not that important of a functionality.

thomas-bc:
Nevermind, it is a breaking change, and not that important. I'll further investigate to see if there are better ways to deal with this.

Closing

PR #3389, opened 2025-03-17
Add How-To guide for integrating external libraries

### LINKED ISSUE #3056: ROSES - Documentation Request: Integrating External Libraries into F Prime

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/3056

Examples can be found here: https://github.com/nasa/fprime-examples/pull/3

## Future Work

- Verify links work once the fprime-examples PR is merged

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -297,6 +297,7 @@ errornum
 ert
 esb
 ethanchee
+etl
 EVENTARRAY
 EVENTBOOL
 eventflags
@@ -638,6 +639,7 @@ ONLN
 OPCODEBASE
 openmct
 openpyxl
+openssldir
 oper
 optarg
 optind
--- a/docs/how-to/index.md
+++ b/docs/how-to/index.md
@@ -26,6 +26,15 @@ How-To guides offer step-by-step instructions for specific development tasks in 
 
     [View Develop F´ Libraries](develop-fprime-libraries.md){ .md-button .md-button--primary }
 
+-   <span class="card-title">__F´ Library Development__</span>
+
+    ---
+
+    This how to will walk you through how to integrate non-F´ libraries into your F´ project. It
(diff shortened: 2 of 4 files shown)


thomas-bc:
Not ready yet – closing

PR #3398, opened 2025-03-19
[Rate Groups and Timeliness] RateGroups Docs Enhancements

pavly-gerges:
| | |
|:---|:---|
|**_Changed Files (y/n)_**|  [fprime/docs/user-manual/rate-groups.md](https://github.com/nasa/fprime/blob/devel/docs/user-manual/design-patterns/rate-group.md) |
| **_Requested Changes_** | A detailed Clock prescaler example algorithm using components from the `Svc` core module. |

---
### The First Change:
I am completely new to F-Prime, but I've some good experience in embedded systems design in general. I was reading the [RateGroups and Timeliness](https://fprime.jpl.nasa.gov/latest/docs/user-manual/design-patterns/rate-group/) documentation when I stepped into a grammatical error, and I've corrected it in this PR.

### The Second Change:
At the first glance, I didn't understand what does it mean to provide a clock divider until I looked for the code. That's why I've added this example.

This change involves an example for the algorithms that could be used to implement a Virtual Clock Prescaler (And, I believe there is more). The example simplifies the operations of Cycling and Sampling performed by the core routine `Svc::RateGroupDriver::CycleIn_handler(...)`. 

## Rationale

I believe this example would be a great add to this page; as it also describes the core algorithm used here to design the `RateGroupDriver` components around the System base component. It may immerse the reader, who might be new to F`-prime, into the core quickly, in addition to providing a good example to the RateGroups-Drivers pattern. Let me know if there are requested changes.

## Future Work

I think referencing more details from the core on the specialized implementation documentation pages would be a good rigorous change, unless these algorithms shall carry breaking future changes. What do you think about that? Also, another thing, maybe examining whether there are other clock prescaler algorithms?

### DIFF
--- a/Svc/FprimeDeframer/FprimeDeframer.cpp
+++ b/Svc/FprimeDeframer/FprimeDeframer.cpp
@@ -58,15 +58,20 @@ void FprimeDeframer ::dataIn_handler(FwIndexType portNum, Fw::Buffer& data, cons
         return;
     }
     // -------- Attempt to extract APID from Payload --------
-    // If PacketDescriptor translates to an invalid APID, let it default to FW_PACKET_UNKNOWN
-    // and let downstream components (e.g. custom router) handle it
-    FwPacketDescriptorType packetDescriptor;
-    status = deserializer.deserializeTo(packetDescriptor);
-    FW_ASSERT(status == Fw::SerializeStatus::FW_SERIALIZE_OK, status);
     ComCfg::FrameContext contextCopy = context;
-    // If a valid descriptor is deserialized, set it in the context
-    if (packetDescriptor < ComCfg::Apid::INVALID_UNINITIALIZED) {
-        contextCopy.set_apid(static_cast<ComCfg::Apid::T>(packetDescriptor));
+    if (deserializer.getBuffLeft() < FprimeProtocol::FrameTrailer::SERIALIZED_SIZE + sizeof(FwPacketDescriptorType)) {
+        // Not enough data to read a valid FwPacketDescriptor, emit event and skip attempting to read an APID
+        this->log_WARNING_LO_PayloadTooShort();
+    } else {
+        // If PacketDescriptor translates to an invalid APID, let it default to FW_PACKET_UNKNOWN
+        // and let downstream components (e.g. custom router) handle it
+        FwPacketDescriptorType packetDescriptor;
+        status = deserializer.deserializeTo(packetDescriptor);
+        FW_ASSERT(status == Fw::SerializeStatus::FW_SERIALIZE_OK, status);
+        // If a valid descriptor is deserialized, set it in the context
+        if ((packetDescriptor < ComCfg::Apid::INVALID_UNINITIALIZED)) {
+            contextCopy.set_apid(static_cast<ComCfg::Apid::T>(packetDescriptor));
+        }
     }
 
     // ---------------- Validate Frame Trailer ----------------
--- a/Svc/FprimeDeframer/FprimeDeframer.fpp
+++ b/Svc/FprimeDeframer/FprimeDeframer.fpp
@@ -30,6 +30,11 @@ module Svc {
       severity warning high \
       format "Frame dropped: The transmitted frame checksum does not match that computed by the receiver"
 
+    @ An invalid frame was received (not enough data to contain a valid FwPacketDescriptor type)
+    event PayloadTooShort \
+      severity warning low \
+      format "The received buffer is too short to contain a valid FwPacketDescriptor"
+
     ###############################################################################
     # Standard AC Ports for Events 
     ###############################################################################
--- a/Svc/FprimeDeframer/test/ut/FprimeDeframerTester.cpp
+++ b/Svc/FprimeDeframer/test/ut/FprimeDeframerTester.cpp
@@ -27,6 +27,10 @@ FprimeDeframerTester ::~FprimeDeframerTester() {}
 // ----------------------------------------------------------------------
 
 void FprimeDeframerTester ::testNominalFrame() {
+    // This tests a nominal frame with 1 byte of data - which per F Prime protocol
+    // does not contain a valid FwPacketDescriptor (2 bytes) and therefore emits a warning event
+    // See testNominalFrameApid() for a nominal frame with a valid FwPacketDescriptor
+
     // Get random byte of data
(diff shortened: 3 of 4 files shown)


pavly-gerges:
Let me know if there is something needs to be done to get this approved.

docs/user-manual/design-patterns/rate-group.md
```diff
@@ -25,6 +25,44 @@ system clock and the port call to rate group driver's `CycleIn` port.
 The reference application calls the `CycleIn` port followed by a sleep for the system clock time within a while loop to
 simulate a system-driven clock.
 
+**_Example_**: Recall a crystal oscillator running at 1000 HZ (or 1 MHZ), then a system clock source would sample each 1MHZ. Recall that a `RateGroupDriver` is registered to this system clock source sampler that requires to be updated at a rate of 100HZ; therefore the following applies:
```
LeStarch: I'd recommend this:

1. "Consider a system clock driven at 1000Hz by a crystal oscillator and a rate group needing to run at 100Hz".

docs/user-manual/design-patterns/rate-group.md
```diff
@@ -25,6 +25,44 @@ system clock and the port call to rate group driver's `CycleIn` port.
...
+* On the API layers level, both drivers; the `SystemSourceDriver`, and the `RateGroupDriver` shall be implementations of F' components; thus they could only communicate via ports (e.g., commands, events, and telemetry channels). In this case, the implementation could be taken towards commanding.
```
LeStarch: 1. Remove "On the API layers level"
2. "Both drivers, `SystemSourceDriver`, and the `RateGroupDriver`, are F´components"
3. Remove "(e.g., commands, events, and telemetry channels). In this case, the implementation could be taken towards commanding."

docs/user-manual/design-patterns/rate-group.md
```diff
@@ -25,6 +25,44 @@ system clock and the port call to rate group driver's `CycleIn` port.
...
+* On the hardware level, the `SystemSourceDriver` shall use a clock divider that drives the `RateGroupDriver` each $$\frac{100HZ}{1MHZ}$$ of the system source sampler (i.e., at a rate of $$\frac{1}{10}$$ of the system source sampler).
```
LeStarch: Remove "On the hardware level".  All rate group items are handled in software.

docs/user-manual/design-patterns/rate-group.md
```diff
@@ -25,6 +25,44 @@ system clock and the port call to rate group driver's `CycleIn` port.
...
+* On the hardware implementation level, this could be done via many approaches; a counter algorithm (that triggers an interrupt when reaching 100 cycles resetting the `numberOfCycles` each second) can suffice or a more complex modular arithmetics algorithm could be used (e.g., $$((ticks\ mod\ rate)\ ==\ offset) \implies CycleOut$$; where `ticks` represents the current system virtual clocks, `rate` represents the `RateGroupDriver` required frequency, and `offset` represents the remainder of the value at which sampling shall occur when the system clock `ticks` reaches the requested `rate`; Zero `offset` means that the system source sampler will sample this driver when it reaches an integer multiple of its requested `rate`).
```
LeStarch: Here I would just state "In the provided Svc/RateGroupDriver this is done by counting up to a configured number of cycles before emiting the divided signal".  Link to the source code.

docs/user-manual/design-patterns/rate-group.md
```diff
@@ -25,6 +25,44 @@ system clock and the port call to rate group driver's `CycleIn` port.
...
+
+> [!NOTE]
```
LeStarch: Delete this whole block with code.  It would be better to provide a link into the code.

docs/user-manual/design-patterns/rate-group.md
```diff
@@ -25,6 +25,44 @@ system clock and the port call to rate group driver's `CycleIn` port.
...
+>               this->CycleOut_out(static_cast<FwIndexType>(entry),cycleStart);
+>            } 
+>        }
+>        // rollover the tick value when the tick count reaches the rollover value
+>        // the rollover value is the product of all the dividers. See comment in constructor.
+>        this->m_ticks = (this->m_ticks + 1) % this->m_rollover;
+>    }
+> ...
+> ```
+> * The algorithm utilized here provides a software-based analogous activity to [the clock prescaler digital circuits](https://en.wikipedia.org/wiki/Prescaler).
```
LeStarch: Good!  This should come right after your example.

LeStarch (changes requested):
I looked through this PR, and I see where you are coming from.  I made some requests for clarifications, because all of this happens in software.

I really like the analog with the prescaler.

Thanks!

docs/user-manual/design-patterns/rate-group.md
pavly-gerges: I understand the architecture of the RateGroupDrivers. In this part, I was willing to describe what happens during the runtime, which is really hardware and not Rate Groups anymore. Do you recommend a better introductory word?

pavly-gerges: So, this explanation is considered a text-based deployment model.

docs/user-manual/design-patterns/rate-group.md
pavly-gerges: So, what do you recommend regarding this bullet point? I'd like to describe each variable used, but simply. It was best to do this via a logical implication formula (which summarizes what the algorithm does).

LeStarch:
Looks like this branch has developed a merge conflict.  Until this is resolved, there is not much we can do here.

pavly-gerges:
> Looks like this branch has developed a merge conflict. Until this is resolved, there is not much we can do here.

No problem. Let's defer it for the next release. I'll fix it.

LeStarch:
@pavly-gerges would you give edit rights to maintainers for this PR?  We need this to help assist with conflict resolution and pushing this through to merge.

pavly-gerges:
> @pavly-gerges would you give edit rights to maintainers for this PR? We need this to help assist with conflict resolution and pushing this through to merge.

I don't have that option anywhere. I tried discarding the changes that caused the conflicts but it seems to be messing up the whole branch.

pavly-gerges:
@LeStarch I'm still interested anyway in contributing to the F-prime documentation. If you're still interested in merging, I've made a clean PR with the same changes, and the same branch. https://github.com/nasa/fprime/pull/4196

PR #3417, opened 2025-03-26
Replace FpConfig type aliases with FPP generated aliases

Kronos3:
This PR replaces typedefs created in `FpConfig.h` with FPP generated type definitions.  This is the diff I've been building https://github.com/nasa/fpp/pull/669 against. It depends on that PR for this to build successfully which is why I'm opening this as a draft until that is merged.


LeStarch:
I have moved this work into my config branch (maintain commits, but not hashes).  See PR #3422.  Authorship has been retained.

PR #3422, opened 2025-03-28
Refactored type organization

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Allows for PlatformTypes to build FPP!


config/FpConfig.fpp
```diff
@@ -1,16 +1,85 @@
...
+
+@ The type of smaller indices internal to the software, used
+@ for array indices, e.g., port indices. Must be signed.
+type FwIndexType = PlatformIndexType
+
+@ The type of arguments to assert functions.
+type FwAssertArgType = PlatformAssertArgType
+
+@ The type of task priorities used.
+type FwTaskPriorityType = PlatformTaskPriorityType;
```
thomas-bc: There's an extra `;` at end of line here, is that ok in FPP ?

cmake/FPrime.cmake
```diff
@@ -204,7 +204,15 @@ function(fprime_setup_included_code)
...
 
-    # Specific configuration handling
+    # Specific configuration module handling:
+    #
+    # add_fprime_subdirectory cannot be run until later in the build process. Otherwise detection
+    # for model specific post processing is messed up. Thus we synthesize the behavior by setting
+    # the current module and then calling stock "add_subdirectory".
+    get_module_name("${FPRIME_PLATFORM_MODULE_DIRECTORY}")
+    set(FPRIME_CURRENT_MODULE "${MODULE_NAME}")
+    string(REPLACE "_" "/" PLATFORM_MODULE_PATH "${FPRIME_CURRENT_MODULE}")
```
thomas-bc, line 214: If I'm following correctly what's happening here, this means that underscores `_` are forbidden for platform module names?

config/FpConfig.fpp
```diff
@@ -1,16 +1,85 @@
...
-type FwPrmIdType
-type FwSignedSizeType
-type FwSizeStoreType
-type FwSizeType
-type FwTimeBaseStoreType
-type FwTimeContextStoreType
-type FwTlmPacketizeIdType
-type FwTraceIdType
+# ======================================================================
+# \title  Fw/FPrimeBasicTypes.hpp
```
bocchino: I think the file name should be .fpp, not .hpp

bocchino (approved):
Looks great!

@garthwatney We should update the STARS tool to conform to this change. Instead of including `FpConfig.hpp` we need to include `Fw/FPrimeBasicTypes.hpp` in the generated code for F Prime.

config/FpConfig.fpp
LeStarch: Good catch!  I'll fix.

thomas-bc (approved):
LGTM! Just a few comments

cmake/FPrime.cmake
LeStarch: Other way around.  Module names are not permitted "/" and so we replace with "_"

LeStarch: This is all going to change anyway, so lets disallow "_" in module names for platforms.

LeStarch:
## Breaking Changes
1. Platforms now build with FPP and have been restructured

Issue #2714, opened 2024-05-01
Consider having `Fw::Buffer::getSerializeRepr` return an object instead of a reference


bocchino:
`Fw::Buffer::getSerializeRepr` currently returns a reference into a serialize buffer stored as a member of `Fw::Buffer`. There are some issues with this:

1. When you call `getSerializeRepr`, you're not sure what state the serialize buffer is in. It depends on what has previously been done with the buffer.
2. The serialize representation state is reset when the buffer object is serialized and deserialized but otherwise persists. This could be an unexpected behavior.
3. Performing deserialize operations on the buffer updates the serialize buffer and therefore updates the `Fw::Buffer`. It would be better for the`Fw::Buffer` to be read-only in this case.

Instead we could have `getSerializeRepr` return an object, say `Fw::ExternalSerializeBufferWithMemberCopy`. That way the user would own the serialize representation, and it could be freshly initialized each time it is requested. An argument to `getSerializeRepr` could specify whether the representation is for serialization (in which case the internal serial pointer should be set to zero) or deserialization (in which case the internal serial pointer should be set to the end of the data). Access to the buffer for deserialization would be read-only.

LeStarch:
I like it!  Thanks for the write-up @bocchino!

LeStarch:
We should do:
1. Separate serialize/deserialize methods. (e.g. `getSerializer` and `getDeserializer`) using Robs suggestions above.
2. Deprecate (`Fw/Deprecate.hpp`) the `getSerializeRepr` method, but don't remove the functionality (yet)

thomas-bc:
Implemented by #3431  🎉

PR #3431, opened 2025-04-02
Deprecate Fw::Buffer::getSerializeRepr in favor of 'getSerializer' and 'getDeserializer'

vincewoo:
| | |
|:---|:---|
|**_Related Issue(s)_**| #2714 |

---
## Change Description

As suggested in #2714 this change deprecates `Fw::Buffer::getSerializeRepr` and introduces two new functions in `Fw::Buffer` which can be used for retrieving a serializer (`Fw::Buffer::getSerializer`) or deserializer (`Fw::Buffer::getDeserializer`). Each call to get a serializer or deserializer will pre-initialize the `Fw::ExternalSerializeBufferWithMemberCopy` returned so that the caller does not have to do anything before using it.

## Rationale

This change reduces the need for the caller to manage the internal state of the serializer. Each call to get serializer/deserlizer will return an object that is ready to operate on.

As part of these changes I have gone ahead and upgraded all the various ares in the F' codebase that used the `Fw::Buffer::getSerializeRepr` and the associated test cases.

There were two design decisions that I wanted to highlight for reviewers:

1. The copy constructor `ExternalSerializeBufferWithMemberCopy(const ExternalSerializeBufferWithMemberCopy& src)` in `Fw::Serializable` was previously explicit which prevented the ability to return an object of `Fw::ExternalSerializeBufferWithMemberCopy` from a function. I have removed the explicit-ness of it.
2. Both `getSerializer()` and `getDeserializer()` will return an empty `ExternalSerializeBufferWithMemberCopy` object if the parent `Buffer` is uninitialized (i.e. does not contain data). The empty `ExternalSerializeBufferWithMemberCopy` will reject serialization requests as expected. 

## Future Work

None.


Fw/Buffer/Buffer.hpp
```diff
@@ -93,7 +93,18 @@ class Buffer : public Fw::Serializable {
...
+    DEPRECATED(SerializeBufferBase& getSerializeRepr(), "Switch to .getSerializer() and .getDeserializer()");
+
+    //! Returns a ExternalSerializeBufferWithMemberCopy representation of the wrapped data for serializing
+    //!
+    //! \return representation of the wrapped data to aid in serializing to it
+    ExternalSerializeBufferWithMemberCopy getSerializer();
+
+
+    //! Returns a ExternalSerializeBufferWithMemberCopy representation of the wrapped data for deserializing
+    //!
```
LeStarch, line 112: Should add a `\warning` say that this assumes the entire buffer (up to `getSize`) is available for deserialization.

Fw/Buffer/Buffer.hpp
```diff
@@ -93,7 +93,18 @@ class Buffer : public Fw::Serializable {
     //! to the wrapped buffer. Once obtained the user should call one of two functions: `sbb.resetSer();` to setup for
     //! serialization, or `sbb.setBuffLen(buffer.getSize());` to setup for deserializing.
     //! \return representation of the wrapped data to aid in serializing to it
-    SerializeBufferBase& getSerializeRepr();
+    DEPRECATED(SerializeBufferBase& getSerializeRepr(), "Switch to .getSerializer() and .getDeserializer()");
+
+    //! Returns a ExternalSerializeBufferWithMemberCopy representation of the wrapped data for serializing
+    //!
```
LeStarch, line 99: Should add a `\warning` saying that this assumes the buffer is entirely empty.

Fw/Buffer/Buffer.cpp
```diff
@@ -104,8 +104,24 @@
...
+        return esb;
+    } else {
+        return ExternalSerializeBufferWithMemberCopy();
+    }
+}
+
+Fw::ExternalSerializeBufferWithMemberCopy Buffer::getDeserializer() {
+    if (this->isValid()) {
+        Fw::ExternalSerializeBufferWithMemberCopy esb(this->m_bufferData, static_cast<Fw::Serializable::SizeType>(this->m_size));
+        esb.setBuffLen(this->getSize());
```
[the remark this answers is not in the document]
LeStarch: Should assert on the return value here.

Fw/Buffer/docs/sdd.md
```diff
@@ -69,34 +64,24 @@ To check validity, you can call the interface function `isValid()`.
...
-of the `Fw::Buffer` and is initially empty.  Users can serialize and deserialize through `sb` to copy to/from the backed
-memory.
-
-The state of `sb` persists as long as the current `Fw::Buffer` object exists as it is stored as a member. However, all
-`Fw::Buffer` constructors initialize `sb` to an empty state including the `Fw::Buffer` copy constructor. Thus, if an
-`Fw::Buffer` is sent through a port call, passed by-value to a function call, or otherwise copied, the state of the new
-buffer's `sb` member is empty.
+Users can obtain a serialization buffer, `sb`, by calling either `getSerializer()` or `getDeserializer()`. 
+Note that both of these methods return a `Fw::ExternalSerializeBufferWithMemberCopy` object that is meant to be 
+managed by the caller and does not inadvertently affect the original `Fw::Buffer` parent object.
```
LeStarch: "only affects the data of the underlying buffer"?

Fw/Types/Serializable.hpp
```diff
@@ -258,7 +258,7 @@ class ExternalSerializeBufferWithMemberCopy final : public ExternalSerializeBuff
         : ExternalSerializeBuffer(buffPtr, size) {}
     ExternalSerializeBufferWithMemberCopy() : ExternalSerializeBuffer() {}
     ~ExternalSerializeBufferWithMemberCopy() {}
-    explicit ExternalSerializeBufferWithMemberCopy(const ExternalSerializeBufferWithMemberCopy& src)
+    ExternalSerializeBufferWithMemberCopy(const ExternalSerializeBufferWithMemberCopy& src)
```
LeStarch, line 257: Don't we need explicit here?  What was the need to remove this?

docs/user-manual/framework/ground-interface.md
```diff
@@ -123,9 +123,9 @@ class MyFrameProtocol : public Svc::FramingProtocol {
 
     void frame(const U8 *const data, const U32 size, Fw::ComPacket::ComPacketType packet_type) {
         Fw::Buffer my_framed_data = m_interface.allocate(size);
-        my_framed_data.getSerializeRepr().serialize(0xdeadbeef); // Some start word
-        my_framed_data.getSerializeRepr().serialize(size);       // Write size
-        my_framed_data.getSerializeRepr().serialize(data, size, true); // Data copied to buffer no length included
+        my_framed_data.getSerializer().serialize(0xdeadbeef); // Some start word
```
LeStarch: This sample code won't work any more, because state is not internalized in the Fw::Buffer.  Subsequent calls to `.serializer` return a fresh serializer.

We should probably add a `\warning` tag to the .hpp on this too.

LeStarch (changes requested):

LeStarch:
@vincewoo you should also look into the failing unit tests

LeStarch:
@bocchino would you review this one?  @vincewoo has convered `Fw::Buffer` to have `.getSerializer()` and `.getDeserializer()` methods that return an independent serializer that doesn't set state on the underlying `Fw::Buffer`.  You requested this change originally.

Fw/Types/Serializable.hpp
vincewoo: Having this copy constructor as explicit prevented the function from creating a `Fw::ExternalSerializeBufferWithMemberCopy` object on the stack and passing it back by value in the return. The copy constructor is called implicitly to copy the returned object into the callers stack.

LeStarch: Ok.  Makes sense to me.  @bocchino is there a better approach here?

Fw/Buffer/Buffer.cpp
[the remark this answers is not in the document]
vincewoo: Added assert as suggested.

Fw/Buffer/Buffer.hpp
vincewoo: Added warning.

Fw/Buffer/Buffer.hpp
vincewoo: Added warning.

Fw/Types/Serializable.hpp
vincewoo: Discussed this at the F' Infra weekly meeting. @bocchino confirmed that it is okay to remove the `explicit` from this copy constructor.

docs/user-manual/framework/ground-interface.md
vincewoo: Corrected the sample code to retrieve only one `serializer` object. Added warning to `Fw::Buffer.getSerializer()` declaration.

Fw/Types/Serializable.hpp
bocchino: This change looks good to me. The C++ rules require explicit constructors only when there is a single argument of fundamental type. Copy constructors don't have to be explicit.

Fw/Buffer/Buffer.cpp
```diff
@@ -104,8 +104,25 @@ void Buffer::set(U8* const data, const SizeType size, const U32 context) {
...
+        return esb;
+    } else {
+        return ExternalSerializeBufferWithMemberCopy();
+    }
+}
+
+Fw::ExternalSerializeBufferWithMemberCopy Buffer::getDeserializer() {
+    if (this->isValid()) {
+        Fw::ExternalSerializeBufferWithMemberCopy esb(this->m_bufferData, static_cast<Fw::Serializable::SizeType>(this->m_size));
+        Fw::SerializeStatus stat = esb.setBuffLen(this->getSize());
```
bocchino: This is a minor point, but it seems inconsistent to use this->getSize() here and this->m_size in the previous line. Maybe use this->m_size consistently.

Svc/FprimeRouter/test/ut/FprimeRouterTester.cpp
```diff
@@ -76,7 +76,8 @@ void FprimeRouterTester::mockReceivePacketType(Fw::ComPacket::ComPacketType pack
     const FwPacketDescriptorType descriptorType = packetType;
     U8 data[sizeof descriptorType];
     Fw::Buffer buffer(data, sizeof(data));
-    buffer.getSerializeRepr().serialize(descriptorType);
+    Fw::ExternalSerializeBufferWithMemberCopy esb = buffer.getSerializer();
+    esb.serialize(descriptorType);
```
bocchino: Should we check the return type here?

FppTest/dp/test/ut/Tester.cpp
```diff
@@ -94,16 +94,16 @@ void Tester::productRecvIn_Container1_SUCCESS() {
                                              DpTest::ContainerPriority::Container1, this->container1Buffer, buffer,
                                              expectedNumElts);
     // Check the data
-    Fw::SerializeBufferBase& serialRepr = buffer.getSerializeRepr();
-    Fw::TestUtil::DpContainerHeader::checkDeserialAtOffset(serialRepr, Fw::DpContainer::DATA_OFFSET);
+    Fw::ExternalSerializeBufferWithMemberCopy deserializer = buffer.getDeserializer();
```
bocchino: Consider using `auto` here and elsewhere instead of the verbose variable type.

bocchino:
Looks good so far. I added a few comments. I'll complete my review after the tests pass.

FppTest/dp/test/ut/Tester.cpp
vincewoo: Thanks for the suggestion, updated to using auto here and anywhere getSerializer and getDeserializer are used.

Fw/Buffer/Buffer.cpp
vincewoo: Corrected for consistency

Svc/FprimeRouter/test/ut/FprimeRouterTester.cpp
vincewoo: Corrected

vincewoo:
@bocchino We've passed the checks! Please feel free to review when you have the chance

Fw/Buffer/Buffer.hpp
```diff
@@ -93,7 +93,22 @@ class Buffer : public Fw::Serializable {
...
+    //! \warning Assumes that this buffer is entirely empty.
```
thomas-bc: It might be good to call out that if the buffer is not empty, it's still possible to manipulate the serialization offsets with `moveSerToOffset` , `serializeSkip` etc. ? Helps users understand how these things can be used

Svc/GenericHub/GenericHubComponentImpl.cpp
```diff
@@ -65,7 +65,7 @@ void GenericHubComponentImpl ::dataIn_handler(const FwIndexType portNum, Fw::Buf
     Fw::SerializeStatus status = Fw::FW_SERIALIZE_OK;
 
     // Representation of incoming data prepped for serialization
-    Fw::SerializeBufferBase& incoming = fwBuffer.getSerializeRepr();
+    auto incoming = fwBuffer.getDeserializer();
     FW_ASSERT(incoming.setBuffLen(fwBuffer.getSize()) == Fw::FW_SERIALIZE_OK);
```
thomas-bc: Two things here:
- Is this `incoming.setBuffLen()` call now not needed?
- The exact same call is duplicated 2 lines below, we might want to fix this. I believe having side effects within `FW_ASSERT` is not recommended, and I'd bet that it being less readable is the reason it's duplicated by mistake.

thomas-bc (commented):
LGTM! Just a few comments

This is amazing!! 🚀

Fw/Buffer/Buffer.hpp
vincewoo: Added additional guidance to warning.

Svc/GenericHub/GenericHubComponentImpl.cpp
vincewoo: You're right, that is no longer needed. This has been cleaned up by deleting the FW_ASSERT with the side effect and the subsequent setBuffLen() further down.

LeStarch (approved):

thomas-bc (approved):

Fw/Buffer/Buffer.hpp
```diff
@@ -93,7 +93,24 @@ class Buffer : public Fw::Serializable {
     //! to the wrapped buffer. Once obtained the user should call one of two functions: `sbb.resetSer();` to setup for
     //! serialization, or `sbb.setBuffLen(buffer.getSize());` to setup for deserializing.
     //! \return representation of the wrapped data to aid in serializing to it
-    SerializeBufferBase& getSerializeRepr();
+    DEPRECATED(SerializeBufferBase& getSerializeRepr(), "Switch to .getSerializer() and .getDeserializer()");
+
+    //! Returns a ExternalSerializeBufferWithMemberCopy representation of the wrapped data for serializing
+    //!
+    //! \warning Assumes that this buffer is entirely empty. If the buffer is not empty it is still possible to
```
bocchino: This warning text is a bit confusing. What is "this buffer" referring to? The ExternalSerializeBuffer or the Fw::Buffer? In the first case, the code doesn't assume the buffer is empty -- it is empty (i.e., the serialization pointer is set to the start of the buffer). In the second case, the buffer is neither empty nor full; it's just a byte array. Maybe rewrite this text to say more directly that the serialization pointer is set to zero, so that serialization will start at the beginning of the memory pointed to by the Fw::Buffer. If that behavior isn't wanted, one can move the offset before serializing.

bocchino (approved):
Looks good! Please look at the comment I made with a suggested wording change.

Fw/Buffer/Buffer.hpp
vincewoo: Updated wording as suggested.

thomas-bc:
## Other Changes / breaking

- `Fw::Buffer::getSerializeRepr()` has been deprecated in favor of two simpler-to-use methods for serialization and deserialization: `Fw::Buffer::getSerializer()` and `Fw::Buffer::getDeserializer()`. Example changesets are found [nasa/fprime#3431](https://github.com/nasa/fprime/pull/3431).

PR #3440, opened 2025-04-02
Update Ref to use FPP telemetry packets

bocchino:
This PR updates the Ref topology to use FPP telemetry packets instead of XML packets.


jwest115 (approved):

thomas-bc (approved):
LGTM! Just one question; why not use TlmPacketizer on Ref ? Especially now that we have that integrated in FPP, it sounds like defining Packets and everything –only to not use them– could be confusing.

LeStarch (approved):

PR #3487, opened 2025-04-11
Config blank slate

### LINKED ISSUE #3499: CMake Refactor: Switch to Using Target Properties
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Switch to using the following "modern" CMake practices:

1. Using TARGET properties to store properties
2. Using function arguments to `register_fprime_*` functions instead of magic variables

Keep backwards compatibility for the timebeing.

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

CMake refactor - pt1


cmake/API.cmake
```diff
@@ -314,25 +304,20 @@ endfunction(register_fprime_module)
...
-    if(NOT DEFINED EXECUTABLE_NAME AND ARGC GREATER 0)
-        set(EXECUTABLE_NAME "${ARGV0}")
-    elseif(NOT DEFINED EXECUTABLE_NAME)
-        # Check to be sure before using
-        if (NOT DEFINED FPRIME_CURRENT_MODULE)
-            message(FATAL_ERROR "FPRIME_CURRENT_MODULE not defined. Please supply name to: register_fprime_executable()")
-        endif()
-        set(EXECUTABLE_NAME "${FPRIME_CURRENT_MODULE}")
+    if (DEFINED EXECUTABLE_NAME)
+        fprime_fatal_cmake_error("EXECUTABLE_NAME variable no longer supported")
```
thomas-bc: name of htis function has been changed? To be confirmed

cmake/API.cmake
```diff
@@ -314,25 +304,20 @@ endfunction(register_fprime_module)
...
-    get_nearest_build_root(${CMAKE_CURRENT_LIST_DIR})
-    generate_executable("${EXECUTABLE_NAME}" "${SOURCE_FILES}" "${MOD_DEPS}")
+    fprime_add_executable_build_target(${ARGN})
+
+    # Set up target/ targets for this module
+    setup_module_targets("${INTERNAL_MODULE_NAME}")
 endfunction(register_fprime_executable)
 
+function(fprime_add_executable_build_target)
+    fprime__internal_add_build_target("Executable" "INTERFACE" ${ARGN})
```
thomas-bc: "INTERFACE" to be deleted

cmake/module.cmake
```diff
@@ -10,95 +10,143 @@
...
+        add_library("${TARGET_NAME}" ${EXTRA_CMAKE_DIRECTIVES} "${SOURCES}")
+    else()
+        fprime_fatal_cmake_error("Cannot register compilation target of type ${TYPE}")
     endif()
-endfunction(generate_ut)
+    # TODO: this is needed because sub-builds still attempt register targets, but without the build target to add back in the
+    #       autocoding output. Thus empty must be substituted. Would it be possible to force the library to be an INTERFACE
+    #       instead?  Or only add empty on sub-builds?
+    target_sources("${TARGET_NAME}" PRIVATE "${FPRIME__INTERNAL_EMPTY_CPP}")
+    target_sources("${TARGET_NAME}" PRIVATE ${SOURCES})
```
thomas-bc: to be deleted?

cmake/target/build.cmake
```diff
@@ -19,103 +19,115 @@ if (FPRIME_ENABLE_UT_COVERAGE)
...
+    # not require sources while still providing the name. 
+    list(REMOVE_ITEM MODULE_SOURCES "${FPRIME__INTERNAL_EMPTY_CPP}")
+    set_target_properties(
+        ${BUILD_MODULE_NAME}
+        PROPERTIES
+        SOURCES "${MODULE_SOURCES}"
+    )
+    #### End Remove empty.cpp ####
 
+    #### Set FPRIME_TARGET_DEPENDENCIES ####
```
thomas-bc: Remove this block

cmake/autocoder/fpp_ut.cmake
```diff
@@ -72,7 +72,9 @@ function(fpp_ut_setup_autocode AC_INPUT_FILES)
     endif()
     set(AUTOCODER_GENERATED ${GENERATED_AI} ${GENERATED_CPP})
     set(AUTOCODER_GENERATED "${AUTOCODER_GENERATED}" PARENT_SCOPE)
-    set(AUTOCODER_DEPENDENCIES "${MODULE_DEPENDENCIES}" PARENT_SCOPE)
+    set(AUTOCODER_DEPENDENCIES "${MODULE_DEPENDENCIES};${FPRIME_CURRENT_MODULE}" PARENT_SCOPE)
```
thomas-bc: Why is FPRIME_CURRENT_MODULE here?

cmake/test/data/test-fprime-library/cmake/autocoder/test.cmake
```diff
@@ -18,4 +18,6 @@ function(test_setup_autocode AC_INPUT_FILE)
     set(GENERATED "${CMAKE_BINARY_DIR}/test-ac-1" "${CMAKE_BINARY_DIR}/test-ac-2")
     set(AUTOCODER_SCRIPT ${CMAKE_COMMAND} -E touch ${GENERATED} PARENT_SCOPE)
     set(AUTOCODER_GENERATED "${GENERATED}" PARENT_SCOPE)
+    set(AUTOCODER_BUILD_SOURCES "${GENERATED}" PARENT_SCOPE)
+    set(AUTOCODER_NEW_AUTOCODER_INPUTS "${GENERATED}" PARENT_SCOPE)
```
thomas-bc, line 22: If an output is not read by either subsequent autocoders or the build; there's an issue.

cmake/autocoder/autocoder.cmake
```diff
@@ -16,35 +16,47 @@ include(autocoder/helpers)
...
-    set(AC_GENERATED "${GENERATED_FILE_LIST}" PARENT_SCOPE)
-    set(AC_SOURCES "${CONSUMED_SOURCES_LIST}" PARENT_SCOPE)
-    set(AC_FILE_DEPENDENCIES "${FILE_DEPENDENCY_LIST}" PARENT_SCOPE)
+    # Append all autocoder properties to the various target properties
+    append_list_property("${GENERATED_FILE_LIST}" TARGET "${BUILD_TARGET_NAME}" PROPERTY AC_GENERATED)
+    append_list_property("${FILE_DEPENDENCY_LIST}" TARGET "${BUILD_TARGET_NAME}" PROPERTY AC_FILE_DEPENDENCIES)
+    append_list_property("${AUTOCODER_INPUT_SOURCES}" TARGET "${BUILD_TARGET_NAME}" PROPERTY AUTOCODER_INPUTS)
+    append_list_property("${BUILD_SOURCES_LIST}" TARGET "${BUILD_TARGET_NAME}" PROPERTY SOURCES)
+    append_list_property("${MODULE_DEPENDENCIES_LIST}" TARGET "${BUILD_TARGET_NAME}" PROPERTY LINK_LIBRARIES)
+    append_list_property("${MODULE_DEPENDENCIES_LIST}" TARGET "${BUILD_TARGET_NAME}" PROPERTY INTERFACE_LINK_LIBRARIES)
```
thomas-bc, line 51: add a comment here

cmake/target/target.cmake
```diff
@@ -110,18 +132,28 @@ function(setup_single_target TARGET_FILE MODULE SOURCES DEPENDENCIES)
...
+                    foreach(ITEM_PRINT IN LISTS ${LIST_PRINT})
+                        message(STATUS "    ${ITEM_PRINT}")
+                    endforeach()
+                endforeach()
+            endif()
+            check_unknown_links("${MODULE}" ${UNKNOWN_LINKS})
+            set_target_properties("${MODULE}" PROPERTIES TRANSITIVE_LINK_LIBRARIES "${KNOWN_TRANSITIVE_LINKS}")
         endif()
-        cmake_language(CALL "${TARGET_NAME}_add_deployment_target" "${MODULE}" "${TARGET_NAME}" "${SOURCES}" "${DEPENDENCIES}" "${RECURSIVE_DEPENDENCIES}")
+        cmake_language(CALL "${TARGET_NAME}_add_deployment_target" "${MODULE}" "${TARGET_NAME}" "${SOURCES}" "${DEPENDENCIES}" "${TRANSITIVE_LINK_LIBRARIES}")
```
thomas-bc: Update TRANSITIVE_LINK_LIBRARIES after it's set, or read it in _add_deployment_target

thomas-bc (commented):
Need to update comments to functions that have changed

thomas-bc (commented):
What happens in the case of a build dependency that is not a link dependency, and does it need to be added to the recursive call?

cmake/test/data/test-fprime-library/cmake/autocoder/test.cmake
LeStarch: We'll fix this in a follow-up issue.  Likely with a custom target to handle byproducts. Will resolve when an issue is put in to cover this work.

LeStarch:
Another issue to add is checking that deployments have UTs.

LeStarch:
## Breaking Change Report

1. All targets used by a deployment must be defined before `register_fprime_deployment`
2. Imported targets linked PUBLICly  must be imported as `GLOBAL`
3. CMake is now installed via PIP
4. EXECUTABLE_NAME variable is deprecated
5. API to `*_setup_autocode` function gets a new build module parameter (1st)
6. `*_setup_autocode` must return AUTOCODER_BUILD_SOURCES || AUTOCODER_NEW_AUTOCODER_INPUTS

## Recommended Changes

1. Update to use new CMake call structure

cmake/autocoder/fpp.cmake
```diff
@@ -253,11 +253,14 @@ function(fpp_setup_autocode AC_INPUT_FILES)
...
                     version
         )
-endif()
+    endif()
+
     set(AUTOCODER_GENERATED ${GENERATED_AI} ${GENERATED_CPP} ${GENERATED_DICT})
     set(AUTOCODER_GENERATED "${AUTOCODER_GENERATED}" PARENT_SCOPE)
     set(AUTOCODER_DEPENDENCIES "${MODULE_DEPENDENCIES}" PARENT_SCOPE)
     set(AUTOCODER_INCLUDES "${FILE_DEPENDENCIES}" PARENT_SCOPE)
+    set(AUTOCODER_BUILD_SOURCES "${GENERATED_CPP};${GENERATED_AI};${GENERATED_DICT}" PARENT_SCOPE)
```
thomas-bc, line 262: Why are `GENERATED_AI`s and `DICT` in BUILD_SOURCES ?

cmake/target/check.cmake
```diff
@@ -69,6 +71,10 @@ function(check_add_module_target MODULE_NAME TARGET_NAME SOURCE_FILES DEPENDENCI
             COMMAND ${CMAKE_CTEST_COMMAND} --verbose
         )
     endif()
-    add_dependencies("${CHECK_TARGET_NAME}" ${UT_EXE_NAME})
-    add_dependencies(check ${UT_EXE_NAME})
+    get_target_property(DEPENDENCY_UTS "${FPRIME_CURRENT_MODULE}" FPRIME_UTS)
+    if (DEPENDENCY_UTS)
+        add_dependencies("${CHECK_TARGET_NAME}" ${DEPENDENCY_UTS})
+        # Dependencies for the check target should be added via UT directories
+        add_dependencies(check ${DEPENDENCY_UTS})
```
thomas-bc, line 78: Are we planning to eventually prefix the targets (global or all) with e.g. `fprime_check` ? Should we ?

cmake/module.cmake
```diff
@@ -10,95 +10,197 @@
...
+    # TODO: this is needed because sub-builds still attempt register targets, but without the build target to add back in the
+    #       autocoding output. Thus empty must be substituted. Would it be possible to force the library to be an INTERFACE
+    #       instead?  Or only add empty on sub-builds?
+    target_sources("${TARGET_NAME}" PRIVATE "${FPRIME__INTERNAL_EMPTY_CPP}")
+    target_link_libraries("${TARGET_NAME}" PUBLIC ${DEPENDENCIES})
+    set_target_properties("${TARGET_NAME}"
+        PROPERTIES
+            SUPPLIED_HEADERS "${HEADERS}"
+            SUPPLIED_SOURCES "${SOURCES}"
+            SUPPLIED_DEPENDENCIES "${DEPENDENCIES}"
```
thomas-bc, line 196: Supplied dependencies is never used - is that for future use?

cmake/API.cmake
```diff
@@ -138,367 +138,294 @@ function(add_fprime_subdirectory FP_SOURCE_DIR)
...
+####
+function(fprime_attach_custom_targets BUILD_TARGET_NAME)
+    setup_module_targets("${BUILD_TARGET_NAME}")
+endfunction()
+
+####
+# Function `register_fprime_library`:
 #
-# ### Non-Autocoded and Autocode-Only Modules Example ###
+# Registers a library using the fprime build system. This comes with dependency management and fprime
```
thomas-bc, line 157: Just to remind myself, is the rationale for moving away from `module` and into `library` terminology because a deployment is also a `module`?

Also if names are to be changed, should these eventually be `fprime_register_*` instead of `register_fprime_*` ?

cmake/API.cmake
```diff
@@ -138,367 +138,294 @@ function(add_fprime_subdirectory FP_SOURCE_DIR)
...
-# Some modules need to pick a specific set of dependencies and link flags. This can be done
-# with the `MOD_DEPS` variable. This feature can be used to pick specific implementations
-# for some fprime modules that implement to a generic interface like the console logger implementation.
-# 
-# ```
-# set(SOURCE_FILE
-#     MyComponent.fpp
-#     SomeFile.cpp
-#     MyComponentImpl.cpp)
+# See `register_fprime_library`. This provides the same capability as `register_fprime_library` using the
```
thomas-bc, line 197: Add warning DEPRECATED, at least in the docstring ?

cmake/API.cmake
```diff
@@ -138,367 +138,294 @@ function(add_fprime_subdirectory FP_SOURCE_DIR)
...
-    generate_library("${MODULE_NAME}" "${SOURCE_FILES}" "${MOD_DEPS}")
-    if (TARGET "${MODULE_NAME}" AND NOT MODULE_NAME MATCHES "cmake_platform.*" )
-        add_dependencies("${MODULE_NAME}" config)
-    endif()
+    register_fprime_library(${ARGN})
 endfunction(register_fprime_module)
 
 ####
-# Function `register_fprime_executable`:
+# Function `fprime_add_library_build_target`:
```
thomas-bc, line 213: Question: what use case do you envision for these calls adding a module without the autocoders ?

cmake/target/target.cmake
```diff
@@ -135,60 +179,34 @@ endfunction(setup_single_target)
 # - SOURCES: sources specified with `set(SOURCE_FILES ...)` in module's CMakeLists.txt
 # - DEPENDENCIES: dependencies and link libraries specified with `set(MOD_DEPS ...)` in module's CMakeLists.txt
 ####
-function(setup_module_targets MODULE SOURCES DEPENDENCIES)
+function(setup_module_targets BUILD_TARGET)
```
thomas-bc, line 182: `target` is super overloaded here - since this is an internal function, thoughts on renaming some of those things?

BUILD_TARGET is actually a module/library name, and `setup_module_targets` actually sets up module with "F´ targets" such as `build`. Maybe we could speak of "target factories/generators(/else?)" instead ?

thomas-bc: Maybe for a follow up ticket

thomas-bc (commented):
A couple of comments / thoughts, but overall this looks great !

cmake/autocoder/fpp.cmake
LeStarch: This is a temporary work around until we fix the ability for autocoders to generate byproducts.

cmake/target/check.cmake
LeStarch: yes, in a follow-up issue

cmake/module.cmake
LeStarch: More like for custom target and AC use.

cmake/API.cmake
LeStarch: Yes, module is more overloaded and library is clear to map to a library artifact.

Maybe we should rename, but also i don't have a deep drive to do so.

cmake/API.cmake
LeStarch: More of separation: add module, do something custom, then call autocoder setup.  `config` needs this flow.

cmake/target/target.cmake
LeStarch: Follow-up ticket.

cmake/API.cmake
LeStarch: Let's do this when we break folder targets from module targets.

thomas-bc (approved):

PR #3488, opened 2025-04-12
Add FW_PACKET_FILE descriptor to FileDownlink and FileUplink logic

### LINKED ISSUE #3467: Add packet type serialization in/out of FileDownlink/FileUplink
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Currently FileDownlink/FileUplink send/receive packets that are raw buffers. The receiver needs to process these buffers and add the `FW_PACKET_FILE` token in front.
We should modify this so that Fw::Buffer that represent file packets are directly stamped with FW_PACKET_FILE out of FileDownlink. In the same fashion, FileUplink should accept buffers that are stamped with FW_PACKET_FILE in front.

## Rationale

Raw buffers that don't have descriptors is not good as this impose additional processing on the components that are connected to it.

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/3467

Packet descriptor is now handled at reception/emission time within FileUplink/FileDownlink.

## Rationale

No custom code in routing / framing needed anymore, etc. File packets are stamped when created.

### DIFF
--- a/Svc/FileDownlink/FileDownlink.cpp
+++ b/Svc/FileDownlink/FileDownlink.cpp
@@ -363,7 +363,7 @@ void FileDownlink ::sendCancelPacket() {
     Fw::SerializeStatus status = buffer.getSerializer().serialize(Fw::ComPacket::FW_PACKET_FILE);
     FW_ASSERT(status == Fw::FW_SERIALIZE_OK);
     Fw::Buffer offsetBuffer(buffer.getData() + sizeof(FwPacketDescriptorType),
-                            buffer.getSize() - sizeof(FwPacketDescriptorType));
+                            buffer.getSize() - static_cast<Fw::Buffer::SizeType>(sizeof(FwPacketDescriptorType)));
     // Serialize the filePacket content into the buffer
     status = filePacket.toBuffer(offsetBuffer);
     FW_ASSERT(status == Fw::FW_SERIALIZE_OK);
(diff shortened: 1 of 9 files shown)


Svc/FileDownlink/FileDownlink.cpp
```diff
@@ -429,7 +430,12 @@ namespace Svc {
     filePacket.fromCancelPacket(cancelPacket);
     this->getBuffer(buffer, CANCEL_PACKET);
 
-    const Fw::SerializeStatus status = filePacket.toBuffer(buffer);
+    // Serialize the packet descriptor FW_PACKET_FILE to the buffer
+    Fw::SerializeStatus status = buffer.getSerializer().serialize(Fw::ComPacket::FW_PACKET_FILE);
+    FW_ASSERT(status == Fw::FW_SERIALIZE_OK);
+    Fw::Buffer offsetBuffer(buffer.getData() + sizeof(FwPacketDescriptorType), filePacket.bufferSize());
```
LeStarch: You are setting the buffer size here to `filePacket.bufferSize();` (just like in `sendFilePacket`), however; unlike in `sendFilePacket` there is no assert to ensure this size is correct.

You should either:
1.  Set the size of `offsetBuffer` to be `buffer.getSize() - sizeof(FwPacketDescriptor()` --or--
2. Add an assert like in `sendFilePacket`
3. Do both

I think 3 is probably the safest.

LeStarch (changes requested):
Just the one minor fix!

Svc/FileDownlink/FileDownlink.cpp
thomas-bc: Done! Option 3

LeStarch (approved):

LeStarch (approved):

thomas-bc:
Adding breaking change in that if a user were to do unexpected things with buffers coming out of FileDownlink/FileUplink, things would break. The common use cases do not break.

PR #3523, opened 2025-04-21
Fix hashes.txt generation by getting build_root

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

https://github.com/nasa/fprime-tools/pull/253 (and specifically CI runs: https://github.com/nasa/fprime-tools/actions/runs/14522278310/job/40902595131?pr=253) showed that hashes.txt wasn't being generated properly.

Hashes.txt file was not listing hashes of full (relative) paths, for example
```
Fw/Types/Assert.cpp: 0x72ad3277
```
but rather only the filename was getting hashed
```
Assert.cpp: 0x68.... 
```

[set_assert_flags()](https://github.com/thomas-bc/fprime/blob/5a5f1f66bbd7474012950598517bf532d2dd7bce/cmake/utilities.cmake#L554-L555) expects FPRIME_CLOSEST_BUILD_ROOT to be set, but it wasn't anymore

This PR fixes this by retrieving FPRIME_CLOSEST_BUILD_ROOT within `set_assert_flags()`


LeStarch (approved):

thomas-bc:
Cookiecutter CI test is failing because of wrong ordering of merges - will bypass this check. It should succeed once merged.

PR #3525, opened 2025-04-23
Revise array allocation in Frame Accumulator

bocchino:
Avoid use of placement new with array types. This is a known defect in C++11/14, per the research conducted by @LeStarch.

Because the allocator interface returns `void*`, we can use `static_cast` to convert the allocated memory to the proper type. No `reinterpret_cast` is required.

### DIFF
--- a/Svc/FrameAccumulator/FrameAccumulator.cpp
+++ b/Svc/FrameAccumulator/FrameAccumulator.cpp
@@ -4,10 +4,9 @@
 // \brief  cpp file for FrameAccumulator component implementation class
 // ======================================================================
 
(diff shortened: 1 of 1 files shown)


thomas-bc (approved):

LeStarch (approved):

LeStarch:
For anyone who stumbles on this: array placement new has a flawed specification (until it was listed as a defect in 2019) that allowed compilers the flexibility to allocate more data than supplied in the placement new call.  This could overrun pre-allocated data.

This only affects placement new of arrays, hence avoiding the pattern.

Issue #3521, opened 2025-04-21
FpySequencer V0.2


zimri-leisher:
This issue covers implementing the following features for the C++ FpySequencer component:
1. Local variables
2. If statements
3. Gotos
**PR** target June 15
4. Simple boolean operators
5. Dynamic command args
6. Telemetry db access
7. Param db access
8. 90% UT coverage
9. Integration tests
**PR** target July 31. Baseline into Ref.

Remaining:
1. Arithmetic operators
2. More boolean operators
3. Global variables
4. Arguments to sequences
5. Return values from sequence
6. Abort/exit/etc.

zimri-leisher:
@LeStarch  can you assign this to me?

zimri-leisher:
@bocchino @timcanham @LeStarch I'm going to use this thread for async discussion of the next version of Fpy. I'll ask async questions about it here.

zimri-leisher:
Decided to add a simple debugging mode to the sequencer, because I'd like it to be designed with debugging in mind.

Adding breakpoints, continue, and some inspection logic

PR #3552, opened 2025-05-03
FpySequencer 0.2

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**|  #3521 |
|**_Has Unit Tests (y/n)_**|  y|

---
### Local variables
  * Byte arrays which can store arbitrary, serialized data, much like prmdb.
  * Cleared after sequence stops.
  * These are the "workspace" of the sequence, which it can use for arguments, variables, or as a go between with prm/tlm/other sequences
  * Currently only accessed by the IF directive
  * By default, each sequencer has 16 local variables

### New commands
| | |
|:---|:---|
| Name | Description |
| DEBUG_SET_BREAKPOINT | Sets the debugging breakpoint which will pause the execution of the sequencer when reached, until unpaused by the DEBUG_CONTINUE command.  |
| DEBUG_BREAK | Pauses the execution of the sequencer, just before it is about to dispatch the next statement, until unpaused by the DEBUG_CONTINUE command. |
| DEBUG_CONTINUE | Continues the execution of the sequence after it has been paused by a debug break. |
| DEBUG_CLEAR_BREAKPOINT | Clears the debugging breakpoint, but does not continue executing the sequence. |

### New directives

| Name | Description |
|------|-------------|
| NO_OP | Does nothing |
| SET_LVAR | Sets a local variable to a byte array value |
| GOTO | Sets the index of the next statement to be executed |
| IF | Interprets an lvar as a boolean. If false, goto a specified statement index, otherwise proceed |

### Misc
* **87% UT coverage**
* A couple bugfixes


zimri-leisher:
@LeStarch this is ready for review

zimri-leisher:
I transitioned all the UTs to using test fixtures. Not sure why we aren't doing this normally. This reduces the number of needed files and boilerplate.

config/FpySequencerCfg.fpp
```diff
@@ -4,5 +4,9 @@ module Svc {
         constant MAX_SEQUENCE_ARG_COUNT = 16
         @ The maximum number of statements a sequence can have
         constant MAX_SEQUENCE_STATEMENT_COUNT = 1024
+        @ The maximum number of local variables a sequence can have
```
LeStarch: Clarifying question:

- Does this imply the storage of the sequencer's local variables is 16 * 500?

Svc/FpySequencer/docs/sdd.md
```diff
@@ -14,4 +14,29 @@ The FpySequencer accepts files in the Fpy Bytecode format, which is defined in `
...
-| CANCEL | Cancels a running or validated sequence. After running CANCEL, the sequence should return to IDLE |
\ No newline at end of file
+| CANCEL | Cancels a running or validated sequence. After running CANCEL, the sequencer should return to IDLE |
+
+## Debugging Commands
+The FpySequencer has a set of debugging commands which can be used to pause and step through sequences. They should not be necessary for nominal use cases.
+
+| Name | Description |
+|-----|-----|
+| DEBUG_SET_BREAKPOINT | Sets the debugging breakpoint which will pause the execution of the sequencer when reached, until unpaused by the DEBUG_CONTINUE command.  |
```
LeStarch, line 24: I assume these would normally not be used in flight?

Svc/FpySequencer/FpySequencerDirectives.fppi
```diff
@@ -6,6 +6,45 @@ struct WaitAbsDirective {
...
\ No newline at end of file
+internal port directive_waitAbs(directive: WaitAbsDirective) priority 6 assert
+
+internal port directive_setLocalVar(directive: SetLocalVarDirective) priority 6 assert
+
+internal port directive_goto(directive: GotoDirective) priority 6 assert
+
+internal port directive_if(directive: IfDirective) priority 6 assert
+
+internal port directive_noOp(directive: NoOpDirective) priority 6 assert
```
LeStarch: Why do we have a NO_OP directive?

Svc/FpySequencer/FpySequencerCommands.fppi
```diff
@@ -1,27 +1,55 @@
...
 async command CANCEL() \
-    opcode 3 priority 8 assert
\ No newline at end of file
+    opcode 3 priority 8 assert
+
+@ Sets the debugging breakpoint which will pause the execution of the sequencer when
+@ reached, until unpaused by the DEBUG_CONTINUE command. Will pause just before 
+@ dispatching the specified statement. This command is valid in all states. Debug 
+@ settings are cleared after a sequence ends execution.
+async command DEBUG_SET_BREAKPOINT(
```
LeStarch, line 33: Are the indexes going to be hard to calculate?

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -131,6 +131,75 @@ void FpySequencer::CANCEL_cmdHandler(FwOpcodeType opCode,  //!< The opcode
...
+                                          bool breakOnce  //!< Whether or not to break only once at this breakpoint
+) {
+    if (!this->isRunningState(this->sequencer_getState())) {
+        // can only break while running
+        this->log_WARNING_HI_InvalidCommand(static_cast<I32>(sequencer_getState()));
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
+        return;
+    }
+    this->sequencer_sendSignal_cmd_DEBUG_BREAK(
+        FpySequencer_DebugBreakpointArgs(true, breakOnce, this->m_runtime.nextStatementIndex));
```
LeStarch, line 167: Follow-Up: can we have a ships passing in the night scenario?

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -31,6 +35,21 @@ using State = FpySequencer_SequencerStateMachineStateMachineBase::State;
 
 class FpySequencer : public FpySequencerComponentBase {
   public:
+    // TODO need to be very careful about how I'm using this. question:
+    // if I set waitrel to something, then later set waitAbs to something else
+    // do I have to do any memory management in between?
+    union DirectiveUnion {
```
LeStarch, line 42: I assume this was done to save space as only one directive can be alive at a time.  However, how do you know if this data is already in use?  Is the efficiency worth it?

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -330,8 +441,27 @@ class FpySequencer : public FpySequencerComponentBase {
         // the absolute time we should wait for until returning
         // a statement response
         Fw::Time wakeupTime = Fw::Time();
+
+        // all the local variables in the sequence
+        struct LocalVariable {
+            // the value buffer of the lvar
+            U8 value[Fpy::MAX_LOCAL_VARIABLE_BUFFER_SIZE] = {};
```
LeStarch, line 457: If the local variables are all primitive types, would it be possible to use PolyType here?

Svc/FpySequencer/FpySequencer.hpp
LeStarch: It looks like this is only used in the implementation as a stack variable.

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -47,4 +68,70 @@ Signal FpySequencer::waitAbs_directiveHandler(const FpySequencer_WaitAbsDirectiv
...
+    }
+
+    U8* conditionalBuf = this->m_runtime.localVariables[directive.getconditionalLocalVarIndex()].value;
+    FwSizeType conditionalBufSize = this->m_runtime.localVariables[directive.getconditionalLocalVarIndex()].valueSize;
+    // create an esb so we don't modify the actual buf
+    Fw::ExternalSerializeBuffer conditionalEsb(conditionalBuf, conditionalBufSize);
+    Fw::SerializeStatus status = conditionalEsb.setBuffLen(conditionalBufSize);
+    FW_ASSERT(status == Fw::SerializeStatus::FW_SERIALIZE_OK);  // coding error if this fails
+    bool conditional;
+    status = conditionalEsb.deserialize(conditional);
```
LeStarch, line 126: There is no type checking here.  Perhaps localVariables should store an ID of the expected type so that we don't read say a string, interpret it as a boolean, and then make decisions on that.

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -1,28 +1,37 @@
...
     // send it to where it needs to go
     if (nextStatement.gettype() == Fpy::StatementType::DIRECTIVE) {
-        result = this->dispatchDirective(nextStatement);
+        // directives need to be deserialized first before dispatching
+        DirectiveUnion directiveUnion;
+        result = this->deserializeDirective(nextStatement, directiveUnion);
+        if (result) {
+            this->dispatchDirective(directiveUnion,
+                                    Fpy::DirectiveId(static_cast<Fpy::DirectiveId::T>(nextStatement.getopCode())));
+        }
```
LeStarch, line 34: What do you do in the failure case: i.e. a bad directive statement came in?

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -25,10 +25,10 @@ void FpySequencer::deallocateBuffer(Fw::MemAllocator& allocator) {
     this->m_sequenceBuffer.clear();
 }
 
-void FpySequencer::updateComputedCRC(const U8* buffer, FwSizeType bufferSize) {
+void FpySequencer::updateCrc(U32& crc, const U8* buffer, FwSizeType bufferSize) {
```
LeStarch, line 28: This is a nice fix.  Less state.

LeStarch (changes requested):
The biggest design issue I see is type-safety of local variables.  Right now that buffer could contain anything.  Consider adding a type enumeration to each, using PolyType, or some other update to allow for not implicitly changing types.

I'd also like to understand more about how you are using UT fixtures.

Otherwise, the implementation and changes look really solid.

zimri-leisher:
* will want to check if we serialize too large of a value in the bytecode.
* also would be nice to write all fpp constants to the dictionary

when we do lvar allocation:
* start out by assuming infinitely many of them, then do a pass to lower them down

big question: probably thinking about changing to stack based only?
no^ will need them for global variables etc


if you have functions, you probably need a function call stack. and if you have a call stack, you might as well use it
primary advantage of registers is hardware speed anyways...

Svc/FpySequencer/docs/sdd.md
zimri-leisher: They could be if a sequence is in a bad state. I just wanted to give more power to the ground to control these sequences

Svc/FpySequencer/FpySequencerDirectives.fppi
zimri-leisher: They are useful in a number of situations, especially testing, but conceivably in other places too. They don't really cost anything to have as well, as we have 2^32 possible directive ids

Svc/FpySequencer/FpySequencerCommands.fppi
zimri-leisher: Ahh, you mean as a user of the high level language, how would you know the index in the low level language?
I will add a "debug symbols" option to the bytecode which allows you to see clearly on the ground which statement index corresponds to which high-level language line number

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: Wow you got me I don't know what that is but it sounds interesting

Svc/FpySequencer/FpySequencer.hpp
zimri-leisher: Correct, I just wanted to be able to pass around an arbitrary directive by value. Happy to look at it more, I know unions are dangerous

Svc/FpySequencer/FpySequencer.hpp
zimri-leisher: They are not all primitive types. Think of a local variable like a param or a telemetry channel. They are just an arbitrary byte array which can contain an arbitrary serialized fprime value

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: Type checking will be left up to the compiler, in the same way that there is no type checking in assembly language, it's all done in the higher level language.

Open to talking more about this, but I think type safety should be the concern of the compiler, and the runtime should just deal with byte arrays

Svc/FpySequencer/docs/sdd.md
LeStarch: Agreed.

Svc/FpySequencer/FpySequencerCommands.fppi
LeStarch: Works for me!

Svc/FpySequencer/FpySequencer.hpp
LeStarch: I think it is appropriately used.  Especially since it is only used on the stack and not stored as a member.

Svc/FpySequencer/FpySequencerDirectives.cpp
LeStarch: By compiler we mean "low level code to bytecode" and not "highlevel to low level".  That plus checksum checking means that the sequence should be safe from corruption.  I am ok with that.

LeStarch:
@zimri-leisher I Agree with with your responses.  Just need to resolve the last two.  Was the last comment for this iteration (as there are no functions yet).

zimri-leisher:
@LeStarch standby on responses to your last two questions. Yes, the last comment is for the next version of FpySequencer.

@LeStarch @timcanham I'm going to put one more feature in this PR as it was rather quick to develop and adds a huge amount of functionality: getting telemetry channel time tags and values. This is a top priority for my team, I hope you won't mind the scope creep. I still think we can get this merged by May 20th.

1. Question about this: if FW_TLM_BUF_MAX_SIZE is greater than our LVAR_BUF_MAX_SIZE, should we allow it to compile or static assert? If we allow it to compile, how should we handle the case in which we get a tlm chan with a value of size greater than we can store in an lvar? Should we fail the "GET" directive or should we truncate the tlm chan? EDIT--I made it fail the "GET" directive.
2. I'm thinking of renaming local variables to "registers" to be more consistent with general compiler terminology. Thoughts?

zimri-leisher:
Second question: noticing some inconsistencies around PrmGet and TlmGet ports.

PrmDb impl of PrmGet does not modify ParamBuffer if it fails to find the prm:
image


But TlmChan impl of TlmGet does set TlmBuffer to 0:
image

I propose we have PrmDb set size to 0. Just be consistent!
Although I guess it's okay if we don't do this because PrmDb has the ParamValid enum as a return code

zimri-leisher:
Another question:
what should we do when the sequencer makes a request for an "ignored" channel? At the moment, the TlmPacketizer doesn't store the bytes of ignored channels. But this means that we may not be able to access all channels at runtime. In the future, perhaps TlmPacketizer could store the value of ignored channels, just not telemeter them to the ground.

zimri-leisher:
Another question:
What should we do when the sequencer requests a valid channel, but the channel hasn't been written into the tlm db yet? For TlmChan it returns tlmbuf of size 0, but for tlmpacketizer it returns an undefined tlmbuf/timetag.
Likely we want both to return a tlmbuf of size 0.

EDIT--I went ahead and implemented this ^

zimri-leisher:
Also while I'm at it I added GET_PRM_VAL. I apologize for the scope creep but these were features I'd already developed on a separate branch and adding them in this PR means we can actually start using this sequencer for non-trivial things! At the moment, all you can do is IF conditions with boolean tlm/prms, but that's enough to really start to hook the sequencer into the wider fprime system.

Promise that's it for scope creep.

config/FpySequencerCfg.fpp
zimri-leisher: Correct.

Svc/FpySequencer/FpySequencerRunState.cpp
zimri-leisher: a few lines down, return dispatchStatement_failure, which causes the sequence to fail

Svc/FpySequencer/FpySequencer.hpp
zimri-leisher: I'm thinking about this more. I can definitely see the benefits to having the lvars all be polytypes. Much less memory usage, and types are known to the sequencer. I'm willing to discuss this more, although it would mean reworking a decent bit of this lvar architecture I have.
I went with this architecture because it seems this is the fprime idiomatic way to do it. For params and tlm chans, you just pass around byte arrays, which are inherently not type safe, but you rely on autocode to manage the type safety. This is much the same.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: Ah, by compiler I meant highlevel to bytecode. The type safety all exists in the high level language, and it's up to the high-level compiler to keep the bytecode it produces type safe.

zimri-leisher:
Note:
TlmChan has weird behavior when calling TlmGet port on it. Because it alternates between two buffers, the value it returns from TlmGet can alternate between two values, or one value and no value. I believe this is a mistake in the implementation of TlmChan::TlmGet so I'll go ahead and fix it.

zimri-leisher:
* UT for TlmPacketizer and TlmChan
* GDS PR
* Change signature of TlmGet to have TlmValid. Also add IGNORED status perhaps?
* cut down on duplication by having one GET_TLM
* ponder TlmChan edge cases

Fw/Tlm/Tlm.fpp
```diff
@@ -13,7 +13,10 @@ module Fw {
   port TlmGet(
                $id: FwChanIdType @< Telemetry Channel ID
                ref timeTag: Fw.Time @< Time Tag
-               ref val: Fw.TlmBuffer @< Buffer containing serialized telemetry value
+               @ Buffer containing serialized telemetry value. 
+               @ Size set to 0 if channel not found, or if no value
+               @ has been received for this channel yet.
+               ref val: Fw.TlmBuffer 
              )
```
LeStarch: As discussed I expect to see a Status added her.  Good catch!

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -21,8 +21,16 @@
...
-static_assert(Svc::Fpy::MAX_SEQUENCE_STATEMENT_COUNT <= std::numeric_limits<U16>::max(), "Sequence statement count must be below U16 max");
+static_assert(Svc::Fpy::MAX_SEQUENCE_ARG_COUNT <= std::numeric_limits<U8>::max(),
+              "Sequence arg count must be below U8 max");
+static_assert(Svc::Fpy::MAX_SEQUENCE_STATEMENT_COUNT <= std::numeric_limits<U16>::max(),
+              "Sequence statement count must be below U16 max");
+static_assert(Svc::Fpy::MAX_LOCAL_VARIABLE_BUFFER_SIZE <= std::numeric_limits<FwSizeType>::max(),
+              "Local variable buffer size must be below FwSizeType max");
+// TODO should we assert on this? or should we just fail to pull in tlm buffers that are greater than
+// a certain size?
+// static_assert(Svc::Fpy::MAX_LOCAL_VARIABLE_BUFFER_SIZE >= FW_TLM_BUFFER_MAX_SIZE,
```
LeStarch: I expect to see this assert in the final, as discussed.

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -31,6 +39,24 @@ using State = FpySequencer_SequencerStateMachineStateMachineBase::State;
 
 class FpySequencer : public FpySequencerComponentBase {
   public:
+    // TODO need to be very careful about how I'm using this. question:
```
LeStarch: Remove TODO, as this looks right.

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -269,6 +375,29 @@ class FpySequencer : public FpySequencerComponentBase {
...
+    void directive_goto_internalInterfaceHandler(const Svc::FpySequencer_GotoDirective& directive) override;
+
+    //! Internal interface handler for directive_if
+    void directive_if_internalInterfaceHandler(const Svc::FpySequencer_IfDirective& directive) override;
+
+    //! Internal interface handler for directive_noOp
+    void directive_noOp_internalInterfaceHandler(const Svc::FpySequencer_NoOpDirective& directive) override;
+
+    //! Internal interface handler for directive_getTlmTime
+    void directive_getTlmTime_internalInterfaceHandler(const Svc::FpySequencer_GetTlmTimeDirective& directive) override;
```
LeStarch: Recommendation: get full time + value and just value.  Time seems less useful on its own.  A separate call has a race condition.

Svc/FpySequencer/FpySequencer.hpp
LeStarch: After discussion, I accept that type safety is handled at the compiler level.

Svc/TlmPacketizer/TlmPacketizer.cpp
```diff
@@ -228,9 +236,91 @@ void TlmPacketizer ::TlmRecv_handler(const FwIndexType portNum,
...
+            FW_ASSERT(val.setBuffLen(entryToUse->channelSize) == Fw::SerializeStatus::FW_SERIALIZE_OK);
             this->m_lock.unLock();
+            return;
         }
     }
+
+    // did not find a packet which stores this channel.
+    // coding error, this was not an ignored channel so it must be in a packet somewhere
+    FW_ASSERT(0, static_cast<FwAssertArgType>(entryToUse->id));
+    // TPP (tim paranoia principle)
```
LeStarch, line 322: To make this acronym official, add it to `expect.txt`

Svc/TlmPacketizer/TlmPacketizer.cpp
```diff
@@ -228,9 +236,91 @@ void TlmPacketizer ::TlmRecv_handler(const FwIndexType portNum,
...
+        this->missingChannel(id);
+        val.resetSer();
+        return;
+    }
+
+    for (FwChanIdType bucket = 0; bucket < TLMPACKETIZER_HASH_BUCKETS; bucket++) {
+        if (entryToUse) {
+            if (entryToUse->id == id) {  // found the matching entry
+                // check to see if the channel is ignored. If so, just return, as
+                // we don't store the bytes of ignored chans
```
LeStarch: I'd spell this out to fix spelling.

Svc/TlmPacketizer/TlmPacketizer.cpp
```diff
@@ -228,9 +236,91 @@ void TlmPacketizer ::TlmRecv_handler(const FwIndexType portNum,
...
+                if (entryToUse->ignored) {
+                    val.resetSer();
+                    return;
```
LeStarch: Expect added IGNORED status.

Svc/TlmPacketizer/TlmPacketizer.cpp
```diff
@@ -228,9 +236,91 @@ void TlmPacketizer ::TlmRecv_handler(const FwIndexType portNum,
...
+            // telemetry channel not in any packets
+            this->missingChannel(id);
+            val.resetSer();
+            return;
+        }
+    }
+
+    if (!entryToUse->hasValue) {
+        // haven't received a value yet for this entry.
+        val.resetSer();
```
LeStarch, line 292: Expect NOT_SET value here

Svc/TlmChan/TlmChan.cpp
```diff
@@ -62,23 +64,58 @@ void TlmChan::TlmGet_handler(FwIndexType portNum, FwChanIdType id, Fw::Time& tim
...
+    if (activeEntry && inactiveEntry) {
+        // two entries. grab the one with the most recent time tag
+        if (Fw::Time::compare(inactiveEntry->lastUpdate, activeEntry->lastUpdate) == Fw::Time::Comparison::GT) {
+            // inactive entry is more recent
+            val = inactiveEntry->buffer;
+            timeTag = inactiveEntry->lastUpdate;
+            return;
+        } else {
+            // active entry is more recent, or they are equal
+            val = activeEntry->buffer;
```
LeStarch, line 105: In the case of UNCOMPARIBLE, both are valid, but no way to infer which is newer.   Thus one was just picked. 

Could the updated flag be used? If one is `updated` and the other is not, that means it has new data.   If both are updated or not, then choose one.

It is a question, I am not an expert on this code.

Svc/FpySequencer/test/ut/FpySequencerTester.cpp
```diff
@@ -13,224 +15,276 @@ namespace Svc {
...
+void FpySequencerTester::allocMem(FwSizeType bytes) {
+    if (bytes > sizeof(internalSeqBuf)) {
+        bytes = sizeof(internalSeqBuf);
     }
+    cmp.m_sequenceBuffer.setExtBuffer(internalSeqBuf, bytes);
+}
+
+void FpySequencerTester::assertQueueMsg(FwEnumStoreType msg) {
+    // TODO I would like to write this function but I can't have access to the
+    // cmpipcbuf and msgtypeenum
```
LeStarch: sp

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -47,4 +86,166 @@ Signal FpySequencer::waitAbs_directiveHandler(const FpySequencer_WaitAbsDirectiv
...
+        // could not find this prm in the DB
+        return Signal::stmtResponse_failure;
+    }
+
+    if (prmValue.getBuffLength() > Fpy::MAX_LOCAL_VARIABLE_BUFFER_SIZE) {
+        // cannot store the prm value in the lvar
+        return Signal::stmtResponse_failure;
+    }
+    // copy value into lvar
+    Runtime::LocalVariable& lvar = this->m_runtime.localVariables[directive.getdestLvarIndex()];
```
LeStarch, line 217: Seems like all these copy/sets could be a helper.  This would save very similat code.

```
helper (FwSize index, const U8* buffer, const FwSizeType size) {
   FW_ASSERT buffer != null
    Runtime::LocalVariable& lvar = this->m_runtime.localVariables[index];
     memcpy(lvar.value, buffer,size);
    lvar.valueSize = size;
}
``

**Review by LeStarch — CHANGES_REQUESTED** [review]

**Comment by LeStarch** [comment]
@zimri-leisher added some comments.  Almost all were repeating your above todos.

You should also merge in the base branch, as configuration changed.

File: Svc/TlmPacketizer/TlmPacketizer.cpp (hunk 17 above)
Comment (zimri-leisher) [reply Svc/TlmPacketizer/TlmPacketizer.cpp]: gonna make the executive decision not to enumerate the status beyond VALID or INVALID for now. It's all I need for my use case, and it leaves the door open for more specific status codes

File: Svc/TlmChan/TlmChan.cpp (hunk 19 above)
Comment (zimri-leisher) [reply Svc/TlmChan/TlmChan.cpp:105]: Good point, I think that works. I made the fix

**Comment by zimri-leisher** [comment]
@LeStarch all feedback is complete I believe. Ready for merge

**Review by LeStarch — APPROVED** [review]
I see the expected changes!

@timcanham would you please review the changes to Svc/Tlm*

Also, please agree/disagree with the algorithm of "comparable":
1. If time comparable, return newest by time
2. If time incomparable, return "updated"
3. If neither updated the return "active" buffer for lack of a better choice.

Note: this algorithm only runs when both buffers are valid.

File: Fw/Tlm/Tlm.fpp [hunk 22]
```diff
@@ -13,7 +18,10 @@ module Fw {
   port TlmGet(
                $id: FwChanIdType @< Telemetry Channel ID
                ref timeTag: Fw.Time @< Time Tag
-               ref val: Fw.TlmBuffer @< Buffer containing serialized telemetry value
-             )
+               @ Buffer containing serialized telemetry value. 
+               @ Size set to 0 if channel not found, or if no value
+               @ has been received for this channel yet.
+               ref val: Fw.TlmBuffer 
+             ) -> Fw.TlmValid
```
timcanham, line 25: Add a note describing what conditions return `INVALID`

Svc/TlmPacketizer/TlmPacketizer.cpp
```diff
@@ -228,11 +236,94 @@ void TlmPacketizer ::TlmRecv_handler(const FwIndexType portNum,
...
+            this->m_lock.unLock();
+            return Fw::TlmValid::VALID;
+        }
+    }
+
+    // did not find a packet which stores this channel.
+    // coding error, this was not an ignored channel so it must be in a packet somewhere
+    FW_ASSERT(0, static_cast<FwAssertArgType>(entryToUse->id));
+    // TPP (tim paranoia principle)
+    val.resetSer();
```
timcanham, line 323: If the assert happens, it can't reach this code.

timcanham (approved):
Minor findings that could be done on the next release, but nothing broken that I can see.

Svc/TlmPacketizer/TlmPacketizer.cpp
LeStarch: That is the TPP, asserts could be disabled.

PR #3562, opened 2025-05-06
UT: Do not add `-g` flag if `CMAKE_BUILD_TYPE` is set to "Release"

pcrosemurgy:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

By default when CMake testing is enabled `CMAKE_BUILD_TYPE` is set to "Debug". This PR will not break any backwards compatibility.

If, in a deployment's top-level CMakeLists.txt, `CMAKE_BUILD_TYPE` is explicitly set to "Release" this should be respected and the debug flag should be omitted

## Rationale

This is needed in order to work around compile times for AC files with very long `toString()` impls (for example, AC array data types with thousands of args). When `-g` is used, the UT builds fail to complete (that is, fail to complete within any human/reasonable time).

### DIFF
--- a/cmake/settings.cmake
+++ b/cmake/settings.cmake
@@ -23,5 +23,12 @@ set(CMAKE_C_EXTENSIONS OFF)
 
 # fprime unit testing methodology requires the following flags
 if (BUILD_TESTING)
-    add_compile_options(-g -DBUILD_UT -DPROTECTED=public -DPRIVATE=public -DSTATIC=)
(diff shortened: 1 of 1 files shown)


LeStarch (approved):
This looks right, although I wonder if the -g is needed at all.  CMake Debug should add that for us.

pcrosemurgy:
Closed, https://github.com/nasa/fpp/issues/689 will fix our underlying issue.

PR #3566, opened 2025-05-07
Clarify allocation failure error

celskeggs:
| | |
|:---|:---|

---
## Change Description

Add a new queue error type for allocation errors.

## Rationale

It's not useful to report an UNKNOWN_ERROR when the cause is known.

N/A

## Future Work

N/A


LeStarch (changes requested):
Would you add a check to this block: https://github.com/nasa/fprime/blob/e2bd5f7dc77b893bf6edebaa26def04bb88a3827/Os/Models/Models.cpp#L186

to ensure the FPP model enum value matches the raw enum.  Once done, we can merge.

LeStarch:
@celskeggs went ahead and made the change.  It was small.

LeStarch (approved):

thomas-bc (approved):

PR #3567, opened 2025-05-07
Improve one possible error on Svc.TlmChan bucket exhaustion

celskeggs:
| | |
|:---|:---|

---
## Change Description

Reduces the chance that we get a confusing error message in TlmChan, like the following, when we run out of telemetry buckets:

```
Assert: "/workspaces/[...]/lib/fprime/Svc/TlmChan/TlmChan.cpp:128"
```

This assertion comes from the following code:

```
    // copy into entry
    FW_ASSERT(entryToUse);
```

With this change, the same scenario will result in this assertion tripping instead, which is more easily understood:

```
        // Make sure that we haven't run out of buckets
        FW_ASSERT(this->m_tlmEntries[this->m_activeBuffer].free < TLMCHAN_HASH_BUCKETS);
```

## Rationale

Makes it easier to understand why we encountered an assertion in TlmChan.

To reproduce the error:

1. Set TLMCHAN_NUM_TLM_HASH_SLOTS to 1 and TLMCHAN_HASH_BUCKETS to 1.
2. Report 2 telemetry items.

## Future Work

N/A


timcanham:
@celskeggs The proposed fix in the comments don't match the file changes. Can you elaborate?

celskeggs:
@timcanham

Normally, if we run out of buckets, `entryToUse` will be null, so the code on [lines 106-115](https://github.com/nasa/fprime/pull/3567/files#diff-2c32a6842d2453c6af9521f0bae3faf5b907efa3b225530db496ca6ce2e09d7eL105-L114) will run:

```
                // Make sure that we haven't run out of buckets
                FW_ASSERT(this->m_tlmEntries[this->m_activeBuffer].free < TLMCHAN_HASH_BUCKETS);
                // add new bucket from free list
                entryToUse =
                    &this->m_tlmEntries[this->m_activeBuffer].buckets[this->m_tlmEntries[this->m_activeBuffer].free++];
                FW_ASSERT(prevEntry);
                prevEntry->next = entryToUse;
                // clear next pointer
                entryToUse->next = nullptr;
                break;
```

If we are indeed out of buckets, the assertion on line 107 will trip:

```
FW_ASSERT(this->m_tlmEntries[this->m_activeBuffer].free < TLMCHAN_HASH_BUCKETS);
```

However, if `entryToUse` doesn't reach null until the end of the TLMCHAN_HASH_BUCKETS-th iteration of the loop, then there won't be an additional iteration for this branch to execute, and execution will fall through to the assertion on [lines 127-128](https://github.com/nasa/fprime/pull/3567/files#diff-2c32a6842d2453c6af9521f0bae3faf5b907efa3b225530db496ca6ce2e09d7eL127-L128):

```
    // copy into entry
    FW_ASSERT(entryToUse);
```

This path is less clear. By adding one additional iteration to the loop, we ensure that we cannot fall through to this second assertion purely due to exhausting all available buckets. Instead, the earlier, clearer assertion will have a chance to trip in the final (TLMCHAN_HASH_BUCKETS+1)-th iteration.

timcanham (approved):
OK - Had to study the code again after a while, but I see your point and agree.  As long as unit tests still pass, I'm good.

thomas-bc:
CI was failing because of a change in `devel` that wasn't in this branch here. Merged `devel`

PR #3574, opened 2025-05-09
Removing CMake from installation guide.  Fixes #3501

### LINKED ISSUE #3501: CMake Refactor: Update Install Instructions to Remove CMake
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

Remove CMake from install instructions.  It is now handled by PIP!

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Removing CMake from installation guide.

## Rationale

CMake is installed through PIP now, so it isn't a specific installation step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,8 @@ Learn more about [F´ key features](https://fprime.jpl.nasa.gov/overview).
 
 1. Linux, Windows with WSL, or macOS operating system
 2. [git](https://git-scm.com/)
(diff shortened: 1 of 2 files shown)


thomas-bc (changes requested):
We probably should remove it from the README's System requirements too ?

https://github.com/nasa/fprime/blob/devel/README.md

LeStarch:
@thomas-bc good catch!  DONE

thomas-bc (approved):

PR #3600, opened 2025-05-13
Fix memory leak in ComStub when reads fail

celskeggs:
| | |
|:---|:---|

---
## Change Description

drvReceiveIn passes ownership of recvBuffer, but ComStub only took care of deallocating the buffer if the read operation succeeded. If a communications driver indicates that a read did not succeed (for example with status RECV_NO_DATA), the buffer is lost and memory is leaked.

## Rationale

Memory leaks cause FSW to fail to operate correctly.

This should probably be unit tested to prevent regressions, but that's not covered in this PR.

## Future Work

It would be worth thinking about how we could avoid these sorts of buffer handling problems in the future.


LeStarch:
@thomas-bc this looks right to me, but I'd prefer you to review.

thomas-bc (approved):
Uh oh! 👀 
Good catch, thank you!

Issue #3605, opened 2025-05-14
Support overriding the time base in the F Prime GDS


celskeggs:
| | |
|:---|:---|
|**_F´ Version_**| devel |
|**_Affected Component_**| GDS |
---
## Feature Description

I'm using a target device that doesn't have a RTC, so I am not able to log channel values using wall clock time. Instead, I use the Svc.OsTime component with Os.RawTime. This means that the GDS sees the times as being on 1970-01-01. This causes problems with graphing.

I request that we either:

1. Support a configuration option to change the time base using a linear shift, OR
2. Add plugin support for providing custom interpretation of times.

## Rationale

Better support for embedded devices without RTCs.

timcanham:
@celskeggs You can use a time base of TB_PROC_TIME:

https://github.com/nasa/fprime/blob/devel/default/config/FpConfig.h#L31

The GDS will just display the time as the upper and lower raw values instead of 1970+. I do agree with providing for plug-ins based on the time base.

celskeggs:
@timcanham Are you _sure_ this is rendered correctly by the GDS? My testbed shift just ended, so I can't check again, but I'm pretty sure I was seeing the 1970+ date, and I _was_ using TB_PROC_TIME.

timcanham:
@celskeggs I admit I'm going from memory, but that's what I remembered. I'll check when I get a chance.

celskeggs:
Just checked the data lingering in my GDS tab. For a particular channel, I could see:

`_datastore.channels[0x4601].time.base` is `1` (TB_PROC_TIME)
`_datastore.channels[0x4601].time.seconds` is `58`
`_datastore.channels[0x4601].time.microseconds` is `608490`

And it was displayed like this:

![Image](https://github.com/user-attachments/assets/d4dbb40e-27c9-44dc-a985-e1b7fb49232f)

Now, it _was_ correctly rendered in the channel log:

`TB_PROC_TIME: 58.608490s`

And that makes sense, since the `timeToString` implementation does not consider TB_PROC_TIME:
https://github.com/nasa/fprime-gds/blob/ba05c6d2059506cfd248292026010136aff15c1b/src/fprime_gds/flask/static/js/vue-support/utils.js#L100-L125

timcanham:
I stand corrected then... I wonder if the handling has changed.

celskeggs:
[...]

PR #3607, opened 2025-05-14
Add command to OsTime to set the current time

celskeggs:
| | |
|:---|:---|

---
## Change Description

Add a command to Svc.OsTime to set the current time.

## Rationale

The current implementation of F Prime GDS only allows for charting telemetry channels if the time associated with those channels is close to the current time. See https://github.com/nasa/fprime/issues/3605. In order to allow channels to be graphed on systems using OsTime, we need the ability to adjust the epoch so that the time is close to the current time. This cannot be hard-coded, as the hard-coded time would only work for less than a day (due to zoom limits), and then a rebuild would be required. So this information needs to be passed from the ground somehow, and the easiest way to do this is via a command.

N/A

## Future Work

Hopefully https://github.com/nasa/fprime/issues/3605 will be resolved favorably.

### DIFF
--- a/Svc/OsTime/OsTime.cpp
+++ b/Svc/OsTime/OsTime.cpp
@@ -33,6 +33,19 @@ void OsTime::set_epoch(const Fw::Time& fw_time, const Os::RawTime& os_time) {
     m_epoch_valid = true;
 }
 
+void OsTime::SetCurrentTime_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, U32 seconds_now) {
+    Os::RawTime time_now;
+    Os::RawTime::Status stat = time_now.now();
+    if (stat != Os::RawTime::OP_OK) {
+        this->log_WARNING_HI_SetCurrentTimeError(stat);
+        return;
+    }
+    Os::ScopeLock lock(m_epoch_lock);
+    m_epoch_fw_time = Fw::Time(seconds_now, 0);
+    m_epoch_os_time = time_now;
+    m_epoch_valid = true;
+}
+
 // ----------------------------------------------------------------------
 // Handler implementations for user-defined typed input ports
 // ----------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


LeStarch (approved):
Looks good to me.  @kubiak-jpl do you agree?

Svc/OsTime/OsTime.cpp
```diff
@@ -33,6 +33,19 @@ void OsTime::set_epoch(const Fw::Time& fw_time, const Os::RawTime& os_time) {
     m_epoch_valid = true;
 }
 
+void OsTime::SetCurrentTime_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, U32 seconds_now) {
+    Os::RawTime time_now;
+    Os::RawTime::Status stat = time_now.now();
+    if (stat != Os::RawTime::OP_OK) {
+        this->log_WARNING_HI_SetCurrentTimeError(stat);
```
kubiak-jpl, line 40: I think there's a chance for a recursive error here. If `time_now.now()` returns a bad result and the EVR gets sent, the Ac implementation for this EVR will try to get the time again, which will likely fail. If that's the case, then it might be better to hardcode a null timestamp

celskeggs: This is a command handler, not the time handler.

kubiak-jpl: Yes, but the time handler will be called when the EVR is generated and will probably return ZERO_TIME as well. Does the AC code interpret this value specially or will it just pass it through?

celskeggs: The zero time is just passed through like any other time.

LeStarch: Yes, zero time should just go through, so I don't believe there is a loop here.

PR #3620, opened 2025-05-15
Making configuration files trigger configure

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

This marks input sources to be cmake configure step drivers.


thomas-bc (approved):
Confirmed it does fix the issue! Thanks a lot

Issue #3445, opened 2025-04-03
UDP Doesn't Support Responses to Received Data


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

The `Udp` component does not allow configuration of the response port based on data received.

e.g. Udp reads from established port.  The incoming packet can inform Udp of the response port.  However, Udp does not read or use this data.  This case would not configure a send port but would just send data back.

e.g. Udp sends to a know port, and should be able to read responses back on an ephemeral port.  This case would not call `bind` to listen, but could listen after a send.

## Expected Behavior

Udp should support ephemeral port.

PR #3691, opened 2025-06-04
Adding support for UDP ephemeral ports

vincewoo:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3445 |

---
## Change Description

This change adds the ability to support ephemeral ports when using UdpSocket.

In the sender case, you could set up your receive port as 0 to be assigned an ephemeral port for responses.
In the receiver case, you will be able to reply to the port that is specified in the UDP datagram of the first received message.

## Rationale

This improves F' compliance to the UDP specification and allows for the use of ephemeral ports.

Added test case to TestUdp:

TestEphemeralPorts - Set up a receiver specifying 0 for the port to be assigned an ephemeral one. Setup a sender that is full duplex (with an ephemeral receive port). Send a message from the sender and ensure it is received properly by the receiver. Send a response back from the receiver and ensure that it is received properly by the sender.

## Future Work

In the current implementation it only supports responding to messages from a single sender. If we need to support being able to reply to multiple senders this would require a different design.


Drv/Ip/UdpSocket.cpp
```diff
@@ -199,7 +199,21 @@
...
+
+    struct sockaddr_in sender_addr;
+    socklen_t sender_addr_len = sizeof(sender_addr);
+    I32 received = static_cast<I32>(::recvfrom(socketDescriptor.fd, data, size, SOCKET_IP_RECV_FLAGS,
+                                              reinterpret_cast<struct sockaddr*>(&sender_addr), &sender_addr_len));
+    // If we have not configured a send address, set it to the sender of the last received packet
+    if (received > 0 && this->m_state->m_addr_send.sin_port == 0) {
+        this->m_state->m_addr_send = sender_addr;
+
+        char ip_str[INET_ADDRSTRLEN] = {0};
```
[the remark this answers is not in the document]
vincewoo: Switched to using CHAR

Drv/Ip/UdpSocket.cpp
```diff
@@ -199,7 +199,21 @@
...
+        ::inet_ntop(AF_INET, &(sender_addr.sin_addr), ip_str, INET_ADDRSTRLEN);
+        this->configureSend(ip_str, ntohs(sender_addr.sin_port), 0, 100);
```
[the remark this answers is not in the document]
vincewoo: Added a FW_ASSERT for checking returned status

Drv/Ip/UdpSocket.cpp
```diff
@@ -199,7 +199,21 @@
...
+    struct sockaddr_in sender_addr;
+    socklen_t sender_addr_len = sizeof(sender_addr);
+    I32 received = static_cast<I32>(::recvfrom(socketDescriptor.fd, data, size, SOCKET_IP_RECV_FLAGS,
+                                              reinterpret_cast<struct sockaddr*>(&sender_addr), &sender_addr_len));
+    // If we have not configured a send address, set it to the sender of the last received packet
+    if (received > 0 && this->m_state->m_addr_send.sin_port == 0) {
+        this->m_state->m_addr_send = sender_addr;
+
+        char ip_str[INET_ADDRSTRLEN] = {0};
+        ::inet_ntop(AF_INET, &(sender_addr.sin_addr), ip_str, INET_ADDRSTRLEN);
```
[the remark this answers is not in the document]
vincewoo: Adding check for return value.

Drv/Ip/UdpSocket.hpp
```diff
@@ -122,8 +123,9 @@
     I32 recvProtocol(const SocketDescriptor& socketDescriptor, U8* const data, const U32 size) override;
   private:
     SocketState* m_state; //!< State storage
-    U16 m_recv_port;  //!< IP address port used
-    char m_recv_hostname[SOCKET_MAX_HOSTNAME_SIZE];  //!< Hostname to supply
+    U16 m_recv_port;  //!< Port to receive on
+    char m_recv_hostname[SOCKET_MAX_HOSTNAME_SIZE]; //!< Hostname to receive on
```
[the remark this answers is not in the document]
vincewoo: Updating to use CHAR

Drv/Ip/UdpSocket.cpp
```diff
@@ -199,7 +197,23 @@ I32 UdpSocket::sendProtocol(const SocketDescriptor& socketDescriptor, const U8*
...
+    I32 received = static_cast<I32>(::recvfrom(socketDescriptor.fd, data, size, SOCKET_IP_RECV_FLAGS,
+                                              reinterpret_cast<struct sockaddr*>(&sender_addr), &sender_addr_len));
+    // If we have not configured a send address, set it to the sender of the last received packet
+    if (received > 0 && this->m_state->m_addr_send.sin_port == 0) {
+        this->m_state->m_addr_send = sender_addr;
+
+        CHAR ip_str[INET_ADDRSTRLEN] = {0};
+        const CHAR* address = ::inet_ntop(AF_INET, &(sender_addr.sin_addr), ip_str, INET_ADDRSTRLEN);
+        FW_ASSERT(address != nullptr);
+        SocketIpStatus confSendStatus = this->configureSend(ip_str, ntohs(sender_addr.sin_port), 0, 100);
```
LeStarch: Store and forward the time out values (0, 100)

LeStarch (commented):
Add update and store and forward

LeStarch:
@zimri-leisher change to UDP, would you take a look?

Drv/Ip/UdpSocket.cpp
vincewoo: Reverted configureSend to do the setting of the timeout values. This section was refactored to simply update the internal state with the new port received.

Drv/Ip/UdpSocket.hpp
```diff
@@ -122,8 +123,9 @@
     I32 recvProtocol(const SocketDescriptor& socketDescriptor, U8* const data, const U32 size) override;
   private:
     SocketState* m_state; //!< State storage
-    U16 m_recv_port;  //!< IP address port used
-    char m_recv_hostname[SOCKET_MAX_HOSTNAME_SIZE];  //!< Hostname to supply
+    U16 m_recv_port;  //!< Port to receive on
+    CHAR m_recv_hostname[SOCKET_MAX_HOSTNAME_SIZE]; //!< Hostname to receive on
```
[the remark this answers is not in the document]
vincewoo: Changed to CHAR

Drv/Ip/UdpSocket.cpp
```diff
@@ -199,7 +199,21 @@
 
 I32 UdpSocket::recvProtocol(const SocketDescriptor& socketDescriptor, U8* const data, const U32 size) {
     FW_ASSERT(this->m_state->m_addr_recv.sin_family != 0); // Make sure the address was previously setup
-    return static_cast<I32>(::recvfrom(socketDescriptor.fd, data, size, SOCKET_IP_RECV_FLAGS, nullptr, nullptr));
+
+    struct sockaddr_in sender_addr;
+    socklen_t sender_addr_len = sizeof(sender_addr);
```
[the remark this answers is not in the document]
vincewoo: Changed to U32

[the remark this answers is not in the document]
vincewoo: Changed to U32

Drv/Ip/UdpSocket.cpp
```diff
@@ -107,7 +107,7 @@ SocketIpStatus UdpSocket::bind(const PlatformIntType fd) {
         return SOCK_FAILED_TO_BIND;
     }
 
-    socklen_t size = sizeof(address);
+    U32 size = sizeof(address);
```
LeStarch: Why use a `U32` here as opposed to an `FwSizeType`?

LeStarch (dismissed):
One minor question, but otherwise looks good.

Drv/Ip/UdpSocket.cpp
vincewoo: You're right, FwSizeType is more appropriate here. I will change it.

vincewoo: I actually had to revert to back to socklen_t, it was originally changed to U32 in response to the static analyzer. I looked into it further and actually socklen_t is the most correct to ensure support for all platforms.

LeStarch: It is.  We also allow the use of API defined  types here because it does not leave this component.

LeStarch (approved):

Drv/Ip/UdpSocket.hpp
```diff
@@ -122,8 +123,9 @@ class UdpSocket : public IpSocket {
     I32 recvProtocol(const SocketDescriptor& socketDescriptor, U8* const data, const U32 size) override;
   private:
     SocketState* m_state; //!< State storage
```
zimri-leisher, line 125: this is confusing, but maybe not within the scope of this PR to fix. Why is there a SocketState which stores send/recv addr/port, a U16 which also stores recv port, and a char buf which also stores recv hostname? Maybe some comments to make it clear which internal state is meant to be used where, under which circumstances.

zimri-leisher:
Aside from the confusing internal state of this component, this PR looks good to me. My highest priority is that this not change any existing behavior, and it doesn't appear to.

Drv/Ip/UdpSocket.hpp
vincewoo: Thanks for the feedback, you're right there is actually no reason to keep a separate copy of these outside of m_state. I would suggest that we refactor the code and remove m_recv_port and m_recv_hostname and manipulate the m_state directly.

@LeStarch Would you like me to handle that within this PR or would you prefer that to be split off into another issue?

zimri-leisher: If I may ask, what is the reason for there to be a separate "state" struct? Why not just make them member variables?

vincewoo: I don't have a strong preference either way. You're right, the SocketState struct is only used within the context of UdpSocket so it's an added layer of indirection doesn't buy us much as far as I can tell. The only slight (maybe) benefit of maintaining the `m_state` struct over new members is a smaller refactor size overall.

zimri-leisher: I bias towards fewer layers of indirection rather than more, so if it were up to me I would remove m_state. Member variables are already state, no need to put them in a struct and call them state!

LeStarch:
Thanks @zimri-leisher see https://github.com/nasa/fprime/issues/3716 for follow-up.

PR #718, opened 2021-06-15
Remove StructSerializable

Joshua-Anderson:
| | |
|:---|:---|
|**_Originating Project/Creator_**| |
|**_Affected Component_**|  Fw/Types |
|**_Affected Architectures(s)_**|  |
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|  |
|**_Builds Without Errors (y/n)_**|  |
|**_Unit Tests Pass (y/n)_**|  |
|**_Documentation Included (y/n)_**|  |

---
## Change Description

Remove StructSerializable class and tests.

## Rationale

StructSerializable is designed to support serializing already defined C++ structs. It and it's associated unit tests are currently disabled and compiled as part of F'. It uses templates, which we currently are not allowing within Fw/Type, is documented as non-functional, and hasn't been touched in 4 years. We should implement support for serializing structs within F' or we should remove this code.

PR #3692, opened 2025-06-04
Add constants to dictionary spec

jwest115:
| | |
|:---|:---|
|**_Related Issue(s)_**| [#718](https://github.com/nasa/fpp/issues/718) |

---
## Change Description

Updated JSON dictionary spec to include constants and constant values. Also cleaned up Array, Enum, and Struct section to reference array values, enum values, and struct values.

## Rationale

Spec updates were needed since we are adding some constants to the dictionary in https://github.com/nasa/fpp/pull/729.

### DIFF
--- a/docs/reference/fpp-json-dict.md
+++ b/docs/reference/fpp-json-dict.md
@@ -15,6 +15,7 @@ This document describes the format of FPP JSON dictionaries.
     - [String Type Descriptors](#string-type-descriptors)
         - [String Types](#string-types)
     - [Qualified Identifier Type Descriptors](#qualified-identifier-type-descriptors)
+- [Constants](#constants)
 - [Type Definitions](#type-definitions)
     - [Array Type Definition](#array-type-definition)
     - [Enumeration Type Definition](#enumeration-type-definition)
@@ -29,6 +30,7 @@ This document describes the format of FPP JSON dictionaries.
     - [Boolean Values](#boolean-values)
     - [String Values](#string-values)
     - [Array Values](#array-values)
+    - [Constant Values](#constant-values)
     - [Enumeration Values](#enumeration-values)
     - [Struct Values](#struct-values)
     - [Invalid Values](#invalid-values)
@@ -175,6 +177,39 @@ Example JSON of qualified name
 }
 ```
 
+## Constants
+
+| Field | Description | Options | Required | 
+| ----- | ----------- | ------- | -------- |
+| `kind` | The kind of type | `constant` | true |
+| `qualifiedName` | Fully qualified name of element in FPP model | Period-separated **String** | true |
+| `type` | The type of the constant value | **[Type Descriptor](#type-descriptors)** | true
+| `value` | Value associated with the constant | **[Constant Value](#constant-values)** | true |
+| `annotation` | User-defined annotation | **String** | false |
+
+Example FPP model with JSON representation:
+```
+module M1 {
+  @ Constant with value 1
+  constant C = 1
+}
+```
+
+```json
+{
+  "kind" : "constant",
+  "qualifiedName" : "M1.C",
+  "type" : {
+    "name" : "U64",
+    "kind" : "integer",
+    "size" : 64,
+    "signed" : false
+  },
+  "value" : 1,
+  "annotation" : "Constant with value 1"
+}
+```
+
 ## Type Definitions
 
 ### Array Type Definition
@@ -185,7 +220,7 @@ Example JSON of qualified name
 | `qualifiedName` | Fully qualified name of element in FPP model | Period-separated **String** | true |
 | `size` | Size of the data structure | **Number** | true |
 | `elementType` | The type of the array's elements | **[Type Descriptor](#type-descriptors)** | true
-| `default` | Default array value | Value of type specified in `elementType` | true |
+| `default` | Default array value |  **[Array Value](#array-values)**  | true |
 | `annotation` | User-defined annotation | **String** | false |
 
 Example FPP model with JSON representation:
@@ -282,7 +317,7 @@ module M1 {
 | `kind` | The kind of type | `struct` | true |
 | `qualifiedName` | Fully qualified name of element in FPP model | Period-separated **String** | true |
 | `members` | The members of the struct | JSON dictionary of Member Name (key) to [Struct Member Descriptor](#struct-member-descriptor) (value) | true |
-| `default` | The default value of the struct | JSON dictionary of Member Name (key) to default value (value) | true |
+| `default` | The default value of the struct | JSON dictionary of Member Name (key) to **[Struct Value](#struct-values)** (value) | true |
 | `annotation` | User-defined annotation | **String** extracted from FPP model | false |
 
 #### Struct Member Descriptor
@@ -452,6 +487,10 @@ Example JSON of an array of type U32 consisting of 10 elements
 ```json
 [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
 ```
+
+### Constant Values
+Constant values include [Primitive Integer](#primitive-integer-values), [Floating-Point](#floating-point-values), 
+[String](#string-values), and [Boolean](#boolean-values) values.
 
 ### Enumeration Values
 String qualified identifier name of enumeration value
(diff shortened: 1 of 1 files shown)


docs/reference/fpp-json-dict.md
```diff
@@ -175,6 +177,39 @@ Example JSON of qualified name
 }
 ```
 
+## Constants
+
+| Field | Description | Options | Required | 
+| ----- | ----------- | ------- | -------- |
+| `kind` | The kind of type | `constant` | true |
```
LeStarch: Perhaps we ought to define "kind" in this context.  I don't know what this is.  I thought "kind" was for components.  Looks like it is just `constant`.  Probably should say that in the description.

docs/reference/fpp-json-dict.md
```diff
@@ -175,6 +177,39 @@ Example JSON of qualified name
...
+| `qualifiedName` | Fully qualified name of element in FPP model | Period-separated **String** | true |
+| `type` | The type of the constant value | **[Type Descriptor](#type-descriptors)** | true
```
LeStarch, line 186: I thought constants lacked types.  Are they always U64/I64 or how is this calculated?

LeStarch (commented):
Some basic questions/suggestions

docs/reference/fpp-json-dict.md
```diff
@@ -453,6 +488,10 @@ Example JSON of an array of type U32 consisting of 10 elements
 [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
 ```
 
+### Constant Values
+Constant values include [Primitive Integer](#primitive-integer-values), [Floating-Point](#floating-point-values), 
+[String](#string-values), and [Boolean](#boolean-values) values.
```
thomas-bc, line 497: Should the spec call out which specific constants are pulled into the dictionary? Or maybe just the rationale on why some are and others aren't?

jwest115: I was thinking of adding that to the FPP user guide under the fpp-to-dict tool section.

docs/reference/fpp-json-dict.md
jwest115: The dictionary only outputs the constants that are required by the dictionary, which so far, are only integers. If the integer is negative, the type will be `I64`, otherwise it will be `U64`.

docs/reference/fpp-json-dict.md
jwest115: I will update the description.

jwest115: After talking with Rob, I will remove the kind field from constants entirely since constants are in their own category.

docs/reference/fpp-json-dict.md
thomas-bc: Makes sense to me!

docs/reference/fpp-json-dict.md
thomas-bc: I think what Michael was pointing out is that constants do not carry any type information, neither in FPP nor in the C++ layer

```
    enum FppConstant_QUEUE_SIZE {
      QUEUE_SIZE = 10
    };
```

My 2-minute googling understanding is that the compiler is free to choose what it likes to represent this, and will most likely choose `int`, which is not necessarily `I64`. 
Does it make sense to include type in the dictionary for constants?

jwest115: I think it might be better to leave the type of constants out of the dictionary because it seems inaccurate. I was thinking we would need the type information for the GDS, but maybe since constants required by the dictionary are always integers that is sufficient (ie: we know that any constants in the dictionary are integers so the type info is not needed).

bocchino: The purpose of assigning the type U64 or I64 is to say that the constant is to be interpreted as an integer value. We use U64 or I64 because those are the types that (1) are available in the GDS dictionary spec and (2) are big enough to hold any integer value. Other things we could do:

1. Add an Integer type that can hold any integer value, without specifying the signedness or bit width. This is how FPP represents integer constants internally. For the dictionary, though, I'm not sure this is needed just to specify the type of a constant; using U64 or I64 as a stand-in for Integer seems OK.
2. Compute a narrower integer type based on the constant value. E.g., the type of -1 would be I8, and the type of 65536 would be U32. I'm not sure this is necessary either, if the constants are used in Python computations where the bit width is not important. Bit width is important when serializing values and storing them in memory.

> I think what Michael was pointing out is that constants do not carry any type information

Integer constants do carry the info that they are integers. In the code you cited above, if a constant is defined as a C++ enum, then it must be an integer constant. So far we have only integer constants in the dictionary, but I think it's good to be future-proof. E.g., we may want to add floating-point constants in the future.

bocchino: Another option would be to infer the type from the form of the literal. This is how it's done in the FPP source. For example, if you see
```
constant a = 3
```
then you know that the type of a is `Integer`, and if you see
```
constant b = "abc"
```
then you know that the type of b is `string`. Maybe we could do this in the dictionary too. Is this what you are suggesting, Thomas?

bocchino: > but maybe since constants required by the dictionary are always integers that is sufficient

As I said above, I don't think we should rely on this argument (our messages crossed). I think a stronger argument is that the explicit type info is not needed because it is in the form of the literal constant. With this approach, though, the analyzer would infer the type of the constant and then throw it away and leave the GDS to infer it again. I think that approach is not ideal in general, but maybe it's OK here since the type inference rule is so simple.

bocchino: According to the specification of values here

https://github.com/nasa/fprime/blob/devel/docs/reference/fpp-json-dict.md#values

without the type information there appears to be no way to distinguish a string value from an enumeration value. So omitting the type info seems to rely on the fact that the constants today all happen to be of a certain kind (integers). I think it's better to keep the design future proof.

jwest115: Good point on futureproofing. I think I am leaning more towards including the type in the dictionary and we can document in the user guide how the type is determined for integers (ie: using max size and signed if the value is negative).

How does that sound @LeStarch @thomas-bc @bocchino?

bocchino: That sounds good to me.

thomas-bc: > Add an Integer type that can hold any integer value, without specifying the signedness or bit width. This is how FPP represents integer constants internally.

Interesting. I don't have a strong opinion on how to put that in the dictionary, I was mostly trying to understand the choice made here. 

<details>

<summary>Although now I have even more questions... (somewhat unrelated, click to expand) </summary>

I didn't know you could have string constants. 
https://github.com/nasa/fprime/blob/devel/docs/reference/fpp-json-dict.md#values
Does this suggest I could have an array constant? 

```
    constant TmFrameFixedSize = 1024  # Needs to be at least COM_BUFFER_MAX_SIZE + (2 * SpacePacketHeaderSize) + 1
    constant c = "abcd"
    constant MY_ARRAY = [0, 1, 2]
    constant MY_FLOAT = 1.7
```

This seems to build fine, but `MY_ARRAY` doesn't get generated in FppConstantsAc.hpp 🤔 
But `MY_FLOAT` does get generated, of type F64! 

```
  extern const char *const c;
  extern const F64 ABCD; // defined to 1.7 in .cpp
```

Anyways, I digress. 

</details>

Let's confirm how to go forward with this at the infrastructure meeting tomorrow ?

bocchino: Yes, any value can be bound to a constant in the FPP model, but only scalar values get generated to C++. This is for practical reasons: generating aggregate constants in C++ seems challenging and not particularly useful. See the explanation here: https://nasa.github.io/fpp/fpp-users-guide.html#Analyzing-and-Translating-Models_Generating-C-Plus-Plus_Constant-Definitions.

bocchino: Note that aggregate constants can be useful in FPP, e.g.,
```
constant s = { x = 1, y = 2.0 }

struct S { x: U32, y: F32 } default s
```

bocchino:
Let's make sure that the JSON spec lists all the FPP symbols that are implicitly required in the model when you generate a dictionary. I think the JSON spec should be the source of truth for this list as it evolves. I've updated the FPP docs under the assumption that this list is available: https://github.com/nasa/fpp/wiki/Construct-Implied-Use-Map#topology-definitions.

LeStarch (approved):

Issue #3446, opened 2025-04-03
Remove the use of PROTECTED and PRIVATE


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Remove the use of `PROTECTED` and `PRIVATE`.  Replace with friend classes.

F Prime currently uses `PROTECTED` and `PRIVATE` custom definitions to allow unit-tests to whitebox test (access protected members).  However, defining these macros have several problems:

1. This is as close to overriding a language feature as can be done without actually overriding a reserved definition.
2. This affects all code (whitebox or otherwise)
3. There are better patterns (Friend classes)

To fix this, one should:
1. Define PRIVATE=private and PROTECTED=protected and remove STATIC altogether.
2. Build UTs and correct errors
3.  Replace `PRIVATE` and `PROTECTED` once all UTs build and pass

You may run into errors doing this.  Here is how to fix them:

1. TesterBase is automatically a friend of the base class in FPP.  Any accesses to test harness features `ASSERT_EVENT...` should work.
2. Accesses to hand implementation from Tester should add a friend class to the hand implementation that points at the hand written tester class.

i.e. add to `CommandDispatcherImpl.hpp`:
```diff
  class CommandDispatcherImpl final : public CommandDispatcherComponentBase {
+ friend  class CommandDispatcherImplTester
}
```

4. Accesses from Tester to BaseClass code not through TesterBase need to go through a newly defined class with the name `<ComponentBaseClass>Friend`
i.e. for command dispatcher, you'd define a class:

```
class CommandDispatcherComponentBaseFriend {

}
```
and use it as an accessor to reach base class functions.

thomas-bc:
Closed by https://github.com/nasa/fprime/pull/3821

PR #3701, opened 2025-06-05
PRIVATE->private and PROTECTED->protected updates in Os/File.hpp (#3446)

m-aleem:
| | |
|:---|:---|
|**_Has Unit Tests (y/n)_**| Yes - existing |

---
## Change Description

Updates to `Os/File.hpp`:

* `PRIVATE` -> `private`

There were issues with existing namespaces and class names conflicting, so I first (1) updated the Os::Test::File namespace from `File` -> `FileTest` and made the corresponding updates where that namespace was used, and then (2) added a `friend struct` until no build errors.

### DIFF
--- a/Os/File.hpp
+++ b/Os/File.hpp
@@ -8,7 +8,11 @@
 #include <Fw/FPrimeBasicTypes.hpp>
 #include <Os/Os.hpp>
 
+// Forward declation for UTs
+namespace Os {namespace Test {namespace FileTest {struct Tester;}}}
+
 namespace Os {
+
 //! \brief base implementation of FileHandle
 //!
 struct FileHandle {};
@@ -213,6 +217,9 @@ class FileInterface {
 };
 
 class File final : public FileInterface {
+
+  friend struct Os::Test::FileTest::Tester;
+
   public:
     //! \brief constructor
     //!
@@ -504,7 +511,8 @@ class File final : public FileInterface {
     Status finalizeCrc(U32& crc);
 
   private:
-    PRIVATE : static const U32 INITIAL_CRC = 0xFFFFFFFF;  //!< Initial value for CRC calculation
+    static const U32 INITIAL_CRC = 0xFFFFFFFF;            //!< Initial value for CRC calculation
+
     Mode m_mode = Mode::OPEN_NO_MODE;                     //!< Stores mode for error checking
     const CHAR* m_path = nullptr;                         //!< Path last opened
 
--- a/Os/File.hpp
+++ b/Os/File.hpp
@@ -8,7 +8,7 @@
 #include <Fw/FPrimeBasicTypes.hpp>
 #include <Os/Os.hpp>
 
-// Forward declation for UTs
+// Forward declaration for UTs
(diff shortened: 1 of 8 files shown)


LeStarch (dismissed):

Os/File.hpp
```diff
@@ -8,7 +8,11 @@
 #include <Fw/FPrimeBasicTypes.hpp>
 #include <Os/Os.hpp>
 
+// Forward declation for UTs
```
thomas-bc: Typo - would you please fix for CI to pass?

thomas-bc (changes requested):

Os/File.hpp
m-aleem: Adressed in [e63ba27](https://github.com/nasa/fprime/pull/3701/commits/e63ba275a81890dc0c31cef76fdbecf699097c93)

Os/Posix/test/ut/PosixFileTests.cpp
```diff
@@ -13,7 +13,7 @@
 #include <csignal>
 namespace Os {
 namespace Test {
-namespace File {
+namespace FileTest {
```
thomas-bc, line 16: Is the reason to rename this to not have potentially ambiguous `File::` references? Or is there something else?

I think most of our OS modules follow the convention `Os::XXX` / `Os::Test::XXX`. Is the plan to switch them all to `Os::XXX`/`Os::Test::XXXTest` ?

thomas-bc (approved):
LGTM! Just one quick question

Os/Posix/test/ut/PosixFileTests.cpp
m-aleem: There was conflict due to both a namespace named File `Os::Test::File` and a class named `File` in Os. 

Since I needed to remove PRIVATE from File.hpp that required making struct `Os::Test::File[Test]::Tester` a friend of [class File](https://github.com/nasa/fprime/blob/7f5710ace0c9bc2b1109b64f6b36c3d31102a933/Os/File.hpp#L215). But without a forward declaration that will not build.

If the forward declaration is made like `namespace Os {namespace Test {namespace File {struct Tester;}}}` this causes other build errors because of Os::Test::File being declared a namespace and then File:: references (meant to refer to the class) being interpreted as namespace and causing build errors, for example like the follow:

```
In file included from /home/aleem/workspace/fprime/Os/test/ut/file/SyntheticFileSystem.cpp:5:
/home/aleem/workspace/fprime/Os/test/ut/file/SyntheticFileSystem.hpp:28:11: error: ‘Mode’ in namespace ‘Os::Test::File’ does not name a type
     File::Mode m_mode = File::OPEN_NO_MODE;
```
So in this particular case I changed the namespace to `Os::Test::File -> Os::Test::FileTest`, but have not/do not plan to modify any others unless there are similar conflicts.

thomas-bc: Ok sounds good! Thanks a lot for the explanation

PR #3684, opened 2025-06-02
Add support for CCSDS TC/TM and SpacePacket protocols

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/3443 |
|**_Has Unit Tests (y/n)_**| yes |
|**_Documentation Included (y/n)_**| yes |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/3443 https://github.com/nasa/fprime/issues/3491 https://github.com/nasa/fprime/issues/3492 

Add Framers and Deframers (and helpers) to support encapsulating F´ data in CCSDS Space Packets + TC and TM frames.

Each protocol is supported as described by the component SDD.

## Rationale

A long standing request by the community.
https://github.com/nasa/fprime/discussions/1397 https://github.com/nasa/fprime/discussions/2507 https://github.com/nasa/fprime/discussions/3026

## Future Work

TBD

PR #3707, opened 2025-06-09
Fix ambiguous error for .serialize() overloads

ethancheez:
| | |
|:---|:---|

---
## Change Description

Add `static_cast<U32>` to the `.serialize()` calls within [`FileDownlink.cpp`](https://github.com/nasa/fprime/blob/7f5710ace0c9bc2b1109b64f6b36c3d31102a933/Svc/FileDownlink/FileDownlink.cpp#L363) and [`FpySequencerRunState.cpp`](https://github.com/nasa/fprime/blob/7f5710ace0c9bc2b1109b64f6b36c3d31102a933/Svc/FpySequencer/FpySequencerRunState.cpp#L53).

## Rationale

Older versions of GCC (i.e. GCC 11.3.1) aren't that clever at picking one of the overloaded functions for `.serialize()`. So it gives the error:
```sh
fprime/Svc/FpySequencer/FpySequencerRunState.cpp:53:48: error: call of overloaded 'serialize(Fw::ComPacket::ComPacketType)' is ambiguous
   53 |     Fw::SerializeStatus stat = cmdBuf.serialize(Fw::ComPacket::FW_PACKET_COMMAND);
      |                                ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from fprime/Fw/Com/ComPacket.hpp:11,
                 from fprime/Svc/FpySequencer/FpySequencerRunState.cpp:2:
fprime/Fw/Types/Serializable.hpp:63:21: note: candidate: 'Fw::SerializeStatus Fw::SerializeBufferBase::serialize(U8)'
   63 |     SerializeStatus serialize(U8 val);  //!< serialize 8-bit unsigned int
      |                     ^~~~~~~~~
fprime/Fw/Types/Serializable.hpp:64:21: note: candidate: 'Fw::SerializeStatus Fw::SerializeBufferBase::serialize(I8)'
   64 |     SerializeStatus serialize(I8 val);  //!< serialize 8-bit signed int
      |                     ^~~~~~~~~
fprime/Fw/Types/Serializable.hpp:67:21: note: candidate: 'Fw::SerializeStatus Fw::SerializeBufferBase::serialize(U16)'
   67 |     SerializeStatus serialize(U16 val);  //!< serialize 16-bit unsigned int
      |                     ^~~~~~~~~

... etc.
```

I chose to cast as `U32` based on unit tests. `Fw.Com` has a `U32` context that causes the UT to fail if casted as an `U8` (specifically [`FpySequencerTestMain.cpp:772`](https://github.com/nasa/fprime/blob/7f5710ace0c9bc2b1109b64f6b36c3d31102a933/Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp#L772).

I tested this on `fprime-arduino` toolchains since some of them use older GCC versions. 

Review:
Make sure it is fine to cast as a `U32`.

## Future Work

N/A

### DIFF
--- a/Svc/FileDownlink/FileDownlink.cpp
+++ b/Svc/FileDownlink/FileDownlink.cpp
@@ -360,7 +360,7 @@ void FileDownlink ::sendCancelPacket() {
               static_cast<FwAssertArgType>(filePacket.bufferSize() + sizeof(FwPacketDescriptorType)));
 
     // Serialize the packet descriptor FW_PACKET_FILE to the buffer
-    Fw::SerializeStatus status = buffer.getSerializer().serialize(static_cast<U32>(Fw::ComPacket::FW_PACKET_FILE));
+    Fw::SerializeStatus status = buffer.getSerializer().serialize(static_cast<FwPacketDescriptorType>(Fw::ComPacket::FW_PACKET_FILE));
     FW_ASSERT(status == Fw::FW_SERIALIZE_OK);
     Fw::Buffer offsetBuffer(buffer.getData() + sizeof(FwPacketDescriptorType),
                             buffer.getSize() - static_cast<Fw::Buffer::SizeType>(sizeof(FwPacketDescriptorType)));
@@ -398,7 +398,7 @@ void FileDownlink ::sendFilePacket(const Fw::FilePacket& filePacket) {
     FW_ASSERT(this->m_buffer.getSize() >= bufferSize, static_cast<FwAssertArgType>(bufferSize),
               static_cast<FwAssertArgType>(this->m_buffer.getSize()));
     // Serialize packet descriptor FW_PACKET_FILE to the buffer
-    Fw::SerializeStatus status = this->m_buffer.getSerializer().serialize(static_cast<U32>(Fw::ComPacket::FW_PACKET_FILE));
+    Fw::SerializeStatus status = this->m_buffer.getSerializer().serialize(static_cast<FwPacketDescriptorType>(Fw::ComPacket::FW_PACKET_FILE));
     FW_ASSERT(status == Fw::FW_SERIALIZE_OK);
     // Serialize the filePacket content into the buffer, offset by the size of the packet descriptor
     Fw::Buffer offsetBuffer(
--- a/Svc/FileDownlink/FileDownlink.cpp
+++ b/Svc/FileDownlink/FileDownlink.cpp
@@ -360,7 +360,7 @@ void FileDownlink ::sendCancelPacket() {
               static_cast<FwAssertArgType>(filePacket.bufferSize() + sizeof(FwPacketDescriptorType)));
 
     // Serialize the packet descriptor FW_PACKET_FILE to the buffer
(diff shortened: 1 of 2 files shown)


codeflight1:
Thanks for looking into this!

Svc/FileDownlink/FileDownlink.cpp
```diff
@@ -360,7 +360,7 @@ void FileDownlink ::sendCancelPacket() {
               static_cast<FwAssertArgType>(filePacket.bufferSize() + sizeof(FwPacketDescriptorType)));
 
     // Serialize the packet descriptor FW_PACKET_FILE to the buffer
-    Fw::SerializeStatus status = buffer.getSerializer().serialize(Fw::ComPacket::FW_PACKET_FILE);
+    Fw::SerializeStatus status = buffer.getSerializer().serialize(static_cast<U32>(Fw::ComPacket::FW_PACKET_FILE));
```
thomas-bc: We may want to use `FwPacketDescriptorType` instead (by default that's a U32)

thomas-bc (changes requested):

Svc/FileDownlink/FileDownlink.cpp
ethancheez: changed.

thomas-bc (approved):

thomas-bc:
@zimri-leisher would you please review and confirm this is ok for FpySequencer ?

zimri-leisher:
The compiler might be able to figure this out without an explicit cast if you changed the declaration of the ComPacketType from this:
```
            typedef enum {
                FW_PACKET_COMMAND, // !< Command packet type - incoming
                FW_PACKET_TELEM, // !< Telemetry packet type - outgoing
                FW_PACKET_LOG, // !< Log type - outgoing
                FW_PACKET_FILE, // !< File type - incoming and outgoing
                FW_PACKET_PACKETIZED_TLM, // !< Packetized telemetry packet type
                FW_PACKET_DP, //!< Data product packet
                FW_PACKET_IDLE, // !< Idle packet
                FW_PACKET_UNKNOWN = 0xFF // !< Unknown packet
            } ComPacketType;
```

To this

```
enum class ComPacketType : FwPacketDescriptorType {
...
}
```

That would be my preferred solution to this problem. However, the PR as it stands is acceptable, but I don't think it's the "future-proof" way to do it.

thomas-bc:
@zimri-leisher you're absolutely right, and indeed it actually simply doesn't work for now if you tried to switch `FwPacketDescriptorType` to a smaller value.

There is somewhat related work going on in https://github.com/nasa/fprime/compare/devel...jwest115:fprime:issue-3529

Then, the [CCSDS work](https://github.com/nasa/fprime/pull/3684) will define that enum in FPP

thomas-bc:
I'll go ahead and merge this since static_cast is the way it's handled in the core framework for now.

https://github.com/nasa/fprime/blob/devel/Fw/Com/ComPacket.cpp#L19

PR #3708, opened 2025-06-09
feat: use config header for CRC filename size

0xb-s:
Replaced hardcoded CRC_MAX_FILENAME_SIZE with value from CRCCheckerConfig.hpp.


Utils/CRCChecker.hpp
```diff
@@ -13,11 +13,12 @@
 #define CRC_CHECKER_HPP
 
 #include <Fw/FPrimeBasicTypes.hpp>
+#include <config/CRCCheckerConfig.hpp>
 
 namespace Utils {
 
   static const FwSignedSizeType CRC_FILE_READ_BLOCK = 2048;
```
0xb-s: Should we apply the same approach for CRC_FILE_READ_BLOCK and move it to the config header as well?

default/config/CRCCheckerConfig.hpp
```diff
@@ -0,0 +1,12 @@
+// ======================================================================
+// \title  CRCCheckerConfig.hpp
+// \brief  Configuration file for CRCChecker module
+// ======================================================================
+
+#ifndef CONFIG_CRC_CHECKER_CONFIG_HPP
+#define CONFIG_CRC_CHECKER_CONFIG_HPP
+
+// Maximum size allowed for filename strings in CRC checking
+#define CONFIG_CRC_MAX_FILENAME_SIZE 128
```
LeStarch: Generally it is better to use a constant in files like this.  While the `#define` will work, we as a project are moving toward constants.

```
constexpr FwSizeType CONFIG_CRC_MAX_FILENAME_SIZE = 128;
```

LeStarch (changes requested):
Additional improvement suggestion. Otherwise this is an excellent fix!

default/config/CRCCheckerConfig.hpp
0xb-s: Thank you! I’ve updated the code to use constexpr.

Utils/CRCChecker.hpp
LeStarch: yes, this should likely be included in the config file in the same manner.   Let me know if you want to do a separate PR or fix here!

LeStarch (dismissed):
Ran CI, may need a cast but CI will tell.

Utils/CRCChecker.hpp
0xb-s: > yes, this should likely be included in the config file in the same manner. Let me know if you want to do a separate PR or fix here!

Would be great to address it in a separate PR—thanks a lot!

Utils/CRCChecker.hpp
```diff
@@ -13,11 +13,12 @@
...
-  static const U32 CRC_MAX_FILENAME_SIZE = 128; // TODO use a config variable
+  static const U32 CRC_MAX_FILENAME_SIZE = CONFIG_CRC_MAX_FILENAME_SIZE;
```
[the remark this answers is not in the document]
LeStarch: @0xb-s looks like this is correct, this is never used.  Want to move the READ_BLOCK into your config file and remove this constant all together?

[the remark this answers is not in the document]
0xb-s: > @0xb-s looks like this is correct, this is never used. Want to move the READ_BLOCK into your config file and remove this constant all together?

Done !

LeStarch (approved):
I removed the extra definition of the size in the config, since it was removed in the code.  Once CI passes, we'll merge!

Thanks!

Utils/CRCChecker.hpp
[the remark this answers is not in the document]
LeStarch: Awesome, thanks for the hard work!

0xb-s:
> I removed the extra definition of the size in the config, since it was removed in the code. Once CI passes, we'll merge!
> 
> Thanks!

Thank you ! let me know if any additional changes are needed.

thomas-bc (approved):

Issue #3529, opened 2025-04-24
Error when building with U64 Fw Types


jwest115:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

Tried building Ref deployment after changing config/FpConfig.fpp's `FwEventIdType` to U64 and received the following error:
```
Svc/ActiveLogger/ActiveLoggerImpl.cpp:189:57: error: implicit conversion loses integer precision: 'FwEventIdType' (aka 'unsigned long long') to 'U32' (aka 'unsigned int') [-Werror,-Wshorten-64-to-32]
  189 |                 this->log_ACTIVITY_HI_ID_FILTER_ENABLED(this->m_filteredIDs[entry]);
      |                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

If I try a smaller type (such as U16), I get different errors (example below is not the complete list):
```
Svc/AssertFatalAdapter/AssertFatalAdapterComponentAc.cpp:292:29: error: implicit conversion loses integer precision: 'U32' (aka 'unsigned int') to 'FwEventIdType' (aka 'unsigned short') [-Werror,-Wimplicit-int-conversion]
  292 |     _id = this->getIdBase() + EVENTID_AF_ASSERT_0;
      |         ~ ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
Svc/AssertFatalAdapter/AssertFatalAdapterComponentAc.cpp:398:29: error: implicit conversion loses integer precision: 'U32' (aka 'unsigned int') to 'FwEventIdType' (aka 'unsigned short') [-Werror,-Wimplicit-int-conversion]
  398 |     _id = this->getIdBase() + EVENTID_AF_ASSERT_1;
```

## Context / Environment

Execute `fprime-util version-check` and share the output.

```
Operating System: Darwin
CPU Architecture: arm64
Platform: macOS-15.4-arm64-arm-64bit
Python version: 3.9.21
CMake version: 3.26.3
Pip version: 24.3.1
Pip packages:
    fprime-tools==3.6.1
    fprime-gds (Running with dev version I am testing)
    fprime-fpp-* (Running with dev version I am testing https://github.com/jwest115/fpp/tree/fw-types-in-dict)

thomas-bc:
[...]

PR #3714, opened 2025-06-10
Support non-default Fw type config (ie: U16, U64)

jwest115:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/3529 |

---
## Change Description

- Added `FwIdType` alias (default is U32), which represents the max ID type size. Updated `PassiveComponentBase` to use this type for the base ID. 
- Updated the following types in `FpConfig.fpp` to be aliases of `FwIdType`:
   - `FwChanIdType`
   - `FwDpIdType`
   - `FwEventIdType`
   - `FwOpcodeType`
   - `FwPacketDescriptorType`
   - `FwPrmIdType`
- Updated component FPP so the proper types are used for IDs (example: `FwOpcodeType` for opcodes instead of `U32`).
- Updated related component cpp/hpp so the Fw type is used instead of U32.
- Updated unit tests so the Fw type is used instead of U32.

## Rationale

Needed when changing the Fw types in FwConfig.fpp to something other than U32.

Confirmed fprime builds successfully and without any warnings. Confirmed that all unit tests pass.

Related FPP change is in this PR https://github.com/nasa/fpp/pull/737, should wait until the FPP PR is merged and the FPP version is updated in this PR _before_ merging.


LeStarch (approved):
Looks like what I'd expect!

Issue #3519, opened 2025-04-21
Os::FileSystem::getPathType not detecting directories


LeStarch:
### Discussed in https://github.com/nasa/fprime/discussions/3516

<div type='discussions-op-text'>

<sup>Originally posted by **LornDMiller** April 18, 2025</sup>
I'm trying to use Os::FileSystem::getPathType to determine if a path is a directory and it's constantly telling me it's a file even though I know it's a directory. I traced this back through the code and saw that it was ultimately doing an "open" on the path (fprime/Os/Posix/File.cpp, line 86) and if that succeeded, calling the path a file.

I did a quick C test program on the same machine, Ubuntu 22.04.5, and saw that the open call does in fact return a valid file descriptor (in my case, 3) and leaves errno set to 0 when attempting to open a directory.

Is this something that's just broken in FPrime? Is this something I need to examine more thoroughly or is this a potential bug report? Am I missing a compiler flag somewhere?</div>

thomas-bc:
Should the F Prime OSAL impose a specific behavior when calling `Os::File::open()` on a directory (has to error / succeed), or should that be up to the OS implementation ?

https://github.com/nasa/fprime/blob/79005109981a7bdb910ccdd473b2b0795c864640/Os/FileSystem.cpp#L133-L146

This code assumes that `Os::File::open()` fails for a directory; which is not the case in the current Posix implementation (delegates to ::open() which succeeds on directories). We could rework the logic by trying to opening a `Os::Directory` first, imposing that `Os::Directory::open()` fails on a file.

vincewoo:
Fix has been merged into devel

PR #3727, opened 2025-06-12
Fix directory detection bug in Os::FileSystem::getPathType

vincewoo:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3519  |

---
## Change Description

Fix directory detection in Os::FileSystem::getPathType. The new code is relying on standard POSIX features `stat()`, `S_ISDIR`, and `S_ISREG` for detection instead of trying to open. 

## Rationale

This fixes a bug that incorrectly detects a directory as a regular file.

New UT, `DetectDirectory`, added to `FileSystemRules` that test the following cases.

## Future Work

None

### DIFF
--- a/Os/FileSystem.cpp
+++ b/Os/FileSystem.cpp
@@ -4,6 +4,7 @@
 // ======================================================================
 #include <Fw/Types/Assert.hpp>
 #include <Os/FileSystem.hpp>
+#include <sys/stat.h>
 
 namespace Os {
 
@@ -130,17 +131,13 @@ FileSystem::Status FileSystem::touch(const char* path) {
 
 FileSystem::PathType FileSystem::getPathType(const char* path) {
     FW_ASSERT(path != nullptr);
-    Os::File file;
-    File::Status file_status = file.open(path, Os::File::OPEN_READ);
-    file.close();
-    if (file_status == File::OP_OK) {
-        return PathType::FILE;
-    }
-    Os::Directory dir;
-    Directory::Status dir_status = dir.open(path, Os::Directory::OpenMode::READ);
-    dir.close();
-    if (dir_status == Directory::Status::OP_OK) {
-        return PathType::DIRECTORY;
+    struct stat path_stat;
+    if (stat(path, &path_stat) == 0) {
+        if (S_ISDIR(path_stat.st_mode)) {
+            return PathType::DIRECTORY;
+        } else if (S_ISREG(path_stat.st_mode)) {
+            return PathType::FILE;
+        }
     }
     return PathType::NOT_EXIST;
 }  // end getPathType
--- a/Os/FileSystem.cpp
+++ b/Os/FileSystem.cpp
@@ -4,7 +4,6 @@
 // ======================================================================
 #include <Fw/Types/Assert.hpp>
 #include <Os/FileSystem.hpp>
-#include <sys/stat.h>
 
 namespace Os {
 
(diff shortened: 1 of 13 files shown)


Os/FileSystem.cpp
```diff
@@ -130,17 +131,13 @@ FileSystem::Status FileSystem::touch(const char* path) {
...
-    if (dir_status == Directory::Status::OP_OK) {
-        return PathType::DIRECTORY;
+    struct stat path_stat;
+    if (stat(path, &path_stat) == 0) {
+        if (S_ISDIR(path_stat.st_mode)) {
+            return PathType::DIRECTORY;
+        } else if (S_ISREG(path_stat.st_mode)) {
+            return PathType::FILE;
+        }
     }
```
thomas-bc, line 143: Should we return `PathType::OTHER` in the branch `stat == 0` but is not directory or regular file - instead of falling through to return `NOT_EXIST` ? 
Seems like this could happen for a couple of things (sockets, symlinks, etc)

https://www.gnu.org/software/libc/manual/html_node/Testing-File-Type.html

vincewoo: That's a good suggestion. I'll add the else branch.

Os/FileSystem.cpp
```diff
@@ -130,17 +131,15 @@ FileSystem::Status FileSystem::touch(const char* path) {
...
-    if (file_status == File::OP_OK) {
-        return PathType::FILE;
-    }
-    Os::Directory dir;
-    Directory::Status dir_status = dir.open(path, Os::Directory::OpenMode::READ);
-    dir.close();
-    if (dir_status == Directory::Status::OP_OK) {
-        return PathType::DIRECTORY;
+    struct stat path_stat;
+    if (stat(path, &path_stat) == 0) {
```
LeStarch: Posix calls should not be added to `Os/FileSystem.cpp` they only belong in the Posix Implementation `Os/Posix/FileSystem.cpp`

Os/test/ut/filesystem/FileSystemRules.cpp
```diff
@@ -6,6 +6,8 @@
 #include "RulesHeaders.hpp"
 #include "FileSystemRules.hpp"
 #include "STest/Pick/Pick.hpp"
+#include <sys/stat.h>
```
LeStarch: Again, posix calls in generic test code.

LeStarch (changes requested):

Os/FileSystem.cpp
vincewoo: This has been refactored to delegate to OS-specific implementation

Os/test/ut/filesystem/FileSystemRules.cpp
vincewoo: This include has been removed after the refactor to delegate to OS-specific implementations

Os/Stub/test/ut/StubFileSystemTests.cpp
```diff
@@ -73,10 +73,17 @@ TEST_F(Interface, GetFreeSpace) {
 
 // Ensure that Os::FileSystem properly calls the implementation getHandle()
 TEST_F(Interface, GetHandle) {
-    ASSERT_EQ(Os::FileSystem::getSingleton().getHandle(), nullptr);
```
LeStarch, line 76: How did this pass before?

LeStarch (approved):
Excellent work.

thomas-bc:
Labeling this so we remember to mention the new implementation-specific function in the release notes

Issue #3716, opened 2025-06-11
Clean Up `Drv::Ip` Dynamic Memory


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

`Drv::Udp` uses a struct and dynamic memory to store state.  It looks like that is no longer needed and should be cleaned-out.

Investigate `Drv::Tcp*` and `Drv::IpSocket` for similar code for removal.

vincewoo:
I have reviewed Drv::TcpServerSocket, Drv::TcpClientSocket, and Drv::IpSocket. They do not have the same state deficiency as Drv::UdpSocket had, so no changes need to be made to those for this issue.

vincewoo:
#3728 Has been merged with the fix

PR #3728, opened 2025-06-12
Refactor Drv::Ip::UdpSocket to clean up state members and remove dependency on dynamic memory

vincewoo:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3716 |
|**_Has Unit Tests (y/n)_**| y|

---
## Change Description

This change cleans up Drv::Ip::UdpSocket to remove the members `m_state`, `m_recv_port`, and `m_recv_hostname` in favour of two new members `m_addr_send` and `m_addr_recv`. As part of this change we are also removing the dependency on dynamic memory for storing state.

This change also improved error handling, edge case handling, and memory safety.

## Rationale

Simplifies code and fixes bug of having duplicate data structures storing the same state.

Existing UTs cover the changes made except for handling 0-length datagrams. A new UT was added specifically to test the behaviour of 0-length datagrams.

## Future Work

None


Drv/Ip/UdpSocket.cpp
```diff
@@ -40,24 +40,12 @@
...
-    SocketState() {
-        ::memset(&m_addr_send, 0, sizeof(m_addr_send));
-        ::memset(&m_addr_recv, 0, sizeof(m_addr_recv));
-    }
-};
-
-UdpSocket::UdpSocket() : IpSocket(), m_state(new(std::nothrow) SocketState), m_recv_port(0), m_recv_configured(false) {
-    FW_ASSERT(m_state != nullptr);
+UdpSocket::UdpSocket() : IpSocket(), m_recv_configured(false) {
+    ::memset(&m_addr_send, 0, sizeof(m_addr_send));
```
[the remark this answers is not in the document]
LeStarch: Should put `(void)` in front of memset to discard result.

Drv/Ip/UdpSocket.cpp
```diff
@@ -66,42 +54,57 @@ SocketIpStatus UdpSocket::configure(const char* const hostname, const U16 port,
 
 
 SocketIpStatus UdpSocket::configureSend(const char* const hostname, const U16 port, const U32 timeout_seconds, const U32 timeout_microseconds) {
-    //Timeout is for the send, so configure send will work with the base class
-    FW_ASSERT(hostname != nullptr);
-    return this->IpSocket::configure(hostname, port, timeout_seconds, timeout_microseconds);
+    if (hostname == nullptr) {
```
LeStarch: Why did this become an if then status as opposed to an assert?

Drv/Ip/UdpSocket.cpp
```diff
@@ -66,42 +54,57 @@ SocketIpStatus UdpSocket::configure(const char* const hostname, const U16 port,
...
-    this->m_recv_port = port;
-    (void) Fw::StringUtils::string_copy(this->m_recv_hostname, hostname, static_cast<FwSizeType>(SOCKET_MAX_HOSTNAME_SIZE));
+    if (hostname == nullptr) {
+        return SOCK_INVALID_IP_ADDRESS;
+    }
+    if (!this->isValidPort(port)) {
+        return SOCK_INVALID_CALL;
+    }
+    // Check hostname length to prevent potential buffer overflow
+    if (strlen(hostname) >= SOCKET_MAX_HOSTNAME_SIZE) {
```
LeStarch: Should not use raw `strlen`, use `StringUtils::string_length` instead to avoid memory overruns.

Drv/Ip/UdpSocket.cpp
```diff
@@ -66,42 +54,57 @@ SocketIpStatus UdpSocket::configure(const char* const hostname, const U16 port,
...
 
 SocketIpStatus UdpSocket::configureRecv(const char* hostname, const U16 port) {
-    FW_ASSERT(this->isValidPort(port));
-    FW_ASSERT(hostname != nullptr);
-    this->m_recv_port = port;
-    (void) Fw::StringUtils::string_copy(this->m_recv_hostname, hostname, static_cast<FwSizeType>(SOCKET_MAX_HOSTNAME_SIZE));
+    if (hostname == nullptr) {
+        return SOCK_INVALID_IP_ADDRESS;
+    }
+    if (!this->isValidPort(port)) {
```
LeStarch: Probably should make this a named status, e.g. `SOCK_INVALID_PORT`

Drv/Ip/UdpSocket.cpp
```diff
@@ -66,42 +54,57 @@ SocketIpStatus UdpSocket::configure(const char* const hostname, const U16 port,
...
+    }
+    return IpSocket::configure(hostname, port, timeout_seconds, timeout_microseconds);
 }
 
 SocketIpStatus UdpSocket::configureRecv(const char* hostname, const U16 port) {
-    FW_ASSERT(this->isValidPort(port));
-    FW_ASSERT(hostname != nullptr);
-    this->m_recv_port = port;
-    (void) Fw::StringUtils::string_copy(this->m_recv_hostname, hostname, static_cast<FwSizeType>(SOCKET_MAX_HOSTNAME_SIZE));
+    if (hostname == nullptr) {
```
LeStarch: Again, why not the assert?

Drv/Ip/UdpSocket.cpp
```diff
@@ -66,42 +54,57 @@ SocketIpStatus UdpSocket::configure(const char* const hostname, const U16 port,
...
 
 
 SocketIpStatus UdpSocket::bind(const PlatformIntType fd) {
-    struct sockaddr_in address;
-    FW_ASSERT(fd != -1);
-
-    // Set up the address port and name
-    address.sin_family = AF_INET;
-    address.sin_port = htons(this->m_recv_port);
+    if (fd < 0) {
```
LeStarch: Another assert changed to status....

Drv/Ip/UdpSocket.cpp
```diff
@@ -66,42 +54,57 @@ SocketIpStatus UdpSocket::configure(const char* const hostname, const U16 port,
...
+        return SOCK_FAILED_TO_BIND;
+    }
+    
+    struct sockaddr_in address = this->m_addr_recv;
```
LeStarch, line 88: Why copy here instead of just using the member?  If you are to copy, could we use a reference?

Drv/Ip/UdpSocket.hpp
```diff
@@ -16,9 +16,18 @@
 #include <Drv/Ip/IpSocket.hpp>
 #include <config/IpCfg.hpp>
 
-namespace Drv {
+// Include system headers for sockaddr_in
+#ifdef TGT_OS_TYPE_VXWORKS
+    #include <socket.h>
+    #include <inetLib.h>
+#elif defined TGT_OS_TYPE_LINUX || TGT_OS_TYPE_DARWIN
```
LeStarch: Perhaps we should just have the `else` clause and VXWORKS.  Most patforms will put the headers as UNIX does.

Drv/Ip/UdpSocket.cpp
```diff
@@ -186,21 +198,31 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
...
 I32 UdpSocket::recvProtocol(const SocketDescriptor& socketDescriptor, U8* const data, const U32 size) {
-    FW_ASSERT(this->m_state->m_addr_recv.sin_family != 0); // Make sure the address was previously setup
+    FW_ASSERT(this->m_addr_recv.sin_family != 0); // Make sure the address was previously setup
+    FW_ASSERT(socketDescriptor.fd >= 0); // File descriptor should be valid
+    FW_ASSERT(data != nullptr); // Data pointer should not be null
+    FW_ASSERT(size > 0); // Size should be positive
 
+    // Initialize sender address structure to zero
     struct sockaddr_in sender_addr;
+    ::memset(&sender_addr, 0, sizeof(sender_addr));
```
LeStarch: void out the return.

Drv/Ip/UdpSocket.cpp
```diff
@@ -186,21 +198,31 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
...
-                    reinterpret_cast<struct sockaddr *>(&this->m_state->m_addr_send), sizeof(this->m_state->m_addr_send)));
+                    reinterpret_cast<struct sockaddr *>(&this->m_addr_send), sizeof(this->m_addr_send)));
 }
 
 I32 UdpSocket::recvProtocol(const SocketDescriptor& socketDescriptor, U8* const data, const U32 size) {
-    FW_ASSERT(this->m_state->m_addr_recv.sin_family != 0); // Make sure the address was previously setup
+    FW_ASSERT(this->m_addr_recv.sin_family != 0); // Make sure the address was previously setup
+    FW_ASSERT(socketDescriptor.fd >= 0); // File descriptor should be valid
+    FW_ASSERT(data != nullptr); // Data pointer should not be null
+    FW_ASSERT(size > 0); // Size should be positive
```
LeStarch: This is a new assert here....We should check if a send of size `0` is legal UDP or not.

LeStarch: It seems for receive a zero-length is allowed (and can indicate a zero-length datagram).  This probably should be `>=`

Drv/Ip/UdpSocket.cpp
```diff
@@ -186,21 +198,31 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
 }
 
 I32 UdpSocket::sendProtocol(const SocketDescriptor& socketDescriptor, const U8* const data, const U32 size) {
-    FW_ASSERT(this->m_state->m_addr_send.sin_family != 0); // Make sure the address was previously setup
+    FW_ASSERT(this->m_addr_send.sin_family != 0); // Make sure the address was previously setup
+    FW_ASSERT(socketDescriptor.fd >= 0); // File descriptor should be valid
+    FW_ASSERT(data != nullptr); // Data pointer should not be null
+    FW_ASSERT(size > 0); // Size should be positive
```
LeStarch: Zero length datagrams are allowed.

Drv/Ip/UdpSocket.cpp
```diff
@@ -186,21 +198,31 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
...
     struct sockaddr_in sender_addr;
+    ::memset(&sender_addr, 0, sizeof(sender_addr));
+
     socklen_t sender_addr_len = sizeof(sender_addr);
     I32 received = static_cast<I32>(::recvfrom(socketDescriptor.fd, data, size, SOCKET_IP_RECV_FLAGS,
                                               reinterpret_cast<struct sockaddr*>(&sender_addr), &sender_addr_len));
     // If we have not configured a send port, set it to the source of the last received packet
-    if (received > 0 && this->m_state->m_addr_send.sin_port == 0) {
-        this->m_state->m_addr_send = sender_addr;
+    if (received > 0 && this->m_addr_send.sin_port == 0) {
```
LeStarch: I don't know that this is right anymore.  a received length can be `0` and would still be a valid datagram.  I am wondering if we should just update the sender on every receive that didn't return `-1`.

I would expect this call to block, timeout, (errors), or return a valid receive

LeStarch (changes requested):
There is a lot more here than I expected.  Also, it seems that many asserts have been changed to if-checks.  The general rule is:

1. If it is clearly a coding error, assert.  (e.g. a char* was passed in as nullptr, or `bind` was called before `open`)
2. Use an `if` to check for things that could be a error with the run.

Drv/Ip/UdpSocket.cpp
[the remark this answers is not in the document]
vincewoo: Thanks for the feedback! Incorporated a cast to void

Drv/Ip/UdpSocket.cpp
```diff
@@ -40,24 +40,12 @@
...
-        ::memset(&m_addr_send, 0, sizeof(m_addr_send));
-        ::memset(&m_addr_recv, 0, sizeof(m_addr_recv));
-    }
-};
-
-UdpSocket::UdpSocket() : IpSocket(), m_state(new(std::nothrow) SocketState), m_recv_port(0), m_recv_configured(false) {
-    FW_ASSERT(m_state != nullptr);
+UdpSocket::UdpSocket() : IpSocket(), m_recv_configured(false) {
+    ::memset(&m_addr_send, 0, sizeof(m_addr_send));
+    ::memset(&m_addr_recv, 0, sizeof(m_addr_recv));
```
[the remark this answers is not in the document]
vincewoo: Cast to void to discard result.

Drv/Ip/UdpSocket.cpp
vincewoo: Reverting this back to FW_ASSERT per general rules above.

Drv/Ip/UdpSocket.cpp
vincewoo: Converted to using StringUtils::string_length

Drv/Ip/UdpSocket.cpp
vincewoo: This was converted to an FW_ASSERT based on above general rules

Drv/Ip/UdpSocket.cpp
vincewoo: Converted back to FW_ASSEERT

Drv/Ip/UdpSocket.cpp
vincewoo: Converted back to an FW_ASSERT

Drv/Ip/test/ut/TestUdp.cpp
```diff
@@ -100,6 +109,47 @@ TEST(SingleSide, TestSingleSideSendUdp) {
...
+    U8 empty_data[1] = {0}; // Buffer is required, but size is 0
+    ASSERT_EQ(sender.send(send_fd, empty_data, 0), Drv::SOCK_SUCCESS);
+
+    // Receive the zero-length datagram
+    U8 recv_buf[1] = {0xFF};
+    U32 recv_buf_len = 1;
+    I32 recv_status = receiver.recv(recv_fd, recv_buf, recv_buf_len);
+
+    // Accept 0 (success), -EAGAIN, or -EWOULDBLOCK (timeout/no data)
+    bool valid_status = (recv_status == 0 ||
```
LeStarch: `EWOULDBLOCK` should not block, as there should have been a packet (zerolength).  Remove that status.  

I think the other tests loop on EAGAIN too....but maybe that is a bad idea.

Drv/Ip/test/ut/TestUdp.cpp
```diff
@@ -100,6 +109,47 @@ TEST(SingleSide, TestSingleSideSendUdp) {
...
+                         recv_status == -EAGAIN ||
+#ifdef EWOULDBLOCK
+                         recv_status == -EWOULDBLOCK ||
+#endif
+                         recv_status == -16 // fallback for platforms where EWOULDBLOCK is not defined but -16 is returned
+    );
+    ASSERT_TRUE(valid_status)
+        << "recv_status=" << recv_status << ", errno=" << errno;
+
+    sender.close(send_fd);
```
LeStarch, line 140: There is not test of the read data.  Should we not assert recv_buf_length == 0?

Drv/Ip/test/ut/TestUdp.cpp
```diff
@@ -3,10 +3,19 @@
 //
 #include <gtest/gtest.h>
 #include <cstring>
+#include <cerrno>
 #include <string>
+
+#ifdef __VXWORKS__
```
LeStarch: Should come in from the socket headers, no?

Drv/Ip/UdpSocket.hpp
LeStarch: Can we move the #include for linux/darwin into else?

Drv/Ip/UdpSocket.cpp
```diff
@@ -66,42 +54,57 @@
...
+        return SOCK_INVALID_IP_ADDRESS;
+    }
+    
+    // Initialize the receive address structure
+    ::memset(&m_addr_recv, 0, sizeof(m_addr_recv));
+    m_addr_recv.sin_family = AF_INET;
+    m_addr_recv.sin_port = htons(port);
+    
+    // Convert hostname to IP address
+    if (IpSocket::addressToIp4(hostname, &m_addr_recv.sin_addr) != SOCK_SUCCESS) {
```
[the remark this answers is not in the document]
LeStarch: Might fix this one.

Drv/Ip/UdpSocket.cpp
LeStarch: Not used anymore?  If so, delete.

LeStarch: nm

LeStarch: This is done to only update receive address once the bind was successful.  Good.

Drv/Ip/UdpSocket.cpp
```diff
@@ -147,18 +155,19 @@
...
         // First IP address to socket sin_addr
         if ((status = IpSocket::addressToIp4(m_hostname, &(address.sin_addr))) != SOCK_SUCCESS) {
+            Fw::Logger::log("Failed to resolve hostname %s: %d\n", m_hostname, static_cast<I32>(status));
             ::close(socketFd);
             return status;
         };
 
         // Now apply timeouts
-        if ((status = IpSocket::setupTimeouts(socketFd)) != SOCK_SUCCESS) {
+        if ((status = this->setupTimeouts(socketFd)) != SOCK_SUCCESS) {
```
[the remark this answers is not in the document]
LeStarch: And fix this.

Drv/Ip/UdpSocket.cpp
```diff
@@ -166,18 +161,23 @@
...
~
~        if (status != SOCK_SUCCESS) {
~            (void) ::close(socketFd); // Closing FD as a retry will reopen send side
~            return status;
~        }
     }
 
     // Log message for UDP
+    char recv_addr[INET_ADDRSTRLEN];
+    if (inet_ntop(AF_INET, &(this->m_addr_recv.sin_addr), recv_addr, INET_ADDRSTRLEN) == nullptr) {
```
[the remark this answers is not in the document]
LeStarch: Again, fix.

LeStarch (changes requested):
Few more cleanup items.

vincewoo:
@LeStarch The latest check-ins include a fix for a latent bug in the UdpSocket implementation for the handling of 0-length datagrams that you pointed out. Due to UDP supporting 0-length data and TCP not technically supporting 0-length data it required a little bit more refactoring to properly detangle the send/recv implementation in IpSocket and allow UdpSocket to override these commands.

Drv/Ip/test/ut/TestUdp.cpp
```diff
@@ -100,6 +99,51 @@ TEST(SingleSide, TestSingleSideSendUdp) {
...
+    Drv::SocketDescriptor send_fd;
+    Drv::SocketDescriptor recv_fd;
+    U16 port = Drv::Test::get_free_port(true);
+    ASSERT_NE(0, port);
+    
+    // Configure receiver and sender
+    ASSERT_EQ(receiver.configureRecv("127.0.0.1", port), Drv::SOCK_SUCCESS);
+    ASSERT_EQ(receiver.open(recv_fd), Drv::SOCK_SUCCESS);
+
+    // Set a custom, longer receive timeout for this specific test to 500ms
```
LeStarch: Curious as to why....

Drv/Ip/test/ut/SocketTestHelper.cpp
```diff
@@ -65,10 +71,23 @@ void receive_all(Drv::IpSocket& receiver, Drv::SocketDescriptor& receiver_fd, U8
...
-        status = receiver.recv(receiver_fd, buffer + received_size, size_in_out);
-        ASSERT_TRUE((status == Drv::SOCK_NO_DATA_AVAILABLE || status == Drv::SOCK_SUCCESS));
-        received_size += size_in_out;
+        U32 bytes_to_request = size - received_size;
+        U32 bytes_actually_received = bytes_to_request; // Will be updated by receiver.recv
+        status = receiver.recv(receiver_fd, buffer + received_size, bytes_actually_received);
+
+        if (status == Drv::SOCK_SUCCESS) {
+            received_size += bytes_actually_received;
+        } else if (status == Drv::SOCK_NO_DATA_AVAILABLE) {
```
LeStarch: The point of this method is to wait (even if no data is available) until such a time as the whole message received (in that the message might span multiple timeouts).

Thus I am confused why this is being commuted to a failure.

Drv/Ip/IpSocket.cpp
```diff
@@ -161,40 +163,53 @@ SocketIpStatus IpSocket::send(const SocketDescriptor& socketDescriptor, const U8
...
-            req_read = static_cast<U32>(size);
-            return SOCK_READ_ERROR;  // Stop recv task on error
+        } else { // bytes_received_or_status == -1, an error occurred
+            if ((errno == EAGAIN) || (errno == EWOULDBLOCK)) {
+                // Non-blocking socket would block, or SO_RCVTIMEO timeout occurred.
+                req_read = 0;
+                return SOCK_NO_DATA_AVAILABLE;
+            } else if (errno == EINTR) {
+                // Interrupted system call. Retry if not the last iteration.
+                if (i == (SOCKET_MAX_ITERATIONS - 1)) {
```
LeStarch: Why doe we need both this and the check at the end?  Can't we just continue again?

Drv/Ip/UdpSocket.cpp
```diff
@@ -186,25 +189,116 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
...
+    if (received >= 0 && this->m_addr_send.sin_port == 0) {
+        this->m_addr_send = sender_addr;
         this->m_port = ntohs(sender_addr.sin_port);
         Fw::Logger::log("Configured send port to %hu as specified by the last received packet.\n", this->m_port);
     }
     return received;
 }
 
+SocketIpStatus UdpSocket::send(const SocketDescriptor& socketDescriptor, const U8* const data, const U32 size) {
+    // Note: socketDescriptor.fd can be -1 in some test cases
```
LeStarch, line 222: Why?  How could this ever be `-1`?

Drv/Ip/UdpSocket.cpp
```diff
@@ -186,25 +189,116 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
...
-        this->m_state->m_addr_send = sender_addr;
+    if (received >= 0 && this->m_addr_send.sin_port == 0) {
+        this->m_addr_send = sender_addr;
         this->m_port = ntohs(sender_addr.sin_port);
         Fw::Logger::log("Configured send port to %hu as specified by the last received packet.\n", this->m_port);
     }
     return received;
 }
 
+SocketIpStatus UdpSocket::send(const SocketDescriptor& socketDescriptor, const U8* const data, const U32 size) {
```
LeStarch, line 221: Why do we need a special send?  Can we not just update the send method on the IpSocket?

It seems odd that:
1. We fixed asserts in the send in IpSocket to allow size==0, but also override the method
2. `recv` is updated to respect 0 length, in Ip, but send is done in UDP.

LeStarch (commented):
I need to understand this comment

```
//TODO: Uncomment FW_ASSERT for socketDescriptor.fd once we fix TcpClientTester to not pass in uninitialized socketDescriptor
```

LeStarch (approved):

PR #3731, opened 2025-06-12
Add CI build for RHEL8 environment

### LINKED ISSUE #3548: Add RHEL8 runner to CI
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

RHEL8 is commonly used as a build host by our users. We've seen that compilers on RHEL8 seem to be catch more issues in test code (see https://github.com/nasa/fprime/pull/3493) than what our current CI suite does.

We need to add a RHEL8 runner in CI to be sure to catch these before we merge in devel

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| [#3548](https://github.com/nasa/fprime/issues/3548) |

---
## Change Description

Fix #3548 

Also standardizes a little more CI job names so they're nicely sorted

## Rationale

RHEL8 compilers seem to catch more warnings than our current CI suite. Some F Prime users use RHEL8, so it's good to test on that.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -152,6 +152,7 @@ DHTML
 diafile
 diles
 dinkel
+dnf
 dnp
 docbook
 docset
@@ -527,6 +528,7 @@ rcvd
 rdwr
 Readback
 Recvd
+redhat
 refspec
 REFTOPOLOGY
 REFTOPOLOGYDEFS
@@ -535,6 +537,7 @@ REMOVEDIRECTORY
 REMOVEFILE
 reprioritize
 RGD
+rhel
 RHH
 Rizvi
 ROOTDIR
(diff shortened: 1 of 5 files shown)


thomas-bc:
Confirmed that this set up would have caught a warning like the one fixed in https://github.com/nasa/fprime/pull/3493

thomas-bc:
BufferLogger UTs seem to fail in a docker container. I was able to reproduce in a rhel8 container. But running on native rhel8 passed ok. I wonder if the UTs (or component itself) has something that doesn't play nice with being in a container... wil investigate

thomas-bc:
So, for posterity ....

The issue had to do with filesystem permissions within a container. What's really weird is that the issue only appeared when running the RHEL8 container on a linux host. The same container on a macOS host would behave as expected.

Basically, `chmod -w dir` would not successfully remove permissions of the root user to create files in `dir`. The fix is to use a non-root user to run UTs. Which arguably is the way to do CI anyways, but the only official RedHat images I've found do not include a non-root end user. So I've added the creation of the user in the job itself.

LeStarch (approved):
LGTM

PR #3753, opened 2025-06-17
PRIVATE -> private for Fw/Dp

m-aleem:
| | |
|:---|:---|
|**_Has Unit Tests (y/n)_**| Yes - existing |

---
## Change Description

Updates to `fprime/Fw/Dp*` to change:

* `PRIVATE` -> `private`
* `PROTECTED` -> `protected`
* And additional corresponding updates, as needed

## Rationale

#3446

N/A

## Future Work

Additional PRs to be opened for further updates corresponding to #3446


LeStarch (dismissed):
I don't like this style as much as your other PRs.  I prefer to have the tester control access to the private members' usage, rather than just adding in getters.

For now I am approving, because it isn't wrong to do it this way, just feels less clean.

Fw/Dp/test/ut/DpContainerTester.hpp
```diff
@@ -0,0 +1,41 @@
...
+// ======================================================================
+
+#ifndef Fw_DpContainerTester_HPP
+#define Fw_DpContainerTester_HPP
+
+#include "Fw/Dp/DpContainer.hpp"
+
+namespace Types {
+
+    class DpContainerTester {
```
thomas-bc, line 14: Should this be namespaced `Fw::DpContainerTester` instead? Or is that the pattern we've been going with for testing types in `Fw` ?

m-aleem: No, I think you're correct this should be in the Fw namespace. I will update this

m-aleem:
> I don't like this style as much as your other PRs. I prefer to have the tester control access to the private members' usage, rather than just adding in getters.
> 
> For now I am approving, because it isn't wrong to do it this way, just feels less clean.

@LeStarch Updated the Tester to actually be doing some of the checks, ready for re-review

Fw/Dp/test/ut/DpContainerTester.hpp
m-aleem: Addressed in [e0e40eb](https://github.com/nasa/fprime/pull/3753/commits/e0e40eb7886266f9c43f9a60e1ede529a337ae47)

LeStarch (approved):

PR #3771, opened 2025-06-20
CMake: Copy hashes.txt to build-artifact

### LINKED ISSUE #3283: Copy hashes.txt to build-artifacts
| | |
|:---|:---|
|**_F´ Version_**| 3.6.1 |
|**_Affected Component_**| fprime |
---
## Feature Description

hashes.txt is an artifact that we want inside of build-artifacts.

valdaarhun:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3283 |

---
## Change Description

Modify `install.cmake` to copy hashes.txt to build-artifacts.

## Rationale

Fixes #3283.

Testing procedure
```sh
$ fprime-util generate
[...]
$ fprime-util build
[...]
$ diff build-fprime-automatic-native/hashes.txt build-artifacts/hashes.txt # No difference
$
```

## Future Work

NA


cmake/target/install.cmake
```diff
@@ -56,6 +56,7 @@ function(install_add_deployment_target MODULE TARGET SOURCES DEPENDENCIES FULL_D
     install(FILES ${FPRIME_CURRENT_DICTIONARY_FILE} ${FPRIME_CURRENT_DICTIONARY_FILE_JSON} DESTINATION ${TOOLCHAIN_NAME}/${MODULE}/dict COMPONENT ${MODULE})
     add_custom_command(TARGET "${MODULE}" POST_BUILD COMMAND "${CMAKE_COMMAND}"
             -DCMAKE_INSTALL_COMPONENT=${MODULE} -P ${CMAKE_BINARY_DIR}/cmake_install.cmake)
+    file(COPY "${CMAKE_BINARY_DIR}/hashes.txt" DESTINATION "${CMAKE_INSTALL_PREFIX}")
```
LeStarch: Can we add this to the line 56 install FILES?  The current command runs before the build and may be in an incomplete state.

This is a good idea, and definitely a needed fix.

LeStarch (changes requested):
I am

cmake/target/install.cmake
valdaarhun: Hi. Thank you for your review. I added another `install(FILES...)` line just below line 56 since the destination is different and I don't think multiple destinations can be specified in the same `install(FILES...)` command.

LeStarch (approved):

LeStarch:
There was a merge conflict so I went ahead and fixed that (and the regression I introduced while fixing it).  Should be good once CI passes.

PR #3772, opened 2025-06-20
Updates for Svc/PrmDb corresponding to UT cmake update for protected/private (#3446)

m-aleem:
| | |
|:---|:---|

---
## Change Description

I locally updated `fprime/cmake/settings.cmake` to change `-DPROTECTED=protected -DPRIVATE=private`. This causes additional build errors beyond all our updates in manual code for PRIVATE->private and PROTECTED->protected due to auto-coded files which use PRIVATE, PROTECTED. This set of changes is associated with fixing `Svc/PrmDb` so that the UTs in this directory can build and pass.

Approach in this case was:

1. mv PrmDbTester.cpp PrmDbTestMain.cpp
2. mv PrmDbImplTester.hpp  PrmDbTester.hpp
3. mv PrmDbImplTester.cpp  PrmDbTester.cpp
4. In UTs, change PrmDbImplTester to PrmDbTester where applicable
5. Make class PrmDbTester a friend of PrmDbImpl

## Rationale

#3446

## Future Work

1) Note that this PR **does not** include the update to `fprime/cmake/settings.cmake` itself - this will be done at the end
2) Note that this PR **does not** include re-naming the flight code `<X>Impl` to `<X>` (separate issue)


m-aleem:
@LeStarch From my initial poking, the updates to use UT_AUTO_HELPERS seems like it would require enough further re-work of the existing Testers that I think it would be worth making that its own issue/task.

Svc/PrmDb/test/ut/PrmDbImplTester.hpp
```diff
@@ -1,78 +1,50 @@
...
 
-    class PrmDbImplTester : public PrmDbGTestBase {
+    /**
+     * @brief Wrapper class for PrmDbImpl to expose private methods for testing
+     *
+     * This class exists to provide access to private methods of PrmDbImpl without
+     * requiring additional friendship declarations. It leverages the existing
+     * friendship between PrmDbImpl and PrmDbImplTester.
+     */
+    class PrmDbImplTester {
```
LeStarch: Wasn't I expecting this to be renamed PrmDbTester?  Or did I misunderstand?

Svc/PrmDb/test/ut/PrmDbTestMain.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+#include <gtest/gtest.h>
+
+#include <Fw/Test/UnitTest.hpp>
+
+#if FW_OBJECT_REGISTRATION == 1
+static Fw::SimpleObjRegistry simpleReg;
+#endif
+
+void connectPorts(Svc::PrmDbImpl& impl, Svc::PrmDbTester& tester) {
```
LeStarch, line 20: We discussed trying to use UT_AUTO_HELPERS

Svc/PrmDb/test/ut/PrmDbImplTester.hpp
LeStarch: It looks like this may be entirely unused.  Perhaps we can remove?

LeStarch (changes requested):

Svc/PrmDb/test/ut/PrmDbTestMain.cpp
m-aleem: See prev comment in PR, I think it's worthwhile make this its own update task

Svc/PrmDb/test/ut/PrmDbImplTester.hpp
m-aleem: We now have both `PrmDbTester` and `PrmDbImplTester`. And what *used to be* PrmDbTester.cpp is now the TestMain file.

`PrmDbTester` is a friend of `PrmDbComponentBase`
`PrmDbImplTester` is a friend of `PrmDbImpl`

We use `PrmDbTester`/`m_impl` to call `doDispatch` which is declared private in PrmDbComponentBase
We use `PrmDbImplTester`/`m_implTester` to call `clearDb` which is declared private in PrmDbImpl

m-aleem: We now have both `PrmDbTester` and `PrmDbImplTester`. (And what *used to be* PrmDbTester.cpp is now the TestMain file.)

- `PrmDbTester` is a friend of `PrmDbComponentBase`
    - We use `PrmDbTester` to call `doDispatch` which is declared private in PrmDbComponentBase (in PrmDbComponentAc.hpp) by doing `this->m_impl.doDispatch()` from PrmDbTester
- `PrmDbImplTester` is a friend of `PrmDbImpl`.
    - We use `PrmDbImplTester` to call `clearDb` which is declared private in PrmDbImpl ([here](https://github.com/nasa/fprime/blob/0ec96fbea7a98d94e5f8d7cc199fc7332a8b2239/Svc/PrmDb/PrmDbImpl.hpp#L63-L108)) by doing `this->m_implTester.clearDb()` from PrmDbTester.
    - We will get errors if trying to do `this->m_impl.clearDb()`

m-aleem:
I am going to move this back to draft to fix the naming inconsistencies. Will try to discuss with CogE on desired name and file pattern going forward.

Svc/PrmDb/test/ut/PrmDbTestMain.cpp
LeStarch: Ok, works for me.

LeStarch (approved):

PR #3794, opened 2025-06-24
Use concurrency groups in CI to prevent overfilling the queue

### LINKED ISSUE #3793: Use concurrency groups in CI
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency

This should allow for PRs not to queue up multiple jobs when multiple commits are pushed in a short time span. Should help reduce the slow queueing up of jobs we've been seeing.

**Goal**: on each PR, only allow a single job to run, on the latest-pushed commit. Cancel previous jobs if a newer commit has been pushed.

Note:
- do NOT enable this on `devel`, as we'd like to prove out every commit on that branch.
- The [spell-checking workflow is already doing this](https://github.com/nasa/fprime/blob/1e0ea6cb20604cf7bae0fe9e75e6fb3fd646f54d/.github/workflows/spelling.yml#L85-L88)

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/3793

This should help with not overloading the CI queue too much, by cancelling previous runs if a newer commit has been pushed to a PR branch. The cancelling does not apply to `devel` or `release/` branches.


thomas-bc:
Shown in the commit above - all runs on the `Rephrase 20dc12b` commit have been cancelled once the merge commit was pushed.

image

LeStarch (approved):
Why not cancel devel jobs too?  Not a blocker.

thomas-bc:
This is simply to get successful checks on every commit on `devel`. History would "appear" questionable if we start getting lots of failure-looking checks simply because they're cancelled. I agree it'd be more efficient to cancel on devel as well... but I kind of like the sea of green checkmarks on devel haha

PR #3802, opened 2025-06-25
Switch Fw::Buffer size to FwSizeType instead of U32

### LINKED ISSUE #3782: Fw::Buffer size should use FwSizeType instead of U32
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Submitted as part of https://github.com/nasa/fprime/discussions/3599 and approved for implementation

Fw::Buffer size should use FwSizeType instead of U32

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix #3782

### DIFF
--- a/Drv/LinuxUartDriver/LinuxUartDriver.hpp
+++ b/Drv/LinuxUartDriver/LinuxUartDriver.hpp
@@ -67,7 +67,7 @@ class LinuxUartDriver final : public LinuxUartDriverComponentBase {
     enum UartParity { PARITY_NONE, PARITY_ODD, PARITY_EVEN };
 
     // Open device with specified baud and flow control.
-    bool open(const char* const device, UartBaudRate baud, UartFlowControl fc, UartParity parity, U32 allocationSize);
+    bool open(const char* const device, UartBaudRate baud, UartFlowControl fc, UartParity parity, FwSizeType allocationSize);
 
     //! start the serial poll thread.
     //! buffSize is the max receive buffer size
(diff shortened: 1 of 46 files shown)


Drv/LinuxUartDriver/LinuxUartDriver.hpp
```diff
@@ -104,7 +104,7 @@ class LinuxUartDriver final : public LinuxUartDriverComponentBase {
                                 ) override;
 
     int m_fd;  //!< file descriptor returned for I/O device
-    U32 m_allocationSize; //!< size of allocation request to memory manager
+    FwSizeType m_allocationSize; //!< size of allocation request to memory manager
```
thomas-bc, line 107: Fix in `open()` call above

Fw/Buffer/Buffer.hpp
```diff
@@ -73,7 +73,7 @@ class Buffer : public Fw::Serializable {
     //! \param data: data pointer to wrap
     //! \param size: size of data located at data pointer
     //! \param context: user-specified context to track creation. Default: no context
-    Buffer(U8* data, SizeType size, U32 context = NO_CONTEXT);
+    Buffer(U8* data, FwSizeType size, U32 context = NO_CONTEXT);
```
thomas-bc, line 76: leave this SizeType so it's consistent

thomas-bc: Went the other way and switched all Fw::Buffer code to directly reference FwSizeType so it's easier to understand in IDEs etc

LeStarch, line 76: I meant everything should use 'SizeType' instead as that would be consistent with all uses internal and external to the file.

If we are going this route (fine) but we should consider deleting SizeType herein....which would cause many problems.

LeStarch (approved):

PR #3807, opened 2025-06-26
Updates for Svc/FpySquencer corresponding to UT cmake update for protected/private

m-aleem:
| | |
|:---|:---|

---
## Change Description

I locally updated `fprime/cmake/settings.cmake` to change `-DPROTECTED=protected -DPRIVATE=private`. This causes additional build errors beyond all our updates in manual code for PRIVATE->private and PROTECTED->protected due to auto-coded files which use PRIVATE, PROTECTED. This set of changes is associated with fixing `Svc/FpySquencer` so that the UTs in this directory can build and pass.

Approach in this case was:

1) Add accessor methods to FpySquencerTester
2) Implementing new friend of the statemachine class ** !! This will require FPP update !! **

## Rationale

#3446

## Future Work

1. Note that this PR **does not** include the update to fprime/cmake/settings.cmake itself - this will be done at the end
2. Note that this PR **does not** include re-naming the flight code <X>Impl to <X> (separate issue)


m-aleem:
Closing in favor of https://github.com/nasa/fprime/pull/3812

PR #3810, opened 2025-06-26
Updates for FPPTest corresponding to UT cmake update for protected/private

m-aleem:
| | |
|:---|:---|

---
## Change Description

I locally updated `fprime/cmake/settings.cmake` to change `-DPROTECTED=protected -DPRIVATE=private`. This causes additional build errors beyond all our updates in manual code for PRIVATE->private and PROTECTED->protected due to auto-coded files which use PRIVATE, PROTECTED. This set of changes is associated with fixing `Fpp` so that the UTs in this directory can build and pass.

Approach/changes in this case was:

1) In components, re-name existing `Tester.hpp/cpp` to follow friendship name pattern and allow private/protected access (example: `ActiveTestTester.hpp/cpp`) then update the existing `Tester.hpp` to point to the new file in order to keep the test structure
2) Re-name Fpp/dp tests to DpTestTester to follow friendship name pattern and allow private/protected access
3) Refactor all `FppTest/state_machine/internal_instance/state` files to move tests into their own `<x>Tester.hpp/cpp` classes, again to follow friendship name pattern and allow private/protected access

## Rationale

#3446

## Future Work

1. Note that this PR **does not** include the update to fprime/cmake/settings.cmake itself


LeStarch:
@bocchino and @Kronos3 would one of you review

FppTest/component/active/test/ut/Tester.hpp
```diff
@@ -1,258 +1,9 @@
...
-
-    // ----------------------------------------------------------------------
-    // Parameter delegates
-    // ----------------------------------------------------------------------
-
-    //! Delegate to serialize/deserialize an externally stored parameter
-    ActiveTestComponentBaseParamExternalDelegate paramTesterDelegate;
-
-};
+// ActiveTestTester is the full implementation - using it as the Tester class
```
LeStarch, line 4: Why do we do this instead of removing the Tester class, and this header?

m-aleem: The cpp files/tests in the `FppTest/component/tests/` path do `#include Tester.hpp` since I believe the (same) tests are done across the various component types, so keeping the Tester.hpp in each component but pointing it to the now-new .hpp allows us to use this existing structure

Example: https://github.com/nasa/fprime/blob/devel/FppTest/component/tests/ExternalParamTests.cpp#L13

m-aleem:
@bocchino Incorporated the changes we discussed at todays tag up, this should be ready for-review/merge (and, once merged, the additional steps to coordinate F-Prime/FPP for this task are in a comment in https://github.com/nasa/fpp/pull/754)

FppTest/component/active/test/ut/Tester.hpp
bocchino: I agree with this approach.

bocchino (approved):
Looks good! I noticed that some of the changes required reformatting, so I reformatted all the code. That caused major changes in some files that had not previously been formatted in the standard way.

@LeStarch @Kronos3 if you want to review this further, I can make a branch of nasa/devel with just the reformatting, for comparison. Let me know.

Issue #3817, opened 2025-06-26
FpySequencer floating point comparisons


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**|  FpySequencer |
---
## Feature Description

Add floating point comparison directives to Fpy.

This was a silly omission in the V0.3 PR.

PR #3820, opened 2025-06-27
FpySequencer floating point comparisons

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**|  #3817 |

---
## Change Description

Adds floating point <, >, <=, >=, == and != operators. 

## Rationale

Users of Fpy need to be able to compare floating point values, not just integers.


zimri-leisher:
@LeStarch  looks like it's failing CI due to some nondeterminism, can you take a look?

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -423,55 +426,72 @@ Signal FpySequencer::binaryCmp_directiveHandler(const FpySequencer_BinaryCmpDire
...
-        sign = false;
-    }
-
     I8 cmpResult;
 
-    if (sign) {
-        cmpResult = (lhs == rhs) ? 0 : (lhs < rhs) ? -1 : 1;
+    if (directive.get_op() >= Fpy::DirectiveId::FEQ && directive.get_op() <= Fpy::DirectiveId::FGE) {
+        F64 frhs, flhs;
+        memcpy(&frhs, &rhs, sizeof(frhs));
```
LeStarch: This seems wrong.  Why isn't the directive using a union type if it has multiple sub-types?  Otherwise, why not use `reinterpret_cast<F64>()` to do assignment.  `memcpy` seems like the most hacky solution.

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -423,55 +426,72 @@ Signal FpySequencer::binaryCmp_directiveHandler(const FpySequencer_BinaryCmpDire
...
     if (cmpResult == 0) {
         // values were equal
         // result is true if equality is okay
-        res = (directive.get_op() == Fpy::DirectiveId::UGE 
-            || directive.get_op() == Fpy::DirectiveId::ULE 
-            || directive.get_op() == Fpy::DirectiveId::SGE 
-            || directive.get_op() == Fpy::DirectiveId::SLE);
+        res = (directive.get_op() == Fpy::DirectiveId::UGE || directive.get_op() == Fpy::DirectiveId::ULE ||
+               directive.get_op() == Fpy::DirectiveId::SGE || directive.get_op() == Fpy::DirectiveId::SLE ||
+               directive.get_op() == Fpy::DirectiveId::FGE || directive.get_op() == Fpy::DirectiveId::FLE);
```
LeStarch: You've already handled `FEQ` and `FNE` above, why handle the GE and LE cases here?

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -423,55 +426,72 @@ Signal FpySequencer::binaryCmp_directiveHandler(const FpySequencer_BinaryCmpDire
...
-    if (directive.get_op() >= Fpy::DirectiveId::ULT && directive.get_op() <= Fpy::DirectiveId::UGE) {
-        sign = false;
-    }
-
     I8 cmpResult;
 
-    if (sign) {
-        cmpResult = (lhs == rhs) ? 0 : (lhs < rhs) ? -1 : 1;
+    if (directive.get_op() >= Fpy::DirectiveId::FEQ && directive.get_op() <= Fpy::DirectiveId::FGE) {
+        F64 frhs, flhs;
```
LeStarch: This code in general would benefit from some helper functions.  e.g.  break out the float processing, the signed processing, and the unsigned processing from one another.

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -423,55 +426,72 @@ Signal FpySequencer::binaryCmp_directiveHandler(const FpySequencer_BinaryCmpDire
...
+        } else if (std::isgreater(flhs, frhs)) {
+            cmpResult = 1;
+        } else if (std::isless(flhs, frhs)) {
+            cmpResult = -1;
+        } else {
+            cmpResult = 0;
+        }
+
+        // handle the floating equality cmps
+        if (directive.get_op() == Fpy::DirectiveId::FEQ) {
```
LeStarch: Why can't FEQ and FNE not be handled by  the EQ or NE portions above?

Svc/FpySequencer/FpySequencerDirectives.cpp
LeStarch: This implies all floats are F64?

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -292,6 +292,12 @@ Fw::Success FpySequencer::deserializeDirective(const Fpy::Statement& stmt, Direc
         case Fpy::DirectiveId::SLT:
         case Fpy::DirectiveId::SLE:
         case Fpy::DirectiveId::SGE: 
+        case Fpy::DirectiveId::FEQ:
```
LeStarch, line 295: Why do floats need special directives for equality, but not signed/unsigned.

LeStarch (commented):
In general it looks good, but I did wonder about some things.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: 1) the directive is an FPP type and FPP doesn't have union types
2) good idea, not really sure why I chose memcpy here... strange decision I agree

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: yeah sorry this code sucks... it started out quite simple but I agree, it's too much. I'm going to rewrite it with some helper functions.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: at the moment yes. for F32s, the bytecode will in the future need to implement an F32<>F64 conversion op. in fact, I was planning on doing that later but I'm realizing I might need that now, otherwise my team may be unable to use F32s at all.

Svc/FpySequencer/FpySequencerRunState.cpp
zimri-leisher: good question--float equality is complicated by the fact that if one of them is NAN, equality is always false. (any float op with NAN as an arg is false). also, there are some values whose bits differ (-0 and +0) which should be equal.
this means we can't just compare the bits of the floats, we have to interpret them as floats first. so floats definitely need their own eq ops.

signed and unsigned equality is more subtle. in general, bitwise comparison is okay, because if two integers of the same bit width have different bits, they are definitely not equal. so what happens if two integers have the same bits? should it depend on if one is signed or unsigned? for instance, should (signed) 0b11111111 (-1) be the equal to (unsigned) 0b11111111 (255)? well, i guess those values when interpreted lead to different numbers, so they shouldn't be equal. but the solution to this shouldn't be in the bytecode. if i just added "unsigned" equality and "signed" equality, if you think about it for a moment, you can see both of those ops would just do the same thing: check if bits are equal. really, this kind of thing has to be enforced at the compiler level. At the moment, I do not handle this in any special way, so I do say that (-1) signed == (255) unsigned. This is consistent with C++ but maybe unexpected and could lead to problems. I will make a ticket to fix this.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: TBD on this comment, pending cleanup

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: actually @LeStarch I know why I didn't use reinterpret cast, it seems people online say it is unsafe in this situation. due to memory alignment issues. i don't know if I buy that, but here's a thread: https://stackoverflow.com/questions/13982340/is-it-safe-to-reinterpret-cast-an-integer-to-float

zimri-leisher:
@LeStarch  this PR is WIP, missing a couple UTs

Svc/FpySequencer/FpySequencerDirectives.cpp
LeStarch: U64 and F64 should have the same alignment, but let us not poke the sleeping bear.  `memcpy` is sufficient here.

Svc/FpySequencer/FpySequencerDirectives.cpp
LeStarch: Is this PR blocking on the helper functions?

Svc/FpySequencer/FpySequencerDirectives.cpp
LeStarch: Do you need F32s?  The language specification guarantees that `double` (aka F64) will exist.  Is there need for dealing with F32s in sequences at all?

Svc/FpySequencer/FpySequencerRunState.cpp
LeStarch: I follow.  Will resolve.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: yes, and now I have finished the helpers, but I'm missing a few UTs

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: yes--telemetry points/channels/command args might be F32s, so I have to at least be able to pass them around. I think it would be prudent to support all the primitive types, also.

zimri-leisher:
@LeStarch  this PR is ready for review again, I think it's good to go as well. code is much cleaner now

LeStarch (approved):
Much, MUCH easier to review.  Thanks!

Issue #3783, opened 2025-06-23
Make Fw.Time and Fw.TimeInterval FPP structs


thomas-bc:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Submitted as part of https://github.com/nasa/fprime/discussions/3599 and approved for implementation

Make Fw.Time and Fw.TimeInterval FPP structs

PR #3834, opened 2025-07-01
Make Fw::TimeInterval an FPP Struct

m-aleem:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3783   |

---
## Change Description

1. Updated Fw::TimeInterval to use FPP Struct, as defined in Fw/Time.fpp.
    a. Defined an FPP struct for ~~TimeIntervalType~~ TimeIntervalValue
    ~~b. The (existing) TimeInterval class now derives from TimeIntervalType and extends it with the additional methods that were already in TimeInterval (that are not included by the autocoder)~~
    c. As discussed with CogE, the TimeInterval class interface is unchanged but the underlying representation of the Time value uses this new FPP-defined struct
2. Added additional UTs for TimeInterval
3. Split up the Time test files into separate tester files for TimeTest, TimeIntervalTest, and a driver TimeTestMain
4. Updated CMake files to resolve build errors

## Rationale

#3783

Added UTs for TimeInterval.

## Future Work

Make Fw::Time an FPP Struct


Fw/Time/TimeInterval.hpp
```diff
@@ -3,42 +3,40 @@
...
-//! for easy storage and transmission.
+//! to manipulate and compare time intervals. It derives from the auto-generated TimeIntervalType
+//! which supports serialization and extends it with additional functionality.
 //!
+
 namespace Fw {
 
-    class TimeInterval: public Serializable {
+    // Extend the auto-generated class with additional functionality
+    class TimeInterval : public TimeIntervalType {
```
LeStarch: My concern is this, if we were to send this type through a port (as TimeIntervalType) the class could be [sliced](https://en.wikipedia.org/wiki/Object_slicing).  While not a problem now, because there are no additional members, it provides an easy way to cause us problems in the future.

I would recommend this:

1. Rename the model (FPP type) as "TimeInterval"
2. Make your existing subclass a utility class that constructs around a time interval and provides the additional operations delegating to the TimeInterval internal memeber.

e.g.
```c++
class TimeIntervalCalculator {
    TimeIntervalCalculator(TimeInterval& my_interval) {m_wrapped_interval = my_interval}

    TimeInterval& add(Secs, Usecs) {
        this->m_wrapped_interval.setseconds...
    } 
}
```

LeStarch: This way, users who need the extra calculation functions can "add on" functionality through composition rather than inheritance....as the inheritance could cause issues in the future (should we add members to the derived class).

LeStarch (changes requested):
Strong start but I recommend a different architecture for "adding in" functions.

Fw/Time/TimeInterval.cpp
```diff
@@ -3,78 +3,66 @@
...
     SerializeStatus TimeInterval::serialize(SerializeBufferBase& buffer) const {
-        // serialize members
-        SerializeStatus stat = Fw::FW_SERIALIZE_OK;
-        stat = buffer.serialize(this->m_seconds);
-        if (stat != FW_SERIALIZE_OK) {
-            return stat;
-        }
-        return buffer.serialize(this->m_useconds);
+        // Use TimeIntervalValue's built-in serialization
+        return this->m_val.serialize(buffer);
```
LeStarch, line 52: Is this serialization format the same?  I think so, but could we confirm?

LeStarch (approved):
Looks good, just had one confirmation question.

Fw/Time/TimeInterval.cpp
m-aleem: Yes, in both cases (old hand-code vs new auto-code) the order is:

* Serialize seconds
* Return bad status if status is not OK
* Serialize useconds
* Return status (good/bad) from usecond serialization


Old Hand Implementation
```C++
SerializeStatus TimeInterval::serialize(SerializeBufferBase& buffer) const {
    // serialize members
    SerializeStatus stat = Fw::FW_SERIALIZE_OK;
    stat = buffer.serialize(this->m_seconds);
    if (stat != FW_SERIALIZE_OK) {
        return stat;
    }
    return buffer.serialize(this->m_useconds);
}

SerializeStatus TimeInterval::deserialize(SerializeBufferBase& buffer) {

    SerializeStatus stat = Fw::FW_SERIALIZE_OK;
    stat = buffer.deserialize(this->m_seconds);
    if (stat != FW_SERIALIZE_OK) {
        return stat;
    }

    return buffer.deserialize(this->m_useconds);
}
```

New Autocode Implementation
```C++
  Fw::SerializeStatus TimeIntervalValue ::
    serialize(Fw::SerializeBufferBase& buffer) const
  {
    Fw::SerializeStatus status;

    status = buffer.serialize(this->m_seconds);
    if (status != Fw::FW_SERIALIZE_OK) {
      return status;
    }
    status = buffer.serialize(this->m_useconds);
    if (status != Fw::FW_SERIALIZE_OK) {
      return status;
    }

    return status;
  }

  Fw::SerializeStatus TimeIntervalValue ::
    deserialize(Fw::SerializeBufferBase& buffer)
  {
    Fw::SerializeStatus status;

    status = buffer.deserialize(this->m_seconds);
    if (status != Fw::FW_SERIALIZE_OK) {
      return status;
    }
    status = buffer.deserialize(this->m_useconds);
    if (status != Fw::FW_SERIALIZE_OK) {
      return status;
    }

    return status;
  }
```

LeStarch: Excellent, thank you for confirming!

PR #3840, opened 2025-07-02
Linux should not default to stubbed drivers

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Linux should not default to stubbed drivers.  This was a regression.


chuynh4duarte (approved):
fixed status return  "Failed to open GPIO pin: 7" in fprime-workshop-led-blinker

kevin-f-ortega (approved):

Issue #3839, opened 2025-07-02
FpySequencer F32/F64 conversion


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**| FpySequencer |
---
## Feature Description

Add a bytecode op to convert an F32 to an F64 and vice versa

## Rationale

This will allow users to use the F64 floating point instructions with F32s

PR #3845, opened 2025-07-03
Add fpext, fptrunc directives

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3839  |

---
## Change Description

Adds FPEXT and FPTRUNC ops, which allow you to "extend" an F32 in a register to an F64, and "truncate" an F64 in a register into an F32. This allows the bytecode to start using F32s instead of just F64s.


Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -403,80 +404,92 @@ I8 floatCmp(F64 lhs, F64 rhs) {
...
-I64 FpySequencer::binaryCmp_sge(I64 lhs, I64 rhs) {
+I64 FpySequencer::binaryRegOp_sge(I64 lhs, I64 rhs) {
     return lhs >= rhs;
 }
-I64 FpySequencer::binaryCmp_feq(I64 lhs, I64 rhs) {
-    F64 left = *reinterpret_cast<F64*>(&lhs);
-    F64 right = *reinterpret_cast<F64*>(&rhs);
+I64 FpySequencer::binaryRegOp_feq(I64 lhs, I64 rhs) {
+    F64 left;
+    memcpy(&left, &lhs, sizeof(left));
```
zimri-leisher, line 445: I had some alignment issues, so I changed to memcpy because it's safer around alignment.

LeStarch (approved):
Surfing that new clean structure!  I want to get into the I64/F64 tradeoffs more and understand why this is done, as the code has the same signature as code we typically avoid.

However, this PR only make that less error prone, so I can defer recommendations there.

Looking good!

Issue #3846, opened 2025-07-04
FpySequencer int<>float conversion


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**|4.0.0 |
|**_Affected Component_**| FpySequencer |
---
## Feature Description

A directive to convert a float to an int and back

## Rationale

This would allow comparing or doing math with floats and ints in Fpy

kevin-f-ortega:
This is Done!

PR #3847, opened 2025-07-04
Add float to int conversion directives, some misc changes

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**|  #3846|

---
## Change Description

* Adds the FPTOI (floating point to integer) and ITOFP directives, which take 64 bit ints and floats and convert between them as if by static cast.

* I also added a small amount of debugging telemetry to certain deserialization/file read errors. This was helpful to me for debugging problems and I think would be helpful to users too.

* I also updated the default register count to 128. This equates to 1024 bytes of memory used for registers. Without lots of registers, many sequences do not run, because currently I do not perform any register allocation optimization. In the next large Fpy update I will remove the concept of statically allocated registers and replace them with a stack, so this constant will go away. For now, in order to be usable out of the box, the register count needs to be high.

* I also fixed a bug in the NOT directive. Previously, it was implemented to take the bitwise complement of whatever was given to it. However, this produced incorrect behavior when trying to get the inverse of the Fprime "TRUE" value (0xFF == 255), as it would result in -256 (0xFFFFFFFFFFFFFF00). Now, it interprets the input register as a C++ boolean, and if true it outputs `static_cast<I64>(false)`, otherwise `true`. This gives the correct behavior under all situations. 

## Rationale

This will allow users of Fpy to compare and perform math where one operand is a float and the other is an int. Otherwise, you wouldn't be able to check if `5.0 > 4`.

### DIFF
--- a/Svc/FpySequencer/FpySequencer.fpp
+++ b/Svc/FpySequencer/FpySequencer.fpp
@@ -13,6 +13,12 @@ module Svc {
             IDLE
         }
 
+        enum FileReadStage {
+            HEADER
+            BODY
+            FOOTER
+        }
+
         include "FpySequencerCommands.fppi"
         include "FpySequencerTelemetry.fppi"
         include "FpySequencerEvents.fppi"
--- a/Svc/FpySequencer/FpySequencer.hpp
+++ b/Svc/FpySequencer/FpySequencer.hpp
@@ -648,6 +648,8 @@ class FpySequencer : public FpySequencerComponentBase {
     I64 unaryRegOp_not(I64 src);
     I64 unaryRegOp_fpext(I64 src);
     I64 unaryRegOp_fptrunc(I64 src);
+    I64 unaryRegOp_fptoi(I64 src);
+    I64 unaryRegOp_itofp(I64 src);
 
     Signal exit_directiveHandler(const FpySequencer_ExitDirective& directive, DirectiveError& error);
 };
--- a/Svc/FpySequencer/FpySequencer.hpp
+++ b/Svc/FpySequencer/FpySequencer.hpp
@@ -23,8 +23,7 @@
 
 static_assert(Svc::Fpy::MAX_SEQUENCE_ARG_COUNT <= std::numeric_limits<U8>::max(),
               "Sequence arg count must be below U8 max");
-static_assert(Svc::Fpy::NUM_REGISTERS <= std::numeric_limits<U8>::max(),
-              "Register count must be below U8 max");
+static_assert(Svc::Fpy::NUM_REGISTERS <= std::numeric_limits<U8>::max(), "Register count must be below U8 max");
 static_assert(Svc::Fpy::MAX_SEQUENCE_STATEMENT_COUNT <= std::numeric_limits<U16>::max(),
               "Sequence statement count must be below U16 max");
 static_assert(Svc::Fpy::MAX_SERIALIZABLE_REGISTER_SIZE <= std::numeric_limits<FwSizeType>::max(),
@@ -41,8 +40,7 @@ using State = FpySequencer_SequencerStateMachineStateMachineBase::State;
 using DirectiveError = FpySequencer_DirectiveErrorCode;
 
 class FpySequencer : public FpySequencerComponentBase {
-
-  friend class FpySequencerTester;
+    friend class FpySequencerTester;
 
   public:
     union DirectiveUnion {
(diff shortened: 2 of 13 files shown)


Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -570,7 +564,10 @@ class FpySequencer : public FpySequencerComponentBase {
     // updates the CRC by default, but can be turned off if the contents
     // aren't included in CRC.
     // return success if successful
-    Fw::Success readBytes(Os::File& file, FwSizeType readLen, bool updateCrc = true);
+    Fw::Success readBytes(Os::File& file,
+                          FwSizeType readLen,
+                          const FpySequencer_FileReadStage& readStage,
```
zimri-leisher, line 569: added "FileReadStage" to this method so that it would be able to generate better warning events

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -104,9 +106,9 @@ Fw::Success FpySequencer::readHeader() {
     // deser header
     Fw::SerializeStatus deserStatus = this->m_sequenceBuffer.deserialize(this->m_sequenceObj.getheader());
     if (deserStatus != Fw::SerializeStatus::FW_SERIALIZE_OK) {
-        this->log_WARNING_HI_FileReadDeserializeError(this->m_sequenceFilePath, static_cast<I32>(deserStatus),
-                                                      this->m_sequenceBuffer.getBuffLeft(),
-                                                      this->m_sequenceBuffer.getBuffLength());
+        this->log_WARNING_HI_FileReadDeserializeError(
```
zimri-leisher, line 109: just added the filereadstage to these events

Svc/FpySequencer/FpySequencer.fpp
```diff
@@ -13,6 +13,12 @@ module Svc {
             IDLE
         }
 
+        enum FileReadStage {
```
zimri-leisher, line 16: this is just used for warning events to make debugging easier

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -565,7 +566,10 @@ Signal FpySequencer::binaryRegOp_directiveHandler(const FpySequencer_BinaryRegOp
     return Signal::stmtResponse_success;
 }
 I64 FpySequencer::unaryRegOp_not(I64 src) {
-    return ~src;
+    if (src) {
```
zimri-leisher, line 569: `not` behavior changed to work better with C++ if statement

Svc/FpySequencer/test/ut/FpySequencerTestSequences.cpp
```diff
@@ -113,5 +113,22 @@ TEST_F(FpySequencerTester, CmpIntTlm) {
     dispatchUntilState(State::IDLE);
     ASSERT_EQ(tester_get_m_statementsDispatched(), 8);
 }
+TEST_F(FpySequencerTester, NotTrueSeq) {
+    // this sequence caught one bug
+    allocMem();
+
+    add_SET_REG(0, 255);
```
zimri-leisher, line 120: with the previous NOT behavior, this test would fail. now it succeeds

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -594,10 +598,41 @@ I64 FpySequencer::unaryRegOp_fptrunc(I64 src) {
...
+    memcpy(&res, &fsrc, sizeof(res));
+    return res;
+}
+I64 FpySequencer::unaryRegOp_fptoui(I64 src) {
+    // first interpret as F64
+    F64 fsrc;
+    memcpy(&fsrc, &src, sizeof(fsrc));
+    // then static cast to unsigned int
+    // then return as a signed int
+    return static_cast<I64>(static_cast<U64>(fsrc));
```
zimri-leisher, line 622: does this technically not do anything? probably yes... it's more explicit about what's actually going on here though. The float is supposed to be converted to a U64, and then it's supposed to be represented in the bits of an I64... to be honest, the CPP standard is confusing and I'm not sure how best to write this. It's all quite subtle...

Svc/FpySequencer/FpySequencerDirectives.cpp
LeStarch, line 569: I assumed this would be a biteize not, and it seems to be a logical not. This maintains the weirdness that logical "not" can be used on non-booleans using C++ behavior.  Is this intended?

LeStarch (commented):
I had one clarifications.   Otherwise looks good.  

I will note the I64 representing all registers is getting very hard to follow.  Memcpys just seem wrong.   I know why it is done, but I am starting to think there must be a better way.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: Yes--the NOT instruction would theoretically work if you passed in a non boolean. However, at the high level language we disallow this--you can only take the NOT of a boolean.

zimri-leisher:
> I had one clarifications. Otherwise looks good.
> 
> I will note the I64 representing all registers is getting very hard to follow. Memcpys just seem wrong. I know why it is done, but I am starting to think there must be a better way.

 There is a better way, and it's the next thing I'm going to work on! Switching over to a stack-based architecture will mean that there won't be any registers any more. That's coming, hopefully before the design review.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: Also consider, at the bytecode level, we don't really know what "is" or "isn't" a boolean--it's all just a series of 8 byte registers of unspecified type.

LeStarch (approved):
I buy the argument

PR #3871, opened 2025-07-10
Add lightweight SpscQueue class

celskeggs:
| | |
|:---|:---|

---
## Change Description

Contribute back an AtomicQueue class developed for a specific project using F Prime.

## Rationale

When communicating with performance-sensitive ISRs on bare metal platforms, it can be very useful to transfer small amounts of data (such as buffer pointers) between different execution contexts without needing to disable interrupts to do so.

This is a simple, safe, lightweight templated queue implementation that supports only a single reader context and a single writer context, relying on std::atomic for synchronization. It uses the technique where indices are stored modulo _twice_ the queue size in order to distinguish between queue empty and queue full conditions. (If stored modulo one times the queue size, then additional information is required to interpret `m_nextEnqueueIdx == m_nextDequeueIdx`.)

I want to get a first review of this and incorporate any requested changes before I retest it, so please don't merge until I give you the OK.

## Future Work

It will become important at some point to add unit tests for this class.


ReggieMarr:
I can see the use case however I'm not sure why you would create an additional Queue class as opposed to an "Atomic" or lightweight implementation that could just be selected as a "chosen" implementation. Is there a specific use case that makes this a better option than that ?

celskeggs:
@ReggieMarr This is a single-reader single-writer queue, but Os::Queue needs to be a multi-writer queue. I am having a hard time imagining implementing a fully lockless fixed-capacity multi-writer queue without a much larger amount of complexity. So this simple, efficient lockless implementation serves a purpose that is different from Os::Queue. It is primarily useful within a driver component, not used to communicate between components.

celskeggs:
On second thought, it looks like this would be a decent solution for a lockless fixed-capacity multi-writer queue that could be used to replace Os::Queue: https://github.com/dbittman/waitfree-mpsc-queue. (There might be some issues with priority inversion if used on a system with multithreading in addition to ISRs, though.) However, I'm not sure that invalidates the utility of this simpler single writer queue.

Utils/Types/AtomicQueue.hpp
```diff
@@ -0,0 +1,90 @@
...
+#define UTILS_TYPES_ATOMIC_QUEUE_HPP
+
+#include <atomic>
+
+namespace Types {
+
+// Note: PlatformSizeType is probably generally larger than we need,
+// but it should still be an efficient size to manipulate,
+// and it's guaranteed to be unsigned, which is crucial.
+template<class E, PlatformSizeType SIZE>
```
LeStarch: Typically I expect `FwSizeType` to be used in contexts such as this.  `PlatformSizeType` is really a default set by the platform developer that flows into the `FwSizeType` if the project chooses not to override the default.

Utils/Types/AtomicQueue.hpp
```diff
@@ -0,0 +1,90 @@
...
+class AtomicQueue {
+public:
+    static_assert(SIZE * 2 <= std::numeric_limits<PlatformSizeType>::max(),
+        "This implementation distinguishes full and empty queues by using indices modulo SIZE * 2,"
```
LeStarch: Thanks for the clean message!

Utils/Types/AtomicQueue.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+    bool isFull() {
+        return countElements(m_nextEnqueueIdx.load(), m_nextDequeueIdx.load()) == SIZE;
+    }
+
+    bool isEmpty() {
+        return countElements(m_nextEnqueueIdx.load(), m_nextDequeueIdx.load()) == 0;
+    }
+
+    // Single writer
```
LeStarch: Does this algorithm only hold with a single reader and writer?

LeStarch (commented):
A few comments.  I'd like to see @bocchino review this, as he is working on our types library.

Utils/Types/AtomicQueue.hpp
celskeggs: Yes. Although there is a variant that works with multiple writers that I mentioned above, which I did not implement.

Utils/Types/AtomicQueue.hpp
celskeggs: I'll change this. Might wait for @bocchino's review first in case there's anything else to change.

Utils/Types/AtomicQueue.hpp
```diff
@@ -0,0 +1,90 @@
...
+// but it should still be an efficient size to manipulate,
+// and it's guaranteed to be unsigned, which is crucial.
+template<class E, PlatformSizeType SIZE>
+class AtomicQueue {
+public:
+    static_assert(SIZE * 2 <= std::numeric_limits<PlatformSizeType>::max(),
+        "This implementation distinguishes full and empty queues by using indices modulo SIZE * 2,"
+        "so SIZE * 2 must fit in the index type");
+
+    bool isFull() {
```
bocchino: I think this function should be `const`.

Utils/Types/AtomicQueue.hpp
```diff
@@ -0,0 +1,90 @@
...
+        return countElements(m_nextEnqueueIdx.load(), m_nextDequeueIdx.load()) == SIZE;
+    }
+
+    bool isEmpty() {
```
bocchino: Should be `const`.

bocchino (commented):
This looks good. I have a few questions:

1. Does this use a standard concurrent algorithm? Is there a reference for it?
2. What is the intended use? I'm guessing that all the functions except `peek` would run on a single thread, and `peek` would run on another concurrent thread?
3. What is the atomicity or concurrency guarantee that's provided? For example, is this allowed?
    ```
    Queue is empty
    Thread 1 enqueues an item. Queue size is 1.
    Thread 2 calls peek and reads the indices.
    Thread 1 dequeues an item. Queue size is 0.
    Thread 2 completes the peek call and reads the value at index 0.
    ```
    This seems like a use-after-free error. But maybe it's allowed?

Utils/Types/AtomicQueue.hpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    bool dequeue() {
+        E ignored;
+        return dequeue(ignored);
+    }
+
+private:
+    E m_elements[SIZE];
+    std::atomic<PlatformSizeType> m_nextEnqueueIdx;
```
bocchino: These variables should be initialized to zero, either here or in the constructor initializers.

celskeggs:
@bocchino This is a fairly mundane, run-of-the-mill concurrent algorithm. I don't have a specific reference for you because it's supposed to be simple enough to not need a lot of analysis.

This is a single-reader single-writer queue. The writer thread (or ISR) can call `enqueue`. The reader thread (or ISR) can call `dequeue` or `peek`. Either thread can call `isFull` or `isEmpty`. I thought this was obvious, but clearly I need to document it better.

The concurrency guarantees when used with a single reader and a single writer should be very robust. The operations in this case are lock-free, wait-free, and thread-safe. The example you give is not permissible because you're having both `thread 1` and `thread 2` act as a reader.

bocchino:
I did some searching on the Web. I found out that this is a standard concurrent data structure implementation, and it's often called a single-producer single-consumer (SPSC) queue. For example:

https://www.codeproject.com/Articles/43510/Lock-Free-Single-Producer-Single-Consumer-Circular
https://docs.zephyrproject.org/latest/kernel/data_structures/spsc_lockfree.html

Can we do the following:
1. Rename the class from AtomicQueue to SPSCQueue. AtomicQueue may tempt users who don't read the comments carefully into thinking that this is a general concurrent queue.
2. In the comments, refer to "producer" and "consumer" instead of "writer" and "reader." A natural interpretation of "writer" and "reader" for a shared data structure is that writers call non-const functions and readers call const functions. This is not the intended interpretation. The terms "producer" and "consumer" are clearer because they don't have this ambiguity.

If these changes seem OK, then it looks good to me!

celskeggs:
@bocchino I like this recommendation! I will call it "SpscQueue" and say "producer" and "consumer" instead of "writer" and "reader."

celskeggs:
@bocchino Would you mind approving so that the CI workflows can get kicked off?

bocchino (approved):
Looks good!

celskeggs:
@bocchino The tests broke because I was behind, and the examples repository didn't match. Can you re-approve so that we can get these retested?

Issue #3451, opened 2025-04-03
Support arguments in C asserts


pcrosemurgy:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description & Rationale

There's currently no equivalent for the "heritage" (C) `FSW_ASSERT_1()` which will report an argument.

https://github.com/nasa/fprime/blob/devel/Fw/Types/CAssert.h should have a `FW_CASSERT_1()` perhaps?

thomas-bc:
Is this something that will always fail if reached, and the purpose is to report a single argument value?

Would `FW_ASSERT(false, <arg>)` be equivalent to this ?

PR #3896, opened 2025-07-17
Adding FW_CASSERT_1 macro for C

vincewoo:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3451  |

---
## Change Description

Adding FW_CASSERT_1 macro to support a 1-arg assert in C

## Rationale

There was no equivalent for the "heritage" (C) FSW_ASSERT_1() which will report an argument

Added UT for testing the assert

## Future Work

None

### DIFF
--- a/Fw/Types/test/ut/CAssertTest.cpp
+++ b/Fw/Types/test/ut/CAssertTest.cpp
@@ -12,41 +12,57 @@
  * in embedded systems.
  */
 
-// Define a simple test assert hook
+// Define an Assert handler
 class TestAssertHook : public Fw::AssertHook {
     public:
-        TestAssertHook() : m_asserted(false), m_numArgs(0), m_arg1(0) {}
+        TestAssertHook() {}
         virtual ~TestAssertHook() {}
-        
         void reportAssert(FILE_NAME_ARG file,
-                         FwSizeType lineNo,
-                         FwSizeType numArgs,
-                         FwAssertArgType arg1,
-                         FwAssertArgType arg2,
-                         FwAssertArgType arg3,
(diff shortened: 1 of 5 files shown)


Fw/Types/test/ut/CAssertTest.cpp
```diff
@@ -0,0 +1,39 @@
...
+    
+    // Test that the macro compiles without errors
+    int testValue = 42;
+    
+    // This should not trigger an assertion since condition is true
+    FW_CASSERT_1(testValue == 42, testValue);
+    
+    // This should compile but not execute assertion in normal builds
+    // In debug builds with assertions enabled, this would trigger
+    FW_CASSERT_1(testValue != 0, testValue);
```
LeStarch: How does this not kill the UT?

LeStarch (commented):
One question on the test.

LeStarch:
You may wish to break those tests into separate test cases.  And you could put in an ASSERT_DEATH to prove that the assert will trip and kill the program.

Fw/Types/test/ut/CAssertTest.cpp
vincewoo: Ah yes, good catch! Removing this test and splitting it off into a separate, correct, test.

vincewoo:
> You may wish to break those tests into separate test cases. And you could put in an ASSERT_DEATH to prove that the assert will trip and kill the program.

I've split up the AssertFailure case into another test. The proper way to do this is actually using a TestAssertHook.

LeStarch (approved):

PR #3898, opened 2025-07-18
Add CMake package file for F Prime

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

This adds `FPrimeConfig.cmake` to allow F Prime to be imported as a CMake package.  This simplifies the import structure of F Prime and brings it inline with other CMake projects like Zephyr

It is the first step to making the F Prime codebase importable via `add_subdirectory` as we still need to include the build system setup somehow.


kevin-f-ortega (approved):
LGTM. Just pending successful CI

PR #3899, opened 2025-07-18
Add a repository dirtiness indicator to the auto-generated version information

ianbrault:
| | |
|:---|:---|

---
## Change Description

This change makes a small modification to the version auto-coder script `generate_version_info.py` to include the `--dirty` flag when calling `git describe`. This will show whether a build has any un-committed changes in its repository or any of its dependencies.

for example:
```bash
# before:
$ git describe --tags --always
v4.0.0a1-104-gb217fc79a
# after:
$ git describe --tags --always --dirty
v4.0.0a1-104-gb217fc79a-dirty
```

## Rationale

This change is useful for an operator to see whether or not the binary that is running was built from a dirty repository or with dirty dependencies, and creates useful breadcrumbs in the version telemetry.

Should hopefully be straightforward to test, create a build from a clean repository and compare to a build created from a dirty repository.

## Future Work

One thing to consider might be this note from the `git describe` manual:

> If a repository is corrupt and Git cannot determine if there is local modification, Git will error out, unless ‘--broken’ is given, which appends the suffix "-broken" instead.

Is it worth supporting this functionality for a corrupt repository? If so, it might be useful to include the `--broken` flag in this command as well.

### DIFF
--- a/cmake/target/version/generate_version_info.py
+++ b/cmake/target/version/generate_version_info.py
@@ -28,7 +28,7 @@ FALLBACK_VERSION = "v3.5.0"  # Keep up-to-date on release tag
 
 def get_version_str(working_dir, fallback=FALLBACK_VERSION):
     """
-    System call to get the version. It uses `git describe --tags --always --dirty` to get a standard version string as used
+    System call to get the version. It uses `git describe --tags --always --dirty --broken` to get a standard version string as used
     across fprime. The working directory should be set to a (project, library, fprime) from which to obtain the version.
 
     Args:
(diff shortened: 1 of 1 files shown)


LeStarch (dismissed):

LeStarch:
I would happily include a `--broken` flag, but that could also be a separate PR.

ianbrault:
@LeStarch I can push a commit to add the `--broken` flag under this, not a problem

LeStarch (dismissed):

LeStarch (approved):

thomas-bc (approved):

PR #3900, opened 2025-07-19
Refactor CircularBuffer to extend Fw::SerializableBufferBase

### LINKED ISSUE #3429: CircularBuffer should implement the SerializeBufferBase interface
| | |
|:---|:---|
|**_F´ Version_**| v3.6.2 |
|**_Affected Component_**| CircularBuffer / SerializeBufferBase |
---
## Feature Description

A CircularBuffer can and likely should implement the same interface as SerializeBufferBase. Currently SerializeBufferBase isn't strictly an interface; but we likely can abstract the interface into a SerializeBufferInterface, and have both SerializeBufferBase and CircularBuffer implement it.

## Rationale

This would allow serializing/deserializing directly from/into a CircularBuffer. This is currently a known limitation/inefficiency in the [FprimeFrameDetector](https://github.com/nasa/fprime/blob/f912e1bd754d946ccec227663f26bc0a56da5834/Svc/FrameAccumulator/FrameDetector/FprimeFrameDetector.cpp#L20-L27)

vincewoo:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3429 |

---
## Change Description

Refactoring `CircularBuffer` to extend from `Fw::SerializableBufferBase` so that it can be used in standard interfaces that expect a `Fw::SerializableBufferBase`.

## Rationale

This change makes `CircularBuffer` aligned with the standard allowing it to be used directly without having to be copied. In particular, it allows us to make [FprimeFrameDetector](https://github.com/nasa/fprime/blob/f912e1bd754d946ccec227663f26bc0a56da5834/Svc/FrameAccumulator/FrameDetector/FprimeFrameDetector.cpp#L20-L27) more efficient by being able to use CircularBuffer directly instead of copying.

Added test cases for all the new methods that have been extended from the base class. Added new tests to exercise special cases for wrap-around of the circular buffer.

## Future Work

Improve `FprimeFrameDetector` to use the new `CircularBuffer`


Utils/Types/CircularBuffer.cpp
```diff
@@ -162,4 +187,585 @@ void CircularBuffer ::clear_high_water_mark() {
...
+        return Fw::FW_SERIALIZE_NO_ROOM_LEFT;
+    }
+    
+    // Extract data from circular buffer into a linear temporary buffer
+    U8 temp_buffer[FW_COM_BUFFER_MAX_SIZE]; // Maximum F' communication buffer size
+    FW_ASSERT(size <= sizeof(temp_buffer)); // Size limitation
+    
+    // Copy data from current deserialization position
+    for (FwSizeType i = 0; i < size; i++) {
+        FwSizeType idx = advance_idx(m_head_idx, m_deser_offset + i);
```
LeStarch, line 712: Why not call serialize byte by byte on dest.  This would avoid a redundant copy, would it not?

Utils/Types/CircularBuffer.cpp
```diff
@@ -42,13 +47,16 @@ void CircularBuffer :: setup(U8* const buffer, const FwSizeType size) {
     FW_ASSERT(size > 0);
     FW_ASSERT(buffer != nullptr);
     FW_ASSERT(m_store == nullptr && m_store_size == 0); // Not already setup
+    FW_ASSERT(size < static_cast<FwSizeType>(static_cast<FwSizeType>(-1) / 2)); // Reasonable size limit
```
LeStarch, line 50: Why?  This seems odd.  Why wouldn't the max size just be `FwSizeType`' max value?

LeStarch: Also, probably shouldn't cast -1, just use numeric limits to get the max FwSizeType

Utils/Types/CircularBuffer.cpp
```diff
@@ -162,4 +187,585 @@ void CircularBuffer ::clear_high_water_mark() {
...
+    return get_allocated_size();
+}
+
+FwSizeType CircularBuffer::getBuffLeft() const {
+    return get_allocated_size() - m_deser_offset;
+}
+
+U8* CircularBuffer::getBuffAddr() {
+    // Return pointer to the underlying storage buffer
+    FW_ASSERT(m_store != nullptr && m_store_size != 0); // setup method was called
```
LeStarch, line 208: Should we delete these methods or make them private or something?  It is erroneous to use the addr of the buffer as-if it is linear.

Utils/Types/CircularBuffer.cpp
```diff
@@ -162,4 +187,585 @@ void CircularBuffer ::clear_high_water_mark() {
...
+            return status;
+        }
+    }
+    
+    // Use the existing serialize method for the buffer
+    return serialize(buff, length);
+}
+
+Fw::SerializeStatus CircularBuffer::serialize(const Fw::Serializable& val) {
+    return val.serialize(*this);
```
LeStarch, line 385: I don't think this will work either, at least as long as the get addr functions exist.

Utils/Types/CircularBuffer.cpp
```diff
@@ -162,4 +187,585 @@ void CircularBuffer ::clear_high_water_mark() {
...
+    return Fw::FW_SERIALIZE_OK;
+}
+
+Fw::SerializeStatus CircularBuffer::serializeSkip(FwSizeType numBytesToSkip) {
+    FW_ASSERT(m_store != nullptr && m_store_size != 0); // setup method was called
+    FW_ASSERT(m_ser_offset <= get_allocated_size()); // current offset should be valid
+    FW_ASSERT(numBytesToSkip <= m_store_size); // skip amount cannot exceed total buffer capacity
+    
+    // Check for potential overflow in addition
+    if ((numBytesToSkip > 0) && (m_ser_offset > (static_cast<FwSizeType>(-1) - numBytesToSkip))) {
```
LeStarch, line 635: Probably don't want a cast of `-1`

LeStarch (changes requested):

LeStarch:
@bocchino would would review this one?

Utils/Types/CircularBuffer.hpp
```diff
@@ -135,6 +135,94 @@ class CircularBuffer {
          */
         void clear_high_water_mark();
 
+        // ----------------------------------------------------------------------
+        // SerializeBufferBase implementation
```
thomas-bc, line 139: Should SerializeBufferBase and CircularBuffer share a parent (virtual) class, rather than CircularBuffer inheriting directly from SerializeBufferBase? Or is there a reason it was done this way (less disruptive of a change?) ?

thomas-bc: If kept this way all the `serialize()` methods here should probably be marked `override`

Utils/Types/CircularBuffer.hpp
```diff
@@ -135,6 +135,94 @@ class CircularBuffer {
...
+        Fw::SerializeStatus deserialize(bool& val);
+        Fw::SerializeStatus deserialize(void*& val);
+        Fw::SerializeStatus deserialize(U8* buff, FwSizeType& length);
+        Fw::SerializeStatus deserialize(U8* buff, FwSizeType& length, Fw::Serialization::t mode);
+        Fw::SerializeStatus deserialize(Fw::Serializable& val);
+
+        Fw::SerializeStatus deserializeSize(FwSizeType& size);
+
+        // Buffer management methods
+        void resetSer();
```
thomas-bc, line 210: Agreed with Michael's comment, some of these Buffer management methods seem questionable to use on a CircularBuffer. These seem like they belong to the "linear" types of buffers. This speaks to my other comment about a parent `SerializeBufferInterface` that would not include them.

Utils/Types/CircularBuffer.cpp
```diff
@@ -42,13 +47,16 @@ void CircularBuffer :: setup(U8* const buffer, const FwSizeType size) {
...
+    FW_ASSERT(size < static_cast<FwSizeType>(static_cast<FwSizeType>(-1) / 2)); // Reasonable size limit
 
     // Initialize buffer data
     m_store = buffer;
     m_store_size = size;
     m_head_idx = 0;
     m_allocated_size = 0;
     m_high_water_mark = 0;
+    m_ser_offset = 0;
+    m_deser_offset = 0;
```
thomas-bc, line 59: Should `m_ser_offset` and `m_deser_offset` not be computed from `m_head_idx` and `m_allocated_size` ? Or is there a use case where they would differ?

If I do
```
circ.rotate(5);
circ.deserializeSkip(3);
```
will the `deserialize()` vs. `peek()` offsets not match?

thomas-bc (commented):
A couple of design-y questions since I was out for this

Utils/Types/CircularBuffer.hpp
vincewoo: You're correct, this approach was chosen because it is less disruptive of a change. If we choose to create the shared parent we would then have to update many of the references from SerializeBufferBase to the new virtual class in order to leverage the new flexibility we are enabling to CircularBuffer.

PR #3904, opened 2025-07-22
Restructure core subtopologies and add Svc_Subtopologies target

### LINKED ISSUE #3905: Ccsds and Com Topologies are Dependent on Tcp
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

We currently `restrict_platforms` for Com and CCSDS because the default configuration of those subtopologies use Tcp, which is invalid.

However, since this is configuration and may be changed we cannot just simply exclude them due to default configuration.

moisesmata:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description
fixes nasa/fprime#3905
- Added `EXCLUDE_FROM_ALL` flag to each subtopology's cmake, to prevent subtopologies being automatically built.  
- Added a `Svc_Subtopologies` target which builds just the subtopologies
- Removed two components from the Coms Subtopologies, `comDriver`, and `cmdSeq`. 

## Rationale

- Not all projects want to build everything in subtopologies, which may be platform specific. 
- `Svc_Subtopologies` is useful when we need to build everything for the `cppcheck` CI test. 
- it made more sense to have `comDriver` defined in the top level topology, gives user more freedom
- `cmdSeq` in top level topology for the same reason, gives user more freedom


Svc/Subtopologies/ComCcsds/ComCcsdsConfig/ComCcsdsComDriverConfig.fpp
```diff
@@ -1,7 +1,25 @@
 module ComCcsds {
```
LeStarch: Why is this file still here?

Svc/Subtopologies/ComFprime/ComFprimeConfig/CMakeLists.txt
```diff
@@ -1,9 +1,22 @@
+# Conditionally select driver config based on socket availability
+if(FPRIME_HAS_SOCKETS)
+    # Use TCP-based driver when sockets are available
+    set(DRIVER_CONFIG_FILE "${CMAKE_CURRENT_LIST_DIR}/ComFprimeTcpDriverConfig.fpp")
```
LeStarch: This isn't going to work right, as the file you need to override will change based on what is selected.

Ref/Top/instances.fpp
```diff
@@ -66,14 +66,39 @@ module Ref {
...
 
-  instance systemResources: Svc.SystemResources base id 0x010F0000
+  instance systemResources: Svc.SystemResources base id 0x01140000
 
-  instance linuxTimer: Svc.LinuxTimer base id 0x01100000
+  instance linuxTimer: Svc.LinuxTimer base id 0x01150000
+
+  instance comDriver: Drv.TcpClient base id 0x01200000 \
+  {
+      phase Fpp.ToCpp.Phases.configComponents """
```
thomas-bc: should be RefTopology.cpp instead of in phases

Ref/Top/instances.fpp
```diff
@@ -66,14 +66,39 @@ module Ref {
   # Passive component instances
   # ----------------------------------------------------------------------
 
-  instance posixTime: Svc.PosixTime base id 0x010C0000
+  instance posixTime: Svc.PosixTime base id 0x01100000
```
thomas-bc: fix IDs

Svc/Subtopologies/FileHandling/FileHandling.fpp
```diff
@@ -45,12 +45,27 @@ module FileHandling {
         """
     }
 
+    instance cmdSeq: Svc.CmdSequencer base id FileHandlingConfig.BASE_ID + 0x0500 \
```
thomas-bc: move to root of Ref

moisesmata:
Notes for me:
Remove Os/Console.hpp and Os::Logger

Ref/Top/instances.fpp
```diff
@@ -13,67 +13,74 @@ module Ref {
   # Active component instances
   # ----------------------------------------------------------------------
 
-  instance blockDrv: Ref.BlockDriver base id 0x01000000 \
+  instance blockDrv: Ref.BlockDriver base id 0x10000000 \
```
thomas-bc, line 29: Why do all these IDs start with `0x100....` and not from the beginning `0x00001000` ?

thomas-bc: If there's a convention we want users to adopt, we should document it. A comment in `instances.fpp` would likely be enough.

Ref/Top/RefTopology.cpp
```diff
@@ -69,6 +69,17 @@ void setupTopology(const TopologyState& state) {
...
     startTasks(state);
+
+    //Configuration of the comDriver component
+    if (state.hostname != nullptr && state.port != 0) {
+        comDriver.configure(state.hostname, state.port);
+    }
+
+    //Initialize socket client communication if and only if there is a valid specification
+    if (state.hostname != nullptr && state.port != 0) {
+        Os::TaskString name("ReceiveTask");
```
thomas-bc, line 83: There's 2 if statements with the same condition and related to the same component. Should they be merged? Looking at the old code https://github.com/nasa/fprime/blob/728a022e9c9a3b9a943609361a9f1dad89d67a7d/Ref/Top/RefTopology.cpp#L159
`comDriver.configure()` used to be done right after `configComponents()`. Could also re-use the same structure. Or maybe just put `.configure()` in configComponents? Up to you

Svc/Subtopologies/ComFprime/ComFprime.fpp
```diff
@@ -9,7 +9,7 @@ module ComFprime {
     # ----------------------------------------------------------------------
     # Active Components
     # ----------------------------------------------------------------------
-    instance comQueue: Svc.ComQueue base id ComFprimeConfig.BASE_ID + 0x0100 \
+    instance comQueue: Svc.ComQueue base id ComFprimeConfig.BASE_ID + 0x000000 \
```
thomas-bc: (This also applies to ComCcsds.fpp)

The phase code below duplicates the `enum{EVENTS, TELEMETRY, FILE_QUEUE}` which is also defined in the `Ports_ComPacketQueue` enum above. There's also a few things that are worth pointing out:
- the order in which the entries go in the `configurationTable` depends on the order in which they are connected in the topology.fpp. This is why we define `Ports_ComPacketQueue`: so that the topology.fpp can use the enum as port indexes. Can we revert to using `Ports_ComPacketQueue` rather than the on-the-fly phases definition ?
- Adding `FILE_QUEUE` kinda works right now but can lead to confusion because the configuration of the ComQueue is a little tricky. If a user adds a third ComPacket port, they might add it as a 4th item under `FILE_QUEUE`, and that won't work (because the file queue is a buffer queue, and indexes are shared with the ComPacket queues). That's why I originially was using the `NUM_CONSTANTS` (see [here](https://github.com/nasa/fprime/blob/728a022e9c9a3b9a943609361a9f1dad89d67a7d/Ref/Top/RefTopology.cpp#L136-L137)) value of the Ports_ComPacketQueue enum. It's not a great solution and I would call it technical debt on the ComQueue side, but I would argue reverting back to `NUM_CONSTANTS` would at least trigger users to wonder "why is this different?" and maybe dig into the code to figure it out, rather than leaving the door open for mistakes.

thomas-bc (commented):
A few questions and nits

Svc/Subtopologies/ComFprime/ComFprime.fpp
thomas-bc: Actually `Ports_ComPacketQueue` as well as the configuration table entries likely need to be configurable... maybe we spin this off to a follow up PR

thomas-bc: Or maybe we just say if you need more than the stock config, you shouldn't use the subtopology

Ref/Top/instances.fpp
moisesmata: Yes there is a convention that Michael sketched out, I'll add a comment.

moisesmata: But the reason we start with a 1, is because the first digit refers to the deployment in this convention. Michael mentioned that `0x00000000` is not a valid base ID, so only for deployment, our first digit must start with 0.

Ref/Top/RefTopology.cpp
moisesmata: Ah yes, it does look a bit weird having the two ifs next to each other with the same condition. I think I'll reuse the same structure to have it how it was before we had phased code for `comDriver`

Ref/Top/instances.fpp
```diff
@@ -1,5 +1,18 @@
...
+  # 
+  # All Base IDs follow the 8-digit hex format: 0xDSSCCxxx
+  #
+  # Where:
+  #   D   = Deployment digit (1 for Ref deployment)
+  #   SS  = Subtopology digits (00 for main topology, 01-05 for subtopologies)
+  #   CC  = Component digits (00, 01, 02, etc.)
+  #   xxx = Reserved for internal component items (events, commands, telemetry)
+  #
+
```
moisesmata, line 15: Think about BaseIds

moisesmata: For Subtopolgies: `01-FF`

Ref/Top/instances.fpp
```diff
@@ -13,67 +26,74 @@ module Ref {
...
+    queue size Default.QUEUE_SIZE \
+    stack size Default.STACK_SIZE \
+    priority 97
 
   # ----------------------------------------------------------------------
   # Queued component instances
   # ----------------------------------------------------------------------
 
-  instance sendBuffComp: Ref.SendBuff base id 0x01060000 \
+  instance sendBuffComp: Ref.SendBuff base id 0x10007000 \
```
moisesmata: Give spack between BaseIds between component types

Svc/Subtopologies/CdhCore/CdhCoreConfig/CdhCoreConfig.fpp
```diff
@@ -1,6 +1,6 @@
 module CdhCoreConfig {
     #Base ID for the CdhCore Subtopology, all components are offsets from this base ID
-    constant BASE_ID = 0x10100000
+    constant BASE_ID = 0x01000000
```
moisesmata, line 3: Fix increments in subtopologies `10` -> `01`....

Svc/Subtopologies/CdhCore/CdhCoreConfig/CdhCoreFatalHandlerConfig.fpp
```diff
@@ -1,6 +1,6 @@
 module CdhCore {
 
     # Update this as a custom fatal handler if needed
-    instance fatalHandler: Svc.FatalHandler base id CdhCoreConfig.BASE_ID + 0x0800
+    instance fatalHandler: Svc.FatalHandler base id CdhCoreConfig.BASE_ID + 0x007000
```
moisesmata: Make it only 5 digits, we don't want a notion of the deployment nor subtopology

thomas-bc (approved):

LeStarch (approved):

Issue #3833, opened 2025-07-01
Add `std::atomic` to the OSAL layer


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Currently, `std::atomic` is used as a way to implement atomic variables in F´.  However, `std::atomic` is not supported by all platforms.

Specifically, Atmel ATMega and ARM Cortex 6M (PICO 1) boards do not support `std::atomic`. 

Adding Atomic to the OSAL layer allows these platforms to provide alternate implementations rather than fail to compile F´.

Steps:
1. Enumerate functions of atomic variables used in F Prime (set, add, swap_if, etc).
2. Design OSAL interface
3. Implement OSAL interface via `std::atomic`

Required functions (and source):

1.  Autocoded Event Throttling: `fetch_add` and `=`

The though here is we'd provide a basic implementation of `std:atomic` that is enough to compile F Prime.  It would have limitations (e.g. it uses `Os::Mutex`) but is sufficient to compile with these limitations.

Note: https://en.cppreference.com/w/cpp/atomic/atomic.html

celskeggs:
My understanding was that `std::atomic` is mandatory for all C++ implementations, even freestanding implementations. See https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/n4950.pdf, Table 27. Am I misunderstanding? How are the Atmel ATMega and ARM Cortex 6M platforms standards-compliant if they don't support `std::atomic`?

LeStarch:
@celskeggs test compilations with the Atmel ATMega don't work with `std::atomic`.  Perhaps these platforms are non-compliant.

bocchino:
An alternative approach is to provide our own implementation of `std::atomic`, or a relevant subset: https://en.cppreference.com/w/cpp/atomic/atomic.html. It's not clear to me why the non-compliant platforms aren't already doing this.

celskeggs:
Yeah, I'd rather fix a non-compliant platform by using a better libc++ that provides `std::atomic` in a standards-compliant way, rather than adding our own additional abstraction layer.

Atomic operations can be incredibly performance sensitive and semantic sensitive, and I'm not excited about the idea of having a wrapper class.

m-aleem:
The following Pull Requests are associated with this issue:

[...]

PR #3909, opened 2025-07-22
Stub std::atomic implementation

m-aleem:
| | |
|:---|:---|
|**_Related Issue(s)_**| [#3833](https://github.com/nasa/fprime/issues/3833) |
|**_Documentation Included (y/n)_**| N (In-line explanations, only) |

---
## Change Description

Added a Stub `std::atomic` implementation that is only single-thread safe because it does *not* actually use any locking mechanism, but will allow compilation/build on systems that do not support std::atomic 

## Rationale

This is in support of [#3833](https://github.com/nasa/fprime/issues/3833). 

UTs were added, note that an extra UT header file was needed so that test could be wrapped into another namespace (above `std` ) so that the tests can compile and run on platforms that *do* have std::atomic already


LeStarch (changes requested):
I don't see a register fprime implementation call.  You should probably add one:

```
    register_fprime_implementation(
        StubStandardAtomic
      IMPLEMENTS
        StandardAtomic
      INTERFACE
   )
   target_include_directories(StubStandardAtomic BEFORE SYSTEM ....path to directory)
```
   
You could also ensure that the UT "CHOSESES_IMPLEMENTATION StubStandardAtomic)" to ensure the directory stuff works.

LeStarch:
Also, perhaps we put the `<atomic>` header in a separate directory so we don't get bleed-over from other implementations.  Such as `Stub/atomic`

thomas-bc (approved):

LeStarch (approved):

Issue #3906, opened 2025-07-22
Regression: Dictionaries no longer generated for flat deployments


celskeggs:
| | |
|:---|:---|
|**_F´ Version_**| latest |
|**_Affected Component_**| build system |
---

## Problem Description

My project uses "flat" deployments, where I haven't created a separate subfolder to store the topology. The topology FPP file is specified in the AUTOCODER_INPUTS section of my register_fprime_deployment call.

The changes to the dictionary autocoder under https://github.com/nasa/fprime/pull/3735 result in dictionaries no longer getting generated for my deployments. This implementation relies on the topology being defined in a module, not a deployment.

## Context / Environment

```
[root@3e0e247fc469 scythe]# fprime-util version-check
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-4.18.0-553.58.1.el8_10.x86_64-x86_64-with-glibc2.28
Python version: 3.9.20
CMake version: 3.26.0
Pip version: 25.1.1
Pip packages:
    fprime-tools==4.0.0a6
    fprime-gds==4.0.0a7
    fprime-fpp-*==3.0.0a14
Project submodules:
    https://github.com/nasa/fprime.git @ v4.0.0a1-118-g3d7eac1c2
    https://github.com/fprime-community/fprime-baremetal @ 5e66e17
```

## How to Reproduce

1. Create a flat deployment
2. Try to generate a dictionary
3. No dictionary

## Expected Behavior

We should support flat deployments, and they should be able to generate dictionaries successfully.

thomas-bc:
See 

https://github.com/nasa/fprime/blob/bbcfb15a3fdfcc0b498649affcd33aae1e3aac2b/cmake/target/dictionary.cmake#L36-L49

which expects a **dependency** of a deployment to be generating a dictionary (for example `Ref_Top` for the `Ref` deployment)

LeStarch:
CCB Note: While not typically a supported user story, since a PR was made to correct the problem we can accept it. However, I advise you prepare a minimalist deployment with the same "flat" structure (we can give you a repo in fprime-community) and let us know so we can add it to CI.

This will formalize the support and prevent future regressions!

PR #3735, opened 2025-06-13
CMake Refactor: Explicity Setting Autocoder Outputs

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|  |
|**_Documentation Included (y/n)_**|  |

---

Autocoder shall now set these variables and only these variables.   Fixes: #3485, #3495.

## Required Variables (At Least One Must Be Set)

| Variable Name | Use |  Note
|---------------|-----|---|
| `AUTOCODER_GENERATED_AUTOCODER_INPUTS` | Files that become inputs to subsequent autocoders in the chain | Passed to all future autocoders |
| `AUTOCODER_GENERATED_BUILD_SOURCES` | Source files (.cpp, .c) that get compiled into the target | Passed to the build step |
| `AUTOCODER_GENERATED_OTHER` | Other generated files (data files, etc.) that don't get compiled. | Must be handled by some `add_custom_target` to ensure generation |

## Optional Variables

| Variable Name | Use | Example |
|---------------|-----|---------|
| `AUTOCODER_DEPENDENCIES` | Module dependencies that the autocoder deduces `Fw_Types`, `Svc_CmdDispatcher` |

## NOTE: THIS CODE WAS DEVELOPED WITH THE HELP OF GENERATIVE AI

PR #3910, opened 2025-07-22
Generate dictionaries in deployment targets

celskeggs:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3906 |

---
## Change Description

This is a potential fix for #3906.

## Rationale

The changes to the dictionary autocoder under https://github.com/nasa/fprime/pull/3735 result in dictionaries no longer getting generated for my deployments. This PR attempts to fix that regression.

I'm not sure I've made the right set of changes to autocoder.cmake, but note that these are required, because otherwise the `if (NOT HASH_SET)` optimization in `run_ac` will sometimes prevent `${AC_SET_HASH}_AUTOCODER_GENERATED_OTHER` from being populated in `run_ac_set`.

## Future Work

N/A


celskeggs:
The main problem with this solution is that the dictionary install step runs on every build and prints this text, even if nothing has changed:

```
[build] [21/21] Running utility command for Deployments_DeploymentName_dictionary
[build] -- Install configuration: "Release"
[build] -- Up-to-date: /workspaces/ProjectName/build-artifacts/PlatformName/Deployments_DeploymentName/dict/DeploymentNameTopologyDictionary.json
```

Normally a build that does nothing should be silent. But I'm not sure what I got wrong to introduce this issue.

celskeggs:
Another issue: This seems to break rebuilding of the dictionaries. A purge rebuild is needed for channels to be added or removed.

cmake/autocoder/autocoder.cmake
```diff
@@ -134,22 +134,24 @@ function(run_ac BUILD_TARGET_NAME AUTOCODER_CMAKE SOURCES GENERATED_FILE_LIST HA
...
-                append_list_property("${VARIABLE_VALUES}" TARGET "${BUILD_TARGET_NAME}" PROPERTY "${HASH}_${VARIABLE_NAME}")
-            endif()
-        endforeach()
-        _describe_autocoder_run("${AUTOCODER_NAME}")
     else()
         # Assert runs are identical for the same autocoder
         fprime_cmake_ASSERT("Hash mismatch for autocoder ${AUTOCODER_NAME}: stored '${STORED_HASH}' vs calculated '${SRCS_HASH}'" 
                             "${STORED_HASH}" STREQUAL "${SRCS_HASH}")
     endif()
+
```
LeStarch, line 142: This fix is correct.  Pulling the "read" of the properties out of the `if (not already run)` block ensures they are read in both cases.

The is a minor change in behavior in which the description of the run is also emitted even when run before, however; this is also likely desirable.

cmake/target/dictionary.cmake
```diff
@@ -30,9 +30,20 @@ endfunction(dictionary_add_global_target)
 # - **FULL_DEPENDENCIES:** MOD_DEPS input from CMakeLists.txt
 ####
 function(dictionary_add_deployment_target MODULE TARGET SOURCES DEPENDENCIES FULL_DEPENDENCIES)
+    run_ac_set("${MODULE}" "autocoder/fpp")
```
LeStarch, line 33: This is necessary for dictionary generation if the dictionary is an output of the deployment module.

cmake/target/dictionary.cmake
```diff
@@ -30,9 +30,20 @@ endfunction(dictionary_add_global_target)
...
+
     # Create deployment level target and remove the module from the list of dependencies
-    add_custom_target("${MODULE}_${TARGET}")
+    add_custom_target("${MODULE}_${TARGET}" DEPENDS ${AUTOCODER_GENERATED_OTHER})
     list(REMOVE_ITEM DEPENDENCIES "${MODULE}")
+
+    # Create a custom target with _dictionary suffix that depends on the generated files
+    if(AUTOCODER_GENERATED_OTHER)
+        # Install the files as a component. This is done here so it is output to the deployment directory
```
LeStarch, line 41: This code is similar, but not an exact duplicate of the code below.

LeStarch (approved):
Here are several concerns I hypothesized with this PR:
1. It looked like identical code.  False: the code is different enough that we need not force it to be unified.
2. Improper (missing) triggering of the autocoder on the deployment module.  False: this was done correctly.

There is one outstanding minor issue:  having a deployment that is also a topology (that is, blending deployment and Top modules together) isn't necessarily a "supported" feature.  While I am happy to merge this PR, at some point we will need a reference for this structure of project that we can build in CI to ensure we do not have future regressions.

Thanks for the AC and dictionary fix!

LeStarch:
I understood both issues (the AC and the dictionary issue) and it seems right.  Once it passes CI we can merge.

LeStarch:
CI is failing because the base branch is out-of-date from the external repos branch.

celskeggs:
@LeStarch Can you take a look at the two concerns I listed above? Do you think those were introduced by this change or also a problem in the ordinary case?

LeStarch:
@celskeggs The first issue I am less concerned with.  It is how `install` works in CMake.  It will print text even when up-to-date (no-op).

The other is more concerning.  However, I don't seem to be able to duplicate it.  There is an unrelated issue where include files (.fppi) are not re-triggering generation.  However, I still see dictionaries updated regardless.

The issue applies to both this branch and Ref.  

I would argue that we spin out those two issues into individual reports:

1. Why is generate not rerun.
2. Why are your dictionaries not remade without a purge.  If you can come up with a minimal test case, I would happily help you debug.

However, this PR is ready to go in my opinion.

PR #3937, opened 2025-07-25
Compile-time asserts for string-vs.-buffer max sizes do not account for serialized string lengths

ianbrault:
| | |
|:---|:---|

---
## Change Description

The compile-time assertions which relate the various string maximum sizes (commands, logs, telemetry, parameters) to the corresponding buffer maximum sizes do not account for the serialization of the string length. These assertions only ensure that the strings themselves will fit in the buffers but the string lengths are also serialized into the buffers as `FwSizeStoreType`s. This change modifies the assertions to ensure that `sizeof(FwSizeStoreType)` can also fit into the buffer along with the string contents.

## Rationale

As an end user tweaking the string max sizes in our project configuration, I adjusted `FW_COM_BUFFER_MAX_SIZE` such that the string buffer sizes would be derived appropriately but running the software and downlinking a maximum-sized string triggered an assertion which said that there was no space left for serialization, since an additional 2 bytes were serialized into the buffer for the string length. Modifying the compile-time assertions would have avoided this scenario.

## Future Work

N/A


Fw/FPrimeBasicTypes.hpp
```diff
@@ -46,13 +46,13 @@ static_assert(FW_FIXED_LENGTH_STRING_SIZE >= FW_PARAM_STRING_MAX_SIZE,
 
 // Check that command/telemetry strings are not larger than an argument buffer
 
-static_assert(FW_CMD_STRING_MAX_SIZE <= FW_CMD_ARG_BUFFER_MAX_SIZE,
+static_assert(FW_CMD_STRING_MAX_SIZE + sizeof(FwSizeStoreType) <= FW_CMD_ARG_BUFFER_MAX_SIZE,
```
bocchino, line 49:
```suggestion
static_assert(Fw::CmdStringArg::SERIALIZED_SIZE <= FW_CMD_ARG_BUFFER_MAX_SIZE,
```

bocchino:
Good catch! I made a suggestion that may clean up the code a bit.

ianbrault:
Thanks @bocchino I will incorporate that suggestion

ianbrault:
@bocchino this seems to be getting me into some `include` trouble. I've included `Fw/Cmd/CmdString.hpp` for `Fw::CmdStringArg` but this causes an error while compiling `Fw/Types/snprintf_format.cpp`:

```
[41/633] Building CXX object F-Prime/Fw/Types/CMakeFiles/Fw_StringFormat_snprintf.dir/snprintf_format.cpp.o
FAILED: F-Prime/Fw/Types/CMakeFiles/Fw_StringFormat_snprintf.dir/snprintf_format.cpp.o 
...
In file included from <path>/fprime/Fw/Cmd/CmdString.hpp:13,
                 from <path>/fprime/Fw/FPrimeBasicTypes.hpp:24,
                 from <path>/fprime/Fw/Types/format.hpp:13,
                 from <path>/fprime/Fw/Types/snprintf_format.cpp:6:
<path>/fprime/Fw/Types/StringBase.hpp:66:5: error: ‘FormatStatus’ does not name a type; did you mean ‘SerializeStatus’?
     FormatStatus format(const CHAR* formatString, ...);  //!< write formatted string to buffer
     ^~~~~~~~~~~~
     SerializeStatus
<path>/fprime/Fw/Types/StringBase.hpp:67:5: error: ‘FormatStatus’ does not name a type; did you mean ‘SerializeStatus’?
     FormatStatus vformat(const CHAR* formatString, va_list args);  //!< write formatted string to buffer using va_list
     ^~~~~~~~~~~~
     SerializeStatus
```
I believe this is due to `Fw/Types/format.hpp` depending on `Fw/FPrimeBasicTypes.hpp` prior to the definition of `FormatStatus`, I'd need some guidance on if there's a way to cleanly untangle those dependencies.

bocchino:
You are right. I think your solution is best, given where those assertions are.

bocchino (approved):
Looks good to me!

thomas-bc (approved):

PR #3956, opened 2025-07-30
Updates `Types.fpp` to use optimized sizes for enumerations

kyleajones:
- Update enumerations in Types.fpp to use U8 instead of the default I32; the full I32 size is not needed for these simple enumerations

| | |
|:---|:---|
|**_Related Issue(s)_**|  n/a |

---
## Change Description

This change request modifies the enumeration size for all enums listed in `Types.fpp` from the default I32 type to the smaller U8 type.

## Rationale

This change will improve the memory footprint for telemetry channels that are defined by structs that contain members of types defined in this file.  Particularly useful in memory-constrained systems that use F-Prime.

The change is straightforward.  One review item worth considering would be using I8 instead of U8 as the type.  In either case, all enums are well within the bounds of either type.  I don't know if F-Prime has a preference to use the signed version of the type for these enums or if unsigned is fine.  

## Future Work

No planned future work associated with this issue.

### DIFF
--- a/Fw/Types/Types.fpp
+++ b/Fw/Types/Types.fpp
@@ -7,14 +7,14 @@ module Fw {
   type PolyType
 
   @ Serialization status
-  enum SerialStatus {
+  enum SerialStatus : U8 {
     OK, @< Serialization operation succeeded
     FORMAT_ERROR @< Data was the wrong format (e.g. wrong packet type)
     NO_ROOM_LEFT  @< No room left in the buffer to serialize data
   }
 
   @ Deserialization status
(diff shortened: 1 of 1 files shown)


celskeggs (approved):

LeStarch (approved):

LeStarch:
@bocchino could you weigh in on the use of an unsigned type as the enumeration backing type?  Is this ok to do?

thomas-bc (approved):

bocchino:
Yes, that seems fine.

Issue #3944, opened 2025-07-28
After FPP updates to new v4.0 serialization methods, mark legacy serialization methods as DEPRECATED


vincewoo:
This ticket is a follow-up to #3781 

As part of the changes to that ticket we chose not to DEPRECATE the now legacy serialization methods, serialize and deserialize. This was done so that we could decouple the change needed to FPP.

Once the FPP ticket (https://github.com/nasa/fpp/issues/763) is closed we should DEPRECATE the legacy methods in fprime to encourage the use of the new style.

vincewoo:
As part of this change we will have to undo the backwards compatibility workarounds that were introduced to decouple the FPP changes:

- [x]  Make deserializeFrom and serializeTo in `Fw::Serializable` pure virtual again and remove the default implementations. https://github.com/nasa/fprime/blob/67be6defa09159da639285cd6b51db61b231f5d3/Fw/Types/Serializable.cpp#L20-L35
- [x] Remove fallback logic of serializeFrom and deserializeTo in `Fw::SerializeBufferBase`. https://github.com/nasa/fprime/blob/67be6defa09159da639285cd6b51db61b231f5d3/Fw/Types/Serializable.cpp#L269-L280 https://github.com/nasa/fprime/blob/67be6defa09159da639285cd6b51db61b231f5d3/Fw/Types/Serializable.cpp#L562-L572 
- [x] Remove fallback logic of serializeTo and deserializeFrom in `Fw::Com::ComPacket`. https://github.com/nasa/fprime/blob/67be6defa09159da639285cd6b51db61b231f5d3/Fw/Com/ComPacket.cpp#L29-L37
- [x] Remove definitions of serialization methods in `Fw::Com::ComPacket` and inherit from `Fw::Serializable` directly. https://github.com/nasa/fprime/blob/67be6defa09159da639285cd6b51db61b231f5d3/Fw/Com/ComPacket.hpp#L27-L33
- [x] Remove fallback logic of serializeTo and deserializeFrom in `Fw::Time`. https://github.com/nasa/fprime/blob/67be6defa09159da639285cd6b51db61b231f5d3/Fw/Time/Time.cpp#L78-L101
- [ ] Remove SerializableStatus `FW_SERIALIZE_UNIMPLEMENTED` and `FW_DESERIALIZE_UNIMPLEMENTED` introduced by #3962

LeStarch:
[...]

PR #3880, opened 2025-07-14
Renaming serialize/deserialize methods to serializeTo/serializeFrom and deserializeTo/deserializeFrom

vincewoo:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3781  |
|**_Has Unit Tests (y/n)_**| y, pre-existing tests updated |
|**_Documentation Included (y/n)_**| n |

---
## Change Description

This change is part of the #3599 suggestions for code consistency. It is primarily a set of function name changes to clarify the direction of serialization and deserialization. In general, F' we has two flavors of serialization methods:

1. `<buffer>.serialize(<element>)`
2. `<element>.serialize(<buffer>)`

These both mean that the element is being serialized to the buffer. Whereas the following two flavors of `deserialize` methods mean that we are deserializing from the buffer to the element.

1. `<buffer>.deserialize(<element>)`
2. `<element>.deserialize(<buffer>)`

To improve the understanding of the direction of the serialization/deserialization this change renames the functions to `serializeTo`/`serializeFrom` and `deserializeTo`/`deserializeFrom`.

## Rationale

Improve readability of F' code to allow quicker assessment of serialization/deserialization direction.

## Testing/Review Recommendations

Unit testing was updated with new function names and re-run, 100% passing.

## Future Work

Will need to update FPP shortly afterwards to auto-code new function names.

PR #3962, opened 2025-07-31
Improve backward compatibility [de]serialize[From/To] fallback on Fw::Serializable and Fw::Buffer

Kronos3:
Related to https://github.com/nasa/fprime/issues/3944

We need this for https://github.com/nasa/fpp/pull/766

In the FPP PR `serialize`/`deserialize` has been ported `[de]serialize[From/To]`. In F Prime Serializable, to keep backwards compatibility we have a fallback clause in the default serializable/buffer:
https://github.com/nasa/fprime/blob/3dc73f9d72658310017e1ace8cac23771213d46e/Fw/Types/Serializable.cpp#L562-L573

The issue is that `FW_DESERIALIZE_FORMAT_ERROR` is overloaded in use case for enums. When an enum deserialization is attempted and the value is not an enumerated constant in the enum type, the deserialization will return `FW_DESERIALIZE_FORMAT_ERROR`. The current fallback code will cause a _double_ deserialization (the same happens in the serialization code). 

This PR adds a new status code that disambiguates not implementing the deserialization with a format error.


Kronos3:
We need https://github.com/nasa/fpp/pull/766 for this to work!

vincewoo (approved):
This looks good to me, thanks for fixing this!

bocchino:
Why are we providing a default `deserializeTo` that returns UNIMPLEMENTED and then catching UNIMPLEMENTED here? This seems convoluted. Wouldn't it be simpler to provide the implementation here as the default implementation of deserializeFrom? Is there some reason that won't work?

bocchino:
Also I have concerns about adding a new enumeration to the F Prime serialization interface in order to support the code above, which looks like a temporary workaround for an inconsistency between F Prime and FPP.

bocchino (changes requested):
I would like to see my concerns addressed before this PR is merged.

LeStarch (dismissed):

bocchino:
I think the problem is here:
https://github.com/nasa/fprime/blob/60db7b2364bee9cb3dc7b0221df020979b170ee4/Fw/Com/ComPacket.cpp#L29-L46
We have `deserialize` calling `deserialzeTo`. This seems backwards, unless and until there is an FPP change. Without the FPP change, I think `deserializeTo` must be a convenience method that calls `deserialize`. It seems we are trying to change the interface without changing code that relies on the interface, and that is leading to a workaround and then another workaround to fix the first workaround.

bocchino:
Maybe the way forward is to merge this PR, merge https://github.com/nasa/fpp/pull/766, and then fix up F Prime to remove the workarounds introduced in #3880 and this PR, including the UNIMPLEMENTED serialization status. Is there a clear path to doing all that before the next release? If so, that would address my concerns.

vincewoo:
@bocchino Yes, that is the plan that we discussed in the team meeting on Wednesday. The follow-up change to remove the work-arounds and DEPRECATE the legacy serialize and deserialize methods would be handled in #3944 . I will specifically note the UNIMPLEMENTED reversal in there now so we don't forget.

Fw/Types/Serializable.hpp
```diff
@@ -18,7 +18,9 @@ typedef enum {
     FW_DESERIALIZE_BUFFER_EMPTY,   //!< Deserialization buffer was empty when trying to read more data
     FW_DESERIALIZE_FORMAT_ERROR,   //!< Deserialization data had incorrect values (unexpected data types)
     FW_DESERIALIZE_SIZE_MISMATCH,  //!< Data was left in the buffer, but not enough to deserialize
-    FW_DESERIALIZE_TYPE_MISMATCH   //!< Deserialized type ID didn't match
+    FW_DESERIALIZE_TYPE_MISMATCH,  //!< Deserialized type ID didn't match
+    FW_SERIALIZE_UNIMPLEMENTED,    //!< Serialize has not be implemented for this type
+    FW_DESERIALIZE_UNIMPLEMENTED,  //!< Deserialize has not be implemented for this type
```
bocchino:
```suggestion
    FW_SERIALIZE_UNIMPLEMENTED,    //!< Serialization function called is not implemented
    FW_DESERIALIZE_UNIMPLEMENTED,  //!< Deserialization function called is not implemented
```

bocchino (dismissed):
Thanks, the comments here and in #3944 have addressed my concerns. Please note my suggested changes to the comments in the code.

LeStarch:
@Kronos3 based on @bocchino's comments, I am expecting:

1. New code comments
2. An FPP Alpha release

Is this accurate?

Kronos3:
Yes, here's how I understand it

1. Update code comment (done in [c26f979](https://github.com/nasa/fprime/pull/3962/commits/c26f9793d2ec8b98e5aa09f0d2bcc3c764d8e21a))
2. Merge this PR
3. Merge https://github.com/nasa/fpp/pull/766 & create FPP alpha
4. Vince to do https://github.com/nasa/fprime/issues/3944#issuecomment-3143586099

thomas-bc (approved):

Issue #3781, opened 2025-06-23
Clarify serialize() method names


thomas-bc:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Submitted as part of https://github.com/nasa/fprime/discussions/3599 and approved for implementation

Confusing naming for serialization functions. <element>.serialize(<buffer>), while sometimes it's <buffer>.serialize(<element>). Maybe rename functions to serializeTo() deserializeFrom() where applicable for clarity

vincewoo:
Completed!

PR #3971, opened 2025-08-01
(De)Serialization clean up of temporary workarounds

vincewoo:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3944 |
|**_Has Unit Tests (y/n)_**| y, pre-existing tests |

---
## Change Description

This change continues the effort for #3781. The FPP change has been completed to be compatible with the new serialization methods. This PR removes the temporary workarounds introduced by #3880 that we had in place to decouple with FPP.

## Rationale

Cleaning up the temporary workarounds left that we had in place to decouple the necessary FPP upgrade completed by https://github.com/nasa/fpp/issues/763.

No new UTs were added as part of this change. Existing were updated where necessary and re-run for regression.

## Future Work

We still need to mark the legacy `serialize` and `deserialize` methods of `Fw::Serializable` and `Fw::SerializeBufferBase` as DEPRECATED.


LeStarch:
@vincewoo I don't want the functions deprecated yet. Can we just remove the temporary variables?  Next release we will DEPRECATE and then remote the release after that.

vincewoo:
> @vincewoo I don't want the functions deprecated yet. Can we just remove the temporary variables? Next release we will DEPRECATE and then remote the release after that.

Yes, I can do that!

Fw/Types/StringBase.hpp
```diff
@@ -66,13 +66,12 @@
     FormatStatus format(const CHAR* formatString, ...);  //!< write formatted string to buffer
     FormatStatus vformat(const CHAR* formatString, va_list args);  //!< write formatted string to buffer using va_list
 
-    SerializeStatus serializeTo(SerializeBufferBase& buffer) const override;
-    SerializeStatus serializeTo(SerializeBufferBase& buffer, SizeType maxLen) const;
-    SerializeStatus deserializeFrom(SerializeBufferBase& buffer) override;
+    virtual SerializeStatus serializeTo(SerializeBufferBase& buffer) const override;
```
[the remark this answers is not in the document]
vincewoo: Removed virtual keyword

Fw/Types/StringBase.hpp
```diff
@@ -66,13 +66,12 @@
...
+    virtual SerializeStatus serializeTo(SerializeBufferBase& buffer, SizeType maxLen) const;
+    virtual SerializeStatus deserializeFrom(SerializeBufferBase& buffer) override;
```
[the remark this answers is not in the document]
vincewoo: Removed virtual keyword

Fw/Types/StringBase.hpp
```diff
@@ -66,13 +66,12 @@
...
 
-    SerializeStatus serialize(SerializeBufferBase& buffer) const override;
-    SerializeStatus serialize(SerializeBufferBase& buffer, SizeType maxLen) const;
-    SerializeStatus deserialize(SerializeBufferBase& buffer) override;
+    virtual SerializeStatus serialize(SerializeBufferBase& buffer) const override;
```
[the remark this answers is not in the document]
vincewoo: Removed virtual keyword

vincewoo:
@thomas-bc Note that this PR is deleting all the serialize/deserialize methods from the child classes so they inherit properly from the parent now. This makes it easier for use to later deprecate as well.

Fw/Types/Serializable.hpp
```diff
@@ -18,9 +18,7 @@ typedef enum {
     FW_DESERIALIZE_BUFFER_EMPTY,   //!< Deserialization buffer was empty when trying to read more data
     FW_DESERIALIZE_FORMAT_ERROR,   //!< Deserialization data had incorrect values (unexpected data types)
     FW_DESERIALIZE_SIZE_MISMATCH,  //!< Data was left in the buffer, but not enough to deserialize
-    FW_DESERIALIZE_TYPE_MISMATCH,  //!< Deserialized type ID didn't match
-    FW_SERIALIZE_UNIMPLEMENTED,    //!< Serialization function called is not implemented
-    FW_DESERIALIZE_UNIMPLEMENTED,  //!< Deserialization function called is not implemented
```
vincewoo, line 23: Note that we are deleting these SerializeStatus enum values that were temporarily introduced in #3962

Kronos3 (commented):
https://github.com/nasa/fprime/blob/fe88650efd081e6c5d98a2d17442ac77f19a853b/Fw/Types/Serializable.hpp#L39-L41

I see serialize/deserialize are still `virtual`, we may not want this since they should not be overridden. Maybe making them inline in the header is better?

Fw/Types/StringBase.cpp
```diff
@@ -153,12 +149,14 @@ SerializeStatus StringBase::serializeTo(SerializeBufferBase& buffer, SizeType ma
     return buffer.serializeFrom(reinterpret_cast<const U8*>(this->toChar()), len, Serialization::INCLUDE_LENGTH);
 }
 
-SerializeStatus StringBase::serialize(SerializeBufferBase& buffer, SizeType maxLength) const {
-    return this->serializeTo(buffer, maxLength);
+// Deprecated method for backward compatibility
+SerializeStatus StringBase::serialize(SerializeBufferBase& buffer) const {
+    return this->serializeTo(buffer);
```
thomas-bc, line 154: Why is this one still needed? Shouldn't it be inheritable from Fw::Serializable ?

vincewoo: This is a hiding issue, because `StringBase::serialize(buffer, maxLen)` version still exists we need `StringBase::serialize(buffer)` still in place as well

Fw/Types/Serializable.hpp
bocchino: I believe `FW_DESERIALIZE_TYPE_MISMATCH` was there before, but was not used, because there are no serialized type IDs. So maybe deleting it is right, but the deletion is not required to undo the workaround introduced in #3962.

Fw/Types/StringBase.cpp
bocchino: Is the idea to change both of those functions to `serializeTo` in a a future PR? Should we open an issue for that? I don't think it has to be handled in this PR.

Kronos3 (approved):
I see the `virtual` on `serialize`/`deserialize` has been addressed, looks great! Thanks for the fast turnaround on this!

Fw/Types/Serializable.hpp
vincewoo: You're right, I missed that MISMATCH was in there before. I can put it back in for now, it shouldn't be removed as a side-effect in this change.

Fw/Types/StringBase.cpp
bocchino: Is this function necessary in light of the comment above? The Serializable interface is already providing this function, isn't it? Also, since the Serializable function is not virtual, it seems this function hides (but does not override) the same function in the base class. The C++ coding standard prohibits this kind of hiding.

Fw/Types/Serializable.hpp
Kronos3: I don't think you missed it, you just removed the `,` (see the addition below)

bocchino:
Looks good, I just had a few questions/comments. @LeStarch @thomas-bc can we remove the Deferred to Next Release label from this PR? I think it is close, and it's important for the release.

Fw/Types/Serializable.hpp
vincewoo: Oh, I see that it was actually left in there. I vote we handle the removal of it in another issue if needed.

Fw/Types/StringBase.cpp
vincewoo: These should be handled in the future when we mark all the legacy serialize/deserialize as DEPRECATED. I expect these two functions will also be DEPRECATED.

bocchino: Now I see that @thomas-bc asked the same question I did. I don't understand the answer. It does not seem either necessary or correct to hide a function `serialize` with another function `serialize` that does the same thing. Maybe I am not understanding.

Fw/Types/Serializable.hpp
bocchino: @Kronos3 You are right, I missed the addition. Looks good!

LeStarch (approved):

Fw/Types/StringBase.cpp
bocchino: After discussion, it seems that `serialize` with two arguments is hiding the inherited `serialize` with one argument. So we need to restate the `serialize` with one argument here. This hiding will go away when we remove the `serialize` functions in a later PR.

bocchino (approved):
Looks good!

thomas-bc (approved):

LeStarch:
Got verbal buy off from @thomas-bc and @bocchino .  Good to merge.

Fw/Types/StringBase.cpp
LeStarch: We deemed it necessary.  There was a new variant of the function with new arguments, that would otherwise hide the inherited implementation.

Issue #3888, opened 2025-07-17
FSW Fatal when Ctrl+C GDS


kevin-f-ortega:
| | |
|:---|:---|
|**_F´ Version_**| devel|
|**_Affected Component_**| IpSocket  |
---
## Problem Description

Get FATAL in FSW when Ctrl+C the GDS

```
0x200fc130 (ReceiveTask): [WARNING] Failed to recv from port with status -8 and errno 0
0x200fa4d0 (comQueue): Assert: "/home/ortega/projects/platforms/fprime-vxworks-reference/lib/fprime/Drv/Ip/IpSocket.cpp:132" 0
0x200fa4d0 (comQueue): FATAL 16897 handled.
```

## Context / Environment

Execute `fprime-util version-check` and share the output.

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-4.18.0-553.16.1.el8_10.x86_64-x86_64-with-glibc2.28
Python version: 3.11.9
CMake version: 3.26.0
Pip version: 25.1.1
Pip packages:
    fprime-tools==4.0.0a6
    fprime-gds==4.0.0a7
    fprime-fpp-*==3.0.0a13
Project submodules:
```

## How to Reproduce

1. Checkout fprime-vxworks-reference on branch `upgrade-to-fprime-4`
2. cd ReferenceDeployment
3. fprime-util generate
4. fprime-util build
5. Load and run FSW on BBB
6. Connect GDS to FSW
7. Send No-op Command
8. Send Version ALL command
9. Ctrl+C GDS
10. Watch FATAL

## Expected Behavior

FSW to not fatal after Ctrl+C GDS

jsilveira1409:
I think this assert should not be there in the first place. After running some tests with a `TcpServer` Component, it manages to get abruptly disconnected by a client and reconnected afterwards. Especially considering that there is an if statement in the `send()` function that gracefully handles the case in which the client was disconnected during a send.

LeStarch:
Fixed thanks to @jsilveira1409 !

PR #3973, opened 2025-08-03
Removed sock fd assert on IpSocket

jsilveira1409:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
#3888 

---
## Change Description

Removed unnecessary assert for the `socketDescription.fd` when sending data out. 

## Rationale

When the FSW streams out data to a _e.g._ TCP client, and the client disconnects abruptly, the assert get's triggered, crashing the FSW instance, instead of letting it fail to send in the `sendProtocol()` function further below in the `IpSocket::send()` function.

### DIFF
--- a/Drv/Ip/IpSocket.cpp
+++ b/Drv/Ip/IpSocket.cpp
@@ -130,7 +130,6 @@ SocketIpStatus IpSocket::open(SocketDescriptor& socketDescriptor) {
 }
 
 SocketIpStatus IpSocket::send(const SocketDescriptor& socketDescriptor, const U8* const data, const U32 size) {
-    FW_ASSERT(socketDescriptor.fd != -1, static_cast<FwAssertArgType>(socketDescriptor.fd));
     FW_ASSERT(data != nullptr);
     FW_ASSERT(size > 0);
 


LeStarch:
I am not sure this is the fix we want.  The intent of these assertions was to ensure that the file descriptor is not an invalid file descriptor in both `send` and `recv`. 

Look at the receive code:

https://github.com/nasa/fprime/blob/e7840dee811c1a2625a8f2ce4ebc99cda8796d0b/Drv/Ip/SocketComponentHelper.cpp#L163-L172

Here, we clearly check to avoid a case where the assert would trip unnecessarily.

The send code is a bit different.

https://github.com/nasa/fprime/blob/e7840dee811c1a2625a8f2ce4ebc99cda8796d0b/Drv/Ip/SocketComponentHelper.cpp#L112-L120

Here it attempts to reopen a failed descriptor, checks status, and then persists another copy.  This leads to a race condition because something else (e.g. `recv`) could close the descriptor between a successful open, and when the new copy is persisted.

This code should instead:
1. Call `reopen`
2. Safely persist a local copy of the descriptor.
3. Check for and return bad statuses
4. Check for fd == -1 and return DISCONNECTED

@vincewoo does this make sense?

LeStarch:
We should also put back in the assert in the `recv` function.

LeStarch:
@jsilveira1409 do you want to attempt the above change, or shall I?

jsilveira1409:
@LeStarch hey thanks for the feedback! I'll give it a shot in the coming days and propose a solution! Thanks

LeStarch:
It turns out this is a whole lot more complicated that first assumed.  The crash comes from the fact that in the TcpServer re-connections are potentially a long process (wait listening for some unknown client).

We need this fix for the release today, so I am going to go ahead and make it as it affects multiple files.

For the time being, I will merge this PR as it will partially address the problem, and will give you credit.

LeStarch (approved):

LeStarch:
We looked at this yet deeper, and decided this was the correct fix for now.  A better fix would be to push re-connection requests onto their own thread, and get rid of this weird inter-threading logic. 

However, for now this code results in an error when threads are reconnecting and haven't completed.  This is the stop-gap solution.

jsilveira1409:
I haven't had time to look more into it yet but will study it deeper when I do! Thanks for the merge!

Issue #3849, opened 2025-07-05
Fpy stack-based virtual machine


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0|
|**_Affected Component_**| FpySequencer |
---
## Feature Description

Transition the FpySequencer and Fpy compiler from register-based to stack-based.

## Rationale

Simpler to implement at compiler and virtual machine level. Register-based machines are often used because they're faster when the real hardware is register based, but in this case, the VM is executing at a much higher level than the actual hardware, so I doubt we get any performance gains. In any case, it's a scripting language, it doesn't need to be high performing. Stack-based machines are simpler to understand, write and compile code for.

zimri-leisher:
@bocchino  I was thinking more about non-constant sized lvars/non-word-aligned stacks, do you know if there are any languages with bytecodes which do this?

bocchino:
LLVM uses bitcode, and it supports allocation of non-scalar variables on the stack.

zimri-leisher:
@LeStarch based on the data type i'm using, i'm going to make it impossible to make the stack larger than 2^16 bytes (65535). that sounds very likely to be big enough for all use cases

LeStarch:
@zimri-leisher why not use one of the project configurable sizes, like `FwSizeType`?

PR #3975, opened 2025-08-04
FpySequencer stack based architecture

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3849  |

---
## Change Description

This PR updates the FpySequencer to be stack based, rather than register based. This is a much simpler architecture to implement and compile for.

* Remove all register manip directives: SET_REG, SET_SER_REG, DESER_SER_REG
* Remove registers and serializable registers
* Adds a stack to the runtime
   * The stack is just a byte array, we keep track of the begin/end of the array
* Change all directives which interacted with registers to push/pop to/from the stack
   * E.g. instead of the IF directive checking a register for its conditional, it pops a byte off the stack and uses that
* Adds stack manip directives: ALLOCATE, STORE, LOAD, PUSH_VAL
   * ALLOCATE pushes empty bytes to the stack
   * STORE pops off of stack into lvar array
   * LOAD pushes from lvar array onto stack
   * PUSH_VAL pushes a const byte array to the stack
* Change some directives to get args from the stack instead of from const args: EXIT, WAIT_ABS, WAIT_REL
   * EXIT pops a U8 off the stack to serve as return code of the seq
   * WAIT_ABS pops the absolute time off the stack, as a Fw::Time obj
   * WAIT_REL pops the seconds/useconds off the stack, as U32s


thomas-bc:
@zimri-leisher not for this PR but putting this on your radar:
we've been slowly putting C++ Code Formatting checks in CI #1984  . Please consider in the near future including in a short PR the following changes:

- adding Svc/FpySequencer to the list of checked dirs https://github.com/nasa/fprime/blob/02afd01545ba8363f9fe6dc5c68b7f7299cd6f3c/.github/workflows/format-check.yml#L32-L35
- running `fprime-util format --dirs Svc/FpySequencer` and opening a PR for these changes

zimri-leisher:
@thomas-bc okay, will do. I can do that as soon as this PR merges!
Also, if you're interested, I would love to have your review on this code. I've got a ton of code coming down the pipeline, it would be a huge help for me to get quick reviews!

Svc/FpySequencer/FpySequencer.hpp
thomas-bc, line 1: `sdd.md` still refers to register-based operations and things. If you're not updating the SDD much because so much is changing during development, may want to add a

>[!WARNING]
> SDD is not maintained during development phase

or something

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -298,75 +185,91 @@ Fw::Success FpySequencer::deserializeDirective(const Fpy::Statement& stmt, Direc
...
+            // check to make sure the value will fit in the FpySequencer_PushValDirective::val buf
+            if (bufSize > Fpy::MAX_DIRECTIVE_SIZE) {
+                this->log_WARNING_HI_DirectiveDeserializeError(stmt.get_opCode(), this->m_runtime.nextStatementIndex - 1,
+                                                               Fw::SerializeStatus::FW_DESERIALIZE_FORMAT_ERROR,
+                                                               argBuf.getBuffLeft(), argBuf.getBuffLength());
+                return Fw::Success::FAILURE;
+            }
+
+            // okay, it will fit. put it in
+            status = argBuf.deserialize(deserializedDirective.pushVal.get_val(), bufSize, Fw::Serialization::OMIT_LENGTH);
```
thomas-bc, line 260: not required for this PR but in the near future make sure to switch to `deserializeTo` here, and all other `serialize()/deserialize()` calls should also be updated to `serializeFrom/To()` `deserializeFrom/To()`

thomas-bc (commented):
Reviewed to the best of my ability - looks good to me! A few notes but nothing blocking

Svc/FpySequencer/FpySequencerRunState.cpp
zimri-leisher: Thanks @thomas-bc ! Really appreciate it, glad it mostly looked good to you. I'd always be happy to call and go over any of the code if you have more questions.
I'll update the SDD.
Noted on the deserialize/serialize updates, I'll make that fix in an upcoming PR.

thomas-bc: It all made sense tbh - you've made the architecture easy to follow for how complex this component is!

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -28,8 +28,128 @@ void FpySequencer::sendSignal(Signal signal) {
...
+
+template <> F64 FpySequencer::pop<F64>() {
+    U64 endianness = this->pop<U64>();
+    F64 val;
+    memcpy(&val, &endianness, sizeof(val));
+    return val;
+}
+
+template <typename T>
+void FpySequencer::push(T val) {
```
bocchino, line 92: This code looks like it may be doing right shifts of signed values. The result of that operation is not defined in C++ and should be avoided. It's better to do the twos-complement arithmetic explicitly, and only shift unsigned values.

bocchino (commented):
Looks good! I just had a comment about shifting signed values -- we should avoid that. Also it would be good to format the code.

Svc/FpySequencer/FpySequencerDirectives.cpp
bocchino: You should be able to fix this by casting between signed and unsigned values, and shifting unsigned values. I don't think any complicated arithmetic is required. You can use `std::make_unsigned` to get the unsigned type with the same bit width as `T`.

zimri-leisher:
Thanks for the feedback! The formatting is coming in the next PR, which I already have ready and waiting for when this gets in.
Good point about signed bit shifts, I'll make the fix.
Perhaps if @LeStarch is busy, we could consider both of your reviews to count and we could merge this PR?

zimri-leisher:
@bocchino feedback resolved, if either you or @thomas-bc could approve that would be great

thomas-bc (approved):

thomas-bc:
@zimri-leisher sounds good to me - merging

PR #3980, opened 2025-08-04
Rate drive sent/receive telemetry

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Moves telemetry to a rate group call for UART send/receive.


timcanham:
I was wondering if we should make `this->m_bytesSent` and `this->m_bytesReceived` be `std::atomic` since multiple threads would access the data. (Hat tip to @bocchino).

LeStarch:
@timcanham agreed.

timcanham:
Otherwise, LGTM! Thanks!

LeStarch:
@timcanham want to give the final review?  They should be std atomic now.

timcanham (approved):

PR #3982, opened 2025-08-05
Add command dispatcher configuration item to toggle opcode registration events

ianbrault:
| | |
|:---|:---|

---
## Change Description

This PR adds a new pre-processor flag `CMD_DISPATCHER_ENABLE_OPCODE_REGISTER_EVENTS` to the command dispatcher configuration in `config/CommandDispatcherImplCfg.hpp` which can be used to enable/disable the command opcode registration (and re-registration) events that are downlinked during command registration. This is enabled by default to preserve existing behavior.

## Rationale

This change is useful for memory-constrained systems where the size of downlink buffers must be restricted. If a deployment contains a sufficient number of commands (which can be exacerbated by the auto-generated parameter commands) the opcode registration causes a large number of diagnostic events which can fill the downlink buffer and crowd out more useful initialization events that occur later on. This change adds a simple compile-time flag which can toggle these events off and clean up the initialization event stream.

Run with the flag set and verify that there is no change to existing behavior, then run with the flag unset and verify that opcode registration events are no longer downlinked during the command registration phase.

## Future Work

N/A

### DIFF
--- a/default/config/CommandDispatcherImplCfg.hpp
+++ b/default/config/CommandDispatcherImplCfg.hpp
@@ -15,6 +15,12 @@ enum {
     CMD_DISPATCHER_SEQUENCER_TABLE_SIZE = 25, // !< The size of the table holding commands in progress
 };
 
-
+// Toggles whether events are triggered during command opcode registration. Disabling this can be
+// useful for memory-constrained systems with limited downlink buffer space as the events can crowd
+// out more useful events during initialization.
+#ifndef CMD_DISPATCHER_ENABLE_OPCODE_REGISTER_EVENTS
+#define CMD_DISPATCHER_ENABLE_OPCODE_REGISTER_EVENTS \
+    1  //!< Indicates whether or not command opcode registration events are triggered
+#endif
 
 #endif /* CMDDISPATCHER_COMMANDDISPATCHERIMPLCFG_HPP_ */
--- a/Svc/CmdDispatcher/CommandDispatcherImpl.cpp
+++ b/Svc/CmdDispatcher/CommandDispatcherImpl.cpp
@@ -33,12 +33,16 @@ void CommandDispatcherImpl::compCmdReg_handler(FwIndexType portNum, FwOpcodeType
             this->m_entryTable[slot].opcode = opCode;
             this->m_entryTable[slot].port = portNum;
             this->m_entryTable[slot].used = true;
+            #if CMD_DISPATCHER_ENABLE_OPCODE_REGISTER_EVENTS
             this->log_DIAGNOSTIC_OpCodeRegistered(opCode, portNum, static_cast<I32>(slot));
+            #endif
             slotFound = true;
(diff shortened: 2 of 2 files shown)


default/config/CommandDispatcherImplCfg.hpp
```diff
@@ -15,6 +15,12 @@ enum {
...
 };
 
-
+// Toggles whether events are triggered during command opcode registration. Disabling this can be
+// useful for memory-constrained systems with limited downlink buffer space as the events can crowd
+// out more useful events during initialization.
+#ifndef CMD_DISPATCHER_ENABLE_OPCODE_REGISTER_EVENTS
+#define CMD_DISPATCHER_ENABLE_OPCODE_REGISTER_EVENTS \
+    1  //!< Indicates whether or not command opcode registration events are triggered
+#endif
```
celskeggs, line 24: What's the possibility of using a `constexpr bool` instead, and then using an ordinary `if` statement? That way, we can avoid more preprocessor conditionals and improve the typechecking, but a smart compiler can still eliminate the code.

ianbrault: That is a good idea, I will modify that

LeStarch:
I don't follow why this is needed.  ActiveLogger has an event filter in it.  If downlink constrained systems wish to avoid this, then why not turn off `DIAGNOSTIC` events there.  This would prevent them from entering the downlink steam.

LeStarch:
Since this wasn't tracked by an issue, our Change Control Board has not weighed in on the work.  I've added an issue here so that we can review it: https://github.com/nasa/fprime/issues/3983

This issue will likely be disapproved unless we can refute the following points:
1. Why is ActiveLogger event filtering insufficient for protecting downlink?
2. An aggregate switch to turn off CommandDispatcher `DIAGNOSTIC` events seems like an onerous solution for handling a general problem of `DIAGNOSTIC` events are voluminous.  This would seem to imply that any component with similar events would require configuration, which could greatly expand the configuration parameters to track.

ianbrault:
@LeStarch is there a way to configure the ActiveLogger event filters at compile time? That was my original goal but I was searching thru the SDD/code and was not able to find anything that could accomplish this, I was only able to find the `SET_ID_FILTER` command to filter them at runtime.

I think disabling all `DIAGNOSTIC` events is too big of a hammer. These command opcode registration events pose an issue for us because they are triggered during the deployment setup so they occur before we can downlink other events from the deployment initialization which we would rather see, but we are still interested in `DIAGNOSTIC` events later on and during runtime. I guess to summarize, we are looking for a way to filter or clean up the downlink during initialization.

ianbrault:
@LeStarch if there isn't an existing way to configure event filters at compile time, or at runtime prior to the deployment `setup()` call, I'd be happy to try to think up a proposal for this

ianbrault:
Closing this in favor of https://github.com/nasa/fprime/pull/4028

Issue #3876, opened 2025-07-12
FpySequencer arithmetic


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**| fpyc, FpySequencer |
---
## Feature Description

Add arithmetic to Fpy.
* Exponentiation
* Multiplication
* Division
* Subtraction
* Unary negation, unary plus

PR #4025, opened 2025-08-13
FpySequencer arithmetic

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3876  |

---
## Change Description

* Added FpySequencer to formatting checks, formatted code
* New directives:
* STACK_CMD dispatches a command opcode + args from the stack. Used for cmds with variable arguments
* DISCARD drops a hardcoded number of bytes from the stack. Currently unused but could be helpful in the future.
* MEMCMP pops two bytearrays of hardcoded size off the stack and compares their contents. If equal, pushes 1 to stack, otherwise 0.
* Full suite of arithmetic ops: mul, div, add, sub, log, pow, mod, floor div, across floats, signed and unsigned ints. All take their args from the stack and push their result back onto the stack.
* Integer truncation and signed/unsigned extension. Previously only had float ext/trunc.


.github/workflows/format-check.yml
```diff
@@ -66,6 +66,7 @@ jobs:
             Svc/FprimeFramer
             Svc/FprimeProtocol
             Svc/FprimeRouter
+            Svc/FpySequencer
```
thomas-bc: All of `Svc/` should now be formatted so you can squash this entire list to a single entry

```
CHECKED_DIRS: >-
            Drv
            Fw
            Os
            Svc
```

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -29,10 +29,43 @@ void FpySequencer::sendSignal(Signal signal) {
     }
 }
 
+Fw::Success FpySequencer::sendCmd(FwOpcodeType opcode, const U8* argBuf, FwSizeType argBufSize) {
+    Fw::ComBuffer cmdBuf;
+    Fw::SerializeStatus stat = cmdBuf.serialize(Fw::ComPacketType::FW_PACKET_COMMAND);
```
thomas-bc: Need to be careful when serializing `Fw::ComPacketType` because the serialization size of `Fw::ComPacketType` is not automatically the same as `FwPacketDescriptorType`. Depending on how the value is deserialized this may cause issues.
The rest of the codebase is doing
```suggestion
    Fw::SerializeStatus stat = cmdBuf.serialize(static_cast<FwPacketDescriptorType>(Fw::ComPacketType::FW_PACKET_COMMAND));
```
We may actually want to put a ticket in to see if we should switch `FwPacketDescriptorType` to be equal to the width of Fw::ComPacketType (which is now U16 because of [this](https://github.com/nasa/fprime/blob/a070ee768aa00d6d433452410a766e3d04f2105c/default/config/ComCfg.fpp#L19C17-L19C20))

zimri-leisher:
@thomas-bc feedback resolved! good point about the packet descriptor type.

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -29,10 +29,44 @@ void FpySequencer::sendSignal(Signal signal) {
     }
 }
 
+Fw::Success FpySequencer::sendCmd(FwOpcodeType opcode, const U8* argBuf, FwSizeType argBufSize) {
+    Fw::ComBuffer cmdBuf;
+    Fw::SerializeStatus stat =
+        cmdBuf.serialize(static_cast<FwPacketDescriptorType>(Fw::ComPacketType::FW_PACKET_COMMAND));
+    // TODO should I assert here? this really shouldn't fail, I should just add a static assert
```
bocchino, line 36: Yes, if this is a "should not fail based on static checks" case, then an assertion seems correct.

bocchino: However, use an assertion only if the static checks are in place. Otherwise you could have a situation where a misconfiguration is not caught at compile time and then causes the FSW to abort.

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -649,10 +675,197 @@ DirectiveError FpySequencer::op_uitofp() {
     this->push(static_cast<F64>(this->pop<U64>()));
     return DirectiveError::NO_ERROR;
 }
-Signal FpySequencer::stackOp_directiveHandler(const FpySequencer_StackOpDirective& directive,
-                                              DirectiveError& error) {
-    // coding error, should not have gotten to this binary reg op handler
-    FW_ASSERT(directive.get__op() >= Fpy::DirectiveId::OR && directive.get__op() <= Fpy::DirectiveId::UITOFP,
+DirectiveError FpySequencer::op_iadd() {
+    if (this->m_runtime.stackSize < 16) {
```
bocchino: I'm seeing the magic number 16 here. Can this be replaced with a symbolic constant?

bocchino: Same comment below with regard to 8, 4, etc.

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -99,9 +102,9 @@ Fw::Success FpySequencer::deserializeDirective(const Fpy::Statement& stmt, Direc
             new (&deserializedDirective.noOp) FpySequencer_NoOpDirective();
             // no op does not need deser
             if (argBuf.getBuffLeft() != 0) {
-                this->log_WARNING_HI_DirectiveDeserializeError(stmt.get_opCode(), this->m_runtime.nextStatementIndex - 1,
-                                                               Fw::SerializeStatus::FW_DESERIALIZE_SIZE_MISMATCH,
-                                                               argBuf.getBuffLeft(), argBuf.getBuffLength());
+                this->log_WARNING_HI_DirectiveDeserializeError(
+                    stmt.get_opCode(), this->m_runtime.nextStatementIndex - 1,
```
bocchino: Could m_runtime have a member function `getCurrentIndex` that returns `nextStatementIndex - 1`? That might make the code a bit cleaner.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: There aren't any static assertions in place. I think I'm just going to leave this one as is actually, and remove the comment. If the user wants to set the combuffer size to less than 4 bytes, go ahead, I won't stop you!

zimri-leisher:
@bocchino feedback has been responded to

zimri-leisher:
Made one change to fix a mistake. Modulo directive needed signed and unsigned variants, but I only had signed.

zimri-leisher:
@thomas-bc @bocchino would either of you be able to provide an approval?

bocchino (approved):
Looks good! I see my comments have been addressed.

zimri-leisher:
@LeStarch if you'd like to look, that would be great, otherwise Rob has approved the changes so I think good to merge

thomas-bc (approved):

Issue #4027, opened 2025-08-14
Add an interface to configure event filters for text loggers


ianbrault:
| | |
|:---|:---|
|**_F´ Version_**| 4.0 |
|**_Affected Component_**| `ActiveTextLogger`, `PassiveTextLogger` |
---
## Feature Description

Add a new public interface to the text logger components that allows projects to configure filters for specific events.

## Rationale

Currently no method exists to configure specific event filters for the active/passive text logger components. Such filters are useful for memory-constrained systems where the size of downlink buffers must be restricted. This gives projects the flexibility to restrict the verbosity that comes from events in built-in components. Our particular use-case involves filtering the `CmdDispatcher` opcode registration events which can be quite numerous, since these are early in initialization they can crowd out more important events of higher priorities.

PR #4028, opened 2025-08-14
Add event ID filters to text logger components

ianbrault:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4027 |

---
## Change Description

Adds new `configure` interfaces to `ActiveTextLogger` and `PassiveTextLogger` components which can be used to add event ID filters. These filters are used to exclude messages from being logged at runtime.

Also adds configuration files which set the maximum amount of filters that can be added. This was done to mimic the filtering in the `EventManager` component.

## Rationale

Currently no method exists to configure specific event filters for the active/passive text logger components. Such filters are useful for memory-constrained systems where the size of downlink buffers must be restricted. This gives projects the flexibility to restrict the verbosity that comes from events in built-in components.

See #4027 for additional details.

Configure the text loggers to filter out specific events during deployment initialization and verify that they are no longer logged to their respective destinations.

None.

### DIFF
--- a/Svc/ActiveTextLogger/ActiveTextLogger.cpp
+++ b/Svc/ActiveTextLogger/ActiveTextLogger.cpp
@@ -21,7 +21,8 @@ ActiveTextLogger::ActiveTextLogger(const char* name) : ActiveTextLoggerComponent
 ActiveTextLogger::~ActiveTextLogger() {}
 
 void ActiveTextLogger::configure(const FwEventIdType* filteredIds, FwSizeType count) {
-    FW_ASSERT(count < ACTIVE_TEXT_LOGGER_ID_FILTER_SIZE, count, ACTIVE_TEXT_LOGGER_ID_FILTER_SIZE);
+    FW_ASSERT(count < ACTIVE_TEXT_LOGGER_ID_FILTER_SIZE,
+              static_cast<FwAssertArgType>(count), ACTIVE_TEXT_LOGGER_ID_FILTER_SIZE);
 
     this->m_numFilteredIDs = count;
     for (FwSizeType entry = 0; entry < count; entry++) {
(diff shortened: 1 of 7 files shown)


Svc/ActiveTextLogger/ActiveTextLogger.cpp
```diff
@@ -18,6 +20,15 @@ ActiveTextLogger::ActiveTextLogger(const char* name) : ActiveTextLoggerComponent
 
 ActiveTextLogger::~ActiveTextLogger() {}
 
+void ActiveTextLogger::configure(FwEventIdType* filteredIds, FwSizeType count) {
```
celskeggs: Argument `filteredIds` can and should be a `const` pointer

Svc/ActiveTextLogger/ActiveTextLogger.hpp
```diff
@@ -8,6 +8,7 @@
 
 #include <Svc/ActiveTextLogger/ActiveTextLoggerComponentAc.hpp>
 #include <Svc/ActiveTextLogger/LogFile.hpp>
+#include <config/PassiveTextLoggerCfg.hpp>
```
thomas-bc:
```suggestion
#include <config/ActiveTextLoggerCfg.hpp>
```

thomas-bc (changes requested):

Svc/ActiveTextLogger/ActiveTextLogger.hpp
ianbrault: Oops copy-paste error, thank you for catching that

thomas-bc:
@timcanham @LeStarch would you please weigh in on the methodology here?

LeStarch (approved):
This was what I was expecting.

LeStarch (approved):

thomas-bc (approved):

LeStarch:
A follow-up suggestion was to add in support for severity filtering too.  We'll track this as separate work.

thomas-bc:
https://github.com/nasa/fprime/issues/4082

PR #4035, opened 2025-08-15
Support disabling Svc/GenericHub

celskeggs:
| | |
|:---|:---|

---
## Change Description

Add FPRIME_ENABLE_SERIAL_HUBS cmake option to disable compilation of Svc/GenericHub.

## Rationale

This allows FW_PORT_SERIALIZATION to be set to 0.

N/A

## Future Work

N/A

Not used.

### DIFF
--- a/Svc/CMakeLists.txt
+++ b/Svc/CMakeLists.txt
@@ -42,7 +42,6 @@ add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/FprimeProtocol/")
 add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/FprimeRouter/")
 add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/FrameAccumulator/")
 add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/FramingProtocol/")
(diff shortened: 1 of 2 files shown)


LeStarch (changes requested):
The concern here is that we are adding a flag specifically to disable hubs, where there is no guarantee that these hubs are the only `serial` types used in the system.  Done this way, it would result in a lot of feature flags.

A possible other solution would be to mark the GenericHub `EXCLUDE_FROM_ALL`.  This would prevent it from building unless it is in the dependency graph of the build.  In this case it would be a legitimate error. 

This is how we updated the sub-topologies that added `restrict_platforms`.  See: https://github.com/nasa/fprime/blob/6915c93c10d4af11ccbb87ac81b6e9be4cd5ec22/Svc/Subtopologies/CdhCore/CdhCoreConfig/CMakeLists.txt#L2.

celskeggs:
`EXCLUDE_FROM_ALL` should work. I'll give it a try.

LeStarch:
@celskeggs did this work?  If so, I'd like to get this PR updated, or closed.

LeStarch:
See #4178.  This is the recommended fix.  If this does not work, reopen this PR.

celskeggs:
Thanks, looks good, sorry for the slowness on our end.

PR #4042, opened 2025-08-18
Add logic for assert failure short circuiting in `AssertFatalAdapter`

kyleajones:
| | |
|:---|:---|
|**_Has Unit Tests (y/n)_**| No additions to current unit tests, tests pass |

---
## Change Description

This change adds guard logic to `AssertFatalAdapter::reportAssert` to prevent an infinite loop of cascading assert reporting in the case where a FW_ASSERT check fails in the log_FATAL port calls.

## Rationale

In the event of a system running low on resources to allocate additional buffer memory for the serialized data used in the log_FATAL port calls, FW_ASSERT checks in those port calls can trip, causing an infinite loop.  This guard check short circuits that infinite loop and forces the second FW_ASSERT check failure to exit early and hand off to the system-level assert handler.

This logic change was tested with a deployment that has constrained memory resources.  Upon triggering a FW_ASSERT check failure, the log_FATAL port call also failed to get buffer space for the filename data serialization, resulting in a second FW_ASSERT check failure.  With this logic in place, both asserts were logged and the system-level assert processing was allowed to proceed.  Notably, the FATAL event was never actually issued, due to the second FW_ASSERT check failure, but the infinite cascade of assert check failures did not occur.

## Future Work

I couldn't see a clear way to test this in the existing unit tests, but that may be worth exploring further.

### DIFF
--- a/Svc/AssertFatalAdapter/AssertFatalAdapterComponentImpl.cpp
+++ b/Svc/AssertFatalAdapter/AssertFatalAdapterComponentImpl.cpp
@@ -44,6 +44,8 @@ AssertFatalAdapterComponentImpl ::AssertFatalAdapterComponentImpl(const char* co
     this->m_adapter.regAssertReporter(this);
     // register adapter
     this->m_adapter.registerHook();
+    // Assert handling not in progress
+    this->m_inAssert = false;
 }
 
 AssertFatalAdapterComponentImpl ::~AssertFatalAdapterComponentImpl() {}
@@ -99,11 +101,13 @@ void AssertFatalAdapterComponentImpl::reportAssert(FILE_NAME_ARG file,
                             sizeof(msg));
     Fw::Logger::log("%s\n", msg);
 
-    // Handle the case where the ports aren't connected yet
-    if (not this->isConnected_Log_OutputPort(0)) {
+    // Handle the case where the ports aren't connected yet or assert handling already in progress
+    if (not this->isConnected_Log_OutputPort(0) || this->m_inAssert) {
         assert(0);
         return;
     }
+    // Since log calls below can also fail assert checks, guard against cascading failures
+    this->m_inAssert = true;
 
     switch (numArgs) {
         case 0:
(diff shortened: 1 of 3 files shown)


LeStarch (commented):
It seems that this code could cause an issue in some edge cases.  Namely, if two asserts happen near each other, the first would move into FATAL handling (which often gives the component and system some time to shut down).  A second assert could then trigger `assert` which would bring the system down immediately.

The existing function is really only there to catch the case where the FATAL handler is not hooked up at all, and so we have not choice but to assert.

However, I also see your point that if the log_FATAL system itself asserts, we go into an undefined state because no proper resolution can be reached.

The only solution I see that could fix both concerns would be to check if an ASSERT is already processing in the stack above (i.e. on the same thread).

kyleajones:
@LeStarch yeah, I agree with your analysis.  My proposed solution would prevent the original FW_ASSERT FATAL handling from completing properly in the edge case you presented.  I can alternatively use something like a `thread_local` variable to track the `inAssert` state locally in a given thread.  If the reportAssert function encountered an additional FW_ASSERT within the same thread, it would off-ramp with the system `assert`.  Otherwise, if separate threads were to encounter a failed FW_ASSERT check very near each other, they could both enter the system-level FATAL handling and complete their separate FATAL handling (or at least attempt to before one caused a system-level reset or other event).

LeStarch:
@kyleajones we talked about it with the leads team.  Thread local variables aren't right either.  The suggestion was to add a configuration for how many times you are allowed through that function before asserting.

Basically make your implementation a counter not a boolean, and compare to a config variable.  Projects wanting your behavior could choose a config of 1, projects wanting exactly todays functionality would choose `4000000` or something.  Most projects probably want ~10 which would prevent a run away, but also reduce the risk of accidental uncontrolled bail-out.

kyleajones:
@LeStarch I've made updates to match your proposed implementation.  I confirmed this repeating assert failure handling by injecting a FW_ASSERT check failure in my specific program testbed and saw the expected off-ramp behavior after the configured number of failures.  Please let me know if you'd like me to make any modifications.

LeStarch (approved):

timcanham (approved):
LGTM!

PR #4062, opened 2025-08-22
Data structure library, phase 2

### LINKED ISSUE #4054: Phase 2 of data structure library
This is a continuation of https://github.com/nasa/fprime/issues/3545. Add the red-black tree implementation of set and map.

bocchino:
Adds set and map implementations based on a red-black tree.

Closes #4054.


bocchino:
I need to fix some comments that clang accepts and that gcc treats as illegal multiline comments.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+    //! data must be aligned according to getByteArrayAlignment().
+    //! data must contain at least getByteArraySize(capacity) bytes.
+    void setStorage(ByteArray data,      //!< The data
+                    FwSizeType capacity  //!< The capacity
+    ) {
+        this->m_nodes.setStorage(data, capacity);
+        const auto nodesSize = Nodes::getByteArraySize(capacity);
+        // Compute the nearest offset at or after nodesSize that is aligned for FreeNodes
+        const auto freeNodesAlignment = FreeNodes::getByteArrayAlignment();
+        const U8 modulus = nodesSize % freeNodesAlignment;
```
LeStarch, line 402: I don't follow this modulus/alignment algorithm.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+                done = true;
+                break;
+            }
+            const auto& entryKey = this->m_nodes[child].m_entry.getKeyOrElement();
+            if (keyOrElement == entryKey) {
+                result = Success::SUCCESS;
+                node = child;
+                done = true;
+                break;
+            } else if (keyOrElement < entryKey) {
```
LeStarch, line 470: Review note for further confirmation, "LEFT" is defined as strictly less than.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+                     Direction& direction     //!< The direction (output)
+    ) const {
+        auto result = Success::FAILURE;
+        auto parent = Node::NONE;
+        auto child = this->m_root;
+        const auto capacity = this->getCapacity();
+        bool done = (capacity == 0);
+        for (FwSizeType i = 0; i < capacity; i++) {
+            if (child == Node::NONE) {
+                node = parent;
```
LeStarch, line 460: Why is this assignment necessary?  In the failure case, do we guarantee anything?

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+    //! at the root of the tree, and direction is ignored.
+    //!
+    //! It is not permissible for node to be NONE.
+    void insertNode(Index node,          //!< The node to insert
+                    Index parent,        //!< The new parent
+                    Direction direction  //!< The direction under the new parent
+    ) {
+        // We assume (1) that the tree is a red-black tree, (2) that parent is NONE or
+        // the child of parent in the direction `direction` is NONE, and (3) that
+        // both children of node are NONE.
```
LeStarch, line 535: Should we assert on assertion 2, assertion 3, here?

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+
+    //! This function performs a left or right rotation on the subtree whose
+    //! root is node. The following invariants must hold on entry to this
+    //! function, or an assertion failure will occur:
+    //!
+    //! 1. node must not be NONE.
+    //!
+    //! 2. The child of node in the direction opposite direction must not be
+    //!    NONE.
+    void rotateSubtree(Index node,          //!< The node index
```
LeStarch, line 1670: Here we do not assert on the preconditions:

1. node must not be none.
2. Child of node in opposite direction must not be none.

Should we therefore add:
```
FW_ASSERT(node != Node::NONE);
const auto newRoot = this->m_nodes[node].getChild(oppositeDirection);
FW_ASSERT(newRoot != Node::NONE);
```

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+            child = this->m_nodes[child].getChild(direction);
+        }
+        FW_ASSERT(done == true);
+        return node;
+    }
+
+    //! Get the parent direction for a node, i.e., the direction (left or
+    //! right) to follow from the parent of node to get to node. node must not be
+    //! NONE. The parent of node must not be NONE.
+    Direction getParentDirection(Index node  //!< The node index
```
LeStarch: Might this be better named `getDirectionFromParent`?

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+        }
+    }
+
+    //! This function removes a node of the tree with exactly one child. node
+    //! stores the node to remove. It must not be NONE. direction stores the
+    //! direction of the child.
+    void removeNodeWithOneChild(Index node,          //!< The node
+                                Direction direction  //!< The direction of the child
+    ) {
+        FW_ASSERT(this->m_nodes[node].m_color == Color::BLACK,
```
LeStarch, line 1618: We assert here that the color must be BLACK, however, this is not specified in the comment above.  We do not assert on the commented invariant.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+                  static_cast<FwAssertArgType>(this->m_nodes[node].m_color));
+        const auto parent = this->m_nodes[node].m_parent;
+        const auto child = this->m_nodes[node].getChild(direction);
+        FW_ASSERT(this->m_nodes[child].m_color == Color::RED,
```
LeStarch, line 1622: Another invariant not described but that makes sense.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+                    //                     |      |
+                    //                     V      V
+                    //             ------------------------
+                    //             |                      |
+                    //             |  black height i + 1  |
+                    //             |                      |
+                    //             ------------------------
+                    */
+                    this->removeBlackLeafNodeHelper2(parent, sibling, distantNephew, direction);
+                    // The subtree has this shape. The entire tree is a valid red-black tree.
```
LeStarch, line 1029: I'd expect a "done" and "break" if we are a valid sub tree per this comment.

LeStarch: After re-review, I believe that the tree is valid, therefore we should mark `done = true` and break at this line

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+                    //                       |      |
+                    //                       V      V
+                    //               ------------------------
+                    //               |                      |
+                    //               |  black height i + 1  |
+                    //               |                      |
+                    //               -----------------------
+                    */
+                    this->removeBlackLeafNodeHelper2(parent, sibling, distantNephew, direction);
+                    // The subtree has this shape. The entire tree is a valid red-black tree.
```
LeStarch, line 1149: Expect `done = true; break`

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+                    //                     |      |                         |      |
+                    //                     V      V                         V      V
+                    //             ------------------------         ------------------------
+                    //             |                      |         |                      |
+                    //             |    black height i    |         |    black height i    |
+                    //             |                      |         |                      |
+                    //             ------------------------         ------------------------
+                    */
+                }
+                done = true;
```
LeStarch: Don't we want a break here?

LeStarch: Technically, the break is at the end, but still with all the comments done && break seems clearer.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
LeStarch: Also, the MD file should say done = true.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+                //           |      |                    |     |
+                //           |      |                    |     |
+                //           V      V                    V     V
+                //   ------------------------   ------------------------
+                //   |                      |   |                      |
+                //   |  black height i + 1  |   |  black height i + 1  |
+                //   |                      |   |                      |
+                //   ------------------------   ------------------------
+                */
+                done = true;
```
LeStarch, line 1290: break?

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
```
LeStarch, line 1384: break

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
```diff
@@ -0,0 +1,1699 @@
...
+            } else if (this->getNodeColor(parent) == Color::RED) {
+                this->m_nodes[sibling].m_color = Color::RED;
```
LeStarch, line 1387: Missed the start diagram in this case.

LeStarch (changes requested):
There are several issues that must be addressed:
1. There are several cases in `removeBlackLeafNode` that result in a valid RBT but do not set done.
2. I prefer to set done and break in each step rather than setting done and deferring break to a further if
3.  Many commented invariants aren't check by FW_ASSERT but some are.  I'd feel more warm-fuzzies if we asserted every invariant as commented.  I assume this was for efficiency and would acquiesce. 

There are some minor problems:
1. The two helper function are hard.  Each time I saw them I had to re-read what they do to ensure it was expected because the name was in-descriptive.

Good new, I am convinced that this is a RBT implementation!  Fantastic!

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
LeStarch: Equality is not an issue, as the key would be in the tree already.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: OK. It's the same as the one below, with two different node colors. I thought it wasn't necessary to replicate the diagram.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: `node` is an output parameter. So this assignment provides function output.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: A node with one child must be black according to the red-black tree properties. The access `m_nodes[node]` implicitly asserts the invariant, because if `node == NONE` then an out-of-bounds access will occur, causing an assertion failure.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: > node must not be none.

This is asserted by the access `m_nodes[none]`.

> Child of node in opposite direction must not be none.

This is asserted by the access `m_nodes[newRoot]`.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: I don't follow this comment.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: Fixed.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: Fixed.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: Fixed.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: Fixed.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: Fixed.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: Fixed.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: Fixed.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: I added a comment stating why the node must be black.

bocchino:
@thomas-bc I replaced the png files with svg per your suggestion.

bocchino:
@LeStarch I closed all the comments except two, which require further discussion.

Fw/DataStructures/RedBlackTreeSetOrMapImpl.hpp
bocchino: After discussion, I'll add a comment to the assertion.

bocchino: I revised the comment to explain why the child must be red.

bocchino:
@LeStarch I have addressed all the comments.

LeStarch (dismissed):

LeStarch:
Nice work!

bocchino:
Thanks, Michael!

LeStarch (dismissed):

LeStarch (approved):

Issue #4033, opened 2025-08-15
FpySequencer U32 stack size


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**| FpySequencer |
---
## Feature Description

Make the maximum stack size U32 max instead of U16 max. This would be more in line with off the shelf alternatives, which support relatively large stacks. However, this would grow the size of many bytecode instructions by 2 bytes. 

## Rationale

65535 bytes is actually possible to hit if you're doing complex work down the stack frame with large types, whereas 2^32 bytes is really not likely.

Issue #3866, opened 2025-07-09
FpySequencer report which directive and line failed in tlm


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**|  FpySequencer|
---
## Feature Description

The sequencer should report which directive failed, and which line it was on, not just what the error code was. This is to differentiate between multiple diff dirs which could raise the same error.

PR #4065, opened 2025-08-22
FpySequencer U32 stack size + some error telemetry

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4033, #3866  |

---
## Change Description

* Switch everywhere that references a size on the stack or a stack offset to a `U32` instead of a `U16`
* Add telemetry on which line failed in the sequence, as well as what directive id it was
* Also add a cast to `FwEnumStoreType` to be more correct

## Rationale

Currently, the biggest you can configure the max stack size to be is 65 kilobytes (U16 max)

This allows the stack to be configurable up to a maximum size of 4 GB. Obviously 4 GB is huge, but it's likely better to have a not-very-restrictive max than a restrictive max. 65 KB is small enough that you could conceivably hit this under normal usage, in a very complex script with many large local variables and function calls.

For the telemetry, this is just some stuff that helps in debugging.


zimri-leisher:
@LeStarch @thomas-bc this is a small PR, hoping to make the review cycle quicker

Svc/FpySequencer/FpySequencerDirectives.fppi
```diff
@@ -35,15 +35,15 @@ struct StoreTlmValDirective {
     @ the tlm channel id to get the time of
     chanId: FwChanIdType
     @ the offset in the lvar array to store the tlm in
-    lvarOffset: U16
+    lvarOffset: U32
```
thomas-bc: Would it be useful to define a type alias representing the "size" type used across FpySequencer ? 

Just thinking about it because there's a lot of U32s and other types and if I wasn't looking at this changeset I wouldn't be able to tell at first glance which is related to which.

Just a thought - maybe from your perspective it makes more sense to keep them U32, which is perfectly fine.

Svc/FpySequencer/FpySequencerTelemetry.fppi
thomas-bc, line 1: It might be worth adding a sentence or two in the SDD with the explanation your gave in the PR description as to why U32 is chosen, and not U16 or even FwSizeType.

thomas-bc (commented):
Two small questions/recommendations, but this LGTM

Svc/FpySequencer/FpySequencerDirectives.fppi
zimri-leisher: hmm, I guess I could do that but I don't think that I would make it configurable. it's kinda baked into the bytecode

zimri-leisher:
@thomas-bc feedback responded to!

zimri-leisher:
@LeStarch @thomas-bc looks like this PR failed due to nondeterminism, can you please rerun?
and @thomas-bc if you're good, can you approve?

thomas-bc (approved):

zimri-leisher:
@LeStarch if you could merge now that changes are approved, that would be great

LeStarch (approved):

PR #4076, opened 2025-08-26
Make APID and FwPacketDescriptorType the same width

### LINKED ISSUE #4026: Make FwPacketDescriptorType and APID types share same width
| | |
|:---|:---|
|**_F´ Version_**| 4.0 |
|**_Affected Component_**|  |
---
## Feature Description

`FwPacketDescriptorType` is defined as U32 but `Fw::ComPacketType` is now defined https://github.com/nasa/fprime/blob/a070ee768aa00d6d433452410a766e3d04f2105c/Fw/Com/ComPacket.hpp#L20
and APID is a U16. We could consider having these 2 sizes match. 

Currently 
```
cmdBuf.serialize(Fw::ComPacketType::FW_PACKET_COMMAND);
```
will serialize as a U16, but all our codebase is serializing as FwPacketDescriptorType (U32)
```
cmdBuf.serialize(static_cast<FwPacketDescriptorType>(Fw::ComPacketType::FW_PACKET_COMMAND));
```

Maybe we don't care... but it sounds odd. To be discussed in CCB. This would ripple through to the GDS and such - but since this is a configurable type there's no reason it shouldn't just work out of the box (it should be a type that's pulled in dictionary).

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix #4026

Make FwPacketDescriptorType and APID type share the same width.

🚨🚨🚨 Default FwPacketDescriptorType changed to U16 🚨🚨🚨

With the recent updates to GDS, this should be transparent to the user.

## Future Work

We can't get rid of FwPacketDescriptorType yet because it still is relevant to the framework's inner working (packet are stamped with it). Long term, we may want to stop creating packets by stamping out the FwPacketDescriptorType in front. Either pass as a different argument, or maybe just use SpacePackets all around.


thomas-bc:
Need https://github.com/nasa/fprime-gds/pull/239 to pass integration tests

thomas-bc:
Released https://github.com/nasa/fprime-gds/releases/tag/v4.0.2a5 for this purpose

thomas-bc:
TODO: fix `fprime-seqgen` (not pulling `DictionaryParser`) and `fprime-examples`

LeStarch (approved):

Issue #3994, opened 2025-08-06
ComQ: Retry Last Failed


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Build a new component that implements `Svc.Framer` and handles retry.

The model would look similar to:

```
module Svc {
    active component ComRetry {
        import Svc.Framer
    }
}
```

dataIn should be a pass through to dataOut
dataReturnIn should store buffer and context
comStatusIn should:
   on success: call dataReturnOut(stored buffer, stored context)
   on failure: call dataOut(stored buffer, stored context)

valdaarhun:
Hi. I would love to work on this.

LeStarch:
This should be optional, based on config.

Edit: In the component design, the optional nature of this component is determined by inclusion/exclusion from the topology.

LeStarch:
I would add a config-retries field too.  If this is exceeded, then report the failure to com-queue.  

Also need to handle the case where comStatus comes in and no buffer is stored.  This would be a pass-through to comStatus out.

valdaarhun:
Hi @LeStarch . For clarification, the state machine in ComQueue will have to be shifted to the new component, right?

And if there's a component that requires a retry mechanism, then that component will have to be connected to the new one. Is my understanding correct?

LeStarch:
@valdaarhun no, no state machine needs to be shifted.  This new ComRetry component should:

1. Implement Svc.Framer (dataIn, dataOut, dataReturnIn, dataReturnOut, comStatusIn, ComStatusOut).
2. On dataIn, send the data directly to dataOut.  Note: it is a violation of protocol to receive dataIn while waiting for data to return.
2. On dataReturnIn store the buffer as a member variable
3. On comStatusIn:
    1. If status "GOOD", then send the stored buffer to dataReturnOut and then pass GOOD to comStatusOut
    2. If status "BAD", then send the stored buffer to dataOut.....this is the retry case
  
ComQueue as it is in mainline can remain untouched.

LeStarch:
Any topology needing retry shall put this component into the pipline right before comStub/radio.

valdaarhun:
> @valdaarhun no, no state machine needs to be shifted.  This new ComRetry component should:
> 
> 1. Implement Svc.Framer (dataIn, dataOut, dataReturnIn, dataReturnOut, comStatusIn, ComStatusOut).
[...]

PR #4081, opened 2025-08-27
Support resending last message on failure

valdaarhun:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3994 |

---
## Change Description

Update Svc::ComQueue to resend the last message on failure. Make this change config-driven.
## Rationale

Enable ComQueue to retry sending the last message when it receives `Fw::Success::FAILURE` instead of simply waiting till it receives `Fw::Success::SUCCESS`.
## Future Work

- Refine/rework implementation
- Add tests
- Modify docs

None

### DIFF
--- a/Svc/ComQueue/ComQueue.cpp
+++ b/Svc/ComQueue/ComQueue.cpp
@@ -33,12 +33,7 @@ ComQueue ::ComQueue(const char* const compName)
       m_state(WAITING),
       m_allocationId(static_cast<FwEnumStoreType>(-1)),
       m_allocator(nullptr),
-      m_allocation(nullptr)
-#if FW_COM_BUFFER_RETRY_ON_FAILURE
-      ,
-      m_hasRetried(true)
-#endif
-{
+      m_allocation(nullptr) {
     // Initialize throttles to "off"
     for (FwIndexType i = 0; i < TOTAL_PORT_COUNT; i++) {
         this->m_throttle[i] = false;
@@ -56,7 +51,8 @@ void ComQueue ::cleanup() {
 
 void ComQueue::configure(QueueConfigurationTable queueConfig,
                          FwEnumStoreType allocationId,
-                         Fw::MemAllocator& allocator) {
+                         Fw::MemAllocator& allocator,
+                         bool retryOnFailure) {
     FwIndexType currentPriorityIndex = 0;
     FwSizeType totalAllocation = 0;
 
@@ -130,6 +126,8 @@ void ComQueue::configure(QueueConfigurationTable queueConfig,
     // Safety check that all memory was used as expected
     FW_ASSERT(allocationOffset == totalAllocation, static_cast<FwAssertArgType>(allocationOffset),
               static_cast<FwAssertArgType>(totalAllocation));
+    
+    this->m_retryOnFailure = retryOnFailure;
 }
 // ----------------------------------------------------------------------
 // Handler implementations for user-defined typed input ports
@@ -159,21 +157,20 @@ void ComQueue::comStatusIn_handler(const FwIndexType portNum, Fw::Success& condi
         // On success, the queue should be processed. On failure, the component should still wait.
         case WAITING:
             if (condition.e == Fw::Success::SUCCESS) {
-#if FW_COM_BUFFER_RETRY_ON_FAILURE
-                this->m_hasRetried = false;
-#endif
                 this->m_state = READY;
-                this->processQueue();
+                // The last buffer will not be resent if retry is not configured.
+                if (this->m_tryBufferResend) {
+                    this->m_tryBufferResend = false;
+                    this->dataOut_out(0, this->m_outBuffer, this->m_context);
+                } else {
+                    this->processQueue();
+                }
                 // A message may or may not be sent. Thus, READY or WAITING are acceptable final states.
                 FW_ASSERT((this->m_state == WAITING || this->m_state == READY),
                           static_cast<FwAssertArgType>(this->m_state));
             } else {
-#if FW_COM_BUFFER_RETRY_ON_FAILURE
-                if (!this->m_hasRetried) {
-                    this->m_hasRetried = true;
-                    this->dataOut_out(0, this->m_outBuffer, this->m_context);
-                }
-#endif
+                // Buffer will never be resent if retry has not been configured.
+                this->m_tryBufferResend = this->m_retryOnFailure;
                 this->m_state = WAITING;
             }
             break;
@@ -277,11 +274,8 @@ void ComQueue::sendComBuffer(Fw::ComBuffer& comBuffer, FwIndexType queueIndex) {
     this->dataOut_out(0, outBuffer, context);
     // Set state to WAITING for the status to come back
     this->m_state = WAITING;
-
-#if FW_COM_BUFFER_RETRY_ON_FAILURE
     this->m_context = context;
-    this->m_outBuffer = buffer;
-#endif
+    this->m_outBuffer = outBuffer;
 }
 
 void ComQueue::sendBuffer(Fw::Buffer& buffer, FwIndexType queueIndex) {
@@ -299,11 +293,8 @@ void ComQueue::sendBuffer(Fw::Buffer& buffer, FwIndexType queueIndex) {
     this->dataOut_out(0, buffer, context);
     // Set state to WAITING for the status to come back
     this->m_state = WAITING;
-
-#if FW_COM_BUFFER_RETRY_ON_FAILURE
     this->m_context = context;
     this->m_outBuffer = buffer;
-#endif
 }
 
 void ComQueue::processQueue() {
--- a/Svc/ComQueue/ComQueue.cpp
+++ b/Svc/ComQueue/ComQueue.cpp
@@ -33,7 +33,12 @@ ComQueue ::ComQueue(const char* const compName)
       m_state(WAITING),
       m_allocationId(static_cast<FwEnumStoreType>(-1)),
       m_allocator(nullptr),
-      m_allocation(nullptr) {
+      m_allocation(nullptr)
+#if FW_COM_BUFFER_RETRY_ON_FAILURE
+      ,
+      m_hasRetried(true)
+#endif
+{
     // Initialize throttles to "off"
     for (FwIndexType i = 0; i < TOTAL_PORT_COUNT; i++) {
         this->m_throttle[i] = false;
@@ -154,12 +159,21 @@ void ComQueue::comStatusIn_handler(const FwIndexType portNum, Fw::Success& condi
         // On success, the queue should be processed. On failure, the component should still wait.
         case WAITING:
             if (condition.e == Fw::Success::SUCCESS) {
+#if FW_COM_BUFFER_RETRY_ON_FAILURE
+                this->m_hasRetried = false;
+#endif
                 this->m_state = READY;
                 this->processQueue();
                 // A message may or may not be sent. Thus, READY or WAITING are acceptable final states.
                 FW_ASSERT((this->m_state == WAITING || this->m_state == READY),
                           static_cast<FwAssertArgType>(this->m_state));
             } else {
+#if FW_COM_BUFFER_RETRY_ON_FAILURE
+                if (!this->m_hasRetried) {
+                    this->m_hasRetried = true;
+                    this->dataOut_out(0, this->m_outBuffer, this->m_context);
+                }
+#endif
                 this->m_state = WAITING;
             }
             break;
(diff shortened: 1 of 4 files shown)


valdaarhun:
Hi. I am not sure if my approach is right. I have added a config flag in `FpConfig.h` and have wrapped my changes in macros. The default behavior is for ComQueue to continue waiting until the next `Success` signal.

Svc/ComQueue/ComQueue.cpp
```diff
@@ -154,12 +159,21 @@ void ComQueue::comStatusIn_handler(const FwIndexType portNum, Fw::Success& condi
...
+#if FW_COM_BUFFER_RETRY_ON_FAILURE
+                this->m_hasRetried = false;
+#endif
                 this->m_state = READY;
                 this->processQueue();
                 // A message may or may not be sent. Thus, READY or WAITING are acceptable final states.
                 FW_ASSERT((this->m_state == WAITING || this->m_state == READY),
                           static_cast<FwAssertArgType>(this->m_state));
             } else {
+#if FW_COM_BUFFER_RETRY_ON_FAILURE
```
thomas-bc: Our coding standards discourage this practice of conditionally building code with the help of the preprocessor. An alternate solution would be to use a `.configure(bool does_retry)` method that can set a member variable into the component as to whether it should retry or not. Then add this to your conditionals.

valdaarhun: Thank you for the review. I'll check this out.

thomas-bc: Once this is done it would also be good to update the UTs to test that retries functions properly.

valdaarhun: Sure thing. I'll do that.

thomas-bc: Awesome. Thanks a lot for contributing! 🚀 
This is a much needed feature.

LeStarch: @valdaarhun we already have a configure method, so following the suggested approach I'd recommend you add a bool arg to it e.g.

```
void ComQueue::configure(..., bool retry_on_failure)
```

Svc/ComQueue/ComQueue.cpp
```diff
@@ -154,12 +159,21 @@ void ComQueue::comStatusIn_handler(const FwIndexType portNum, Fw::Success& condi
...
+                if (!this->m_hasRetried) {
```
LeStarch: We technically should only send a retry when "SUCCESS" is received.  When a failed message comes in, we should set a `m_need_retry` boolean here. 

If `m_need_retry` is set we should read your member variable and send there.  I'll leave that note there too.

Svc/ComQueue/ComQueue.cpp
```diff
@@ -154,12 +159,21 @@ void ComQueue::comStatusIn_handler(const FwIndexType portNum, Fw::Success& condi
         // On success, the queue should be processed. On failure, the component should still wait.
         case WAITING:
             if (condition.e == Fw::Success::SUCCESS) {
+#if FW_COM_BUFFER_RETRY_ON_FAILURE
+                this->m_hasRetried = false;
+#endif
                 this->m_state = READY;
                 this->processQueue();
```
LeStarch: This should likely now read:

```
if (m_need_retry) {
   this->dataOut_out(0, this->m_outBuffer, this->m_context);
} else {
 processQueue();
}

**Review by LeStarch — CHANGES_REQUESTED** [review]
I agree with @thomas-bc on the awesomeness of this change.  However, we need to move where sends happen.

Thanks!

File: Svc/ComQueue/ComQueue.cpp (hunk 2 above)
Comment (valdaarhun) [reply Svc/ComQueue/ComQueue.cpp]: Thank you for the review. I think I misunderstood how this is supposed to work. I'll fix it.

**Comment by valdaarhun** [comment]
Before testing my changes, I thought I would run the unit tests in the devel branch to see what it looks like.

I ran the following in the ComQueue directory:

```sh
fprime-util generate --ut
fprime-util check
```

But it fails on the first test with an ASAN error - `AddressSanitizer: stack-use-after-return`.

I am not sure if this is a known issue, or if there's an issue in my setup.

```
Output:
----------------------------------------------------------
[==========] Running 8 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 8 tests from Nominal
[ RUN      ] Nominal.Send
=================================================================
==30175==ERROR: AddressSanitizer: stack-use-after-return on address 0x7b36a8d03cf8 at pc 0x56cac320d807 bp 0x7ffefd298f50 sp 0x7ffefd298f40
READ of size 1 at 0x7b36a8d03cf8 thread T0
    #0 0x56cac320d806 in testing::AssertionResult testing::internal::CmpHelperEQ<unsigned char, unsigned char>(char const*, char const*, unsigned char const&, unsigned char const&) ~/fprime/googletest/googletest/include/gtest/gtest.h:1358
    #1 0x56cac320a13f in testing::AssertionResult testing::internal::EqHelper::Compare<unsigned char, unsigned char, (void*)0>(char const*, char const*, unsigned char const&, unsigned char const&) ~/fprime/googletest/googletest/include/gtest/gtest.h:1377
    #2 0x56cac31df022 in Svc::ComQueueTester::emitOneAndCheck(short, unsigned char*, unsigned long) ~/fprime/Svc/ComQueue/test/ut/ComQueueTester.cpp:72
    #3 0x56cac31e0194 in Svc::ComQueueTester::testQueueSend() ~/fprime/Svc/ComQueue/test/ut/ComQueueTester.cpp:88
...
Address 0x7b36a8d03cf8 is located in stack of thread T0 at offset 248 in frame
    #0 0x56cac32fb7f1 in Svc::ComQueue::processQueue() ~/fprime/Svc/ComQueue/ComQueue.cpp:285

  This frame has 3 object(s):
    [48, 72) 'temp' (line 328)
    [112, 184) 'buffer' (line 308)
    [224, 760) 'comBuffer' (line 304) <== Memory access at offset 248 is inside this variable
HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
      (longjmp and C++ exceptions *are* supported)
SUMMARY: AddressSanitizer: stack-use-after-return ~/fprime/googletest/googletest/include/gtest/gtest.h:1358 in testing::AssertionResult testing::internal::CmpHelperEQ<unsigned char, unsigned char>(char const*, char const*, unsigned char const&, unsigned char const&)
```

The issue seems to arise [here](https://github.com/nasa/fprime/blob/devel/Svc/ComQueue/test/ut/ComQueueTester.cpp#L72).

LeStarch:
@valdaarhun let's spin this off as a separate issue.  For now, just verify the tests pass.

LeStarch:
You can run 

```
fprime-util generate --ut --disable-sanitizers
```

to turn them off.

valdaarhun:
> You can run 
> 
> ```
> fprime-util generate --ut --disable-sanitizers
> ```
> 
> to turn them off.

Thank you. I'll focus on the tests for now, and will open an issue for the failure with sanitizers enabled.

valdaarhun:
While working on the tests I realized that my implementation is incorrect. Based on what I have understood, the following won't work because the assignment performs a shallow copy of the underlying data.

```
this->m_outBuffer = buffer; // or outBuffer
```

Once `Fw::Buffer`  goes out of scope, the underlying data pointed to by `m_outBuffer->m_bufferData`  will not exist.

I was thinking of saving the message in a U8 array instead and wrapping it in a `buffer`  when attempting a resend. What are your thoughts on this?

LeStarch:
First, great catch! There are two types of of data here: 

1. `Fw::Buffer` data.  This data is owned by an external manager, and isn't invalidated until this line:  https://github.com/nasa/fprime/blob/9871fd560832937172f5f96a5cc9c912f7fd785a/Svc/ComQueue/ComQueue.cpp#L206

2. `Fw::Com` this data is invalidated when this buffer goes out of scoper. https://github.com/nasa/fprime/blob/9871fd560832937172f5f96a5cc9c912f7fd785a/Svc/ComQueue/ComQueue.cpp#L304  This itself may be a bug...because this happens after the call.

I'd recommend making these changes:

1. Moving item 2 above to be a member variable
2. You can guarantee that item 1 happens **before** the comStatus return because it is `sync` and `comStatus` is async.  Therefore you could set state to track the returned buffer, and then call the buffer return linked about in the comStatus handler if and only if you don't need to retry.

this is a bit annoyingly complicated.

LeStarch:
@thomas-bc do we need comStatus anymore?  Now that we have DataWithContext, could the status be in the context?   

For now we could add status to context in addition.  Then only deallocate (line 206) if the status is good or if the component is "not retry".

LeStarch:
I confirmed with @thomas-bc that we should maintain comStatus for the time being.  That being said, we should also specify the order of which these ports should be called.

Since data return should be synchronous, and comStatus must be asynchronous, we should assume that data return always happens first.

@valdaarhun are you up to continue working on this, or would you like some assistance?  You've done a great job exposing issues and details, but I don't want to overwhelm you with managing all the details.

Let me know if you want some help.

valdaarhun:
> @valdaarhun are you up to continue working on this, or would you like some assistance? You've done a great job exposing issues and details, but I don't want to overwhelm you with managing all the details.
> 
> Let me know if you want some help.

Hi @LeStarch . Sorry for the late reply. Yes, I am still very interested in continuing this.

There are a few things that I am a little unclear on.

> 2\. You can guarantee that item 1 happens **before** the comStatus return because it is `sync` and `comStatus` is async.

> Since data return should be synchronous, and comStatus must be asynchronous, we should assume that data return always happens first.

I haven't really understood the points above. Is it already guaranteed that data return will happen before comStatus return? Or does that need to be handled separately? I think my confusion stems from my lack of understanding of the following 2 ports - `dataReturnIn` and `bufferReturnOut`.

In section 4 of the [documentation](https://github.com/nasa/fprime/blob/devel/Svc/ComQueue/docs/sdd.md), they are described as ports used for returning "ownership of buffers". I haven't really understood this part.

LeStarch:
@valdaarhun gotcha!  When you send out an (Fw.Buffer) buffer from com queue, you should not touch it anymore.  By sending the buffer, you've transferred ownership out to another component.  Ownership is returned to you when 'dataReturnIn' is called and you may work with the data.

For this PR you may assume 'dataReturnIn' is called before 'comStatusIn', always, as defined by the protocol between this and downstream components. 

Therefore,  the following steps should be taken:

In 'dataReturnIn_handler' simply save the returned buffer in a member variable for later.

In 'comStatusIn_handler' do the following:
1. If status is bad, and the stored buffet is valid then resend the stored buffer.  
2. If status is good and stored buffer is valid,  and original data was a few buffer, then deallocate by sending stored buffer to 'bufferReturnOut'
3. In all cases invalidate the stored buffer (set to empty Fw.Buffer)

Finally,  move the dequeue of the com buffer to be a member variable so the data remains in scope.

LeStarch:
I know this is complicated.   Data ownership usually is.  Please let me know if I can offer more explanation

valdaarhun:
> @valdaarhun gotcha! When you send out an (Fw.Buffer) buffer from com queue, you should not touch it anymore. By sending the buffer, you've transferred ownership out to another component. Ownership is returned to you when 'dataReturnIn' is called and you may work with the data.
> 
> For this PR you may assume 'dataReturnIn' is called before 'comStatusIn', always, as defined by the protocol between this and downstream components.

Thank you for the explanation. This makes sense now.

> In 'comStatusIn_handler' do the following:
>     1. If status is bad, and the stored buffet is valid then resend the stored buffer.
>     2. If status is good and stored buffer is valid,  and original data was a few buffer, then deallocate by sending stored buffer to 'bufferReturnOut'
>     3. In all cases invalidate the stored buffer (set to empty Fw.Buffer)

Understood. I think I am in a position to make some progress now. I'll let you know in case I get stuck again.

LeStarch:
@valdaarhun I had an insight this morning.  This would work better if we just built a new "retry" component.  It would import the Framer interface and implement only the retry logic.

```
active component ComRetry {
        import Framer
}
```

Here are the main points:

1. On `dataIn` send to `dataOut` (pass-through)
2. On `dataReturnIn` store buffer and context.
3. On `comStatusIn`:
   1. If good, send stored buffer and context to `dataReturnOut`, then call `comStatusOut`
   2. If bad, retry by sending stored buffer and context to `dataOut`

This has the major advantage that we can put the same Retry component anywhere in the pipeline.  We should therefore update the issue, and close this PR.

Would you like to build this new component?

LeStarch:
@valdaarhun Also, I apologize for all the back and forth.  I was building another com component yesterday and thought "I need the same retry logic here as in comQueu" then I realized it would stand better as a separate component.

We do really appreciate your work on this issue.  It has exposed many insights into this process that are refining and making the code better.  We are learning on this end too.

valdaarhun:
Hi @LeStarch . I agree. I think this approach would be better especially if it can be made extensible so other components can make use of it too.

I would love to work on building this new component.

LeStarch:
@valdaarhun thanks for your understanding.  I will close this PR.  Please feel free to make a PR with this new component.  We can then review together!

PR #4083, opened 2025-08-27
DataProduct Catalog Runtime Insertion

Willmac16:
| | |
|:---|:---|
|**_Generative AI was used in this contribution (y/n)_**|  n |

---
## Change Description

Modifies DpCatalog to allow for runtime insertion of data products onto the queue through the DpWriter notifier port

## Rationale

Long running missions with the need for automated file downlink would benefit from simultaneous catalog transmission and construction.

## Progress towards Draft Completion
- [x] Refactor file addition into separate function
- [x] Call into `processFile` from addToCatalog port
- [x] Allow adding to catalog while Transmiting
- [x] Add UT coverage for addToCatalog port
- [x] Fix running the 2 recently enabled UTs back to back
- [x] Fix/Remove that final `DISABLED` unit test
- [x] Round out coverage

## Nice to Haves
- [x] Clean all artifacts run to run
- [x] Catalog Navigation w/o the stack
  * would make auto-balancing easier
  * reduces memory req
  * probably doable w/ a current node + traversal direction (up/down) since I've added the parent link
- [ ] Self balancing (prob AVL or Red-Black)
  * Seems prudent given how carved up the tree will be with simultaneous additions and removals
- [ ] Random Sequence order validation @ the addToCat -> fileOut level

N/A


Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -779,6 +799,13 @@ void DpCatalog ::pingIn_handler(FwIndexType portNum, U32 key) {
     this->pingOut_out(0, key);
 }
 
+void DpCatalog ::addToCat_handler(FwIndexType portNum,
+                                  const Fw::StringBase& fileName,
+                                  FwDpPriorityType priority,
+                                  FwSizeType size) {
+    // TODO
+}
```
Willmac16, line 1054: Since nodes are only added to leaves (& no rebalancing), I should be able to add a node without needing to update any of the `m_current` pointers.

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -534,6 +455,134 @@ Fw::CmdResponse DpCatalog::fillBinaryTree() {
...
+            // Compare both strings up to location of final slash
+            // memsafe since both are fixed width strings
+            // and loc is before the fixed width
+            if (strncmp(dir_string, fullFile.toChar(), static_cast<FwSizeType>(loc)) == 0) {
+                break;
+            }
+        }
+
+        if (dir == this->m_numDirectories) {
+            this->log_WARNING_HI_DirectoryOpenError(fullFile, Os::FileSystem::Status::OTHER_ERROR);
```
timcanham: Create a new event specifically for this, perhaps something like "Unable to add file {}; directory not managed"

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -534,6 +455,134 @@ Fw::CmdResponse DpCatalog::fillBinaryTree() {
...
+    if (not insertedOk) {
+        this->log_WARNING_HI_DpInsertError(entry.record);
+        // clean up and return
+        this->resetBinaryTree();
+        this->resetStateFileData();
+        return -1;
+    }
+
+    if (entry.record.get_state() == Fw::DpState::UNTRANSMITTED) {
+        m_pendingFiles++;
```
timcanham: minor: use `this->` to be consistent with existing code.

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -779,6 +834,47 @@ void DpCatalog ::pingIn_handler(FwIndexType portNum, U32 key) {
...
+    }
+
+    // check that initialization got memory
+    if (0 == this->m_numDpSlots) {
+        this->log_WARNING_HI_NoDpMemory();
+        return;
+    }
+
+    // make sure a downlink is not in progress
+    if (this->m_xmitInProgress) {
```
timcanham: I think we should allow it to be added if the xmit is in progress. The downloads can potentially cover many hourse with a slow link.

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -779,6 +834,47 @@ void DpCatalog ::pingIn_handler(FwIndexType portNum, U32 key) {
...
+        this->log_WARNING_HI_NotInitialized();
+        return;
+    }
+
+    // check that initialization got memory
+    if (0 == this->m_numDpSlots) {
+        this->log_WARNING_HI_NoDpMemory();
+        return;
+    }
+
```
timcanham, line 1017: We should add a check to see if the catalog has already been loaded. If it hasn't, just drop this call, since the new file will be noticed when the catalog is loaded.

timcanham (changes requested):

timcanham:
Looks good so far!

Svc/DpCatalog/DpCatalog.cpp
Willmac16: Good call. My copy paste of all the guards from `doCatalogBuild` would defeat the goal of these changes

Willmac16:
Do you have a recommendation for a cmake way to clean up the test artifacts consistently regardless of if the test passes?

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -534,7 +446,148 @@ Fw::CmdResponse DpCatalog::fillBinaryTree() {
...
+    // if all is well, don't need the file any more
+    dpFile.close();
+
+    // give buffer to container instance
+    container.setBuffer(hdrBuff);
+
+    // reset header deserialization in the container
+    Fw::SerializeStatus desStat = container.deserializeHeader();
+    if (desStat != Fw::FW_SERIALIZE_OK) {
+        this->log_WARNING_HI_FileHdrDesError(fullFile, desStat);
```
zimri-leisher, line 531: return 0?

.github/actions/spelling/expect.txt
```diff
@@ -631,6 +631,7 @@ strnlen
 subgrouping
 subhist
 subhistory
+subsecond
 subseconds
```
Willmac16:
```suggestion
```

Willmac16: Bc subsecond globs any instances of subseconds

Fw/Types/StringUtils.cpp
```diff
@@ -73,3 +73,42 @@ FwSignedSizeType Fw::StringUtils::substring_find(const CHAR* source_string,
...
+FwSignedSizeType Fw::StringUtils::substring_find_last(const CHAR* source_string,
+                                                      FwSizeType source_size,
+                                                      const CHAR* sub_string,
+                                                      FwSizeType sub_size) {
+    FW_ASSERT(source_string != nullptr);
+    FW_ASSERT(sub_string != nullptr);
+
+    // zero size sub-strings should always match
+    if ((source_size > 0) && (0 == sub_size)) {
+        return 0;
```
LeStarch: Shouldn't this return `source_size - 1` since this is the last match?

Fw/Types/StringUtils.cpp
```diff
@@ -73,3 +73,42 @@ FwSignedSizeType Fw::StringUtils::substring_find(const CHAR* source_string,
...
+    }
+
+    // Cannot find a substring larger than the source
+    if (source_size < sub_size) {
+        return -1;
+    }
+    // Confirm that the output type can hold the range of valid results
+    FW_ASSERT(static_cast<FwSignedSizeType>(source_size - sub_size) <= std::numeric_limits<FwSignedSizeType>::max());
```
LeStarch: I think this still has a roll-over problem because you cast to `FwSignedSizeType` thus the result by definition will always be less than max.  I recommend:

```
FW_ASSERT(
    (( source_size < std::numeric_limits<FwSignedSizeType>::max()) &&
    ((std::numeric_limits<FwSignedSizeType>::max()) - source_size )> sub_size)
)
```

Fw/Types/StringUtils.cpp
```diff
@@ -73,3 +73,42 @@ FwSignedSizeType Fw::StringUtils::substring_find(const CHAR* source_string,
...
+
+    // Loop from source_size - sub_size to zero (inclusive)
+    for (FwSizeType source_index = (source_size - sub_size + 1); source_index-- > 0;) {
```
LeStarch: Should make this three clauses.  This is difficult to read.

Fw/Types/test/ut/TypesTest.cpp
```diff
@@ -1409,6 +1415,72 @@ TEST(OffNominal, sub_string_substring_zero_size) {
...
+TEST(OffNominal, sub_string_last_source_zero_size) {
+    const char* source_string = "";
+    const char* sub_string_last = "234";
+    ASSERT_EQ(Fw::StringUtils::substring_find_last(source_string, 0, sub_string_last, 3), -1);
+}
+
+TEST(OffNominal, sub_string_last_substring_zero_size) {
+    const char* source_string = "abc123";
+    const char* sub_string_last = "";
+    ASSERT_EQ(Fw::StringUtils::substring_find_last(source_string, 6, sub_string_last, 0), 0);
```
LeStarch: I still contend this should probably be 5 but perhaps it doesn't matter as long as it is well defined.

Svc/DpCatalog/DpCatalog.fpp
```diff
@@ -105,15 +109,15 @@ module Svc {
 
     @ Processing directory
     event ProcessingDirectory (
-                            directory: string size 80 @< The directory 
+                            directory: string size 80 @< The directory
```
LeStarch: At some point these file strings should switch to: `FileNameStringSize` instead of `80`

Svc/DpCatalog/DpCatalog.hpp
```diff
@@ -113,6 +125,40 @@ class DpCatalog final : public DpCatalogComponentBase {
...
+            if (left.record.get_priority() == right.record.get_priority()) {
+                // check time. Older is higher priority
+                if (left.record.get_tSec() == right.record.get_tSec()) {
+                    // check subsecond time. Older is higher priority
+                    if (left.record.get_tSub() == right.record.get_tSub()) {
+                        // check ID. Lower is higher priority
+                        if (left.record.get_id() == right.record.get_id()) {
+                            return 0;
+                        } else {  // if ids are not equal. smaller is higher priority
+                            return left.record.get_id() < right.record.get_id() ? -1 : 1;
```
LeStarch: perfer () around the clauses to ternary.

Svc/DpCatalog/DpCatalog.hpp
```diff
@@ -113,6 +125,40 @@ class DpCatalog final : public DpCatalogComponentBase {
         friend class DpCatalogTester;
         FwIndexType dir;  //!< index to m_directories entry that has directory name where DP exists
         DpRecord record;  //!< data product metadata
+
+        /// @brief insert an entry into the sorted list; if it exists, update the metadata
+        /// @param left an entry to compare
+        /// @param right other entry to compare
+        /// @return -1 if left is higher priority, 0 if equal, and 1 if right is higher priority
+        static int CompareEntries(const DpStateEntry& left, const DpStateEntry& right) {
```
LeStarch: I don't follow why these functions are inlined into the HPP, is there a reason?  Should we move to CPP?

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -26,10 +26,8 @@ DpCatalog ::DpCatalog(const char* const compName)
       m_dpTree(nullptr),
```
LeStarch, line 27: We should move to the RBT implementation rather than this inline one.  Eventually.

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -131,12 +125,13 @@ void DpCatalog::resetBinaryTree() {
             this->m_freeListHead[slot - 1].left = &this->m_freeListHead[slot];
         }
     }
-    // set the foot of the free list
-    this->m_freeListFoot = &this->m_freeListHead[this->m_numDpSlots - 1];
+    // set the foot of the free list (one past what we can use)
+    this->m_freeListFoot = &this->m_freeListHead[this->m_numDpSlots];
```
LeStarch: Either we are indexing into an array past it's length or we are wasting some space.  Either way this isn't great.

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -434,93 +420,19 @@ Fw::CmdResponse DpCatalog::fillBinaryTree() {
...
-            this->getFileState(entry);
-
-            // insert entry into sorted list. if can't insert, quit
-            bool insertedOk = this->insertEntry(entry);
-            if (not insertedOk) {
-                this->log_WARNING_HI_DpInsertError(entry.record);
-                // clean up and return
-                this->resetBinaryTree();
-                this->resetStateFileData();
+            int ret = processFile(fullFile, dir);
```
LeStarch, line 422: I appreciate this.

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -534,7 +446,148 @@ Fw::CmdResponse DpCatalog::fillBinaryTree() {
...
+
+    // Working buffer for DP headers
+    U8 dpBuff[Fw::DpContainer::MIN_PACKET_SIZE];  // Header buffer
+    Fw::Buffer hdrBuff(dpBuff, sizeof(dpBuff));   // buffer for container header decoding
+    Fw::DpContainer container;                    // container object for extracting header fields
+
+    this->log_ACTIVITY_LO_ProcessingFile(fullFile);
+
+    // Check if file is in one of our directories
+    if (dir >= DP_MAX_DIRECTORIES) {
```
LeStarch: This will loop event when dir==DP_MAX_DIRECTORIES

LeStarch: Which is intended, as this is the case when we need to detect a directory.

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -534,7 +446,148 @@ Fw::CmdResponse DpCatalog::fillBinaryTree() {
...
+        // Grab the directory string (up until the final slash)
+        // Could be found directly w/ a dirname func or regex
+        FwSignedSizeType loc = Fw::StringUtils::substring_find_last(
```
LeStarch: Break this into a helper function.

LeStarch: Specifically the directory determination chunk.  Just so it is clear: when directory is out of bounds, we run a search function.

LeStarch (changes requested):
I reviewed as much as I am able.  @timcanham I am not familiar enough with this component to be really confident in the changes, so we will need your review too.

Fw/Types/StringUtils.cpp
Willmac16: That does seem like a problem. I'll tweak both `substring_find` and `substring_find_last`

Willmac16: I think technically this new check might assert in certain cases that could still be represented. The only place the signed version is used is at return time, and even there, `source_size` doesn't need to be represented, only `source_size - sub_size`. Since we know that `source_size - sub_size >= 0` I propose doing the old comparison, but in the unsigned range of `FwSizeType`

Fw/Types/StringUtils.cpp
Willmac16: The rationale for the ugly form is to index with the same unsigned type as the string lengths (so strings longer than `FwSignedSizeType` could be used), while stopping after evaluating zero.

Assuming we stick with the `FwSizeType` index, this is the cleanest for loop form that doesn't require separate handling of zero & doesn't infinitely iterate.

Svc/DpCatalog/DpCatalog.hpp
Willmac16: LMK if the parentheses around the condition isn't all you wanted. I believe the operator precedence means the old code is equivalent, but I get that this is more clear.

Svc/DpCatalog/DpCatalog.hpp
Willmac16: There was no good reason. Moving them to cpp

Svc/DpCatalog/DpCatalog.cpp
Willmac16: Its now in a helper function & processFile is always called with a valid directory index

Svc/DpCatalog/DpCatalog.cpp
Willmac16: The existing implementation doesn't actually use the foot. I had assumed that in a worst case the head might == the foot, but one more look revealed that the head will point to nullptr, so I'm removing the foot

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -81,23 +82,16 @@ void DpCatalog::configure(Fw::FileNameString directories[DP_MAX_DIRECTORIES],
...
-    // end of the free list.
-    //
-    // 4) Place the state file data in memory after the binary tree traverse
-    // stack by indexing the traverse stack to one element past the end of
-    // the traverse tree.
+    // 3) Place the state file data in memory after the binary free list
+    // by indexing the free list to one element past the end of
+    // the free list.
 
     if ((this->m_memSize >= sizeof(DpBtreeNode)) and (this->m_memPtr != nullptr)) {
```
timcanham: I think this is a mistake that goes back to the original - it should have at minimum one slot, not just one `DpTreeNode`, since we also need room for at least one `DpStateFileEntry` to make one slot.

Svc/DpCatalog/DpCatalog.cpp
timcanham: @Willmac16 Where is this checked an dropped?

timcanham (dismissed):
The usage pattern seems correct to me. A few small actions to take.

Fw/Types/StringUtils.cpp
bocchino: After discussion with @LeStarch, we concurred that we should revise this function and also `substring_find` to return a status value (found or not found) and return a separate unsigned index value (via an out parameter). However, we will defer this change to a separate issue, since it requires a change to `substring_find`, which is outside the scope of this PR.

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -618,27 +702,144 @@ bool DpCatalog::allocateNode(DpBtreeNode*& newNode, const DpStateEntry& newEntry
...
+        // bounded while loop (in case we're linked onto the free list somehow)
+        for (FwSizeType record = 0; record < this->m_numDpSlots && rightmostNode->right != nullptr; record++) {
+            rightmostNode = rightmostNode->right;
+
+            // I really hope these never fire
+            FW_ASSERT(rightmostNode != this->m_freeListHead);
+            FW_ASSERT(rightmostNode != this->m_freeListFoot);
+        }
+
+        // We can stich its left branch onto its parent in its place
```
bocchino:
```suggestion
        // We can stitch its left branch onto its parent in its place
```

Fw/Types/StringUtils.cpp
bocchino: I have an alternative suggestion: Could we write the function body so it produces `FwSizeType` and a status, and then at the end pack the result into `FwSignedSizeType`, using a checked cast? I think the code would be cleaner, and it would be closer to what we eventually want.

bocchino: > Assuming we stick with the FwSizeType index, this is the cleanest for loop form that doesn't require separate handling of zero & doesn't infinitely iterate.

In my opinion, when iterating backwards over a loop with an unsigned index, it's cleaner to iterate forward and subtract:
```
for i in [0, size)
  do_something(size - i - 1)
```

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -534,7 +444,188 @@ Fw::CmdResponse DpCatalog::fillBinaryTree() {
...
+        return DP_MAX_DIRECTORIES;
+    }
+
+    for (FwSizeType dir = 0; dir < this->m_numDirectories; dir++) {
+        const char* const dir_string = this->m_directories[dir].toChar();
+
+        // Compare both strings up to location of final slash
+        // memory safe since both are fixed width strings
+        // and loc is before the fixed width
+        if (strncmp(dir_string, fullFile.toChar(), static_cast<FwSizeType>(loc)) == 0) {
```
bocchino: We should use `StringUtils` instead of `strncmp` here if possible. If it's not possible, does `StringUtils` need to be updated?

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -534,7 +444,188 @@ Fw::CmdResponse DpCatalog::fillBinaryTree() {
...
+        this->m_currentNode = addedEntry;
+    }
+
+    return 1;
+}
+
+// ----------------------------------------------------------------------
+// DpStateEntry Comparison Ops
+// ----------------------------------------------------------------------
+int DpCatalog::DpStateEntry::compareEntries(const DpStateEntry& left, const DpStateEntry& right) {
```
bocchino, line 593: I think the comparison code could be simplified. In general, if we are comparing `(a, b, c)` to `(d, e, f)` could we do the following:
```
result = {}
if (a < d) result = -1
else if (a > d) result = 1
else if (b < e) result = -1
...
else result = 0
return result
```
I think this approach would avoid the staircase of `if` statements.

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -618,27 +702,144 @@ bool DpCatalog::allocateNode(DpBtreeNode*& newNode, const DpStateEntry& newEntry
...
-void DpCatalog::deleteEntry(DpStateEntry& entry) {}
+void DpCatalog::deallocateNode(DpBtreeNode* node) {
+    DpBtreeNode* parent = node->parent;
+
+    // since nodes are deallocated after xmit, left should be gone
+    // However, left node could be added during xmit
+    if (node->left != nullptr) {
+        // To try to only deallocate nodes at fileDone,
+        // finished nodes w/ children are trimmed out
+        // Simple shifts only work with branches that don't branch
```
bocchino: I don't understand this comment. Can it be clarified? What are "simple shifts"? What is being shifted where? What are "branches that don't branch"? That sounds like a riddle.

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -618,27 +702,144 @@ bool DpCatalog::allocateNode(DpBtreeNode*& newNode, const DpStateEntry& newEntry
...
+
+        // Since we aren't limited to adding just 1 node during file transfer
+        // left might not be a leaf
+        // Instead, find the lowest priority node on the left branch
```
bocchino:
```suggestion
        // Instead, find the highest priority node on the left branch
```
I think we want the rightmost node under the left child. I'm assuming priorities go up from left to right.

bocchino (commented):
Looks good! Please see my comments. Please run code coverage analysis to make sure the unit tests are covering the newly added code.

Svc/DpCatalog/DpCatalog.cpp
Willmac16: substring_find can cover that behavior. With a few lines in a wrapper function we could make something with a more similar api to strncmp, but the current is plenty capable (I just missed that I could use it). or a new function could be made if lexical comparison (not just equality evaluation) is needed.

Svc/DpCatalog/DpCatalog.cpp
Willmac16: It's a bit convoluted. I believe my original comment is actually correct.
Here is my understanding of the priority system:

For all compared fields (i.e. priority, tSec, tSub, id), a lower numerical value is treated as higher priority. To keep with this scheme, my compareEntries function returns a -1 if the first entry passed in is less than the second entry (meaning it is numerically lower and thus higher priority). My comparison operators return numerical rather than priority comparisons (e.g. an entry filled with all zeros is "less than" a node filled with all ones even though it is higher priority). The insertion scheme places higher priority nodes (numerically lesser) on the left branch.

This step is looking to replace the just transmitted (and shortly deallocated) node w/ a node of immediately neighboring priority. Since the left (higher priority) side can only be added to by the addToCat port, there should be fewer nodes to traverse (than the lower priority half of the catalog) making the lowest priority child of the left node the best candidate.

Svc/DpCatalog/DpCatalog.cpp
Willmac16: I'm pretty sure I wrote this comment before I had a solution for deallocating a node w/ more than a single, linear branch on its left side. The purpose of this comment was to point out that I still had a problem to resolve, but a conversation w/ a colleague got me a solution. I'm going to cut this comment and bolster the comment block that immediately follows this. To answer the "riddle":

My goal (that the current code achieves) was to immediately deallocate a node as soon as its file completed transmission; rather than setting a flag and later traversing to clean up. A "simple shift" in my parlance is replacing a node w/ its right descendant by removing the deallocated node & pointing the parent node at the right child. This does not work for a non-linear left branch (aka. a branching left branch) for which the deallocated node's immediate left child has more than one child node.

Svc/DpCatalog/DpCatalog.cpp
Willmac16: I believe I added a check at one point for the dpTree pointer being non-null; however, I later removed that check since nodes could be added after tree completion (and pruning) with the remainActive arg. I'm adding a state file initialization check, since that is the primary thing a runtime added file needs from the catalog build.

Willmac16: I am now using a catalogBuilt flag since you can't guarantee the state file that is loaded will have any items.

Willmac16:
I am finally ready for re-review!

kmk142789 (approved):

LeStarch (approved):

PR #4096, opened 2025-09-02
Add FPP state machine How-To

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

A How-To guide for FPP State Machines!

### DIFF
--- a/docs/how-to/define-state-machines.md
+++ b/docs/how-to/define-state-machines.md
@@ -21,7 +21,7 @@ Before starting, you should have:
 State machines are useful when your component has distinct **modes** or **operational states** with different behavior in specific states and rules for the transitions between states. Examples include:
 
 * A radio with `OFF`, `IDLE`, and `TRANSMITTING` states.
-* A sensor with a series of start-up states
+* A sensor with a series of start-up states.
 
 By modeling these as a state machine, you make system behavior explicit, verifiable, and easier to test.
 
--- a/docs/how-to/define-state-machines.md
+++ b/docs/how-to/define-state-machines.md
@@ -1,4 +1,3 @@
-
 # How-To: Define State Machines in F Prime
 
 This guide shows how to define and use state machines in F Prime using the F Prime Modeling Language (FPP). State machines help capture component behavior by modeling modes (states) and transitions explicitly, making complex logic easier to implement, test, and maintain. FPP provides autocoding capabilities to allow users to quickly implement state-defined behavior.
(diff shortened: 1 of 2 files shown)


docs/how-to/define-state-machines.md
```diff
@@ -0,0 +1,427 @@
...
+
+```
+module MpuImu {
+    @ Define ImuStateMachine State Machine
+    state machine ImuStateMachine {
+        @ Initial state: reset the device
+        initial enter RESET
+
+        @ Reset the Imu
+        state RESET {
```
bocchino, line 108: You may want to omit the braces `{}` here and add them later, when they are needed. Otherwise it may create the impression that braces are required in all cases, even when they are empty.

docs/how-to/define-state-machines.md
```diff
@@ -0,0 +1,427 @@
...
+* A working build of F Prime on your system (`fprime-util` runs successfully).
+
+---
+
+## When to Use State Machines
+
+State machines are useful when your component has distinct **modes** or **operational states** with different behavior in specific states and rules for the transitions between states. Examples include:
+
+* A radio with `OFF`, `IDLE`, and `TRANSMITTING` states.
+* A sensor with a series of start-up states
```
bocchino:
```suggestion
* A sensor with a series of start-up states.
```

docs/how-to/define-state-machines.md
```diff
@@ -0,0 +1,427 @@
...
+```
+
+This state machine is implemented in the [MpuImu](TODO) component. This component drives the "next" signal from a rate group handler.
+
+> [!NOTE]
+> The [MpuImu](TODO) component implements a further `reconfigure` signal allowing the machine to return to reconfigure state.
+
+## Designing a State Machine in FPP
+
+We will model the Imu state machine in FPP.  Typically users would define state machines in a new file.  We should use `ImuStateMachine.fpp` defined in the same folder (i.e. module) as the `ImuManager` this will support.  Define the basic module and state machine with:
```
bocchino: >Typically users would define state machines in a new file.

This statement seems a bit strong. In my opinion:

* If the state machine is part of the `ImuManager` component implementation, the standard practice is to define it inside the component (possibly via inclusion from an `.fppi` file).
* If it's a more general state machine for use by multiple components, then it would normally go in its own directory.

Using a separate `.fpp` file in the same directory as the component is possible but doesn't seem standard. Maybe at least mention the other options.

docs/how-to/define-state-machines.md
```diff
@@ -0,0 +1,427 @@
...
+That should complete the definition of our state machine. However, we still have yet to bind it to our `ImuManager` component.
+
+## Instantiating the State Machine
+
+Every state machine can be used multiple times. To define a single instance attached to the `ImuManager` component, we use the following in our component definition:
+
+```
+queued component ImuManager {
+    @ Use the ImuStateMachine
+    state machine instances imuStateMachine: ImuStateMachine
```
garthwatney: Should be `instance `not `instances`

kevin-f-ortega (approved):

PR #4113, opened 2025-09-04
Add Framing subtopologies and use them within Com subtopologies

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Adds in `ComCcsds/ComFprime.FramingSubtopology` that does not include `ComStub`. 

## Rationale

So users can use the Comms stack with another Com Interface (e.g. radio)

Copilot autocomplete


Svc/Subtopologies/ComCcsds/ComCcsds.fpp
```diff
@@ -90,38 +90,49 @@ module ComCcsds {
         """
     }
 
-    instance fprimeRouter: Svc.FprimeRouter base id ComCcsdsConfig.BASE_ID + 0x03000 \
-    
-    instance comStub: Svc.ComStub base id ComCcsdsConfig.BASE_ID + 0x04000 \
+    instance fprimeRouter: Svc.FprimeRouter base id ComCcsdsConfig.BASE_ID + 0x03000
 
-    instance tcDeframer: Svc.Ccsds.TcDeframer base id ComCcsdsConfig.BASE_ID + 0x05000 \
+    instance tcDeframer: Svc.Ccsds.TcDeframer base id ComCcsdsConfig.BASE_ID + 0x04000
```
LeStarch, line 95: Why wasn't this called `defamer` for the same reason as framer below.

LeStarch: Frame accumulator. Gotcha.

LeStarch (dismissed):
I saw changes in the fprime subtopo to buffer connections but not in CCSDS.  Why?

Svc/Subtopologies/ComCcsds/ComCcsds.fpp
thomas-bc: Yes exactly, the FrameAccumulator is the component that needs to be hooked into from user main topologies. So didn't need to standardize on names for deframer, and would rather not.

LeStarch (approved):

Issue #3993, opened 2025-08-06
TCP Reconnections


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

Automatic reconnection in TCP should likely be moved to a dedicate thread with dedicated reconnect behavior.  `recv` and `send` should produce an error until reconnections are established.

Currently, `send` and `recv` can both trigger re-connections.  However, these calls work on different threads.  This proves a bit of a hassle to sort out.  First, there is a race condition between threads as to which is causing re-connections.  Second if the other thread is reconnecting, the current thread can error-out.  It probably should wait with a timeout.

PR #4136, opened 2025-09-10
Socket automatic reconnections in dedicated thread

m-aleem:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/3993 |
|**_Has Unit Tests (y/n)_**| Y - Existing |
|**_Documentation Included (y/n)_**| Y - Existing (updated) |

---
## Rationale

Send and recv for a socket can introduce a race condition because they both use their own thread (send=caller thread, recv = component defined thread) to reconnect. Having a dedicated reconnect thread avoid this.

Existing UTs will exercise the new thread logic

Debugging UT failures


Drv/Ip/SocketComponentHelper.cpp
```diff
@@ -221,4 +238,127 @@ void SocketComponentHelper::readTask(void* pointer) {
...
+            else if (status == SOCK_SUCCESS) {
+                Os::ScopeLock scopedLock(this->m_reconnectLock);
+                this->m_reconnectState = ReconnectState::NOT_RECONNECTING;
+            }
+            // Reopen Case 3: Keep trying to reconnect - NO reconnect
+            // state change
+            else {
+                Fw::Logger::log("[WARNING] Failed to open port with status %d and errno %d\n", status, errno);
+                // (void)Os::Task::delay(SOCKET_RETRY_INTERVAL);
+                m_reconnectTask._delay(SOCKET_RETRY_INTERVAL);  // FIXME review
```
m-aleem: @LeStarch In the current implementation of SocketComponentHelper, there is use of `(void)Os::Task::delay(SOCKET_RETRY_INTERVAL);` to delay before a retry attempt. In this (PR) new implementation we now have 3 interacting 'threads' and places where a wait may be done

1) receiveThread: can request a reconnect and then waits
2) sendThread: can request a reconnect and then waits
3) reconnectThread: checks for reconnect requests, briefly waiting `SOCKET_RETRY_INTERVAL` between attempts when in a 'RECONNECT_REQUESTED' state or `m_reconnectCheckInterval` between checks to see if there is a new request while in the NOT_RECONNECTING state

I initially used `Os::Task::delay` in all the wait calls, e.g. 

1) Called by reconnectThread: https://github.com/nasa/fprime/pull/4136/files#diff-67a68b4daab941bc994e0bbd5b961fc72933e623a821915d2d598b01923968e6R299
2) Called by reconnectThread: https://github.com/nasa/fprime/pull/4136/files#diff-67a68b4daab941bc994e0bbd5b961fc72933e623a821915d2d598b01923968e6R304
3) Called by send or recv thread: https://github.com/nasa/fprime/pull/4136/files#diff-67a68b4daab941bc994e0bbd5b961fc72933e623a821915d2d598b01923968e6R346

But I was getting intermittent but frequent failures locally in the Drv/Udp UTs, and some debugging led to the fact that that `Os::Task::delay` uses a singleton, so in the cases of (1) & (2) where we have the `Os::Task` handler (for the dedicated reconnection thread) I now use `_delay` and still use `Os::Task::delay` in (3), but I would like to know if there's another recommended way to approach this?

m-aleem:
@LeStarch This is ready for a preliminary review and I have left a question in-line with one of the FIXMEs (they're all related), but I am keeping this PR in draft until that discussion is resolved/all the FIXMEs are addressed (and while I review these CI issues...)

Drv/Ip/SocketComponentHelper.hpp
```diff
@@ -218,13 +265,27 @@ class SocketComponentHelper {
      */
     SocketIpStatus reopen();
 
+    void threadSafeDelay(const Fw::TimeInterval& interval);
```
LeStarch: Was this defined anywhere?

Os/Task.cpp
```diff
@@ -94,6 +94,11 @@ Task::Status Task::start(const Fw::StringBase& name,
 }
 
 Task::Status Task::start(const Task::Arguments& arguments) {
+    Task::init();
+    // init call above is to ensure singleton is initialized in a thread-safe
+    // manner and such that the address sanitizer does not inadvertently
+    // result in a stack overflow when multiple calls to getSingleton are made
+    // simultaneously from different threads. (As was observed in UT runs.)
```
m-aleem, line 101: @LeStarch Moved the init() call to this start method, since it can be called directly or via the other above start call (SocketComponentHelpers actually calls this version), so it'll run in either case - and that seemed to do the trick to get consistent, not-failing UT runs (locally did 7 iterations)

Drv/Ip/SocketComponentHelper.cpp
```diff
@@ -25,6 +25,18 @@ void SocketComponentHelper::start(const Fw::StringBase& name,
                                   const FwTaskPriorityType priority,
                                   const Os::Task::ParamType stack,
                                   const Os::Task::ParamType cpuAffinity) {
+    // Reconnect Thread
+    FW_ASSERT(m_reconnectTask.getState() ==
+              Os::Task::State::NOT_STARTED);  // It is a coding error to start this task multiple times
+    this->m_reconnectStop = false;
+    Fw::String reconnectName;
+    reconnectName.format("%s_reconnect", name.toChar());
```
LeStarch, line 36: This implies the reconnect threads has the same priority as the receive thread.

LeStarch (approved):

LeStarch (changes requested):
Minor issues:

1. Address priority
2. Remove undefined function

Drv/Ip/SocketComponentHelper.hpp
m-aleem: This was leftover from a previous iteration. Removed. Thanks!

Drv/Ip/SocketComponentHelper.cpp
m-aleem: Great catch, I forgot to add these to the constructor. Updated with defaults.

LeStarch (approved):

Issue #3329, opened 2025-03-04
Provide ability to update parameters via a file.


timcanham:
| | |
|:---|:---|
|**_F´ Version_**|v3.6 |
|**_Affected Component_**| `Svc/PrmDb`  |
---
## Feature Description

Update `PrmDb` with a command that loads a file from storage and updates the RAM values for components. The file could have selective IDs and not the whole set. A subsequent `PRM_SAVE_FILE` would write the values to the `PrmDb` file.

## Rationale

In operations, sending a bunch of commands with individual parameter updates can be burdensome. This approach can do it in a much more coordinated way.

LeStarch:
Merged!

PR #4165, opened 2025-09-17
Provide ability to update parameters via a file

m-aleem:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3329  |
|**_Documentation Included (y/n)_**| Y (comments) |

---
## Change Description

1) Added a command to set parameters from file.
2) Expanded PrmDb to have a Prime and a Backup database (rather than single Db). Note that I did not modify the _external_ interface beyond addition of the new command, but see Future Work for recommendations related to this.

## Rationale

1) This item was requested in #3329 
2) This item was added because we need a way to 'roll back' changes if some aspect of the parameter files error checking

* Added UTs for the new features

## Future Work

There are various updates to the Parameter DB service that I think can be spun off into 1 or multiple more issues

1) Revisit design of database itself, currently a simple array, but this will not scale
2) Consider updating 'save file' command to explicitly take in which database (prime or backup) to save to a file
3) Add integration tests to exercise the new command
4) Add CRC checking to parameter file logic (this was already in comments as a to do)

* UT Generation!


Svc/PrmDb/PrmDbCmdDict.fppi
```diff
@@ -1,3 +1,10 @@
 @ Command to save parameter image to file. Uses file name passed to constructor
 async command PRM_SAVE_FILE \
-  opcode 0
+  opcode 0x00
+
+
+@ Loads a file from storage and updates the volatile values for the parameters defined in the file. The file could have selective IDs and not the whole set.
+async command PRM_SET_FILE(
+  fileName: string size 100 @< The name of the on-board file to set parameters format
```
timcanham: Isn't this "to set parameter values"?

Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -37,18 +35,21 @@ class WorkingBuffer : public Fw::SerializeBufferBase {
...
 void PrmDbImpl::configure(const char* file) {
     FW_ASSERT(file != nullptr);
     this->m_fileName = file;
 }
 
-void PrmDbImpl::clearDb() {
+void PrmDbImpl::clearDb(PrmDbType prmDbType) {
+    t_dbStruct* db;
+    getDbPtr(prmDbType, &db);
     for (FwSizeType entry = 0; entry < PRMDB_NUM_DB_ENTRIES; entry++) {
```
timcanham: Agree with code warning; pointer should be asserted.

Svc/PrmDb/PrmDbImpl.hpp
```diff
@@ -99,20 +168,40 @@ class PrmDbImpl final : public PrmDbComponentBase {
...
     //!  This function clears all entries from the RAM database
     //!
+    //!  \param dbType The type of database to read into (primary or backup)
+    void clearDb(PrmDbType prmDbType);  //!< clear the parameter database
+
+    //!  \brief PrmDb get db pointer function
+    //!  This function returns a pointer to the requested database
+    //!  \param dbType The type of database requested (primary or backup)
+    //!  \param dbPtr Pointer to the database array to be set
+    void getDbPtr(PrmDbType dbType, t_dbStruct** dbPtr);
```
timcanham: If the return is void, why not return the pointer?

Svc/PrmDb/PrmDbImpl.hpp
```diff
@@ -79,6 +135,19 @@ class PrmDbImpl final : public PrmDbComponentBase {
...
+    //!  This function does the underlying parameter update
+    //!
+    //!  \param id identifier for parameter being used.
+    //!  \param val buffer where value to be saved is stored.
+    //!  \param dbType The type of database to read into (primary or backup)
+    //!  \param index pointer to index of updated or added parameter
+    PrmDbImpl::paramUpdateType updateAddPrm(FwPrmIdType id,
+                                            Fw::ParamBuffer& val,
+                                            PrmDbType prmDbType,
+                                            FwSizeType* index);
```
timcanham: make this a `FwSizeType&` instead. Safer.

Svc/PrmDb/PrmDbImpl.hpp
```diff
@@ -58,8 +63,58 @@ class PrmDbImpl final : public PrmDbComponentBase {
     //!
     virtual ~PrmDbImpl();
 
+    enum paramUpdateType { NO_SLOTS, PARAM_ADDED, PARAM_UPDATED, MAX_PARAM_UPDATE_TYPES };
```
timcanham: To keep with the convention, this should be `ParamUpdateType`. Lower case is for variable.

Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -77,48 +78,87 @@ Fw::ParamValid PrmDbImpl::getPrm_handler(FwIndexType portNum, FwPrmIdType id, Fw
...
+
+    FW_ASSERT(dbType == PrmDbType::DB_PRIME or dbType == PrmDbType::DB_BACKUP);
     FW_ASSERT(this->m_fileName.length() > 0);
+
+    t_dbStruct* db;
+    if (dbType == PrmDbType::DB_PRIME) {
+        db = this->m_dbPrime;
+    } else if (dbType == PrmDbType::DB_BACKUP) {
+        db = this->m_dbBackup;
+    } else {
```
timcanham: This should be caught by the `FW_ASSERT` above.

Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -338,36 +422,96 @@ void PrmDbImpl::readParamFile() {
         desStat = buff.deserializeTo(parameterId);
         FW_ASSERT(Fw::FW_SERIALIZE_OK == desStat);
 
-        // copy parameter
-        this->m_db[entry].used = true;
-        this->m_db[entry].id = parameterId;
+        // copy parameter value from file into a temporary buffer
+        Fw::ParamBuffer tmpParamBuffer;  // temporary param buffer to read parameter value from file
```
timcanham, line 426: We should look at how these buffers are instantiated on the stack. If they are larger, it can cause stack problems on VxWorks.

Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -338,36 +422,96 @@ void PrmDbImpl::readParamFile() {
...
+        FwSizeType index;  // Which index was used for this parameter
+        paramUpdateType updateStatus = updateAddPrm(parameterId, tmpParamBuffer, dbType, &index);
+        if (updateStatus == PARAM_ADDED) {
+            recordNumAdded++;
+        } else if (updateStatus == PARAM_UPDATED) {
+            recordNumUpdated++;
+        }
+
+        if (updateStatus == NO_SLOTS) {
+            this->log_FATAL_PrmDbFull(parameterId);
```
timcanham: I know it was in the original, but we shouldn't make this FATAL, since the ground could upload a valid but incorrectly sized file.

timcanham (changes requested):
I think we need to discuss the conops for the file loading before merging. The sequence of operations should be as follows:

1) Ground uplinks the file
2) The file load command loads the data into a backup set of buffers. This could take a while for larger files. The "prime" set of buffers would be active during this time in a read-only state and would reject commands to update parameters. Flight rules should make sure that component parameter update commands were not issued during this time, but if they are pushed to ParamDb, they should be rejected with WARNING_HI EVRs.
3) There should be a "commit" command to switch a pointer from the "prime" set of buffers to the newly loaded backup set. It would now be the active one, and any component queries would go to this one. Any components that need new values prior to reboot would need a command to reload the parameters.
4) A command to write the new values to the parameter file could happen any time after this.
5) The command to load a file should have options to reset or merge from a file. That way the ground could do a set of files and completely overwrite the values on board. If the request was for a merge, the values in the prime would be copied first. If not, the "backup" would start empty, and the file would fill new values.

I think it's close, but not quite there.

Svc/PrmDb/PrmDbCmdDict.fppi
m-aleem: Changed the names of the commands and updated descriptions.

Svc/PrmDb/PrmDbImpl.hpp
m-aleem: Changed -> PrmUpdateType

Svc/PrmDb/PrmDbImpl.hpp
m-aleem: Updated getDbPtr to return pointer

Svc/PrmDb/PrmDbImpl.hpp
m-aleem: Updated updateAddPrmImpl and we actually no longer use this argument so removed it

Svc/PrmDb/PrmDbImpl.cpp
m-aleem: PrmDbFull changed from FATAL to WARNING_HI

Svc/PrmDb/PrmDbImpl.cpp
m-aleem: Updated getDbPtr and getDbString to now rely on assert and just do if/else return

Svc/PrmDb/PrmDbImpl.cpp
m-aleem: FW_ASSERT is in the getDbPtr, which is the first call in this function

m-aleem:
Updates following first review and additional ConOps information:

1) Rather than using the nomenclature "prime"/"backup" which indicates the two parameter databases (prime and backup) should have the same content, I am switching to an 'active' and 'staging' database nomenclature.
    * The **active** parameter database is loaded upon init (if applicable), all get/set queries are sent to this database
    * The **staging** parameter database is only used for file loads
2)  Loading parameters from a file is a two-command process
    1) **PRM_LOAD_FILE** to load the file-based parameter updates into the staging database
        - argument merge (bool) indicates whether the parameter file updates should be merged into the existing parameters (true) or weather the parameters should be wiped and the file load represents the full set of parameters (false); at the end of this command the parameter updates are considered staged (in the staging database)
    2) **PRM_COMMIT_STAGED** to apply the staged changes by swapping pointers such that staging becomes active
3) After the PRM_LOAD_FILE has been initiated, PrmDb will NOT allow any of the following until we successfully commit the updates (following the second command being sent) OR if we clean up and exit early
    * PRM_LOAD_FILE cmd (rationale: only 1 file load can be active at a time)
    * PRM_SAVE_FILE cmd (rationale: we should not save to file in the middle of a file update)
    * setPrm_handler (rationale: don't edit active database in the middle of the file-based update)

Operation Flow:

* PrmDb states:
    * IDLE
    * LOADING_FILE_UPDATES
    * FILE_UPDATES_STAGED
   
* Nominal Flow: `IDLE -cmd(1)-> LOADING_FILE_UPDATES -> FILE_UPDATES_STAGED -cmd(2)-> [Active database updated via active/staging pointer-swap] -back to-> IDLE`
* Off-Nominal Flow: `IDLE -cmd(1)-> LOADING_FILE_UPDATES [Staging database cleared, due to some file error, Active database not impacted] -back to-> IDLE`

Future To Do/Considerations:
* Consider channelizing the parameter db file load state
* Consider the time it could take for the parameter file load/how is reentrancy handled should that need to be split up over multiple PrmDb cycles?
* Consider the comment from Tim on https://github.com/nasa/fprime/pull/4165/files#r2377661723

Svc/PrmDb/PrmDbCmdDict.fppi
```diff
@@ -1,3 +1,15 @@
 @ Command to save parameter image to file. Uses file name passed to constructor
 async command PRM_SAVE_FILE \
-  opcode 0
+  opcode 0x00
+
+
+@ Loads a file from storage into the staging database. The file could have selective IDs and not the whole set.
+async command PRM_LOAD_FILE(
+  fileName: string size 100 @< The name of the on-board file to set parameters from
```
LeStarch: Should use the file name constant

Svc/PrmDb/PrmDbCmdDict.fppi
```diff
@@ -1,3 +1,15 @@
...
+  merge: bool @< Whether to merge (true) or fully reset (false) the parameter database from the file contents
```
LeStarch: Best to make this an FPP enum such that we aren't sending raw booleans.

Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -78,47 +95,42 @@ Fw::ParamValid PrmDbImpl::getPrm_handler(FwIndexType portNum, FwPrmIdType id, Fw
...
-                this->m_db[entry].val = val;
-                this->m_db[entry].id = id;
-                this->m_db[entry].used = true;
-                noSlots = false;
-                break;
-            }
-        }
+    // Reject parameter updates during non-idle file load states
+    if (m_state != PrmDbFileLoadState::IDLE) {
+        this->log_WARNING_LO_PrmDbFileLoadInvalidAction(m_state, Fw::String("setPrm"));
```
LeStarch: Prefer an ENUM over a string arg.

Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -78,47 +95,42 @@ Fw::ParamValid PrmDbImpl::getPrm_handler(FwIndexType portNum, FwPrmIdType id, Fw
...
+    if (m_state != PrmDbFileLoadState::IDLE) {
+        this->log_WARNING_LO_PrmDbFileLoadInvalidAction(m_state, Fw::String("SAVE_FILE"));
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::BUSY);
+        return;
+    }
+
     FW_ASSERT(this->m_fileName.length() > 0);
+
+    t_dbStruct* db = getDbPtr(PrmDbType::DB_ACTIVE);
+
```
LeStarch, line 131: Assert db not null here

Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -233,24 +245,94 @@ void PrmDbImpl::PRM_SAVE_FILE_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
...
+void PrmDbImpl::PRM_COMMIT_STAGED_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
+    // Verify we are in the correct state
+    if (m_state != PrmDbFileLoadState::FILE_UPDATES_STAGED) {
+        this->log_WARNING_LO_PrmDbFileLoadInvalidAction(m_state, Fw::String("COMMIT_STAGED"));
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::VALIDATION_ERROR);
+        return;
+    }
+
+    // Swap active and staging databases
```
LeStarch: This code needs to be protected from the "get" handler.

Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -233,24 +245,94 @@ void PrmDbImpl::PRM_SAVE_FILE_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
...
-    FW_ASSERT(this->m_fileName.length() > 0);
     // load file. FIXME: Put more robust file checking, such as a CRC.
     Os::File paramFile;
 
-    Os::File::Status stat = paramFile.open(this->m_fileName.toChar(), Os::File::OPEN_READ);
+    Os::File::Status stat = paramFile.open(fileName.toChar(), Os::File::OPEN_READ);
     if (stat != Os::File::OP_OK) {
         this->log_WARNING_HI_PrmFileReadError(PrmReadError::OPEN, 0, stat);
-        return;
+        return false;
```
LeStarch: prefer a status enum

Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -338,36 +421,133 @@ void PrmDbImpl::readParamFile() {
...
+    this->unLock();
+    return updateStatus;
+}
+
+//! ----------------------------------------------------------------------
+//! Helpers for database management
+//! ----------------------------------------------------------------------
+
+void PrmDbImpl::clearDb(PrmDbType prmDbType) {
+    t_dbStruct* db = getDbPtr(prmDbType);
```
LeStarch, line 501: Lock here, I think?

LeStarch (commented):
Partial review.  Waiting on @timcanham to confirm the con-ops.

Svc/PrmDb/PrmDbImpl.cpp
timcanham: It would make sense to have a private data member that is a working buffer. If this is the only function using it, it should be safe.

timcanham (dismissed):
Looks good so far - I think we need to add https://github.com/nasa/fprime/issues/4270 ASAP. I had thought this feature was in standard F Prime, but we can add it after the merge.

timcanham (dismissed):
LGTM!

LeStarch (dismissed):
I corrected the minor nits, and @timcanham approved.

LeStarch (approved):
Approving w/ @timcanham's approval and I updated the minor nits I had.

PR #4166, opened 2025-09-18
Add new PassiveCmdDispatcher component

ianbrault:
| | |
|:---|:---|

---
## Change Description

This PR adds a new component `PassiveCmdDispatcher` which is designed to be a pared-down, passive replacement for the active `CmdDispatcher` component. The port handlers are implemented identically, minus some extraneous events and telemetry.

## Rationale

We are running in a memory-constrained environment and need to limit our data size. One of our largest impacts comes from the `CmdDispatcher` component; see [CmdDispatcher.txt](https://github.com/user-attachments/files/22396391/CmdDispatcher.txt) for a breakdown of its memory footprint (given our configured constants). The active component base takes up a significant amount of this data size, and we do not need it to be active for our uses, so after a discussion with members of the fprime team we determined that the best solution was to replace `CmdDispatcher` with a passive equivalent. 

This is a drop-in replacement for `CmdDispatcher`, swap it out for a `PassiveCmdDispatcher` instance in the topology and send commands thru the GDS to verify that it works.

## Future Work

Add UTs/additional documentation.

Not used.

### DIFF
--- a/Svc/PassiveCmdDispatcher/PassiveCmdDispatcher.cpp
+++ b/Svc/PassiveCmdDispatcher/PassiveCmdDispatcher.cpp
@@ -4,15 +4,26 @@
 // \brief  cpp file for PassiveCmdDispatcher component implementation class
 // ======================================================================
 
-#include "Svc/PassiveCmdDispatcher/PassiveCmdDispatcher.hpp"
+#include <Fw/Cmd/CmdPacket.hpp>
+#include <Svc/PassiveCmdDispatcher/PassiveCmdDispatcher.hpp>
 
 namespace Svc {
+
+// Check the CMD_DISPATCHER_DISPATCH_TABLE_SIZE and CMD_DISPATCHER_SEQUENCER_TABLE_SIZE constants for overflow
+static_assert(CMD_DISPATCHER_DISPATCH_TABLE_SIZE <= std::numeric_limits<FwOpcodeType>::max(),
+              "Opcode table limited to opcode range");
+static_assert(CMD_DISPATCHER_SEQUENCER_TABLE_SIZE <= std::numeric_limits<U32>::max(),
+              "Sequencer table limited to range of U32");
 
 // ----------------------------------------------------------------------
 // Component construction and destruction
 // ----------------------------------------------------------------------
 
-PassiveCmdDispatcher::PassiveCmdDispatcher(const char* const compName) : PassiveCmdDispatcherComponentBase(compName) {}
+PassiveCmdDispatcher::PassiveCmdDispatcher(const char* const compName)
+    : PassiveCmdDispatcherComponentBase(compName), m_seq(0) {
+    memset(this->m_entryTable, 0, sizeof(this->m_entryTable));
+    memset(this->m_sequenceTracker, 0, sizeof(this->m_sequenceTracker));
+}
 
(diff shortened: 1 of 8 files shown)


ianbrault:
CC @celskeggs and @kevin-f-ortega

Svc/PassiveCmdDispatcher/PassiveCmdDispatcher.cpp
```diff
@@ -0,0 +1,168 @@
...
+              "Sequencer table limited to range of U32");
+
+// ----------------------------------------------------------------------
+// Component construction and destruction
+// ----------------------------------------------------------------------
+
+PassiveCmdDispatcher::PassiveCmdDispatcher(const char* const compName)
+    : PassiveCmdDispatcherComponentBase(compName), m_seq(0) {
+    memset(this->m_entryTable, 0, sizeof(this->m_entryTable));
+    memset(this->m_sequenceTracker, 0, sizeof(this->m_sequenceTracker));
```
celskeggs, line 25: For the sake of reducing code size... can we find another way to initialize these arrays so that we don't need a `memset`?

Svc/PassiveCmdDispatcher/PassiveCmdDispatcher.hpp
```diff
@@ -0,0 +1,102 @@
...
+    //!
+    //! Clear command tracking info to recover from components that are not returning status
+    void CMD_CLEAR_TRACKING_cmdHandler(FwOpcodeType opCode,  //!< The opcode
+                                       U32 cmdSeq            //!< The command sequence number
+                                       ) override;
+
+    struct DispatchEntry {
+        bool used;            //!< if entry has been used yet
+        FwOpcodeType opcode;  //!< opcode of entry
+        FwIndexType port;     //!< which port the entry invokes
```
celskeggs, line 79: Can we find a way to avoid needing the `bool used` in these two data structures? It's a lot of overhead.

Svc/PassiveCmdDispatcher/docs/sdd.md
```diff
@@ -0,0 +1,71 @@
+# Svc::PassiveCmdDispatcher
+
+A passive component for dispatching commands. Designed as a pared-down, passive replacement for
+[`Svc::CmdDispatcher`](../../CmdDispatcher/docs/sdd.md).
```
celskeggs, line 4: I think we should document that the memory coherency of the data structures in this component assume that only one invocation of any port is made at a time. This is true in our use case, but it might not be true for all possible users.

celskeggs (commented):
Looks pretty good to me, just a few notes.

LeStarch:
We should probably replace the inline map-like data structure with our actual map implementations: https://github.com/nasa/fprime/blob/b28709cbdbf1603c06252e61f07bd233686583b9/Fw/DataStructures/MapBase.hpp#L18

We should try to avoid code-duplication with command dispatcher (e.g. refactor into a common library).

I'd also like your opinion: should this PR add the component to F Prime core, or should we add it to fprime-baremetal where we have other minimalist components for use on restricted systems?

ianbrault:
@LeStarch I think it makes more sense to add this component to `fprime-baremetal`, I will submit a PR to that repo

LeStarch:
@ianbrault I will close this one.

PR #4177, opened 2025-09-22
Fix Intermittent UT failure in FprimeDeframer

### LINKED ISSUE #4158: Intermittent unit test failure for F Prime defamer
See here: https://github.com/nasa/fprime/actions/runs/17803483588/job/50609269104?pr=4157.

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/4158

### UT failure explanation

`FprimeDeframer` would go ahead and attempt to deserialize an APID from the data regardless of how much data is available. In cases where there's less data in the frame than `sizeof(FwPacketDescriptor)`, then it would deserialize garbage data (the CRC value). This was never an operational scenario because we always have at least a full `FwPacketDescriptor` in our packet data, but this is still a quirk that needs fixing.

It started being intermittent recently because the garbage-deserialized `FwPacketDescriptor` data is checked to be `< ComCfg::Apid::INVALID_UNINITIALIZED` (=2048), which for a random U32 was pretty hard to get. We switched to `FwPacketDescriptor` to U16 and the intermittence started being noticeable. 

Was rightfully caught by random values in UTs. Go `STest::Random` !!

### Proposed fix 
Only attempt to harvest the APID from the encapsulated data if there is enough valid data to begin with.

### DIFF
--- a/Svc/FprimeDeframer/FprimeDeframer.cpp
+++ b/Svc/FprimeDeframer/FprimeDeframer.cpp
@@ -58,15 +58,18 @@ void FprimeDeframer ::dataIn_handler(FwIndexType portNum, Fw::Buffer& data, cons
         return;
     }
     // -------- Attempt to extract APID from Payload --------
-    // If PacketDescriptor translates to an invalid APID, let it default to FW_PACKET_UNKNOWN
-    // and let downstream components (e.g. custom router) handle it
-    FwPacketDescriptorType packetDescriptor;
-    status = deserializer.deserialize(packetDescriptor);
-    FW_ASSERT(status == Fw::SerializeStatus::FW_SERIALIZE_OK, status);
     ComCfg::FrameContext contextCopy = context;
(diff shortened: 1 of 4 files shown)


LeStarch:
@thomas-bc the contract with the deframer is that it shall never be handed an incomplete packet. However, since the engine that produces these packets might fail or error, this scenario is possible.  Therefore it is the correct answer to skip deframing (as opposed to asserting, or making blind assumptions).

Should we put a WARNING event in this case?

LeStarch (commented):
See comment.

thomas-bc:
Updated to emit the event, and reflected the expectation in the F Prime framing protocol SDD.

LeStarch (approved):

vijayts2704-hub:
dell bill inviocr now

Issue #4192, opened 2025-09-25
File Unit Test RegEx bug


kevin-f-ortega:
| | |
|:---|:---|
|**_F´ Version_**| `devel` |
|**_Affected Component_**| `Os/test/ut/file`  |
---
## Problem Description

The regular expression has a slight error. It doesn't handle relative-path assertions.

## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-4.18.0-553.16.1.el8_10.x86_64-x86_64-with-glibc2.28
Python version: 3.11.9
CMake version: 3.26.0
Pip version: 25.2
Pip packages:
    fprime-tools==4.0.2a1
    fprime-gds==4.0.2a6
    fprime-fpp==3.1.0a1
Project submodules:
```

## How to Reproduce

1. Run entire FileSystem unit tests with relative-path asserts

## Expected Behavior

I expect the unit tests to pass

kevin-f-ortega:
Addressed here https://github.com/nasa/fprime/pull/4193

PR #4193, opened 2025-09-25
regex is a bit wrong. It never would work with 'Os/File.hpp'

kevin-f-ortega:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/4192  |
|**_Generative AI was used in this contribution (y/n)_**|n  |

---
## Change Description

Fixed regular expression when matching with ASSERT messages

## Rationale

Sometimes we use relative-path asserts which provide relative paths to the files. The regex was not working in this case.

Rerun File unit tests

## Future Work

None

None


LeStarch (approved):
Approved.

Issue #4155, opened 2025-09-16
Add a dedicated Component port queue for Commands for all F' Components


bitWarrior:
| | |
|:---|:---|
|**4.0**| |
|**All F' Components**|  |
---
## Feature Description

Add a dedicated port queue for command processing for all F' Components

## Rationale

Currently all ports of an F' component share the same queue. This includes the cmdIn port. Since the number of commands an F' component will receive and process varies greatly based on the functionality of the component, this is not ideal. If a component receives a series of commands within a short time-span, then the queue will be filled with commands, potentially starving out queue space for the other ports such as the tlmOut and eventOut ports. It also leads to a cyber vulnerability where an adversary can flood a component with commands until the queue is filled and an ASSERT is generated, causing the FSW to stop executing. This is known as an Denial of Service (DOS) attack. 

My recommendation is to have a dedicated queue for the cmdIn, cmdRegOut, and cmdResponseOut ports. This will free up queue space for the other ports defined for the component. 

This recommendation would be a modification to the auto code generator.  When a "fprime-util new --component" is processed, this feature will be done for the user automatically. 

I understand this modification will add to F's baseline resource allocation, but my option is the improved robustness to F' counters that limitation. Using the default queue size of 10, the appropriate allocation for a components port queue is only ~800 bytes. Implementing this feature, assuming the default size of 10 for commands, will add 800 bytes per component. 

The cyber resiliency adding this feature to F' cannot be understated.

thomas-bc:
CCB: this could be implemented with a single queue by using a behavior of "drop commands after queue is X % full"

Issue #4154, opened 2025-09-16
Add Timeout to Command Dispatcher for commands in-process


bitWarrior:
| | |
|:---|:---|
|**4.0**| |
|**CommandDispatcher**|  |
---
## Feature Description

Request to add a timeout to Command Dispatcher for commands in-process

## Rationale

One of the modifications we made to the "Up the Hill" F' version is a timeout in Command Dispatcher for command "in process". The rationale for this feature this is to prevent the condition where the FSW becomes uncommandable because CommandDispatcher is waiting for a response from the Component being commanded that it will never receive due to an off-nominal condition. 

Currently, Command Dispatcher will wait forever for a response and will not process any commands. This is a classic "lock out" condition. The only way to clear this condition is to reboot the Hardware/FSW. Although this condition should not occur, Murphy's Law is real and this feature adds robustness to F'. 

The default timeout we are using is 5 seconds, but its a parameter so its adjustable without code modification.

timcanham:
Note that the Command Dispatcher does not stop responding if a single command doesn't return. It has a dispatch table that stores the command sequence number, and it can have multiple outstanding commands. If the table fills up, it will assert. One command doesn't block subsequent commands. A sequencer will block until a command is complete, because it specifically is designed to wait for each command.

LeStarch:
Closed.

PR #4198, opened 2025-09-26
CWE 453 cmd buffer assert

bitWarrior:
| | |
|:---|:---|
|Related Issue(s) CWE-617 |  |
|Has Unit Tests (y/n) Y |  |
|Documentation Included (y/n) N |  |
|Generative AI was used in this contribution (y/n) N |  |

---
## Change Description
The modifications to the F' baseline are as follows:
1) Behavior modification to Command Dispatcher's seqCmdBuf port queue overflow from ASSERT to event generation (added new hook)
2) New event to record seqCmdBuf port queue overflow (CommandDroppedQueueOverflow)
3) New telemetry monitor (CommandsDropped)
4) New hook handler function (seqCmdBuff_overflowHook)

## Rationale
As was documented in CWE-617, if Command Dispatcher's seqCmdBuf port queue overflows due to a rapid rate of received commands, the default action was to ASSERT. This causes any FSW  using F' to cease to execute.  

A new unit test was created to verify generation of the CommandDroppedQueueOverflow event and increment of the CommandsDropped telemetry channel if the seqCmdBuf port queue overflows. 

## Future Work
There are other vulnerabilities with command flooding in F'. They have been recorded under the following PR's:
1) #4155 -  Add a dedicated Component port queue for Commands for all F' Components
                       * All F' components will have a dedicated port queue for command processing with
                          default "hook's" to handle overflow conditions.
2) #4154 - Add Timeout to Command Dispatcher for commands in-process 
                      * Prevent an off-nominal condition where the FSW becomes noncommunicable 

N/A


TestMyFprime/Top/CMakeLists.txt
thomas-bc, line 1: Need to get rid of this test dir

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -188,4 +188,20 @@ void CommandDispatcherImpl::pingIn_handler(FwIndexType portNum, U32 key) {
     this->pingOut_out(0, key);
 }
 
+void CommandDispatcherImpl::seqCmdBuff_overflowHook(FwIndexType portNum, Fw::ComBuffer& data, U32 context){
+    // Extract command opcode
+    Fw::CmdPacket cmdPkt;
+    Fw::SerializeStatus stat = cmdPkt.deserialize(data);
+    U32 opcode = 0xDEADBEEF; // Note: 0xDEADBEEF = unable to extract cmd opcode
```
thomas-bc: Probably should use `FwOpcodeType` instead of U32

Svc/CmdDispatcher/test/ut/CommandDispatcherTester.cpp
```diff
@@ -879,6 +878,87 @@ void CommandDispatcherTester::runClearCommandTracking() {
...
+
+        this->clearEvents();
+        this->invoke_to_seqCmdBuff(0, buff, 12);
+        ASSERT_EQ(Fw::QueuedComponentBase::MSG_DISPATCH_OK, this->m_impl.doDispatch());
+    }
+
+    // Verify at least one Queue Overflow event was generated and the
+    // CommandsDropped telemetry channel is not 0
+    ASSERT_GT(this->eventHistory_CommandDroppedQueueOverflow->size(), 0);
+    ASSERT_GT(this->tlmHistory_CommandsDropped->size(), 0);
```
thomas-bc: Do we not want to just assert `== 1` on both of these?

thomas-bc (commented):
few nits but LGTM

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
thomas-bc: CI is also pointing out we now need `deserializeFrom()` instead of `deserialize()`

CMakeLists.txt
```diff
@@ -34,3 +34,4 @@ add_compile_options(
     -Wno-vla
 )
 include("${CMAKE_CURRENT_LIST_DIR}/cmake/FPrime-Code.cmake")
```
LeStarch, line 36: Remove this Test deployment.

Svc/CmdDispatcher/CmdDispatcher.fpp
```diff
@@ -184,6 +184,16 @@ module Svc {
       id 10 \
       format "Opcode 0x{x} is already registered to port {}"
 
+    @ This log event reports the Command Sequence Buffer port queue has overflowed.
+    event CommandDroppedQueueOverflow(
+                              OpCode: U32 @< The command opcode dropped
```
LeStarch: Should use FwOpcodeType in the event

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
LeStarch: Can we use 0x00000000 as the invalid opcode?  EventManager already uses 0x00000000 as a reserved opcode, and so it would be natural to "reserve" the same id.

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -188,4 +188,20 @@ void CommandDispatcherImpl::pingIn_handler(FwIndexType portNum, U32 key) {
...
+    U32 opcode = 0xDEADBEEF; // Note: 0xDEADBEEF = unable to extract cmd opcode       
+
+    if (stat == Fw::FW_SERIALIZE_OK){
+        opcode = cmdPkt.getOpCode();
+    }
+    
+    // Log Cmd Buffer Overflow and increment CommandsDroppedBufOverflow counter
+    this->log_ACTIVITY_HI_CommandDroppedQueueOverflow(opcode,context);
+    this->m_numCmdsDropped++;
+    this->tlmWrite_CommandsDropped(this->m_numCmdsDropped);
```
LeStarch: Just a note, this is not throttled until post telemetry manager.  Probably ok, but this will put some pressure on tlm component, but since that component is designed for efficient high-rate processing....it should be ok. 

We could also argue that we should telemeter this value on a rate group tick rather than on update.  This is how most component slow down telemetry with an unbounded update rate.

Svc/CmdDispatcher/test/ut/CommandDispatcherTester.cpp
LeStarch: Also, this test does not test the throttle behavior.  It might be better to push a number of commands past the overflow limit and ensure that the throttle works.  Why?  If the throttle fails, then another part of the system (eventManager) is DDOSed.

LeStarch (changes requested):

Svc/CmdDispatcher/CmdDispatcher.fpp
bitWarrior: Updated to FwOpcodeType

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
bitWarrior: I like to use a specific value for an off-nominal condition, but 0 is fine so I made that update.

bitWarrior: I updated the UT to verify the throttle behavior for CommandDroppedQueueOverflow

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
bitWarrior: Updated to deserializeFrom() and changed to FwOpcodeType

Svc/CmdDispatcher/test/ut/CommandDispatcherTester.cpp
bitWarrior: Updated ASSERTS to ASSERT_EVENTS and ASSERT_TLM

TestMyFprime/Top/CMakeLists.txt
bitWarrior: D'oh! Removed test directory

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -188,4 +188,20 @@ void CommandDispatcherImpl::pingIn_handler(FwIndexType portNum, U32 key) {
...
+    FwOpcodeType opcode = 0; // Note: 0 = Reserved opcode    
+
+    if (stat == Fw::FW_SERIALIZE_OK){
+        opcode = cmdPkt.getOpCode();
+    }
+    
+    // Log Cmd Buffer Overflow and increment CommandsDroppedBufOverflow counter
+    this->m_numCmdsDropped++;
+    this->log_WARNING_HI_CommandDroppedQueueOverflow(opcode,context);
+    this->tlmWrite_CommandsDropped(this->m_numCmdsDropped);
```
LeStarch: Two more things:
1. We should report tlmWrite_CommandsDropped on a sched call to prevent a tlm flood in response to a cmd flood
2. We should  add CommandDroped telemetry to `Ref/Top/RefPackets.fppi` this will   prevent the CI failure.

Happy to clean these nits up if you'd like help.

LeStarch (changes requested):

LeStarch (approved):

thomas-bc (approved):

Issue #4218, opened 2025-09-29
Read Telemetry Time


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Add a directive to read telemetry time.

PR #4230, opened 2025-09-30
FpySequencer Add push tlm val and time dir

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4218  |

---
## Change Description

This PR adds a PUSH_TLM_VAL_AND_TIME directive, which takes in one hardcoded argument, the tlm channel, and pushes the value of the tlm chan and the time to the stack.

Also updates the sdd with the latest directives.

## Rationale

This is needed for situations in which the time a telemetry measurement was made is important.

Used for generating updated sdd.md with all the directives.

### DIFF
--- a/Svc/FpySequencer/FpySequencerDirectives.cpp
+++ b/Svc/FpySequencer/FpySequencerDirectives.cpp
@@ -451,9 +451,7 @@ Signal FpySequencer::pushTlmValAndTime_directiveHandler(const FpySequencer_PushT
     FW_ASSERT(stat == Fw::SerializeStatus::FW_SERIALIZE_OK, static_cast<FwAssertArgType>(stat));
 
     // check that our stack won't overflow if we put both val and time on it
-    // cast to u64 first so that we don't get u32 overflow
-    if (static_cast<U64>(this->m_runtime.stackSize) + tlmValue.getBuffLength() + timeEsb.getBuffLength() >
-        Fpy::MAX_STACK_SIZE) {
+    if (Fpy::MAX_STACK_SIZE - tlmValue.getBuffLength() - timeEsb.getBuffLength() < this->m_runtime.stackSize) {
         error = DirectiveError::STACK_OVERFLOW;
         return Signal::stmtResponse_failure;
     }
--- a/Svc/FpySequencer/FpySequencerDirectives.cpp
+++ b/Svc/FpySequencer/FpySequencerDirectives.cpp
@@ -237,6 +237,14 @@ void FpySequencer::directive_storeTlmVal_internalInterfaceHandler(
     handleDirectiveErrorCode(Fpy::DirectiveId::STORE_TLM_VAL, error);
 }
 
+//! Internal interface handler for directive_pushTlmValAndTime
+void FpySequencer::directive_pushTlmValAndTime_internalInterfaceHandler(
+    const Svc::FpySequencer_PushTlmValAndTimeDirective& directive) {
+    DirectiveError error = DirectiveError::NO_ERROR;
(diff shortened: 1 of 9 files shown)


zimri-leisher:
@LeStarch this PR is ready for review

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -418,6 +426,47 @@ Signal FpySequencer::storeTlmVal_directiveHandler(const FpySequencer_StoreTlmVal
...
 }
 
+Signal FpySequencer::pushTlmValAndTime_directiveHandler(const FpySequencer_PushTlmValAndTimeDirective& directive,
+                                                        DirectiveError& error) {
+    if (!this->isConnected_getTlmChan_OutputPort(0)) {
+        error = DirectiveError::TLM_GET_NOT_CONNECTED;
+        return Signal::stmtResponse_failure;
+    }
+
+    Fw::Time tlmTime;
```
LeStarch, line 436: I feel like I have seen this code in the tlm directive before.  Could this be a helper, where we just include the extra bit of time processing?

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -418,6 +426,47 @@ Signal FpySequencer::storeTlmVal_directiveHandler(const FpySequencer_StoreTlmVal
...
+    U8 tlmTimeBuf[Fw::Time::SERIALIZED_SIZE];
+    Fw::ExternalSerializeBuffer timeEsb(tlmTimeBuf, Fw::Time::SERIALIZED_SIZE);
+    Fw::SerializeStatus stat = timeEsb.serializeFrom(tlmTime);
+
+    // coding error if this failed, we should have enough space
+    FW_ASSERT(stat == Fw::SerializeStatus::FW_SERIALIZE_OK, static_cast<FwAssertArgType>(stat));
+
+    // check that our stack won't overflow if we put both val and time on it
+    // cast to u64 first so that we don't get u32 overflow
+    if (static_cast<U64>(this->m_runtime.stackSize) + tlmValue.getBuffLength() + timeEsb.getBuffLength() >
```
LeStarch: I prefer to write these as subtractions, because then you don't need to deal with casting.

```
// Make sure the operation will ever be valid
FW_ASSERT(tlmValue.getBuffLength() + timeEsb.getBuffLength() < Fpy::MAX_STACK_SIZE);
if (Fpy::MAX_STACK_SIZE - lmValue.getBuffLength() - timeEsb.getBuffLength() < this->m_runtime.stackSize) { // overflow}

LeStarch (changes requested):
Two nits.  I'd like you to respond, but otherwise looks good.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: I think it would be more trouble than it's worth to make a helper here. Thinking about it for a sec, I don't think it would really save any lines. The problem is that right now the other tlm dir behaves differently, writing it to lvar array whereas this one pushes it to stack. Maybe when I change the other dir to push to stack then we can commonize them.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: why the assert first? shouldn't that just be part of the condition?
EDIT - ah is it because otherwise it could wrap around?

LeStarch (approved):

PR #4157, opened 2025-09-17
Revise GenericHub model

bocchino:
Start to work on #4156.

PR #4244, opened 2025-10-01
Generic Hub updates

timcanham:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Updated Generic Hub component to use updated driver interfaces.

## Rationale

Generic Hub was broken with incorrect interfaces

## Future Work

Need to catch up unit tests.

No used.


timcanham:
I'll re-run the unit tests after resolving the conflicts with `devel`.

bocchino:
@timcanham We need to discuss how this work relates to #4157. It looks to me that this code is assuming context is never used with the GenericHub. It's not clear to me that we always want that; for example, we might want to hook up a com queue mechanism. And with the buffer-plus-context interface, one still can't connect the generic hub directly to a byte stream driver, which is what we need for another project. An alternative pattern is to use adapters or "tinkertoys" to connect the different interfaces. For example, one could have adapters for adapting the simple "send and receive raw Fw::Buffer objects" interface to the sync and async byte stream driver interfaces.

LeStarch:
Closing this in favor of our new issues.

PR #4252, opened 2025-10-02
Clear up encoding and newline errors

celskeggs:
| | |
|:---|:---|

---
## Change Description

Add missing newlines to ends of files. Enable automatic check for missing newlines in clang-format config. Fix UTF-8 encoding of ci/config.xml (was previously latin1 but claimed it was utf-8).

## Rationale

Adding missing newlines and fixing encodings clears up warnings in our line counting tool. Updating the clang-format config allows us to avoid having the issue with missing newlines recur.

Running CI is recommended.

## Future Work

No future work.


ci/config.xml
```diff
@@ -1,9 +1,9 @@
 <?xml version='1.1' encoding='UTF-8'?>
 <project>
   <actions/>
-  <description>&lt;h2&gt;F� Continuous Integration&lt;/h2&gt;&#xd;
+  <description>&lt;h2&gt;F′ Continuous Integration&lt;/h2&gt;&#xd;
```
LeStarch, line 4: The prime got scrabled here.

LeStarch (changes requested):

LeStarch:
In the diff view the prime was scrambled.  I cannot edit it so I cannot fix it.

ci/config.xml
celskeggs: The `0xe2 0x80 0xb2` UTF-8 sequence used here is exactly the same sequence as was used in the F Prime Nomenclature document: https://github.com/nasa/fprime/blob/devel/docs/reference/nomenclature.md?plain=1#L12. I checked using a hex editor.

image

If it works there, it should work here. Unless you're saying that you want me to use a different symbol than was listed in the nomenclature document?

image

LeStarch: Apparently it was a quirk in the GH rendering because upon reload I see the expected symbol.

LeStarch (approved):

Issue #4151, opened 2025-09-15
PosixFileFailure


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

on macOS:

```
76/95 Test: PosixFileTest
Directory:Os/Posix
"PosixFileTest" start time: Sep 15 14:35 PDT
Output:
----------------------------------------------------------
[STest::Random] Generated seed 906639 from system time
````

Resulted in a failure.

LeStarch:
This was merged.

thomas-bc:
New failures, unclear if related: https://github.com/nasa/fprime/actions/runs/23554929905/job/68578950898?pr=4859

PR #4254, opened 2025-10-02
Fix for PosixFileTest UT failure

ianbrault:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4151 |

---
## Change Description

The seed value 906639 (see #4151) creates a test suite where there is an `OpenFileCreate` rule and the randomly-generated filename collides with a previously-created file. This causes the file `open` call to fail with status `Os::Status::FILE_EXISTS` which fails the UT.

This change adds a short snippet to `Os::Test::FileTest::Tester::OpenBaseRule::action` to ensure that this collision cannot happen by re-generating the filename until it is unique when filenames are random and this is not an overwrite test.

## Rationale

PosixFileTest UTs were failing previously.

Seed value 906639 caused the failure, replace the call to `STest::Random::seed` in main inside `Os/Posix/test/ut/PosixFileTests.cpp` with `STest::Random::SeedValue::set(906639)` and run the PosixFileTest UTs.

## Future Work

N/A

N/A

### DIFF
--- a/Os/test/ut/file/FileRules.cpp
+++ b/Os/test/ut/file/FileRules.cpp
@@ -272,6 +272,13 @@ void Os::Test::FileTest::Tester::OpenBaseRule::action(Os::Test::FileTest::Tester
     printf("--> Rule: %s mode %d\n", this->getName(), this->m_mode);
     // Initial variables used for this test
     std::shared_ptr<const std::string> filename = state.get_filename(this->m_random);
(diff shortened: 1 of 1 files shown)


LeStarch (approved):

thomas-bc (approved):

raj4036 (commented):
[Ask Intel transcript.pdf](https://github.com/user-attachments/files/26002424/Ask.Intel.transcript.pdf)

Issue #4049, opened 2025-08-20
FpySequencer "flags" system


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**| FpySequencer |
---
## Feature Description

The sequencer should have a set of boolean flags that control the sequencer's behavior in various ways. These flags should be configurable at sequence runtime via directives.

PR #4259, opened 2025-10-03
FpySequencer flag system

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4049  |

---
## Change Description

* Adds a "flag" system to the sequencer. 
  * Flags can be modified at runtime of a sequence, by the sequence itself, or by ground command
  * Flags get set to a default value at the beginning of running a sequence.
    * The default value is configurable via param
* Adds SET_FLAG and GET_FLAG directives
  * These allow sequences to access and modify the value of flags during execution of a sequence
* Add the EXIT_ON_CMD_FAIL flag
  * Default to false
  * If true, the sequence will immediately end if a command fails


zimri-leisher:
@LeStarch  this is ready for review

LeStarch:
@zimri-leisher merge of the last PR created conflicts here.  Would you take a look?  They aren't trivial.

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -200,6 +200,31 @@ void FpySequencer::DEBUG_CLEAR_BREAKPOINT_cmdHandler(FwOpcodeType opCode,  //!<
...
 
+//! Handler for command SET_FLAG
+//!
+//! Sets the value of a flag. See Fpy.FlagId docstrings for info on each flag.
+//! This command is only valid in the RUNNING state.
+void FpySequencer::SET_FLAG_cmdHandler(FwOpcodeType opCode,  //!< The opcode
+                                       U32 cmdSeq,           //!< The command sequence number
+                                       Svc::Fpy::FlagId flag,
+                                       bool value) {
+    if (!this->isRunningState(this->sequencer_getState())) {
```
LeStarch, line 227: Why?  The comment discusses breaking, which is not this.  Copy-paste error, or is the intent to lock-out flags on "not running"

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -200,6 +200,31 @@ void FpySequencer::DEBUG_CLEAR_BREAKPOINT_cmdHandler(FwOpcodeType opCode,  //!<
...
+        // can only break while running
+        this->log_WARNING_HI_InvalidCommand(static_cast<I32>(sequencer_getState()));
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
+        return;
+    }
+
+    // this is a sanity check, we shouldn't even get here if this isn't true
+    // because the enum should check for validity and raise a format err if not valid.
+    // actually what this really catches is an incorrect FLAG_COUNT value
+    FW_ASSERT(static_cast<I32>(flag.e) < Fpy::FLAG_COUNT, static_cast<FwAssertArgType>(flag.e));
```
LeStarch, line 237: Why use an I32 here, I see 8-bits below.

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -31,6 +33,7 @@ static_assert(Svc::Fpy::MAX_STACK_SIZE >= FW_TLM_BUFFER_MAX_SIZE,
               "Max stack size must be greater than max tlm buffer size");
 static_assert(Svc::Fpy::MAX_STACK_SIZE >= FW_PARAM_BUFFER_MAX_SIZE,
               "Max stack size must be greater than max prm buffer size");
+static_assert(Svc::Fpy::FLAG_COUNT < std::numeric_limits<U8>::max(), "Flag count must be less than U8 max");
```
LeStarch, line 36: USed I32 abvoce, not U8???

Svc/FpySequencer/FpySequencerTypes.fpp
```diff
@@ -3,6 +3,16 @@ module Svc {
         @ the current schema version (must be representable in U8)
         constant SCHEMA_VERSION = 2;
 
+        @ the number of runtime configurable flags. flags modify the sequencer behavior and can be set by the sequence
+        # should be equal to (last flag id) + 1
+        constant FLAG_COUNT = 1
+
+        enum FlagId {
```
LeStarch: Type = U8?  e.g. : U8

LeStarch (changes requested):
Additionally some comments to answer.

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: ah, yes, the comment is wrong. however, the code is right:  we can only set the value of a flag while running a sequence. flags are only valid while running a sequence, so it wouldn't make sense to set the value of a flag when not running a sequence. if you want to modify the value of a flag before running a sequence, you can use the FLAG_DEFAULT parameters. let me know if this makes sense

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: the args to < will be converted to 32 bit types implicitly, this just makes it explicit.

Svc/FpySequencer/FpySequencerTypes.fpp
zimri-leisher: good call

zimri-leisher:
@LeStarch feedback responded to

LeStarch (approved):

PR #4264, opened 2025-10-06
Add Communication Aggregator Component

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Documentation Included (y/n)_**|y  |

---
## Change Description

This adds an aggregator for communications.  It is by-default added to CCSDS to cut down on Idle packets fills.

## Rationale

CCSDS has a fixed-size frame, which is only reasonably well utilized when passed aggregated frames.

CI.

## Future Work

This would be a great candidate for rules-based testing.  I structured the test in a similar way.

Autocomplete, and GPT for help with requirements.   All reviewed by contributor.


Fw/Buffer/Buffer.hpp
```diff
@@ -47,6 +47,15 @@ namespace Fw {
 class Buffer : public Fw::Serializable {
     friend class Fw::BufferTester;
 
+  public:
+    //! Buffer ownership state
+    //!
+    //! A convenience enumeration to help users implement ownership tracking of buffers.
+    enum class OwnershipState {
+        NOT_OWNED,  //!< The buffer is currently not owned by the TmFramer
+        OWNED,      //!< The buffer is currently owned by the TmFramer
```
thomas-bc:
```suggestion
        NOT_OWNED,  //!< The buffer is currently not owned
        OWNED,      //!< The buffer is currently owned
```

Svc/ComAggregator/docs/sdd.md
```diff
@@ -0,0 +1,22 @@
+# Svc::ComAggregator
```
thomas-bc, line 1: Maybe a nit, but is there a different meaning in your opinion between `Aggregator` and `Accumulator` ? Seems like they could potentially be used interchangeably, and we've used Accumulator in the past so we could keep the same terminology for consistency ?
But maybe FrameAccumulator accumulates **something** until it has a frame, where ComAggregator aggregates **Com** until it has something, which would be the difference? idk :shrug:

Svc/ComAggregator/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+# root when not specifically overridden by the developer. i.e. The module defined by
+# `Ref/SignalGen/CMakeLists.txt` will be named `Ref_SignalGen`.
+
+register_fprime_library(
+    AUTOCODER_INPUTS
+        "${CMAKE_CURRENT_LIST_DIR}/ComAggregator.fpp"
+    SOURCES
+        "${CMAKE_CURRENT_LIST_DIR}/ComAggregator.cpp"
+#   DEPENDS
+#       MyPackage_MyOtherModule
```
thomas-bc:
```suggestion
```

Svc/ComAggregator/docs/sdd.md
```diff
@@ -0,0 +1,22 @@
...
+
+Aggregates buffers in the downlink chain. This is for use with systems that have fixed size frames (e.g. CCSDS TM) that needed internal aggregation.
+
+## Requirements
+
+| ID                    | Description                                                                                                                                                   | Verification |
+| --------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------ |
+| Svc-ComAggregator-001 | ComAggregator shall accept incoming downlink data as `Fw::Buffer`, `ComCfg::FrameContext` pairs and append the buffer into the aggregate space permitting     | Unit-Test    |
+| Svc-ComAggregator-002 | ComAggregator shall hold the incoming buffer when there is insufficient space in the aggregate buffer.                                                        | Unit-Test    |
+| Svc-ComAggregator-003 | ComAggregator shall send the current aggregate buffer when the incoming buffer is held dur to overflow.                                                       | Unit-Test    |
```
thomas-bc:
```suggestion
| Svc-ComAggregator-003 | ComAggregator shall send the current aggregate buffer when the incoming buffer is held due to overflow.                                                       | Unit-Test    |
```

Svc/ComAggregator/docs/sdd.md
```diff
@@ -0,0 +1,22 @@
...
+| Svc-ComAggregator-004 | ComAggregator shall send the current aggregate buffer when it receives a timeout trigger if and only if the aggregate is non-empty.                           | Unit-Test    |
+| Svc-ComAggregator-005 | ComAggregator shall clear aggregation state when a SUCCESS communication status is received back.                                                             | Unit-Test    |
+| Svc-ComAggregator-006 | ComAggregator shall preserve the order of received buffers when forming each aggregate and across aggregate sends.                                            | Unit-Test    |
+| Svc-ComAggregator-007 | ComAggregator shall inter operate with the [Communication Adapter Interface comStatus protocol](https://fprime.jpl.nasa.gov/latest/docs/reference/communication-adapter-interface/) | Unit-Test    |
```
thomas-bc:
```suggestion
| Svc-ComAggregator-007 | ComAggregator shall inter operate with the [Communication Adapter Interface comStatus protocol](../../../docs/reference/communication-adapter-interface.md)   | Unit-Test    |
```

Svc/ComAggregator/ComAggregator.hpp
```diff
@@ -0,0 +1,146 @@
...
+    //! Implementation for guard isGood of state machine Svc_AggregationMachine
+    //!
+    //! Check if last status is good
+    bool Svc_AggregationMachine_guard_isGood(SmId smId,                              //!< The state machine id
+                                             Svc_AggregationMachine::Signal signal,  //!< The signal
+                                             const Fw::Success& value                //!< The value
+    ) const override;
+
+  private:
+    U8 m_frameBufferStore[ComCfg::AggregationSize];  //!< Buffer to hold the frame data
```
thomas-bc, line 134: not required for this PR, but should this be a `configure(FwSizeType size)` configurable value instead ? For example what if this is needed twice with different aggregation sizes ? Otherwise is this really a `SpacePacketAggregator` instead?

Svc/ComAggregator/ComAggregator.cpp
```diff
@@ -0,0 +1,121 @@
...
+        this->m_bufferState = Fw::Buffer::OwnershipState::NOT_OWNED;
+        this->m_frameBuffer.setSize(this->m_frameSerializer.getBuffLength());
+        this->dataOut_out(0, this->m_frameBuffer, this->m_lastContext);
+    }
+}
+
+void ComAggregator ::Svc_AggregationMachine_action_doHold(SmId smId,
+                                                          Svc_AggregationMachine::Signal signal,
+                                                          const Svc::ComDataContextPair& value) {
+    FW_ASSERT(not this->m_held.get_data().isValid());
```
thomas-bc, line 89: So does this crash if a second packet comes in before the comStatus ready comes in? Should it not attempt to fill, and crash (or just drop?) if the aggregation is full instead?

Svc/ComAggregator/ComAggregator.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// ----------------------------------------------------------------------
+// Implementations for internal state machine actions
+// ----------------------------------------------------------------------
+
+void ComAggregator ::Svc_AggregationMachine_action_doClear(SmId smId, Svc_AggregationMachine::Signal signal) {
+    this->m_frameSerializer.resetSer();
+    this->m_frameBuffer.setSize(sizeof(this->m_frameBufferStore));
+    Fw::Success good = Fw::Success::SUCCESS;
```
thomas-bc: `good` unused ?

Svc/ComAggregator/ComAggregator.cpp
```diff
@@ -0,0 +1,121 @@
...
+    this->dataReturnOut_out(0, const_cast<Fw::Buffer&>(value.get_data()), value.get_context());
+    this->comStatusOut_out(0, good);
+}
+
+void ComAggregator ::Svc_AggregationMachine_action_doSend(SmId smId, Svc_AggregationMachine::Signal signal) {
+    // Send only when the buffer will be valid
+    if (this->m_frameSerializer.getBuffLength() > 0) {
+        this->m_bufferState = Fw::Buffer::OwnershipState::NOT_OWNED;
+        this->m_frameBuffer.setSize(this->m_frameSerializer.getBuffLength());
+        this->dataOut_out(0, this->m_frameBuffer, this->m_lastContext);
```
thomas-bc, line 82: This `m_lastContext` will contain values that aren't quite valid for this aggregated frame. E.g. APID will be that of the last packet, but the frame may contain many packets with different APIDs. Would it make sense to create a new context object and fill out only the values that are valid ? Not a big deal because nothing downstream should use these values, but you never know... Or maybe worth just calling it out in the SDD.

thomas-bc (commented):
Few notes and questions, but overall looks great! A much needed improvement!

Svc/ComAggregator/ComAggregator.cpp
thomas-bc: Discussed offline, this is guaranteed not to be hit by the ComQueue/ComInterface protocol. Could be worth calling out in a comment in the state machine?

Svc/ComAggregator/docs/sdd.md
LeStarch: Accumulation is for uplink to assemble a frame, Aggregation is on downlink for a set of frames.  We should not use the same name, as they operation differently.

Svc/ComAggregator/ComAggregator.hpp
LeStarch: Agreed.   Should be a MemAllocator pattern too.

LeStarch: https://github.com/nasa/fprime/issues/4300

Svc/ComAggregator/ComAggregator.cpp
LeStarch: Fixed SDD, added issue to track better pattern: https://github.com/nasa/fprime/issues/4301

thomas-bc (dismissed):

thomas-bc (approved):

PR #4267, opened 2025-10-06
Add ErrorNotify port to CCSDS deframers

### LINKED ISSUE #4175: Add error notification port for CCSDS deframer
| | |
|:---|:---|
|**_F´ Version_**|4.x |
|**_Affected Component_**| https://github.com/nasa/fprime/blob/devel/Svc/Ccsds/SpacePacketDeframer/SpacePacketDeframer.fpp |
---
## Feature Description

A notification interface for errors encountered during deframing.

## Rationale

If there are errors during deframing, the deframer "fails silent," i.e. it doesn't inform any other software of the error. It emits counters and events, but that isn't the best way to inform other software components. For a one-way uplink via a radio, fail silent probably makes sense, but if CCSDS is used for a payload interface that expects an error response for framing errors, there is no way to do this. This update would add a notification port so that other project-specific components can react to deframing errors. It could report status like:
1. Checksum error
2. Size error
3. Token error

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/4175 |
|**_Generative AI was used in this contribution (y/n)_**| autocomplete |

---
## Change Description

Fixes https://github.com/nasa/fprime/issues/4175

Adds optional ports to CCSDS deframers for other components to receive notifications when a deframing error occurs.


Svc/Ccsds/Types/Types.fpp
```diff
@@ -1,10 +1,15 @@
...
-# ------------------------------------------------ 
-
 module Svc {
 module Ccsds {
 
+    @ Enum representing an error during framing/deframing in the CCSDS protocols
+    enum FrameError: U8 {
+        SP_INVALID_LENGTH = 0
+        TC_INVALID_SCID = 1
+        TC_INVALID_LENGTH = 2
```
thomas-bc, line 8: There's two `INVALID_LENGTH`, one per protocol. Because knowing which protocol failed may be useful information. And I didn't want to have one enum (and port type) per protocol, especially given that SpacePacket only has one failure mode.
Thoughts?

Svc/Ccsds/SpacePacketDeframer/SpacePacketDeframer.cpp
```diff
@@ -52,6 +52,9 @@ void SpacePacketDeframer ::dataIn_handler(FwIndexType portNum, Fw::Buffer& data,
     if (pkt_length > data.getSize() - SpacePacketHeader::SERIALIZED_SIZE) {
         FwSizeType maxDataAvailable = data.getSize() - SpacePacketHeader::SERIALIZED_SIZE;
         this->log_WARNING_HI_InvalidLength(pkt_length, maxDataAvailable);
+        if (this->isConnected_errorNotify_OutputPort(0)) {
+            this->errorNotify_out(0, Svc::Ccsds::FrameError::SP_INVALID_LENGTH);
```
thomas-bc, line 56: Could passing the `data, context` pair be useful as well? Maybe... but need to keep in mind that (1) it's likely corrupted and (2) it's about to be deallocated in the next line

LeStarch (approved):

PR #4269, opened 2025-10-06
Add new Fw::StringBase type for strings backed by immutable string literals

ianbrault:
| | |
|:---|:---|

---
## Change Description

Provides a new `Fw::StaticString` type which is a string backed by a `const char*` string literal. This mirrors the design of `Fw::ExternalString` but this type can be initialized with a `const char*` instead of `Fw::ExternalString` which requires a mutable `char*` pointer.

## Rationale

This was devised during work to optimize string-handling critical sections and reduce data copying. We needed a string type that could wrap a string literal without requiring internal storage or copying. `Fw::ExternalString` nearly fit the bill but could not be initialized with constant string literals. This new type fills that gap.

Use in place of any `Fw::StringBase` type. Example usage:
```c++
const char* FOO = "fprime_test_string";
// ...
Fw::StaticString foo(FOO, Fw::StringUtils::string_length(FOO, FW_LOG_STRING_MAX_SIZE));
this->log_ACTIVITY_LO_Foo(foo);
```

## Future Work

Only marginally-related here but wanted to note that, for a string type like this which should have a constant length, I was not able to figure out a way to override the `length` function for use in serializing. Instead, the serialization code is forced to use `StringBase::length` which uses `Fw::StringUtils::string_length` and it becomes an O(n) operation when it could be O(1).

N/A


Fw/Types/StaticString.hpp
```diff
@@ -0,0 +1,59 @@
...
+    // ----------------------------------------------------------------------
+
+    //! Constructor (uninitialized buffer)
+    StaticString() : StringBase(), m_bufferPtr(nullptr), m_bufferSize(0) {}
+
+    //! Constructor (bufferPtr and bufferSize)
+    StaticString(const char* bufferPtr,           //!< The buffer pointer
+                 StringBase::SizeType bufferSize  //!< The buffer size
+                 )
+        : StringBase(), m_bufferPtr(bufferPtr), m_bufferSize(bufferSize + 1) {}
```
celskeggs: We need to make clear in a comment that this class cannot be used in all cases where a StringBase can be used. This is because StringBase will sometimes cast the result of `toChar()` to a mutable pointer, and the caller may have provided a string that cannot safely be modified.

celskeggs (changes requested):

celskeggs (approved):

LeStarch:
@bocchino, will you review? If needed, we can discuss at the CCB on Friday.

bocchino (changes requested):
We cannot initialize an object of type `Fw::StringBase` with an object of type `const char*`. The design of `Fw::StringBase` assumes that the underlying string buffer is mutable.

> We needed a string type that could wrap a string literal without requiring internal storage or copying.

If there is a need for this, then I recommend the following:

1.  Factor the read-only operations of `Fw::StringBase` into a new abstract superclass `Fw::ConstStringBase`.
2. Make a `Fw::ConstExternalString` that inherits from `Fw::ConstStringBase` and can be initialized with a `const char*`.

celskeggs (approved):

ianbrault:
Ready for further review @LeStarch @bocchino

Fw/Types/ConstExternalString.hpp
```diff
@@ -0,0 +1,59 @@
...
+    // ----------------------------------------------------------------------
+
+    //! Constructor (uninitialized buffer)
+    ConstExternalString() : ConstStringBase(), m_bufferPtr(nullptr), m_bufferSize(0) {}
+
+    //! Constructor (bufferPtr and bufferSize)
+    ConstExternalString(const char* bufferPtr,                //!< The buffer pointer
+                        ConstStringBase::SizeType bufferSize  //!< The buffer size
+                        )
+        : ConstStringBase(), m_bufferPtr(bufferPtr), m_bufferSize(bufferSize + 1) {}
```
bocchino:
```suggestion
        : ConstStringBase(), m_bufferPtr(bufferPtr), m_bufferSize(bufferSize) {}
```
I think we should not add one to the buffer size here.

Fw/Types/StringBase.cpp
```diff
@@ -18,21 +18,34 @@
 
 namespace Fw {
 
-StringBase::StringBase() {}
+// ConstStringBase interfaces
 
-StringBase::~StringBase() {}
+ConstStringBase::ConstStringBase() {}
```
bocchino: I think it would be clearer if we moved the `ConstStringBase` implementation to `ConstStringBase.cpp`.

Fw/Types/StringBase.hpp
```diff
@@ -22,15 +23,20 @@
 
 namespace Fw {
 
-class StringBase : public Serializable {
+//! A read-only abstract superclass for StringBase
+class ConstStringBase : public Serializable {
```
bocchino: I think `ConstStringBase` should have its own header file `ConstStringBase.hpp`.

Svc/BufferLogger/BufferLogger.hpp
```diff
@@ -76,8 +76,8 @@ class BufferLogger final : public BufferLoggerComponentBase {
         );
 
         //! Set base file name
-        void setBaseName(
-            const Fw::StringBase& baseName  //!< The base file name; used with prefix, unique counter value, and suffix
+        void setBaseName(const Fw::ConstStringBase&
```
bocchino, line 79: Formatting

bocchino (commented):
Looks good! Just a few small comments.

Fw/Types/StringBase.cpp
```diff
@@ -18,21 +18,34 @@
...
 
-const CHAR* StringBase::operator+=(const CHAR* src) {
-    this->appendBuff(src, static_cast<SizeType>(StringUtils::string_length(src, this->getCapacity())));
-    return this->toChar();
+ConstStringBase::~ConstStringBase() {}
+
+ConstStringBase::SizeType ConstStringBase::length() const {
+    // The length of the string should be 1 less than its capacity (string + null terminator)
```
bocchino: Given the current implementation, I don't think we can assume that `capacity == length + 1`. If we want to enforce this assumption, then we should assert in the constructor that all the bytes of the buffer except the last one are nonzero, and the last one is zero. If we don't do this, then this implementation of `length` is dangerous. If we do assume this, then I think `length` can just return `maxLength`.

ianbrault: What do you think should be done here @bocchino ? Personally, that seems like a strict assumption to make in `ConstStringBase`, perhaps `length` can be left pure virtual and that assumption can be made for `ConstExternalString` instead?

bocchino: Can we do the following:
1. Move the standard implementation of `length` (search for zero and count) to `ConstStringBase` and make it virtual.
2. Make `StringBase` inherit that implementation.
3. In `ConstExternalString`
    a. In the constructor, assert that `length == buffer size + 1`. We can use the superclass `length` implementation for this.
    b. Override the `length` implementation to return `buffer size - 1`

ianbrault: @bocchino updated in latest commit

Fw/Types/ConstExternalString.hpp
```diff
@@ -0,0 +1,69 @@
...
+    //! Constructor (uninitialized buffer)
+    ConstExternalString() : ConstStringBase(), m_bufferPtr(nullptr), m_bufferSize(0) {}
+
+    //! Constructor (bufferPtr and bufferSize)
+    ConstExternalString(const char* bufferPtr,                //!< The buffer pointer
+                        ConstStringBase::SizeType bufferSize  //!< The buffer size
+                        )
+        : ConstStringBase(), m_bufferPtr(bufferPtr), m_bufferSize(bufferSize) {
+        // Assert that the string length plus the null terminator fills the buffer
+        FW_ASSERT(ConstStringBase::length() + 1 == bufferSize);
```
bocchino:
```suggestion
        FW_ASSERT(bufferPtr != nullptr);
        FW_ASSERT(bufferSize > 0, static_cast<FwAssertArgType>(bufferSize));
        // Assert that the string length plus the null terminator fills the buffer
        FW_ASSERT(ConstStringBase::length() + 1 == bufferSize, static_cast<FwAssertArgType>(ConstStringBase::length() + 1), static_cast<FwAssertArgType>(bufferSize));
```

Fw/Types/ConstExternalString.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+  public:
+    // ----------------------------------------------------------------------
+    // ConstStringBase interface
+    // ----------------------------------------------------------------------
+
+    //! Get the length of the string
+    ConstStringBase::SizeType length() const override {
+        // The length of the string is 1 less than its capacity (string + null terminator)
+        return this->m_bufferSize - 1;
```
bocchino:
```suggestion
        return this->m_bufferSize == 0 ? 0 : this->m_bufferSize - 1;
```

bocchino (commented):
The coding looks good to me. Please see my latest comments. It would be good to add unit tests for `ConstExternalString`.

ianbrault:
@bocchino pushed some basic UTs for `ConstExternalString` in my latest commit, I also modified the implementation of `ConstStringBase::maxLength` in accordance with the change that you suggested above for the `ConstExternalString::length` implementation, it can assert for a zero-capacity string which is possible for `ConstExternalString` and was previously possible with `ExternalString` as well

bocchino:
> I also modified the implementation of ConstStringBase::maxLength in accordance with the change that you suggested above for the ConstExternalString::length implementation

Sounds good! I made the `length` implementation in `ConstStringBase` conform to this as well. @LeStarch agrees that not asserting here is good.

bocchino (dismissed):
Looks good! All comments have been addressed. Great work, @ianbrault!

ianbrault:
Thanks @bocchino! I appreciate all the help

LeStarch (dismissed):

thomas-bc (approved):

PR #4287, opened 2025-10-09
Bump FPP with unified tool

Kronos3:
This PR bumps FPP the new alpha release which unifies the FPP executable tools into a single binary. It greatly decreases the package size so we now only need a single pip package rather than a package for each FPP tool.


thomas-bc (approved):
🚀🚀🚀

Issue #4084, opened 2025-08-27
Create a new parent interface class that SerializeBufferBase and CircularBuffer can implement/extend


vincewoo:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**| SerializeBufferBase |
---
## Feature Description

Part of the effort to complete #3429 

We should make `SerializeBufferBase` an abstract parent class that only contains the common functions for serializable buffers, both linear and non-linear. A new `LinearBufferBase` class will be created as a child of `SerializeBufferBase` that extends the common functionality to add functions that are specific to linear serializable buffers.

We will need to evaluate the F' code base to see where we need to use the new `LinearBufferBase` directly or leave as `SerailizeBufferBase`.

## Rationale

The original request suggested introducing a new `SerializeBufferInterface`. This would be a very disruptive change as any references in F' to `SerializeBufferBase` would then have to be updated to `SerializeBufferInterface`.

thomas-bc:
Merged

PR #4288, opened 2025-10-09
Created new SerialBufferBase as a parent of SerializeBufferBase (now renamed LinearBufferBase). 

vincewoo:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4084 |
|**_Has Unit Tests (y/n)_**| y, existing |

---
## Change Description

A pre-cursor to the work for #3429 

This change introduces a parent class to SerializeBufferBase, named tentatively as SerialBufferBase. The intent is to have SerialBufferBase be the parent of both SerializeBufferBase (for linear buffers) and CircularBuffer.

As part of this change we will be improving the naming of some of the members of SerializeBufferBase that were confusing:

```c
    DEPRECATED(Serializable::SizeType getBuffCapacity() const, "Use getCapacity() instead");
    DEPRECATED(Serializable::SizeType getBuffLength() const, "Use getSize() instead"); 
    DEPRECATED(Serializable::SizeType getBuffLeft(), "Use getDeserializeSizeLeft() instead");

    virtual Serializable::SizeType getCapacity() const = 0;     //!< returns capacity, not current size, of buffer
    Serializable::SizeType getSize() const;                     //!< returns current buffer size
    Serializable::SizeType getDeserializeSizeLeft() const;      //!< returns how much deserialization buffer is left
    Serializable::SizeType getSerializeSizeLeft() const;        //!< returns how much serialization space is left
```

## Rationale

Having a common parent allows CircularBuffer to be used in a more standard way in the future and unlock the ability to use it interchangeably with use-cases where currently only linear buffers can be passed.

This change has touched 100+ files but they are almost entirely due to the renaming of methods:

1. getBuffCapacity --> getCapacity
2. getBuffLength --> getSize
3. getBuffLeft --> getDeserializeSizeLeft

The reviewers should focus on Serializable.hpp and Serializable.cpp where the bulk of the refactoring is. 

This change will also require a refactoring of auto-coded classes in FPP. I have done tandem development of the FPP change and have that PR out on the FPP project here: https://github.com/nasa/fpp/pull/850

## Future Work

Follow-up work captured by Michael in design/review meeting on 10/21.

AI assistant was used to generate docstrings for Serializable.hpp. All generated documentation was reviewed thoroughly for correctness.

Systems used was the JPL ChatHPC and VSCode + Cline


timcanham:
Is there an alternate way to do this now? I think I used it on a previous project to copy a serialized buffer to a new one while reserving space at the beginning of the new one.

vincewoo:
> Is there an alternate way to do this now? I think I used it on a previous project to copy a serialized buffer to a new one while reserving space at the beginning of the new one.

Thanks @timcanham we were hoping to sus out the use-case for this function. Would you happen to have a copy of the code so I can see how it is being used?

vincewoo:
In the case of copyRaw, I see two uses of it within the fprime project:

https://github.com/nasa/fprime/blob/e1098018e234ce819964cd831a7dc258f35afb33/Fw/Cmd/CmdPacket.cpp#L43-L47

https://github.com/nasa/fprime/blob/e1098018e234ce819964cd831a7dc258f35afb33/Fw/Dp/DpContainer.cpp#L206-L218

The `copyRaw` function is different than the `copyRawOffset` function in that `copyRaw` resets the destination buffer and copies the data. Note that both `copyRaw*` functions advances the deserialization pointer.

LeStarch:
@vincewoo isn't this equivalent to:

```
buffer.serializeTo(args_buffer);
```

Fw/Types/Serializable.hpp
```diff
@@ -40,154 +45,1248 @@
...
+    DEPRECATED(Serializable::SizeType getBuffLeft(), "Use getDeserializeSizeLeft() instead");
+
+    //! \brief Get buffer capacity
+    //!
+    //! This method returns the total capacity of the buffer, which is the maximum
+    //! amount of data that can be stored in the buffer. This is not the same as
+    //! the current size, which indicates how much data is currently in the buffer.
+    //!
+    //! \return The capacity of the buffer in bytes
+    virtual Serializable::SizeType getCapacity() const override = 0;
```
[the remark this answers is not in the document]
vincewoo: Fixed this!

vincewoo:
@LeStarch I've made the changes that were requested from our meeting. This is ready for your review now.

Fw/Dp/DpContainer.cpp
```diff
@@ -212,7 +212,7 @@ void DpContainer::setDataHash(Utils::HashBuffer hash) {
               static_cast<FwAssertArgType>(dataHashOffset + HASH_DIGEST_LENGTH),
               static_cast<FwAssertArgType>(bufferSize));
     ExternalSerializeBuffer serialBuffer(dataHashAddr, HASH_DIGEST_LENGTH);
-    hash.resetSer();
+    hash.resetDeser();
```
LeStarch, line 215: Was this a bug?

Fw/Dp/DpContainer.cpp
```diff
@@ -212,7 +212,7 @@ void DpContainer::setDataHash(Utils::HashBuffer hash) {
...
     const Fw::SerializeStatus status = hash.copyRaw(serialBuffer, HASH_DIGEST_LENGTH);
```
LeStarch, line 216: Why not do hash.serializeTo(serialiBuffer)?

Fw/Types/Serializable.hpp
```diff
@@ -40,154 +45,1218 @@ class Serializable {
...
+    //! Ensures proper cleanup of derived classes.
+    virtual ~SerialBufferBase();
+
+    // Serialization for built-in types
+
+    //! \brief Serialize an 8-bit unsigned integer value
+    //!
+    //! This method serializes a single 8-bit unsigned integer value into the buffer.
+    //! The endianness of the serialization can be controlled via the mode parameter.
+    //!
```
LeStarch, line 125: Maybe note that this is done for compatibility (endianness) as U8s don't have endianness.

Fw/Types/Serializable.hpp
```diff
@@ -40,154 +45,1218 @@ class Serializable {
...
+    //! \param val The 8-bit unsigned integer value to serialize
+    //! \param mode Endianness mode for serialization (default is Endianness::BIG)
+    //! \return SerializeStatus indicating the result of the operation
+    virtual SerializeStatus serializeFrom(U8 val, Endianness mode = Endianness::BIG) = 0;
+
+    //! \brief Serialize an 8-bit signed integer value
+    //!
+    //! This method serializes a single 8-bit signed integer value into the buffer.
+    //! The endianness of the serialization can be controlled via the mode parameter.
+    //!
```
LeStarch, line 138: same note.

Fw/Types/Serializable.hpp
```diff
@@ -40,154 +45,1218 @@ class Serializable {
...
+    //! \brief Copy raw bytes from the source (this) into a destination buffer and advance source offset
+    //!
+    //! Copies exactly `size` bytes starting at the current deserialization pointer of `this` into `dest`.
+    //! This operation does not prepend a length field and does not interpret the data.
+    //!
+    //! Preconditions:
+    //! - `size` bytes must remain in the source (`getDeserializeSizeLeft() >= size`).
+    //! - Destination must have sufficient capacity (`dest.getCapacity() >= size`).
+    //!
+    //! Postconditions on success:
```
LeStarch: Post-conditions I believe is what is wanted here.

Fw/Types/Serializable.hpp
```diff
@@ -40,154 +45,1218 @@ class Serializable {
...
+    //! \brief Append raw bytes to destination (no length) and advance source offset
+    //!
+    //! Appends exactly `size` bytes from the current deserialization pointer of `this` into `dest` using
+    //! `Serialization::OMIT_LENGTH`, preserving any existing bytes already serialized in `dest`.
+    //!
+    //! Preconditions:
+    //! - `size` bytes must remain in the source (`getDeserializeSizeLeft() >= size`).
+    //! - Destination must have space for the append (`dest.getCapacity() >= dest.getSize() + size`).
+    //!
+    //! Postconditions on success:
```
LeStarch: Post-conditions

Fw/Types/Serializable.hpp
```diff
@@ -40,154 +45,1218 @@ class Serializable {
...
+    //!         `FW_DESERIALIZE_SIZE_MISMATCH` if source does not contain `size` bytes remaining
+    virtual SerializeStatus copyRawOffset(SerialBufferBase& dest, Serializable::SizeType size) = 0;
+
+    //! \brief Reset serialization pointer to beginning of buffer
+    //!
+    //! This method resets the serialization pointer to the beginning of the buffer,
+    //! allowing the buffer to be reused for new serialization operations. Any
+    //! data that was previously serialized in the buffer will be overwritten.
+    //!
+    //! \return None
```
LeStarch: Return None is not needed, as this is c++

Fw/Types/Serializable.hpp
```diff
@@ -40,154 +45,1218 @@ class Serializable {
...
+    virtual void resetSer() = 0;
+
+    //! \brief Reset deserialization pointer to beginning of buffer
+    //!
+    //! This method resets the deserialization pointer to the beginning of the buffer,
+    //! allowing the buffer to be reused for new deserialization operations. The buffer
+    //! contents are not modified, but the pointer is reset to allow reading from the
+    //! start of the data.
+    //!
+    //! \return None
```
LeStarch: same

Fw/Types/Serializable.hpp
```diff
@@ -40,154 +45,1218 @@ class Serializable {
...
+    //! equal to the buffer's capacity.
+    //!
+    //! \return The current size of the buffer in bytes
+    virtual Serializable::SizeType getSize() const = 0;
+
+    //! \brief Get remaining deserialization buffer size
+    //!
+    //! This method returns the amount of data that remains to be deserialized
+    //! from the buffer. It indicates how much data is left starting from the
+    //! current deserialization pointer to the end of the buffer.
```
LeStarch: The the end of the valid size (returned by getSize())

Fw/Types/Serializable.hpp
```diff
@@ -40,154 +45,1218 @@ class Serializable {
...
 
     // Serialization for built-in types
 
-    SerializeStatus serializeFrom(U8 val, Endianness mode = Endianness::BIG);  //!< serialize 8-bit unsigned int
-    SerializeStatus serializeFrom(I8 val, Endianness mode = Endianness::BIG);  //!< serialize 8-bit signed int
+    //! \brief Serialize an 8-bit unsigned integer value
+    //!
+    //! This method serializes a single 8-bit unsigned integer value into the buffer.
+    //! The endianness of the serialization can be controlled via the mode parameter.
+    //!
```
LeStarch, line 694: Same changes as noted above.

LeStarch (changes requested):
Minor changes, but the most of it is good.

Fw/Dp/DpContainer.cpp
vincewoo: Yes, this is a bug. Since we are only deserializing from hash it doesn't make sense to reset serialization here.  What's strange is that changing this back to hash.resetSer() actually causes Fw_Dp_ut_exe and Svc_DpWriter_ut_exe UTs to fail, as I would expect. Looking at Git Blame it has been resetSer for the last year.

Fw/Dp/DpContainer.cpp
vincewoo: In this case yes, because serialBuffer was just initialized doing hash.deserializeTo(serialBuffer, HASH_DIGEST_LENGTH, Serialization::OMIT_LENGTH) would be equivalent to a copyRaw. If serialBuffer wasn't a fresh buffer though the functionality is actually closer to copyRawOffset which would append.

It's a bit awkwardly named but it is the same functionality as you suspected.

As an aside: If you do a serialBuffer.serializeFrom(hash, HASH_DIGEST_LENGTH, Serialization::OMIT_LENGTH) it would create the same buffer but the pointers for hash would not be left in the same place as copyRaw. This is an interesting discussion point on if these operations should be completely symmetrical, both source and dest pointers being updated. Currently I'm matching pre-existing behaviour.

Fw/Types/Serializable.hpp
vincewoo: Added Note

Fw/Types/Serializable.hpp
vincewoo: Added Note

Fw/Types/Serializable.hpp
vincewoo: Corrected

Fw/Types/Serializable.hpp
vincewoo: Corrected

Fw/Types/Serializable.hpp
vincewoo: Removed return

Fw/Types/Serializable.hpp
vincewoo: Removed return

Fw/Types/Serializable.hpp
vincewoo: Reworded

Fw/Types/Serializable.hpp
vincewoo: Added Note

vincewoo: Adding note to the below function for I8 as well

requirements.txt
```diff
@@ -20,7 +20,7 @@ Flask-Compress==1.15
 Flask-RESTful==0.3.10
 fprime-fpl-layout==1.0.3
 fprime-fpl-write-pic==1.0.3
-fprime-fpp==3.1.0a7
+fprime-fpp==v3.1.0a8
```
thomas-bc: Both work but for consistency...
```suggestion
fprime-fpp==3.1.0a8
```

thomas-bc (changes requested):

LeStarch (approved):

Os/Queue.hpp
```diff
@@ -248,7 +248,7 @@ class Queue final : public QueueInterface {
     //! \param priority: priority of the message
     //! \param blockType: BLOCKING to block for space or NONBLOCKING to return error when queue is full
     //! \return status of the send
-    Status send(const Fw::SerializeBufferBase& message, FwQueuePriorityType priority, BlockingType blockType);
+    Status send(const Fw::LinearBufferBase& message, FwQueuePriorityType priority, BlockingType blockType);
```
thomas-bc, line 251: Why require a `LinearBufferBase` here and not just `SerialBufferBase` ?

thomas-bc (approved):
LGTM! Just one question

bocchino (approved):
Looks good!

Issue #4182, opened 2025-09-22
Add Little-Endian Serialization Options


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Allow struct serialization to support little-endian format.

There are two options:
1. Mark the FPP `struct` as little-endian
2. Add a LITTLE_ENDIAN mode to the `serialization` and `deserialization` calls.

## Rationale

External interfaces to F Prime code often use little-endian format.  Users modeling these interfaces as FPP structs need to write custom serialization code to use these interfaces.  Allowing users to specify the mode would help F Prime bridge across these interfaces.

LeStarch:
@Willmac16

Willmac16:
I tried to end user implement this feature w/ fpp abstract serializable types that have custom `serializeTo` and `deserializeFrom` methods; however, it does not work due to the following:
* Command Args must be Displayable types and abstract types are not displayable (breaking my desired implementation of 1 fprime struct per external cmd w/ an fprime command that passes in the struct and handles the serialization though `serializeTo`/`serializeFrom`
* Enums can only work with base integer types (not including abstract ones). This is potentially non-blocking as most enums are U8

thomas-bc:
CCB: use `serializeTo(mode=LITTLE_ENDIAN)`

Willmac16:
I'm going to take a stab at this, but might drop off if I hit a Scala wall in the fpp tooling

bocchino:
> I'm going to take a stab at this, but might drop off if I hit a Scala wall in the fpp tooling

I don't see that any changes to FPP are required here. The proposed feature is a capability in the framework (add a mode parameter to serialization functions). If you provide a default argument `BIG_ENDIAN` to the mode parameter, then the generated C++ code will not need to change.

bocchino:
Instead of adding a mode argument, I wonder if it's better to add a parallel set of functions `serializeToBigEndian`. That seems to give better separation between this new feature and the core F Prime behavior, which is big endian. Also, this is a natural way to implement a mode argument:
```
serialize(arg, mode) {
  switch (mode) {
    case BIG_ENDIAN:
      serializeBigEndian(arg)
      break
    case LITTLE_ENDIAN:
      serializeLittleEndian(arg)
      break
  }
}
```
If we inline all the big endian and little endian behavior into the same functions, the code could get ugly.

Willmac16:
[...]

PR #4294, opened 2025-10-10
Little Endian Serialization

Willmac16:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/4182 |
|**_Has Unit Tests (y/n)_**|  y |

---
## Change Description

Adds Little Endian serialization of fpp data types

## Rationale

Autocoded serialization of complex fpp data types enables easy interfacing w/ external byte level streams/packets. Unfortunately some devices require little endian serialization, making fpp serialization not useful for these applications without these modification.

* Verify that all types with little-endian switched behavior make sense (mostly the swap on pointer types)
* Ensure calling of Unsigned Ser/Des functions from Signed Ser/Des functions agrees with library guarantees (I believe my implementation is closer to the verbiage in rule 1b of the Serialization of FPP Values section of the Writing-C-Plus-Plus-Implementations page of the User Guide).

Requires my accompanying PR against FPP
https://github.com/nasa/fpp/pull/842

### DIFF
--- a/Fw/Time/Time.hpp
+++ b/Fw/Time/Time.hpp
@@ -37,7 +37,7 @@ class Time : public Serializable {
         const;  // !< Time base of time. This is project specific and is meant for indicating different sources of time
     FwTimeContextStoreType getContext() const;  // !< get the context value
     SerializeStatus serializeTo(SerializeBufferBase& buffer,
-                                Fw::Endianness mode) const override;  // !< Serialize method
+                                Fw::Endianness mode = Fw::Endianness::BIG) const override;  // !< Serialize method
     SerializeStatus deserializeFrom(SerializeBufferBase& buffer,
                                     Fw::Endianness mode) override;  // !< Deserialize method
     bool operator==(const Time& other) const;
--- a/Fw/Time/Time.hpp
+++ b/Fw/Time/Time.hpp
@@ -35,9 +35,11 @@ class Time : public Serializable {
     U32 getUSeconds() const;  // !< Gets microseconds part of time
     TimeBase getTimeBase()
         const;  // !< Time base of time. This is project specific and is meant for indicating different sources of time
-    FwTimeContextStoreType getContext() const;                                // !< get the context value
-    SerializeStatus serializeTo(SerializeBufferBase& buffer, Fw::Endianness mode) const override;  // !< Serialize method
-    SerializeStatus deserializeFrom(SerializeBufferBase& buffer, Fw::Endianness mode) override;    // !< Deserialize method
+    FwTimeContextStoreType getContext() const;  // !< get the context value
+    SerializeStatus serializeTo(SerializeBufferBase& buffer,
+                                Fw::Endianness mode) const override;  // !< Serialize method
+    SerializeStatus deserializeFrom(SerializeBufferBase& buffer,
+                                    Fw::Endianness mode) override;  // !< Deserialize method
     bool operator==(const Time& other) const;
     bool operator!=(const Time& other) const;
     bool operator>(const Time& other) const;
--- a/Fw/Time/Time.hpp
+++ b/Fw/Time/Time.hpp
@@ -36,8 +36,8 @@ class Time : public Serializable {
     TimeBase getTimeBase()
         const;  // !< Time base of time. This is project specific and is meant for indicating different sources of time
     FwTimeContextStoreType getContext() const;                                // !< get the context value
-    SerializeStatus serializeTo(SerializeBufferBase& buffer, Fw::Serialization::Endianness mode) const override;  // !< Serialize method
-    SerializeStatus deserializeFrom(SerializeBufferBase& buffer, Fw::Serialization::Endianness mode) override;    // !< Deserialize method
+    SerializeStatus serializeTo(SerializeBufferBase& buffer, Fw::Endianness mode) const override;  // !< Serialize method
+    SerializeStatus deserializeFrom(SerializeBufferBase& buffer, Fw::Endianness mode) override;    // !< Deserialize method
     bool operator==(const Time& other) const;
     bool operator!=(const Time& other) const;
     bool operator>(const Time& other) const;
--- a/Fw/Time/Time.hpp
+++ b/Fw/Time/Time.hpp
@@ -39,7 +39,7 @@ class Time : public Serializable {
     SerializeStatus serializeTo(SerializeBufferBase& buffer,
                                 Fw::Endianness mode = Fw::Endianness::BIG) const override;  // !< Serialize method
     SerializeStatus deserializeFrom(SerializeBufferBase& buffer,
-                                    Fw::Endianness mode) override;  // !< Deserialize method
+                                    Fw::Endianness mode = Fw::Endianness::BIG) override;  // !< Deserialize method
     bool operator==(const Time& other) const;
     bool operator!=(const Time& other) const;
     bool operator>(const Time& other) const;
--- a/Fw/Time/Time.hpp
+++ b/Fw/Time/Time.hpp
@@ -36,8 +36,8 @@ class Time : public Serializable {
     TimeBase getTimeBase()
         const;  // !< Time base of time. This is project specific and is meant for indicating different sources of time
     FwTimeContextStoreType getContext() const;                                // !< get the context value
-    SerializeStatus serializeTo(SerializeBufferBase& buffer) const override;  // !< Serialize method
-    SerializeStatus deserializeFrom(SerializeBufferBase& buffer) override;    // !< Deserialize method
+    SerializeStatus serializeTo(SerializeBufferBase& buffer, Fw::Serialization::Endianness mode) const override;  // !< Serialize method
+    SerializeStatus deserializeFrom(SerializeBufferBase& buffer, Fw::Serialization::Endianness mode) override;    // !< Deserialize method
     bool operator==(const Time& other) const;
     bool operator!=(const Time& other) const;
     bool operator>(const Time& other) const;
--- a/Fw/Types/Serializable.hpp
+++ b/Fw/Types/Serializable.hpp
@@ -65,6 +65,11 @@ class Serialization {
         INCLUDE_LENGTH,  //!< Include length as first token in serialization
         OMIT_LENGTH      //!< Omit length from serialization
     };
+
+    enum Endianess {
+        BIG,   //!< Include length as first token in serialization
+        LITTLE //!< Omit length from serialization
+    };
 };
 
 class SerializeBufferBase {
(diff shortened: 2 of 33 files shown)


Fw/Types/Serializable.hpp
```diff
@@ -22,15 +22,28 @@ typedef enum {
...
+        INCLUDE_LENGTH,  //!< Include length as first token in serialization
+        OMIT_LENGTH      //!< Omit length from serialization
+    };
+
+    enum Endianness {
+        BIG,   //!< Include length as first token in serialization
+        LITTLE //!< Omit length from serialization
+    };
+};
+
```
bocchino:
```suggestion
struct Serialization {
    enum t {
        INCLUDE_LENGTH,  //!< Include length as first token in serialization
        OMIT_LENGTH      //!< Omit length from serialization
    };
}

//! Endianness for serialization
enum class Endianness {
    BIG,   //!< Big endian serialization
    LITTLE //!< Little endian serialization
};

```

bocchino: 1. Going forward, I suggest that we use enum classes instead of the `struct...enum...` pattern that was required in C++98 to get name qualification for enumerated constants.
2. I suggest that the names of the enumerators for endianness should be `Endianness::BIG` and `Endianness::LITTLE`.

requirements.txt
```diff
@@ -20,7 +20,7 @@ Flask-Compress==1.15
 Flask-RESTful==0.3.10
 fprime-fpl-layout==1.0.3
 fprime-fpl-write-pic==1.0.3
-fprime-fpp==3.1.0a3
+fprime-fpp==v3.1.0a3-2-gb5e4c7b23
```
bocchino: Please put the FPP version back to `v3.1.0a3` so we can run CI. If I understand what is being done in this PR, CI should pass with that FPP version.

Willmac16 (commented):
> Please put the FPP version back to `v3.1.0a3` so we can run CI. If I understand what is being done in this PR, CI should pass with that FPP version. 

I believe my changes in here will fail since the autocoded classes fail to generate serialization code that accepts the mode parameter. There might be a way to modify the c++ to prevent that, but my desire is that every fpp coded data type supports both endianness where it makes sense.

bocchino:
> I believe my changes in here will fail

You are right. I tried it. What happens is the compiler complains because the `serialize{To,From}` functions in the framework base class each have a new argument that is not present in the auto-generated code that overrides the functions. So this is an F Prime tandem development: https://github.com/nasa/fpp/wiki/F-Prime-and-FPP-Tandem-Development.

Let me do a little work to bring your two branches in sync and we can go from there.

bocchino:
One thing I noticed when working on the `check-cpp` tests in `nasa/fpp` is that this is a breaking change: Any hand-coded classes that derive from `Fw::Serializable` will have to be changed to support the new endianness argument. @LeStarch @thomas-bc FYI.

Willmac16:
> > I believe my changes in here will fail
> 
> You are right. I tried it. What happens is the compiler complains because the `serialize{To,From}` functions in the framework base class each have a new argument that is not present in the auto-generated code that overrides the functions. So this is an F Prime tandem development: https://github.com/nasa/fpp/wiki/F-Prime-and-FPP-Tandem-Development.
> 
> Let me do a little work to bring your two branches in sync and we can go from there.

I'm currently tweaking both repos to agree w/ the new enum class style. LMK if you want me to hold off on pushing that

bocchino:
> I'm currently tweaking both repos to agree w/ the new enum class style. LMK if you want me to hold off on pushing that

Go ahead and push it. I also made those changes but the changes should be consistent. So far I have the `check-cpp` tests working in my local copy of your FPP branch. It required some changes to handcoded `.hpp` files used by the tests. Note that those tests are not run automatically by CI; you have to run `./check-cpp` by hand in the `fpp-to-cpp/test` directory. Here are the files that I changed:
```
./alias/T.hpp
./array/include/T.hpp
./struct/include/T.hpp
./port/include/T.hpp
```
I had to add the endianness argument to the overloading functions.

bocchino:
We have to get `check-cpp` working in the FPP repo, and we have to get all the unit tests to pass in the F Prime repo, including the tests in `FppTestProject`. Then we can make an alpha release of FPP to submit to F Prime CI (as it says in the notes on tandem development).

Willmac16:
> One thing I noticed when working on the `check-cpp` tests in `nasa/fpp` is that this is a breaking change: Any hand-coded classes that derive from `Fw::Serializable` will have to be changed to support the new endianness argument. @LeStarch @thomas-bc FYI.

I'm happy to try any suggestions; however, from what I tried initially, the behavior with the default parameter prevented things from working with unmodified custom serialization logic.

Maybe both a virtual 

> > I'm currently tweaking both repos to agree w/ the new enum class style. LMK if you want me to hold off on pushing that
> 
> Go ahead and push it. I also made those changes but the changes should be consistent. So far I have the `check-cpp` tests working in my local copy of your FPP branch. It required some changes to handcoded `.hpp` files used by the tests. Note that those tests are not run automatically by CI; you have to run `./check-cpp` by hand in the `fpp-to-cpp/test` directory. Here are the files that I changed:
> 
> ```
> ./alias/T.hpp
> ./array/include/T.hpp
> ./struct/include/T.hpp
> ./port/include/T.hpp
> ```
> 
> I had to add the endianness argument to the overloading functions.

Is that script covered by the top level `./test` command in `fpp/compiler`?

bocchino:
> Is that script covered by the top level ./test command in fpp/compiler?

No, you have to set the `FPRIME` environment variable to your local `fprime` directory, and you have to run `./check-cpp` manually in `fpp-to-cpp/test`. The top-level `./test` just runs all the CI tests.

Willmac16:
Could we implement a top level implementation of the Endianness parameter with a default value that calls the regular serialize function if Big & asserts if Little?

It didn't seem to me like a version with and without the endianness parameter would compile, but that might have been due to me leaving one of the implementations as virtual|

Edit:
e.g. both
```cpp
     virtual SerializeStatus serializeTo(SerializeBufferBase& buffer) const = 0;  //!< serialize contents to buffer
     
     SerializeStatus serializeTo(SerializeBufferBase& buffer, Endianness mode = Endianness::BIG) const {
         FW_ASSERT(mode == Endainness::BIG);
         serializeTo(buffer);
      };  //!< serialize contents to buffer
```

bocchino:
I don't see a way to keep backwards compatibility without a lot more complicated code generation.

Let me know when you've pushed all your changes. I'll verify that it works locally, and then I can make an FPP alpha release.

Willmac16:
I believe all my changes are up and all the tests passSent from my Apple WatchOn Oct 10, 2025, at 19:38, Rob Bocchino ***@***.***> wrote:bocchino left a comment (nasa/fprime#4294)
I don't see a way to keep backwards compatibility without a lot more complicated code generation.
Let me know when you've pushed all your changes. I'll verify that it works locally, and then I can make an FPP alpha release.

requirements.txt
bocchino: We will bump the FPP version to `v3.1.0a4`.

bocchino:
@LeStarch @thomas-bc Everything is working except fprime-zephyr. Someone with write permission to that repo needs to update it on a branch `pr-4294`. The file that needs to be updated is `fprime-zephyr/Os/RawTime.hpp`. It needs the new `mode` argument with the default big endianness.

bocchino (dismissed):
Looks good to me, once `fprime-zephyr` is updated.

bocchino:
Here is a proposed patch for that file. Note that the original file has Windows line terminators in it.
```
diff --git a/fprime-zephyr/Os/RawTime.hpp b/fprime-zephyr/Os/RawTime.hpp
index 4dbea49..11073d9 100644
--- a/fprime-zephyr/Os/RawTime.hpp
+++ b/fprime-zephyr/Os/RawTime.hpp
@@ -5,19 +5,17 @@
 #ifndef OS_ZEPHYR_RAWTIME_HPP
 #define OS_ZEPHYR_RAWTIME_HPP
 
-#include <zephyr/kernel.h>     
+#include <zephyr/kernel.h>
 #include <Os/RawTime.hpp>
 
-
-
 namespace Os {
 namespace Zephyr {
 namespace RawTime {
 
 // Zephyr provides a POSIX-compatible struct timespec
 struct ZephyrRawTimeHandle : public RawTimeHandle {
-        U32 m_seconds;
-        U32 m_micros;
+    U32 m_seconds;
+    U32 m_micros;
 };
 
 //! \brief Zephyr implementation of Os::RawTime
@@ -26,7 +24,6 @@ struct ZephyrRawTimeHandle : public RawTimeHandle {
 //!
 class ZephyrRawTime : public RawTimeInterface {
   public:
-
     //! \brief constructor
     ZephyrRawTime() = default;
 
@@ -61,16 +58,18 @@ class ZephyrRawTime : public RawTimeInterface {
     //! \brief Serialize the contents of the RawTimeInterface object into a buffer.
     //!
     //! This function serializes the contents of the RawTimeInterface object into the provided
-    //! buffer. 
+    //! buffer.
     //!
     //! \note The serialization must fit within `FW_RAW_TIME_SERIALIZATION_MAX_SIZE` bytes. This value is
     //! defined in FpConfig.h. For example, Posix systems use a pair of U32 (sec, nanosec) and can therefore
-    //! serialize in 8 bytes. Should an OSAL implementation require more than this, the project must increase 
+    //! serialize in 8 bytes. Should an OSAL implementation require more than this, the project must increase
     //! that value in its config/ folder.
     //!
     //! \param buffer The buffer to serialize the contents into.
+    //! \param mode The endianness mode.
     //! \return Fw::SerializeStatus indicating the result of the serialization.
-    Fw::SerializeStatus serializeTo(Fw::SerializeBufferBase& buffer) const override;
+    Fw::SerializeStatus serializeTo(Fw::SerializeBufferBase& buffer,
+                                    Fw::Endianness mode = Fw::Endianness::BIG) const override;
 
     //! \brief Deserialize the contents of the RawTimeInterface object from a buffer.
     //!
@@ -79,13 +78,14 @@ class ZephyrRawTime : public RawTimeInterface {
     //!
     //! \note The serialization must fit within `FW_RAW_TIME_SERIALIZATION_MAX_SIZE` bytes. This value is
     //! defined in FpConfig.h. For example, Posix systems use a pair of U32 (sec, nanosec) and can therefore
-    //! serialize in 8 bytes. Should an OSAL implementation require more than this, the project must increase 
+    //! serialize in 8 bytes. Should an OSAL implementation require more than this, the project must increase
     //! that value in its config/ folder.
     //!
     //! \param buffer The buffer to deserialize the contents from.
+    //! \param mode The endianness mode.
     //! \return Fw::SerializeStatus indicating the result of the deserialization.
-    Fw::SerializeStatus deserializeFrom(Fw::SerializeBufferBase& buffer) override;
-
+    Fw::SerializeStatus deserializeFrom(Fw::SerializeBufferBase& buffer,
+                                        Fw::Endianness mode = Fw::Endianness::BIG) override;
 
   private:
     //! Handle for ZephyrRawTime
@@ -95,4 +95,4 @@ class ZephyrRawTime : public RawTimeInterface {
 }  // namespace RawTime
 }  // namespace Zephyr
 }  // namespace Os
-#endif  // OS_ZEPHYR_RAWTIME_HPP
\ No newline at end of file
+#endif  // OS_ZEPHYR_RAWTIME_HPP
```

Fw/Time/Time.hpp
```diff
@@ -35,9 +35,11 @@ class Time : public Serializable {
     U32 getUSeconds() const;  // !< Gets microseconds part of time
     TimeBase getTimeBase()
         const;  // !< Time base of time. This is project specific and is meant for indicating different sources of time
-    FwTimeContextStoreType getContext() const;                                // !< get the context value
-    SerializeStatus serializeTo(SerializeBufferBase& buffer) const override;  // !< Serialize method
-    SerializeStatus deserializeFrom(SerializeBufferBase& buffer) override;    // !< Deserialize method
+    FwTimeContextStoreType getContext() const;  // !< get the context value
+    SerializeStatus serializeTo(SerializeBufferBase& buffer,
+                                Fw::Endianness mode) const override;  // !< Serialize method
```
LeStarch: missing the default parameters here.

LeStarch (dismissed):

Fw/Time/Time.hpp
LeStarch:
```suggestion
                                Fw::Endianness mode = Fw::Endianness::BIG) const override;  // !< Serialize method
```

Fw/Time/Time.hpp
```diff
@@ -35,9 +35,11 @@ class Time : public Serializable {
...
+    SerializeStatus deserializeFrom(SerializeBufferBase& buffer,
+                                    Fw::Endianness mode) override;  // !< Deserialize method
```
LeStarch: Default parameters.

```suggestion
                                    Fw::Endianness mode = Fw::Endianness::BIG) override;  // !< Deserialize method
```

Fw/Time/Time.hpp
Willmac16: I believe the default in the virtual will be applied, but I might be mistaken/not sure what best style is

LeStarch:
Note: I noticed some other issues: https://github.com/nasa/fprime/issues/4302

But these are outside the scope of the change.

Fw/Time/Time.hpp
LeStarch: According to here, it does not: https://stackoverflow.com/questions/3533589/can-virtual-functions-have-default-parameters.  Also, other serializables defined in the PR had defaults specified.

LeStarch:
@Willmac16 I went ahead and merged the suggestions for consistency, and also based on the link provided.  I updated fprime-zephyr so that should pass too.  If all is good, we will merge.

LeStarch (approved):

bocchino (approved):

PR #4304, opened 2025-10-13
Mark SwAssert declarations as noinline

celskeggs:
| | |
|:---|:---|

---
## Change Description

Mark `SwAssert` functions with `__attribute__((noinline))` where available.

## Rationale

Without this change, a compiler using LTO may inline the zeros that are inserted as additional parameters for the call to defaultSwAssert. This increases code size to optimize for the off-nominal condition where an assertion is tripped.

In my testing, this single change reduces my largest deployment's code size by 11%.

I only tested this on a clang compiler where the noinline attribute was available. It may be worth testing on other compilers.

## Future Work

None.

### DIFF
--- a/Fw/Types/Assert.hpp
+++ b/Fw/Types/Assert.hpp
@@ -63,19 +63,36 @@
 #endif
 #endif
 
+#ifndef NOINLINE
+#ifndef __has_attribute
+#define __has_attribute(x) 0
+#endif
+#if __has_attribute(noinline)
+#define NOINLINE __attribute__((noinline))
+#else
+#define NOINLINE
+#endif
+#endif
+
 namespace Fw {
 //! Assert with no arguments
-I8 SwAssert(FILE_NAME_ARG file, FwSizeType lineNo) CLANG_ANALYZER_NORETURN;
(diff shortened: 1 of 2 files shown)


LeStarch:
@bocchino this looks ok to me, thoughts?

Fw/Types/Assert.hpp
```diff
@@ -63,27 +63,44 @@
 #endif
 #endif
 
+#ifndef NOINLINE
```
bocchino, line 69:
```suggestion
// Define NOINLINE as __attribute__((noinline)) if that attribute is available.
// Marking assertion functions as NOINLINE can reduce code size without sacrificing performance
// in the common case that the function is not called.
#ifndef NOINLINE
```

bocchino (dismissed):
Looks good! I made a suggestion for a code comment to clarify what is going on.

Fw/Types/Assert.hpp
celskeggs: Seems like a good idea!

LeStarch (approved):

PR #4330, opened 2025-10-21
Fix RawTime UT oversight on initialization

### LINKED ISSUE #4238: PosixRawTime Random UT Failure
| | |
|:---|:---|
|**_F´ Version_**| fprime-4.0.0 |
|**_Affected Component_**|  PosixRawTime |
---
## Problem Description

PosixRawTimeTest fails when using random seed 227892. This failed on an Ubuntu machine.

```
56/95 Test #83: PosixRawTimeTest .................................***Failed    0.60 sec
[STest::Random] Generated seed 227892 from system time
[==========] Running 7 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 7 tests from Functionality
[ RUN      ] Functionality.Now
[       OK ] Functionality.Now (0 ms)
[ RUN      ] Functionality.SelfDiffIsZero
[       OK ] Functionality.SelfDiffIsZero (0 ms)
[ RUN      ] Functionality.GetTimeDiffU32
[       OK ] Functionality.GetTimeDiffU32 (0 ms)
[ RUN      ] Functionality.GetTimeInterval
[       OK ] Functionality.GetTimeInterval (0 ms)
[ RUN      ] Functionality.Serialization
[       OK ] Functionality.Serialization (19 ms)
[ RUN      ] Functionality.DiffU32Overflow
[       OK ] Functionality.DiffU32Overflow (0 ms)
[ RUN      ] Functionality.RandomizedTesting
/home/runner/work/fprime/fprime/Os/test/ut/rawtime/RulesHeaders.hpp:77: Failure
Value of: result < Fw::TimeInterval(0, INTERVAL_DIFF_THRESHOLD)
  Actual: false
Expected: true
Interval difference: (0s,530us)
Ran 5000 steps for RawTime.
```

A description of the problem with sufficient detail to understand the issue.

## Context / Environment

```
--INSERT HERE--
```

## How to Reproduce

1. 
2.
3.

## Expected Behavior

A description of the expected behavior.

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/4238

Round 2 of attempting to fix https://github.com/nasa/fprime/issues/4238

I believe this should be it! 😓 
There was an oversight on my end, the shadow state initialization was using the old pattern of taking close-enough samples, rather than forcing the values to be the same

I have been testing on RHEL8 under CPU load. I believe we are good to go this time. I will re-run the RHEL8 workflow on GitHub here a few times to confirm further.


Os/test/ut/rawtime/RulesHeaders.hpp
```diff
@@ -83,10 +83,9 @@ struct Tester {
         if (duration < std::chrono::system_clock::duration::zero()) {
             duration = -duration;
         }
-
-        U32 microseconds = std::chrono::duration_cast<std::chrono::microseconds>(duration).count();
-        U32 seconds = std::chrono::duration_cast<std::chrono::seconds>(duration).count();
-
+        U32 total_microseconds = std::chrono::duration_cast<std::chrono::microseconds>(duration).count();
```
LeStarch: Why change this? Seems like we might have introduced an overflow problem, as `U32_MAX` is only 4000 seconds. No?

LeStarch (dismissed):
Minor question.

Os/test/ut/rawtime/RulesHeaders.hpp
thomas-bc: There was a miscalculation that could happen if the interval was >1s, because it was missing a `%1000000`, which I originally thought was the cause of the issue. Thought we could fix still, but you're right. I'll add a modulo to fix the calculation without overflowing.

LeStarch (approved):

PR #4328, opened 2025-10-20
Clarify float operation behavior in documentation

LeStarch:
Updated float operations to specify handling of NaN and infinity.  Fixes #3892, #3893.

| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|  |
|**_Documentation Included (y/n)_**|  |
|**_Generative AI was used in this contribution (y/n)_**| y |

---
## Change Description

Fixes #3892, #3893.

## Review Note

I updated the definition of `fne` to obey the property: `fne = !(feq(v, v))` which is consistent with C++.

## AI Usage (see [policy](https://github.com/nasa/fprime/blob/devel/AI_POLICY.md))

Commit Message.

Issue #4325, opened 2025-10-20
FpySequencer Undefined Behavior


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

FpySequencer exposes raw signed operations to the user and does not guard against signed integer overflow and underflow.  In C/C++ signed integer overflow and underflow are undefined behaviors. 

The FpySequencer must provide proper error handling for these cases lest a program crash occur.

PR #4331, opened 2025-10-21
Add more error checking to arithmetic directives

### LINKED ISSUE #4327: FpySequencer: Remove floatCmp and Use Standard Macros
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

FpySequencer defines a `floatCmp` whose helper function appears to do exactly as the functions here do: https://linux.die.net/man/3/isunordered

Why not implement all comparison with those directly?

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Companion to #4328.  Fixes `fne` to be `!feq`.  Fixes #4327, #4325.

## Rationale

Code quality.

Double check the overflow underflow operations.

Autocomplete.


Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
```diff
@@ -748,7 +748,7 @@ TEST_F(FpySequencerTester, fne) {
     tester_push<F64>(std::numeric_limits<F64>::quiet_NaN());
     tester_push<F64>(std::numeric_limits<F64>::quiet_NaN());
     ASSERT_EQ(tester_op_fne(), DirectiveError::NO_ERROR);
-    ASSERT_EQ(tester_get_m_runtime_ptr()->stack[0], 0);
+    ASSERT_EQ(tester_get_m_runtime_ptr()->stack[0], 1);
```
zimri-leisher, line 751: nan == nan? i think this should be false

LeStarch: This is the fne case (float not equal).  In the float not equal case NaN is deemed not equal to NaN. So that != is not ==

LeStarch:
@zimri-leisher would you review this?

zimri-leisher:
Sorry for the late review. I think maybe this PR would be the right place to add UTs to check these overflows, unless you'd really rather split it up, which is fine.

LeStarch:
@zimri-leisher, I'd like to break these into two steps.  Easier for me to track.

zimri-leisher (approved):

thomas-bc (approved):

Issue #4261, opened 2025-10-03
Length of telemetered enums in PacketsAc.cpp hardcoded to 4 bytes


nathancheek:
| | |
|:---|:---|
|**_F´ Version_**| v3.6.3|
|**_Affected Component_**|`Svc::TlmPacketizer`|
---
## Problem Description

In F´ v3.6.3 and prior, enum sizes other than 4 bytes are not supported by the telemetry packet autocoder for the `PacketsAc.cpp` file.

The `TopologyAppDictionary.xml` and `TopologyDictionary.json` files are generated with the proper enum length, which the fprime-gds correctly reads. But `Autocoders/Python/bin/tlm_packet_gen.py` uses a hardcoded `4` for the length of enums when generating `PacketsAc.cpp`.

If I add a single `U8` enum to a telemetry packet, this issue causes an extra 3 bytes to be inserted into the packet. The fprime-gds is then unable to properly parse the packet (although interestingly it doesn't seem upset that the packet is longer than its definition).

I have a fix patch here: https://github.com/nathancheek/fprime/commit/a20e3cb74ea999f586c5b40625468504842bb1a9

It looks like F´ v4 is using a newly written autocoder, which may not have this issue. Do you have a mechanism for accepting patches for previous versions of F´?

## How to Reproduce

1. Create a `U8` enum in a component's fpp file
2. Create a telemetry channel of that enum's type in the fpp file
3. Modify the deployment to send the telemetry channel (in my case I'm using `TlmPacketizer`)
4. Note the extra 3 bytes `0x00 00 00` in the telemetry packet (using fprime-gds, comlogger files, etc)
5. Modify the telemetry channel definition to be a `U8` instead of the enum type, then repeat, showing that those 3 extra bytes are no longer included in that telemetry packet

## Expected Behavior

`tlm_packet_gen.py` shouldn't assume that all enums are 4 bytes long. It should pull the `serialize_type` from the enum's XML definition by calling `enum_model.get_serialize_type()`. See patch here: https://github.com/nathancheek/fprime/commit/a20e3cb74ea999f586c5b40625468504842bb1a9

LeStarch:
Patches should be contributed to: https://github.com/nasa/fprime/tree/release/v3.6.0-hotfixes.  This branch is at the tag of: v3.6.3 and records hot-fix maintenance of the v3.6 codebase.

thomas-bc:
Fixed in #4333

PR #4333, opened 2025-10-21
Use correct enum size when generating tlm packet structures

nathancheek:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4261 |
|**_Has Unit Tests (y/n)_**| no new unit tests |
|**_Documentation Included (y/n)_**| no new documentation |
|**_Generative AI was used in this contribution (y/n)_**| no |

---
## Change Description

This PR modifies the `tlm_packet_gen.py` autocoder to pull in the correct size of each enum type when generating `PacketsAc.cpp`.

## Rationale

Prior to this change, the autocoder assumes a hardcoded 4-byte length for each enum channel, even though `TopologyAppDictionary.xml` and `TopologyDictionary.json` files are generated with the proper enum length. This leads to a discrepancy, where a GDS using the above Dictionary files will have a different packet definition than the fprime deployment is using to generate telemetry packets.

This change brings the fprime telemetry packet generation process into consistency with the Dictionary files, at least where enum lengths are concerned.

## Notes

The primary fix is to the `process_enum_files()` function, which is modified to grab the `serialize_type` from the parsed XML via `enum_model`, then evaluate that string (likely 'U8', 'U16', etc) using `get_type_size()`, rather than assuming a hardcoded 4-byte length.

A secondary fix is to remove what I believe to be dead code in other parts of the file. It looks like a past version of the autocoder stored enum types in a tuple rather than a string, so there are 3 instances (processing typical channels, processing structs/serializables, and processing arrays) where the type was being checked against `type(tuple())` and in `True` cases, the type was assumed to be an enum with size of 4 bytes. Since enum types are currently stored as strings, these 3 hardcoded `if` statements never resolve `True`, so their hardcoded values don't really matter. But I've removed them for clarity.

I've confirmed the following:

- Tested successfully on internal deployment:
  - The enum length disparity is resolved. `fprime-gds` now happily displays packetized 1-byte, 2-byte, and 4-byte enum channels
  - Other tlm channel types work properly:
    - Numeric values (U8, U16, U32, etc)
    - Integer arrays (U8, U16, U32)
    - Enum arrays (U8, U16, U32)
    - Structs/serializables packed with integers, enums (U8, U16, U32), and strings
    - Strings
- Ref deployment compiles, seems to work correctly via fprime-gds, and passes unit tests

## Future Work

There are some other hardcoded and possible out-of-date assumptions in this autocoder script that could be looked at, but it seems the autocoder was rewritten for F Prime 4.0, so there may be no need to improve this script further.

N/A

### DIFF
--- a/.github/workflows/build-test-rpi.yml
+++ b/.github/workflows/build-test-rpi.yml
@@ -58,6 +58,7 @@ jobs:
     - uses: actions/checkout@v4
       with:
         sparse-checkout: 'requirements.txt'
+        sparse-checkout-cone-mode: false
     - name: "Setup environment"
       run: |
         python -m venv venv
--- a/.github/workflows/cookiecutters-test.yml
+++ b/.github/workflows/cookiecutters-test.yml
@@ -26,12 +26,14 @@ jobs:
     uses: ./.github/workflows/reusable-get-pr-branch.yml
     with:
       target_repository: nasa/fprime-tools
+      default_target_ref: v3.6.1
 
   get-bootstrap-branch:
     name: "Get fprime-bootstrap target branch"
     uses: ./.github/workflows/reusable-get-pr-branch.yml
     with:
       target_repository: nasa/fprime-bootstrap
+      default_target_ref: v1.4.0
 
   # -------- Install target versions of the cookiecutter templates and validate -------
   Validate:
--- a/.github/workflows/ext-build-hello-world.yml
+++ b/.github/workflows/ext-build-hello-world.yml
@@ -20,6 +20,7 @@ jobs:
     uses: ./.github/workflows/reusable-get-pr-branch.yml
     with:
       target_repository: fprime-community/fprime-tutorial-hello-world
+      default_target_ref: v3.6.0
 
   run:
     needs: get-branch
--- a/.github/workflows/ext-build-led-blinker.yml
+++ b/.github/workflows/ext-build-led-blinker.yml
@@ -20,6 +20,7 @@ jobs:
     uses: ./.github/workflows/reusable-get-pr-branch.yml
     with:
       target_repository: fprime-community/fprime-workshop-led-blinker
+      default_target_ref: v3.6.0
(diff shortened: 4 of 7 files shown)


LeStarch (approved):

LeStarch:
The CI errors are caused by a checkout of F Prime devel, not the current branch.

nathancheek:
@LeStarch Do you need anything more from me in order to get this PR merged?

LeStarch:
As you can see, the CI situation is not working.  Need to determine what to do about it.

nathancheek:
I just pushed a commit that should fix the CI issues, by manually specifying v3.6-compatible refs for the relevant repositories. All the actions ran successfully in my test branch, except for the two waiting for a Raspberry Pi to pick up the job.

nathancheek:
I see 2 CI failures.

The `RPI LedBlinker` failure I think was due to not specifying `sparse-checkout-cone-mode: false` within single-file sparse checkouts, which was fixed later in 8ead55a. I have made the same fix and re-pushed.

The `macOS-CI` failure I'm less sure about. The action seemed to run fine [under my repo](https://github.com/nathancheek/fprime/actions/runs/18795329182), but is failing on `fprime/Drv/TcpServer/test/ut/TcpServerTester.cpp:209` with `client.open(client_fd)` returning `Drv::SOCK_SUCCESS` (0), when it's expected to return `Drv::SOCK_FAILED_TO_CONNECT` (-4).

LeStarch:
Might be one of the transient errors that occurred during this era of the cod base.

SterlingPeet:
@LeStarch those errors both look suspect to me, does poking them to re-run tend to work?  We can't try that, I'm curious if theres a race condition that can be overcome by hitting the re-run button a few times.

LeStarch:
I cannot explain this RPI error, but I am of the belief we should merge over this objection.  If it does not pass now, I will do this.

Issue #3985, opened 2025-08-05
Fix Priorities


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

Can we fix the priorities in the default cookie-cutter topology template?  We should choose:

1. Valid on Darwin and Linux
2. Valid on Linux

Right now we are non-compliant with both.

thomas-bc:
Also need to fix cookiecutters

https://github.com/nasa/fprime-tools/blob/devel/src/fprime/cookiecutter_templates/cookiecutter-fprime-deployment/%7B%7Bcookiecutter.deployment_name%7D%7D/Top/instances.fpp

thomas-bc:
@pjromano

pjromano:
Just to make sure I'm clear: Given @LeStarch's overview at the tagup, this is only about fixing thread priorities on  ComponentInstance, right? (There are a lot of other priorities in the framework, like command priorities, port instance priorities, ...)

Given this applies only to ComponentInstance priority: There are assignments of ComponentInstance `priority` in framework code besides the cookie cutters, e.g. the value of [`FwTaskPriorityType Os::Task::TASK_PRIORITY_DEFAULT`](https://github.com/nasa/fprime/blob/475f32352554898fff379859f59093ff1f255165/Os/Task.hpp#L38) which is used for the default task priority in at least `Os::Task::start()` and `Fw::ActiveComponentBase::start()`. `TASK_PRIORITY_DEFAULT` is set to the numeric limit max of its type, which seems like it would not be within the OS-valid priority ranges... I'm trying to trace out the actual type of `FwTaskPriorityType`, but ended up at a dead-end at:

https://github.com/nasa/fprime/blob/475f32352554898fff379859f59093ff1f255165/default/config/FpConfig.fpp#L36

since I don't find `PlatformTaskPriorityType` defined anywhere.

**Edit 10/21/2025:** The rest of this comment was misled. See [my later comment](https://github.com/nasa/fprime/issues/3985#issuecomment-3427306444) for a correction.

---

Unless there's some magic I'm missing, that type must be at least 8 bits, and thus the std::numeric_limits<FwTaskPriorityType>::max() would be 127 or 255 depending on signedness. I anticipate that's outside the valid priority range of the most-restrictive OS we're supporting (which @LeStarch put at ~32-49 for Darwin, but I still have to research that).

[...]

Issue #292, opened 2020-12-01
Rename ActiveLogger


LeStarch:
ActiveLogger isn't really a logger and is more of a filter.  Can we update names and references?

LeStarch:
I don't think this is a major issue.  Closing.

PR #4337, opened 2025-10-21
Adjust task priorities to fit within supported platforms' priority ranges

pjromano:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3985 |

---
## Change Description

Adjusted all task priorities throughout the framework so that they fit within the most-restrictive task priority range among supported platforms, which was determined to be Darwin's [15,47]. See the linked issue for that analysis. Within the fprime repository, this resulted in changes in framework-provided subtopologies and in the Ref project.

Priority changes in the fprime-tools repository are handled in [fprime-tools PR #292](https://github.com/nasa/fprime-tools/pull/292).

In-code documentation was added to clarify the intended meaning of constant `Os::Task::TASK_PRIORITY_DEFAULT`, as well as `Os::Task::TASK_DEFAULT` which follows the same pattern. The meaning has implications for derived classes' implementations which are not immediately obvious.

## Rationale

Previous task priorities were out of the valid range for the Linux and Darwin platforms, clamping tasks to the maximum priority and preventing differentiation of priority between some threads. Accordingly, priority clamp warnings were being (correctly) produced to the Logger when starting F Prime.

No unit tests were added. The changes were verified by running the CI Integration tests and visually inspecting that the priority clamp warnings are no longer produced in the log. We should repeat this verification a final time before merging.

## Future Work

This pull request does not include documentation for recommendations on task priority. A style guide update to keep newly-added or modified threads in the framework within Darwin's range could help to prevent future code changes from reintroducing the priority clamping and provide guidance to users for their own threads.

None.

### DIFF
--- a/Ref/Top/RefTopology.cpp
+++ b/Ref/Top/RefTopology.cpp
@@ -32,7 +32,7 @@ U32 rateGroup2Context[Svc::ActiveRateGroup::CONNECTION_COUNT_MAX] = {};
 U32 rateGroup3Context[Svc::ActiveRateGroup::CONNECTION_COUNT_MAX] = {};
 
 enum TopologyConstants {
-    COMM_PRIORITY = 100,
+    COMM_PRIORITY = 23,
 };
 
 /**
--- a/Ref/Top/RefTopology.cpp
+++ b/Ref/Top/RefTopology.cpp
@@ -32,7 +32,7 @@ U32 rateGroup2Context[Svc::ActiveRateGroup::CONNECTION_COUNT_MAX] = {};
 U32 rateGroup3Context[Svc::ActiveRateGroup::CONNECTION_COUNT_MAX] = {};
 
 enum TopologyConstants {
-    COMM_PRIORITY = 23,
+    COMM_PRIORITY = 34,
 };
 
 /**
--- a/Svc/Subtopologies/ComCcsds/ComCcsdsConfig/ComCcsdsConfig.fpp
+++ b/Svc/Subtopologies/ComCcsds/ComCcsdsConfig/ComCcsdsConfig.fpp
@@ -13,8 +13,8 @@ module ComCcsdsConfig {
     }
 
     module Priorities {
-        constant comQueue   = 24
-        constant aggregator = 23
+        constant aggregator = 30
+        constant comQueue   = 29
     }
 
     # Queue configuration constants
--- a/Svc/Subtopologies/ComCcsds/ComCcsdsConfig/ComCcsdsConfig.fpp
+++ b/Svc/Subtopologies/ComCcsds/ComCcsdsConfig/ComCcsdsConfig.fpp
@@ -13,8 +13,8 @@ module ComCcsdsConfig {
     }
 
     module Priorities {
-        constant comQueue   = 101
-        constant aggregator = 100
+        constant comQueue   = 24
+        constant aggregator = 23
     }
 
     # Queue configuration constants
--- a/Svc/Subtopologies/ComFprime/ComFprimeConfig/ComFprimeConfig.fpp
+++ b/Svc/Subtopologies/ComFprime/ComFprimeConfig/ComFprimeConfig.fpp
@@ -11,7 +11,7 @@ module ComFprimeConfig {
     }
 
     module Priorities {
-        constant comQueue   = 24
+        constant comQueue   = 29
     }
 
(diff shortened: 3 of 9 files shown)


Ref/Top/RefTopology.cpp
```diff
@@ -32,7 +32,7 @@ U32 rateGroup2Context[Svc::ActiveRateGroup::CONNECTION_COUNT_MAX] = {};
 U32 rateGroup3Context[Svc::ActiveRateGroup::CONNECTION_COUNT_MAX] = {};
 
 enum TopologyConstants {
-    COMM_PRIORITY = 100,
+    COMM_PRIORITY = 23,
```
LeStarch: Comm (uplink) should be the highest priority under the rate groups, so like 34 (leaving some space between 41 and the comm).  CmdDispatch would be one step higher (35) as it consumes the output of comm, and thus should run at a higher priority.

Svc/Subtopologies/ComCcsds/ComCcsdsConfig/ComCcsdsConfig.fpp
```diff
@@ -13,8 +13,8 @@ module ComCcsdsConfig {
     }
 
     module Priorities {
-        constant comQueue   = 101
-        constant aggregator = 100
+        constant comQueue   = 24
```
LeStarch: Com queue is highest priority downlink, so it should probably be at 29, with aggregator at 30 as it "consumes" from comQueue.

Svc/Subtopologies/ComFprime/ComFprimeConfig/ComFprimeConfig.fpp
```diff
@@ -11,7 +11,7 @@ module ComFprimeConfig {
     }
 
     module Priorities {
-        constant comQueue   = 101
+        constant comQueue   = 24
```
LeStarch: 29 to be consistent with the CCsds

LeStarch (commented):
This is exactly what I was hoping to see!!!

I made some comments (even though it is still in draft).  Notably in the high-priority, but w/o deadline set (uplink, downlink) I recommend adjusting the priorities such that consumers of data are higher priority than producers of data.  I left specifics.

This review was easy because you've established a clear scheme for priorities, so all I needed to do was nudge a bit. 

Thanks!

pjromano:
> I made some comments (even though it is still in draft). Notably in the high-priority, but w/o deadline set (uplink, downlink) I recommend adjusting the priorities such that consumers of data are higher priority than producers of data. I left specifics.
> 
> This review was easy because you've established a clear scheme for priorities, so all I needed to do was nudge a bit.

@LeStarch, thanks for the review! I'll take your updates into account in a new commit before final submission.

Note that in the changes as I originally submitted, I kept the relative order between the F Prime threads as they were before. My approach was to subtract 77 from every priority, to place the originally-used range [95,120] at [18,43], leaving some room above and below within Darwin's range [15,47]. The one exception was Ref project's blockDrv (was 140), which needed to be brought down a bit further to reach the valid range, but I still kept this as highest priority of the threads in that project, i.e. 47.

Just noting that your review makes changes to the relative priorities between the threads, while I was originally maintaining the relative priority. But your recommendation makes sense to me that consumers should be higher priority than producers to make sure consumers get to service queues so they don't sit in overflow.

Ref/Top/RefTopology.cpp
pjromano: Adjusted uplink and cmdDispatch priorities in new commit 70cca82.

Svc/Subtopologies/ComCcsds/ComCcsdsConfig/ComCcsdsConfig.fpp
pjromano: Adjusted aggregator to be higher priority than comQueue, and both of these downlink tasks to be lower priority than uplink tasks, in new commit 70cca82.

Svc/Subtopologies/ComFprime/ComFprimeConfig/ComFprimeConfig.fpp
pjromano: Adjusted in new commit 70cca82.

LeStarch (approved):
Exactly what I expected.  Thanks for the extra comments as well!

PR #4340, opened 2025-10-21
Refactor FrameAccumulator not to assert on incorrect size_out

### LINKED ISSUE #4320: FrameAccumulator Asserts on Size Out
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

Frame accumlator asserts that size_out is less than ring buffer size.  However, this assert can be triggered by invalid (garbage) data that comes with an invalid size.

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4320  |
|**_Generative AI was used in this contribution (y/n)_**|  No |

---
## Change Description

Fix #4320

Putting in draft for now cause the cognitive complexity of all these if-else is becoming quite high and I'm trying to see if I can make it more readable somehow.

### DIFF
--- a/Svc/FrameAccumulator/FrameAccumulator.cpp
+++ b/Svc/FrameAccumulator/FrameAccumulator.cpp
@@ -118,17 +118,16 @@ void FrameAccumulator ::processRing() {
                   static_cast<FwAssertArgType>(m_inRing.get_allocated_size()), static_cast<FwAssertArgType>(remaining));
         // On successful detection, consume data from the ring buffer and place it into an allocated frame
         if (status == FrameDetector::FRAME_DETECTED) {
-            // size_out must be set to the size of the buffer and must fit within the existing data
+            // size_out must be set (non-zero) and must fit within the remaining data. Otherwise would mean
+            // a coding error in the detector (returned FRAME_DETECTED without proper validation)
             FW_ASSERT(size_out != 0);
             FW_ASSERT(size_out <= remaining, static_cast<FwAssertArgType>(size_out),
                       static_cast<FwAssertArgType>(remaining));
-            // check for overflow before casting down to U32
-            FW_ASSERT(size_out <= std::numeric_limits<U32>::max());
-            Fw::Buffer buffer = this->bufferAllocate_out(0, static_cast<U32>(size_out));
+            Fw::Buffer buffer = this->allocate_out(0, size_out);
             if (buffer.isValid()) {
                 // Copy data out of ring buffer into the allocated buffer
                 Fw::SerializeStatus serialize_status = this->m_inRing.peek(buffer.getData(), size_out);
-                buffer.setSize(static_cast<Fw::Buffer::SizeType>(size_out));
+                buffer.setSize(size_out);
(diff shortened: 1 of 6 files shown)


Svc/FrameAccumulator/FrameAccumulator.cpp
```diff
@@ -118,17 +118,16 @@ void FrameAccumulator ::processRing() {
                   static_cast<FwAssertArgType>(m_inRing.get_allocated_size()), static_cast<FwAssertArgType>(remaining));
         // On successful detection, consume data from the ring buffer and place it into an allocated frame
         if (status == FrameDetector::FRAME_DETECTED) {
-            // size_out must be set to the size of the buffer and must fit within the existing data
+            // size_out must be set (non-zero) and must fit within the remaining data. Otherwise would mean
```
LeStarch: Theoretically, a well-formed (but to large) packet could be sent from the ground.  Thus I would not assert here either.

LeStarch (commented):
I think you need to prevent an assert on a valid, but huge packet.  Otherwise we have a command-to-crash that an adversary could produce.

Svc/FrameAccumulator/FrameAccumulator.cpp
thomas-bc: Looking at the CcsdsFrameDetector, this shouldn't be possible, because `MORE_DATA_NEEDED` will always be returned if the `size_out` is bigger than what's available in the ring buffer

https://github.com/nasa/fprime/blob/6bdd79f29071408c455c07e5b8fb53d5b197e100/Svc/FrameAccumulator/FrameDetector/CcsdsTcFrameDetector.cpp#L49-L53

thomas-bc: But I agree, defensive programming and all, I will make the change. I should even be able to hit the branch in UTs with the mock detector...

Svc/FrameAccumulator/FrameAccumulator.cpp
```diff
@@ -116,19 +116,31 @@ void FrameAccumulator ::processRing() {
         // Detect must not consume data in the ring buffer
         FW_ASSERT(m_inRing.get_allocated_size() == remaining,
                   static_cast<FwAssertArgType>(m_inRing.get_allocated_size()), static_cast<FwAssertArgType>(remaining));
+
+        // Drop frames that are too large to handle (can't fit in the accumulation circular buffer)
+        if (size_out > ringCapacity) {
+            // Detector reports a size_out larger than the accumulation buffer capacity, we will never be able
```
thomas-bc, line 122: I did consider asserting `FW_ASSERT(status != NO_FRAME_DETECTED);` here because size_out really shouldn't be modified if nothing is detected, but since this check is trying to protect against asserts, I figured probably not. Thoughts?

thomas-bc: That really is a "coding error" in our current setup, but who knows maybe some project-specific detector/accumulator could use that return value to e.g. signal "you're good to skip N bytes" or something

LeStarch (approved):

Issue #3921, opened 2025-07-23
Use U8 representation for FPP enums defined in the framework


celskeggs:
| | |
|:---|:---|
|**_F´ Version_**| latest |
|**_Affected Component_**| several |
---
## Feature Description

Currently, enums like `Fw.Logic` use an implied 4-byte representation, which is overkill, since it is basically storing a boolean.

I recommend that all enums that don't require a larger representation should use a U8 representation.

## Rationale

Oversized enums do not pack well, particularly when serializing structs and arrays, resulting in data packet bloat that conflicts with strict memory sizing requirements on embedded systems.

thomas-bc:
@pjromano

PR #4342, opened 2025-10-22
Change FPP enums to smallest representation

pjromano:
| | |
|:---|:---|
|**_Related Issue(s)_**| #3921 |

---
## Change Description

For all FPP enums in the framework, set the representing type to the minimum width integer that can represent the enum's member constants, if representing type was not already specified.

## Rationale

Most enums in the framework only have a few member values that could all fit in U8 but had no representing type specified, resulting in the default I32 representation and unused bytes in the serialized representation.

No tests were updated.

Note: I didn't modify FppTestProject since it is exercising the range of features in FPP, including enum serialized width, and this is not part of framework code affecting user projects. I also didn't modify the Ref project for similar reasons; enums here do not affect users of the framework.

## Future Work

A style guide update to specify enums' representing type to the smallest type possible could help minimize the framework's footprint in downlink/uplink data bandwidth.

None.

### DIFF
--- a/Drv/ByteStreamDriverModel/ByteStreamDriverModel.fpp
+++ b/Drv/ByteStreamDriverModel/ByteStreamDriverModel.fpp
@@ -1,7 +1,7 @@
 module Drv {
 
     @ Status returned by the send call
-    enum ByteStreamStatus {
+    enum ByteStreamStatus : U8 {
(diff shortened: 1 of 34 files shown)


LeStarch (approved):
LGTM!  I will try to get CI passing.

thomas-bc (approved):

PR #4347, opened 2025-10-22
Fix macOS CI intermittent failure

### LINKED ISSUE #4346: macOS CI intermittent failure
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

macOS integration tests have been failing pretty often. Need to fix. It is suspected that the issue may be to do with the ComAggregator.
E.g. https://github.com/nasa/fprime/actions/runs/18727982381/job/53430704489

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4346 |

---
## Change Description

Fix #4346

There was a check that was made on how many events were received, but the check was waiting `0.5s`. In the cases where this didn't line up properly with `ComAggregator`'s timeouts, this would fail. 

Fixed by waiting `1.5s`, with `ComAggregator` being 1Hz by default, this should be plenty of margin.

## Review Note

Why was this only failing on macOS? I have no clue 🤷 
Maybe the macOS runners are just slower so we noticed it there more often?
I'll run this a bunch of times to make sure it actually fixed something

- [x] Agreed ? check this box, merge and make this my 200th commit to this repo! 🤡

### DIFF
--- a/Ref/test/int/ref_integration_test.py
+++ b/Ref/test/int/ref_integration_test.py
@@ -235,7 +235,8 @@ def test_active_logger_filter(fprime_test_api):
         fprime_test_api.send_and_assert_command("CdhCore.cmdDisp.CMD_NO_OP")
         fprime_test_api.send_and_assert_command("CdhCore.cmdDisp.CMD_NO_OP")
 
-        time.sleep(0.5)
+        # Wait >1s for ComAggregator's 1Hz timeout to flush all downlink
+        time.sleep(1.5)
 
         fprime_test_api.assert_event_count(pred, cmd_events)
         fprime_test_api.assert_event_count(pred, actHI_events)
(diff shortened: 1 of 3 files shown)


Ref/test/int/ref_integration_test.py
```diff
@@ -235,7 +235,8 @@ def test_active_logger_filter(fprime_test_api):
         fprime_test_api.send_and_assert_command("CdhCore.cmdDisp.CMD_NO_OP")
         fprime_test_api.send_and_assert_command("CdhCore.cmdDisp.CMD_NO_OP")
 
-        time.sleep(0.5)
+        # Wait >1s for ComAggregator's 1Hz timeout to flush all downlink
+        time.sleep(1.5)
```
thomas-bc, line 239: This one may not be needed since we're using `send_and_assert_command` but I figured it wouldn't hurt

thomas-bc: But likely using only `send_and_assert_command` and not `time.sleep()` _would_ work, and would be a better fix... let me know if you'd rather try that

LeStarch (approved):

thomas-bc:
image

Issue #4149, opened 2025-09-15
fpyc loops


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.1.0 |
|**_Affected Component_**| fpyc |
---
## Feature Description

Add `for` and `while` loops to Fpy

Syntax for loop over range?
```
for i in range(0, x):
    ...

# thinking this one, force them to add + 1 if needed
for i: U32 in <expr> .. <expr>:
    ...

for i in [0, 5):
    ...

while <expr>:
    pass
    break
    continue
```

Should loop define new scope? Probs not

zimri-leisher:
will the ending loop constant need space in the lvar array? i think it must

zimri-leisher:
thinking about
```
for i: U32 start <expr> length <expr> stride <expr>
```

also thinking about
```
var: SomeArrayType = ...

for idx: U32 in 0 .. len(var):
    ...
```
could add a `len` macro later

zimri-leisher:
@bocchino @Kronos3 opinions on whether for loops and while loops should enclose a scope? or should they just use the scope in which they are defined.

if we have for loops use their own scope, then:
cons:
```
x: U8 = 0
for x: U8 in 0..8: # this is legal
    pass
```
pros:
```
for x: U8 in 0..8: 
    pass
x = 5 # this is not legal
```


i'm leaning towards the latter because that's how python does it

zimri-leisher:
Closed by #4149

PR #4356, opened 2025-10-25
FpySequencer new directives for loops

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4149  |

---
## Change Description
Adds 4 new directives:

* GET_FIELD | Pops an offset (StackSizeType) off the stack. Takes a hard-coded number of bytes from top of stack, and then inside of that a second array of hard-coded number of bytes. The second array is offset by the value previously popped off the stack, with offset 0 meaning the second array starts furthest down the stack. Leaves only the second array of bytes, deleting the surrounding bytes
  * Used to get members of structs and elements of arrays at variable locations during runtime
* PEEK | Pops a StackSizeType `offset` off the stack, then a StackSizeType `byteCount`. Let `top` be the top of the stack. Takes the region starting at `top - offset - byteCount` and going to `top - offset`, and pushes this region to the top of the stack.
  * Used to do some array access checks at runtime
* ASSERT | Pops one byte for a condition and one byte for an error code off the stack. If condition is false, raise the error code as an event
  * Allows the `assert` keyword in Fpy, also used in array access checks
* STORE | Pops an offset (StackSizeType) off the stack. Pops a hardcoded number of bytes from the top of the stack, and moves them to the start of the lvar array plus the offset previously popped off the stack
  * Used in variable array index modifications

Also makes the following changes:
* Exit takes an error code instead of a bool, and raises an event
* Pushes tlm/prms to the stack instead of saving to lvar array
* Renames the old STORE directive to STORE_CONST_OFFSET

## Rationale
* Each of these features was necessary for implementing loops, or something related to loops.

used to generate boilerplate, tests and draft implementations


Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -1329,4 +1348,88 @@ Signal FpySequencer::getFlag_directiveHandler(const FpySequencer_GetFlagDirectiv
...
+    if (Fpy::MAX_STACK_SIZE - this->m_runtime.stackSize < byteCount) {
+        error = DirectiveError::STACK_OVERFLOW;
+        return Signal::stmtResponse_failure;
+    }
+    if (this->m_runtime.stackSize < byteCount) {
+        error = DirectiveError::STACK_ACCESS_OUT_OF_BOUNDS;
+        return Signal::stmtResponse_failure;
+    }
+    // copy from top - byteCount to top
+    memcpy(this->top(), this->top() - byteCount, byteCount);
```
LeStarch: This doesn't do what the PR text says.  It does not pop, just copies.  

Also, what is the purpose of this?  It seems like it might be so that you can avoid poping data, and the repushing it after you read.....or to avoid a peak operation.  

Either way, it seems strange.

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -1329,4 +1348,88 @@ Signal FpySequencer::getFlag_directiveHandler(const FpySequencer_GetFlagDirectiv
...
+    Fpy::StackSizeType stackOffset = this->lvarOffset() + this->pop<Fpy::StackSizeType>();
+    // if we popped these bytes off, and put them in lvar array, would we go out of bounds
+    if (stackOffset + directive.get_size() > this->m_runtime.stackSize - directive.get_size()) {
+        // write into lvar array would go out of bounds
+        error = DirectiveError::STACK_ACCESS_OUT_OF_BOUNDS;
+        return Signal::stmtResponse_failure;
+    }
+    // i believe we can be sure the regions are not overlapping, due to the above check
+    memcpy(this->m_runtime.stack + stackOffset, this->top() - directive.get_size(), directive.get_size());
+    this->m_runtime.stackSize -= directive.get_size();
```
LeStarch: I assume this is correct because I believe that `pop` should adjust the stack pointer.  However, these directives are really hard to understand because sometimes they are using formal stack mechanics (i.e. pop) and other times they do raw copies with deferred stack mechanics.

I would recommend that stack operations be pulled into a helper class for managing the stack, where the necessary functions could be exposed.  Separation of stack offset mechanics and directive implementation is a need for future maintainability.

LeStarch (approved):

LeStarch (changes requested):
The DUPLICATE directive does not pop any data off the stack, yet both the PR and the documentation says it does.

I'd prefer a statement like "Add a copy of X number of bytes from the top of the stack".

Alternatively, "pop X bytes of the stack, and then push 2 copies of those bytes onto the stack".

Although It seems to me that this could be a pop, and then a push twice.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: You're right, this could be implemented with a PEEK directive. That would be more general.
I went over the usage of this directive in last monday's meeting. It is currently used when checking array access out of bounds, to copy the value of the index expression and use it in a comparison against the array length. I can explain further but perhaps it would be better to just discuss this next monday. @bocchino and @Kronos3 were both there, maybe they have some feedback?

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: That makes sense, I could work on implementing a stack class.

Svc/FpySequencer/FpySequencerDirectives.cpp
Kronos3: We had a similar thought with the duplicate directive but I think it may be needed. From what I understand, it essentially boils down to there not being a way to use the result of some operation multiple times without the duplicate or introducing an additional local variable since all subsequent operations will remove their operands from the stack.

LeStarch (dismissed):

LeStarch:
@zimri-leisher, need to fix merge conflicts

zimri-leisher:
@LeStarch merge conflicts fixed

LeStarch (approved):
LGTM.  I appreciate the new stack data structure.  It will allow us to clean up some code moving forward.

LeStarch (approved):

PR #4367, opened 2025-10-28
Add new Svc::ComRetry component

valdaarhun:
Add new component that can be used by other components to resend a message on failure.

| | |
|:---|:---|
|**_Related Issue(s)_**| #3994 |

---
## Change Description

Add a new component - `ComRetry`.

## Rationale

Other components such as `ComQueue` can be connected to `ComRetry` to resend a message on failure.

## Future Work

- [x] Add tests
- [x] Add CMakeLists.txt
- [x] Add docs

NA


valdaarhun:
Hi. Apologies, it took much longer to draft this change. A lot has to be refined and the tests have not been added. I had a few questions and I thought I would get them clarified before continuing.

- I am not sure if this is the right way to store the buffer in `dataReturnIn_handler`.
```cpp
this->m_buffer = buffer
```

I was initially thinking of using `DpContainer`. Based on what I have understood from the [documentation](https://fprime.jpl.nasa.gov/latest/docs/user-manual/framework/data-products/) though, it would be an overkill to use it here.

Should `Fw::ExternalSerializeBufferWithMemberCopy` be used instead to store the buffer (similar to what's done in `DpContainer::setBuffer()`)?

- Once I get more clarity on the question above, I'll add in a check in `comStatusIn_handler` to test the validity of the buffer. I have marked it as TODO for now.
- I am not sure how the following would work:
```cpp
    else {
        this->dataReturnOut_out(0, this->m_buffer, this->m_context);
        this->m_retry_count = 0;
        Fw::Success condition = Fw::Success::FAILURE;
        this->comStatusOut_out(0, condition);
    }
```
Based on this [comment](https://github.com/nasa/fprime/pull/4081#issuecomment-3324013931):
1. Will `dataReturnOut` here be connected to `dataReturnIn` in the ComQueue?
2. Similarly, will `comStatusOut` be connected to ComQueue's `comStatusIn`?
4. Given that `comStatus` is an async port while `dataReturn` is a sync port, can it be said that `dataReturnIn` will always be called before `comStatusIn`?
5. How does this work under the hood? I tried reading through the relevant functions in `ComQueueComponentAc.cpp` but couldn't really grasp it.
6. If I placed `comStatusOut_out` before `dataReturnOut_out` in the snippet above, would that make a difference?

Svc/ComRetry/ComRetry.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// ----------------------------------------------------------------------
+// Handler implementations for typed input ports
+// ----------------------------------------------------------------------
+
+void ComRetry ::comStatusIn_handler(FwIndexType portNum, Fw::Success& condition) {
+    // TODO: Check if m_buffer is valid before attempting delivery
+    if (condition == Fw::Success::SUCCESS) {
+        this->dataReturnOut_out(0, this->m_buffer, this->m_context);
```
LeStarch, line 45: Should reset `retry_count` in the success case too.

Svc/ComRetry/ComRetry.fpp
```diff
@@ -0,0 +1,14 @@
+module Svc {
+    @ A component for retrying message delivery on failure
+    active component ComRetry {
```
LeStarch: This need not be active.  Make it `passive`

Svc/ComRetry/ComRetry.fpp
```diff
@@ -0,0 +1,14 @@
...
+        import Svc.Framer
+
+        # One async command/port is required for active components
```
LeStarch: Remove the comments and below.

LeStarch (changes requested):
Minor changes, but this is what I was expecting.

LeStarch:
First, you are storing things as I would expect.  No need to do more.   Second, your validity check isn't necessary either, as this component doesn't alter the validity.

To answer your enumerated questions:
1. yes
2. also yes
3. yes, given that you have comstatus after data return in.
4. You know that dataReturnOut finishes before comstatus is ever sent because dataReturnOut is synchronous and thus finishes all the work before the end of the function call.
5. Yes it would. You have the expected order and should not reorder those calls.

valdaarhun:
> First, you are storing things as I would expect.  No need to do more.   Second, your validity check isn't necessary either, as this component doesn't alter the validity.
> 
> To answer your enumerated questions:
> 1. yes
> 2. also yes
> 3. yes, given that you have comstatus after data return in.
> 4. You know that dataReturnOut finishes before comstatus is ever sent because dataReturnOut is synchronous and thus finishes all the work before the end of the function call.
> 5. Yes it would. You have the expected order and should not reorder those calls. 
> 

Hi. Thank you for the answers.

Regarding the validity check, I was thinking that would take care of the case when `ComRetry` receives `Fw::Success`, but there's no buffer stored in the component.

valdaarhun:
Hi. I am ready with the core changes and unit tests.

Regarding the docs, I have added some preliminary documentation. I am not entirely sure about how to refine it further. I wrote it manually by referring to the docs for `ComQueue`, `ComStub` and `ComAggregator`. I don't know how the figures in those docs were created.

Svc/ComRetry/ComRetry.cpp
```diff
@@ -0,0 +1,67 @@
...
+        this->m_retry_count = 0;
+        this->dataReturnOut_out(0, this->m_buffer, this->m_context);
+        this->comStatusOut_out(0, condition);
+    }
+    // Delivery of last message failed
+    else if (this->m_retry_count < this->m_num_retries) {
+        FW_ASSERT(this->m_bufferState == Fw::Buffer::OwnershipState::OWNED);
+        this->m_retry_count++;
+        this->m_bufferState = Fw::Buffer::OwnershipState::NOT_OWNED;
+        this->dataOut_out(0, this->m_buffer, this->m_context);
```
LeStarch, line 55: We cannot retry on a failed message. This violates the expected protocol.  Instead, we need to wait until the next "SUCCESS" after a fail before returning.

LeStarch (commented):
The comment reveals the deep issue with this component.  Namely that sending to the radio is only allowed after a `SUCCESS`.  Thus you must hold onto the buffer.   I fixed the implementation here as I needed it in a pinch.  The SDD/UTs are as they are here.

https://github.com/LeStarch/fprime-extras/tree/main/FprimeExtras/Utilities/ComRetry

We can fold that implementation in here, to merge this PR.  We'd need to copy that over, and then correct tests and SDD.  You up for that work?

valdaarhun:
> The comment reveals the deep issue with this component. Namely that sending to the radio is only allowed after a `SUCCESS`. Thus you must hold onto the buffer.

Hi @LeStarch . Sorry about this, I must have misunderstood the requirement.

> I fixed the implementation here as I needed it in a pinch. The SDD/UTs are as they are here.
> 
> https://github.com/LeStarch/fprime-extras/tree/main/FprimeExtras/Utilities/ComRetry
> 
> We can fold that implementation in here, to merge this PR. We'd need to copy that over, and then correct tests and SDD. You up for that work?

Yes, I'll do it.

valdaarhun:
Could you please explain why std::atomic is being used for `m_retry_state` [here](https://github.com/LeStarch/fprime-extras/blob/main/FprimeExtras/Utilities/ComRetry/ComRetry.hpp#L69)? I haven't really understood where a data race might occur. Is it because there may be multiple threads running `comStatusIn_handler`?

LeStarch:
@valdaarhun excellent question.  I think I put that in there as a default implementation because:

1. This component is `passive` (could be called by multiple threads) and...
2. This is a state variable which controls implementation and...
3. There could be thread contention (in theory)

Practically speaking, I don't think thread contention would ever happen. `dataReturnIn` and `comStatusIn` are a response to `dataOut` which itself is a response to `dataIn`.  That `dataIn` is only triggered by `dataReturnOut`/`comStatusOut`.

So, as long as the state variable is updated **before** calling `dataReturnOut`/`comStatusOut` and `dataOut`, then there should never be a case where two threads are working on this.

Thus, I think you can safely remove that `std::atomic`.

valdaarhun:
> @valdaarhun excellent question. I think I put that in there as a default implementation because:
> 
>     1. This component is `passive` (could be called by multiple threads) and...
>     2. This is a state variable which controls implementation and...
>     3. There could be thread contention (in theory)
> 

Thank you for the clarification. I think another reason for my confusion was the usage of `m_retry_count` without atomic updates/checks.

> Thus, I think you can safely remove that `std::atomic`.

Understood. I'll remove that.

valdaarhun:
I haven't updated the UTs and SDD yet. I'll do that soon.

valdaarhun:
Hi @LeStarch , my changes are ready for review. Please let me know if I have missed something.

LeStarch (approved):
The only substantial difference I see here is that m_retry_state is not atomic....but it is guarded by the protocol and so I'd say this is good!

LeStarch:
@valdaarhun thanks for powering through this!  It is a good contribution!

valdaarhun:
> @valdaarhun thanks for powering through this!  It is a good contribution!

Thank you for your help and reviews, @LeStarch . I enjoyed working on this and got to learn a lot. I would love to continue contributing to F'.

PR #4393, opened 2025-11-04
Replace FpConfig.h header with FPP model elements

### LINKED ISSUE #3288: Replace configuration settings in FpConfig.h with FPP model elements
- [ ] Convert all the macro constant definitions in `FpConfig.h` to FPP constant definitions.
- [x] Convert all the typedefs in `FpConfig.h` to FPP type aliases.

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/3288 |
|**_Documentation Included (y/n)_**| updated |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/3288

Replaces values in `FpConfig.h` with constants defined in FPP.

## Future Work

Some macros still exist in `FpConfig.h` that can't be supported in FPP. Need to decide how to dispose of those, if the plan is to remove FpConfig.h altogether.

No


thomas-bc:
Note: needs `pr-xxxx` branches for Zephyr's https://github.com/fprime-community/fprime-zephyr-reference/blob/devel/FprimeZephyrReference/project/config/FpConfig.h

default/config/FpConstants.fpp
```diff
@@ -0,0 +1,86 @@
...
+
+@ Specifies the size of the buffer that contains the serialized log arguments
+constant FW_LOG_BUFFER_MAX_SIZE = (FW_COM_BUFFER_MAX_SIZE - 4 - 4)  # TODO: 4 is for `- sizeof(FwEventIdType) - sizeof(FwPacketDescriptorType))`
+
+@ Specifies the maximum size of a string in a log event
+@ Note: This constant truncates file names in assertion failure event reports
+constant FW_LOG_STRING_MAX_SIZE = 200
+
+@ Specifies the size of the buffer that contains the serialized telemetry value
+constant FW_TLM_BUFFER_MAX_SIZE = (FW_COM_BUFFER_MAX_SIZE - 4 - 4)  # TODO: 4's are sizeof TBD
```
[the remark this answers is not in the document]
thomas-bc: It's possible this is accurate but I don't want to dig to confirm and this is outside the scope of this ticket.

Fw/FPrimeBasicTypes.hpp
```diff
@@ -35,24 +35,28 @@ extern "C" {
 // Check that Fw::String is big enough to hold any string that can be given a
 // default value in FPP.
 
-static_assert(FW_FIXED_LENGTH_STRING_SIZE >= FW_CMD_STRING_MAX_SIZE,
+static_assert(FW_FIXED_LENGTH_STRING_SIZE >= static_cast<FwSizeType>(FW_CMD_STRING_MAX_SIZE),
```
thomas-bc, line 38: For posterity: we are casting only on one side to resolve compiler warnings that appear on the GitHub runners (Ubuntu 22) if you cast on both sides of the comparison

```
In file included from /home/runner/work/fprime/fprime/Fw/Types/Assert.hpp:4,
                 from /home/runner/work/fprime/fprime/build-fprime-automatic-native/F-Prime/default/config/TimeBaseEnumAc.cpp:10:
/home/runner/work/fprime/fprime/Fw/FPrimeBasicTypes.hpp:38:39: error: the result of the conversion is unspecified because ‘FW_FIXED_LENGTH_STRING_SIZE’ is outside the range of type ‘FppConstant_FW_CMD_STRING_MAX_SIZE’ [-Werror=conversion]
   38 | static_assert(static_cast<FwSizeType>(FW_FIXED_LENGTH_STRING_SIZE) >= static_cast<FwSizeType>(FW_CMD_STRING_MAX_SIZE),
      |                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

With 
```
enum FppConstant_FW_FIXED_LENGTH_STRING_SIZE {
  FW_FIXED_LENGTH_STRING_SIZE = 256
};
enum FppConstant_FW_CMD_STRING_MAX_SIZE {
  FW_CMD_STRING_MAX_SIZE = 40
};
```



Discussed with Rob and Michael and we don't really have an explanation; the error message doesn't make any sense to us. We are casting a value to FwSizeType, but the error message makes a reference to the other side of the comparison. Quite odd...

thomas-bc: Potentially relevant https://stackoverflow.com/questions/71520512/enumeration-conversion-and-undefined-behavior

Yes one enum likely has a range different than the other, but again we are casting each literal to `FwSizeType` independently of the other. So the fact the the error message speaks to the value of one with the type of the other makes little sense. Could be a compiler bug?

thomas-bc:
TODO:

- [x] Need to review and merge https://github.com/fprime-community/fprime-zephyr-reference/pull/19 before merging this

Fw/FPrimeBasicTypes.hpp
bocchino: It looks like a compiler bug to me. It seems to be analyzing the comparison `a as FwSizeType <= b as FwSizeType` as if it were `a as type of b <= b`.

default/config/FpConstants.fpp
```diff
@@ -0,0 +1,86 @@
...
+constant FW_TASK_NAME_BUFFER_SIZE = 80
+
+@ Specifies the size of the buffer that contains a communications packet
+constant FW_COM_BUFFER_MAX_SIZE = 512
+
+@ Specifies the size of the buffer attached to state machine signals
+constant FW_SM_SIGNAL_BUFFER_MAX_SIZE = 128
+
+@ Specifies the size of the buffer that contains the serialized command arguments
+constant FW_CMD_ARG_BUFFER_MAX_SIZE = (FW_COM_BUFFER_MAX_SIZE - 4 - 4)  # TODO: switch `- 4 - 4` with `- sizeof(FwOpcodeType) - sizeof(FwPacketDescriptorType))`
```
bocchino: I think we should do the following:

1. Define a constant `SIZE_OF_FwOpcodeType = 4`. Ideally this should appear right after the type definition of `FwOpcodeType`.
2. Define a constant `SIZE_OF_FwPacketDescriptorType`. Ideally this should appear right after the type definition of `FwPacketDescriptorType`.
3. Revise the code as suggested below. Same for the other cases of literal 4.

```suggestion
constant FW_CMD_ARG_BUFFER_MAX_SIZE = (FW_COM_BUFFER_MAX_SIZE - SIZE_OF_FwPacketDescriptor_type - SIZE_OF_FwOpcodeType)
```

This way if someone wants to change `FwOpcodeType` to `U16` they just have to change one 4 to a 2. If the literal 4's are sprinkled around the code, it's hard to make this change. Which 4's need to be changed and where? The chance for errors is very high.

Also, this is a style issue, but why are there parentheses around this and other constant expressions? These aren't macros.

thomas-bc: Implemented as recommended in https://github.com/nasa/fprime/pull/4393/commits/0ff62a3c2c5bb371610c987f825ab2c50265f913

bocchino (approved):
Looks good!

LeStarch (approved):

PR #4428, opened 2025-11-17
Set the Name of Fprime Tasks on Linux

Willmac16:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Os::Task for linux will set the thread name to the task's name.

## Rationale

When attempting to profile fprime code, it would be beneficial to see which fprime tasks are eating up CPU resources from outside tooling.

The current implementation does not prefix the task name w/ the original process name as 16 characters is rather restrictive.

## Future Work

N/A

N/A


Os/Posix/Task.cpp
```diff
@@ -115,6 +116,26 @@ int set_cpu_affinity(pthread_attr_t& attributes, const Os::Task::Arguments& argu
...
 
+int set_task_name(pthread_t thread, const Os::Task::Arguments& arguments) {
+    int status = 0;
+// pthread_setname_np is a non-POSIX function.
+// Limit its use to builds that involve glibc, on Linux, with _GNU_SOURCE defined.
+// That's the circumstance in which we expect this feature to work.
+#if defined(TGT_OS_TYPE_LINUX) && defined(__GLIBC__) && defined(_GNU_SOURCE)
+
+    // Construct a sixteen char long version of the task name
+    const FwSizeType PTHREAD_NAME_LENGTH = 16;
```
LeStarch: Why is this limited to 16 characters?

Os/Posix/Task.cpp
```diff
@@ -115,6 +116,26 @@ int set_cpu_affinity(pthread_attr_t& attributes, const Os::Task::Arguments& argu
...
+    char name_sixteen_capped[PTHREAD_NAME_LENGTH];
+    Fw::StringUtils::string_copy(name_sixteen_capped, arguments.m_name.toChar(), PTHREAD_NAME_LENGTH);
+
+    status = pthread_setname_np(thread, name_sixteen_capped);
+#else
+    Fw::Logger::log("[WARNING] %s setting thread name is only available with GNU pthreads\n",
```
LeStarch: May want to limit this warning to once per deployment.

LeStarch (commented):
A few questions, but otherwise looks good!

Os/Posix/Task.cpp
Willmac16: `man pthread_setthread_np` says:
> DESCRIPTION
       By  default,  all  the threads created using pthread_create() inherit the program name.  The pthread_setΓÇÉ
       name_np() function can be used to set a unique name for a thread, which can be useful for debugging  mulΓÇÉ
       tithreaded  applications.   The thread name is a meaningful C language string, whose length is restricted
       to 16 characters, including the terminating null byte ('\0').  The thread argument specifies  the  thread
       whose name is to be changed; name specifies the new name.

Os/Posix/Task.cpp
Willmac16: What would the idiomatic way to throttle a Fw::Logger call? static var for the Task class that is set once we send the first time?

Willmac16: Also doesn't look like we are limiting any of the other messages of this sort to once

thomas-bc (dismissed):

default/config/FpConfig.h
```diff
@@ -148,6 +148,12 @@ extern "C" {
 #define FW_AMPCS_COMPATIBLE 0  //!< Whether or not JPL AMPCS ground system support is enabled.
 #endif
 
+// Posix thread names are limited to 16 characters, this can lead to collisions. In the event of a
+// collision, set this to 0.
+#ifndef POSIX_THREADS_ENABLE_NAMES
+#define POSIX_THREADS_ENABLE_NAMES 0  //!< Enable/Disable assigning names to threads
```
thomas-bc: Seems to be off by default; did you mean to have it on instead?

thomas-bc (changes requested):

default/config/FpConfig.h
LeStarch: Yes, but first I need to correct the reconnect thread issue.  Working on that.

Os/Posix/Task.cpp
```diff
@@ -115,6 +116,27 @@ int set_cpu_affinity(pthread_attr_t& attributes, const Os::Task::Arguments& argu
...
+
+    // Construct a sixteen char long version of the task name. This length is fixed by the posix thread
+    // specification and is a constant.
+    char name_sixteen_capped[PosixTask::PTHREAD_NAME_LENGTH];
+    Fw::StringUtils::string_copy(name_sixteen_capped, arguments.m_name.toChar(), sizeof(name_sixteen_capped));
+
+    status = pthread_setname_np(thread, name_sixteen_capped);
+#else
+
+    Fw::Logger::log("[WARNING] %s setting thread name is only available with GNU pthreads\n",
```
thomas-bc: Also could be worth updating this warning in case it is deliberately switched off.

thomas-bc (approved):

thomas-bc (dismissed):

Os/Posix/Task.hpp
```diff
@@ -22,12 +22,16 @@ namespace Task {
...
 //!
 struct PosixTaskHandle : public TaskHandle {
+    static constexpr FwSizeType PTHREAD_NAME_LENGTH = 16;  //!< Length of pthread name
     static constexpr int SUCCESS = 0;
 
     //! Posix task descriptor
     pthread_t m_task_descriptor;
     //! Is the above descriptor valid
     bool m_is_valid = false;
+#if defined(POSIX_THREADS_ENABLE_NAMES) && POSIX_THREADS_ENABLE_NAMES
```
timcanham, line 32: You have the same macro checked twice?

Os/Posix/Task.cpp
```diff
@@ -22,9 +23,17 @@ static const int SCHED_POLICY = SCHED_RR;
 
 typedef void* (*pthread_func_ptr)(void*);
 
+// Forward declaration
+int set_task_name(pthread_t thread, char* name);
```
timcanham, line 27: Minor: If these functions are used just in this file, I would declare them `static`

Os/Posix/Task.cpp
```diff
@@ -22,9 +23,17 @@ static const int SCHED_POLICY = SCHED_RR;
...
+
 void* pthread_entry_wrapper(void* wrapper_pointer) {
     FW_ASSERT(wrapper_pointer != nullptr);
+    // Both downcasts are safe because we know the types
     Os::Task::TaskRoutineWrapper& wrapper = *reinterpret_cast<Os::Task::TaskRoutineWrapper*>(wrapper_pointer);
+    auto handle = reinterpret_cast<Os::Posix::Task::PosixTaskHandle*>(wrapper.m_task.getHandle());
+    FW_ASSERT(handle != nullptr);
+    // Task name is on a best effort basis
+    (void)set_task_name(handle->m_task_descriptor, handle->m_name);
```
timcanham, line 36: You can only set the name after it's been started from the new thread?

timcanham (commented):
Minor comments, but LGTM

Os/Posix/Task.hpp
LeStarch: I check if it is defined, and the ensure it is set enabled.  This makes sure it won't crash on older configurations.

Os/Posix/Task.cpp
LeStarch: Yes.  This is the cause of the CI errors as it requires the thread to start and the os to allocate a procfs directory before the call will pass

Os/Posix/Task.cpp
LeStarch: We could. The other functions in this file that are helpers are not declared static.

thomas-bc (approved):

PR #4451, opened 2025-11-19
Make Os::Queues use Fw::MemAllocator pattern for memory

### LINKED ISSUE #4039: Replace queue `malloc` with MemAllocator
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

Currently, memory allocation for queues use `malloc` to establish the base memory.

It would be more flexible to use `Fw::MemAllocator` to allocate these queues. There are several ways for us to handle this.

We could update queued and active component queue setup calls to pass-in a mem allocator.  The deployment cookie cutter could supply malloc allocator.  However, users could override on each call by phase, or change the default via the deployment top-level.

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix #4039 

Makes queues use registry memory for allocations.

## Rationale

Raw `new`/`malloc` are difficult.

CI

## Future Work

Other uses of new/malloc in framework

Autocomplete

### DIFF
--- a/Os/Generic/PriorityQueue.cpp
+++ b/Os/Generic/PriorityQueue.cpp
@@ -2,11 +2,11 @@
 // \title Os/Generic/PriorityQueue.cpp
 // \brief priority queue implementation for Os::Queue
 // ======================================================================
-
-#include "PriorityQueue.hpp"
+#include "Fw/LanguageHelpers.hpp"
 #include "Fw/Types/Assert.hpp"
 #include "Fw/Types/MemAllocator.hpp"
 #include "config/MemoryAllocatorTypeEnumAc.hpp"
+#include "Os/Generic/PriorityQueue.hpp"
 #include <cstring>
 
 namespace Os {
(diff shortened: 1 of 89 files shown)


timcanham (dismissed):
LGTM!

Svc/FileDownlink/FileDownlink.cpp
```diff
@@ -46,8 +46,8 @@ void FileDownlink ::configure(U32 timeout, U32 cooldown, U32 cycleTime, U32 file
     this->m_configured = true;
 
     Os::Queue::Status stat =
-        m_fileQueue.create(Os::QueueString("fileDownlinkQueue"), static_cast<FwSizeType>(fileQueueDepth),
-                           static_cast<FwSizeType>(sizeof(struct FileEntry)));
+        m_fileQueue.create(this->getInstance(), Os::QueueString("fileDownlinkQueue"),
```
thomas-bc, line 49: Would this not conflict with FileDownlink's internal queue (being an active component), if we were to use a MemAllocator that uses the identifier for something meaningful?

LeStarch: No, mem allocators allow multiple allocations per id.  Since both the internal and this allocation come from the same component, it is considered ok.

thomas-bc: I see. Is that comment wrong then? Or am I reading it wrong?

https://github.com/LeStarch/fprime/blob/issue-4039/Fw/Types/MemAllocator.hpp#L58

thomas-bc: @LeStarch

LeStarch: It is an isn't wrong, but definitely unclear.  It should be used by a single component, but may allocate multiple times.  I will update.

Os/Generic/PriorityQueue.cpp
```diff
@@ -38,62 +39,137 @@ void PriorityQueueHandle ::load_data(FwSizeType index, U8* destination, FwSizeTy
...
+
+void PriorityQueue::teardownInternal() {
+    if (this->m_handle.m_data != nullptr) {
+        const FwEnumStoreType identifier = this->m_handle.m_id;
+        Fw::MemAllocator& allocator = Fw::MemAllocatorRegistry::getInstance().getAnAllocator(
+            Fw::MemoryAllocation::MemoryAllocatorType::OS_GENERIC_PRIORITY_QUEUE);
+        allocator.deallocate(identifier, this->m_handle.m_data);
+        allocator.deallocate(identifier, this->m_handle.m_indices);
+        allocator.deallocate(identifier, this->m_handle.m_sizes);
+        this->m_handle.m_heap.~MaxHeap();
```
timcanham: Add deinit() instead of destructor.

timcanham (changes requested):

Os/Generic/Types/MaxHeap.cpp
```diff
@@ -51,6 +50,15 @@ void MaxHeap::create(FwSizeType capacity, Fw::ByteArray heap_allocation) {
     this->m_capacity = capacity;
 }
 
+void MaxHeap::teardown() {
+    Fw::arrayPlacementDestruct<Node>(this->m_heap, this->m_capacity);
```
thomas-bc: This would assert if called twice.

thomas-bc (approved):

timcanham (approved):

LeStarch:
@copilot, would you please write release notes indicating how to change code to support the new ID in the `create` function?

PR #4473, opened 2025-11-21
FPP v3.1.0a11

bocchino:
Bump FPP version to v3.1.0a11. This should be the final release candidate, barring any issues that come up.


bocchino:
There are two issues:
1. The tracing for the native build of fpp-to-json is not working properly.
2. On RHEL, there is a problem with {} initialization vs. () initialization. I think we should use () here -- it seems to work more consistently.

Issue #4533, opened 2025-12-09
FpySequencer function calling directives


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.1.0 |
|**_Affected Component_**| FpySequencer |
---
## Feature Description

* Change existing directives:
* STORE -> STORE_LOCAL
* STORE_CONST_OFFSET -> STORE_LOCAL_CONST_OFFSET
* LOAD -> LOAD_LOCAL
* New directives:
* STORE_GLOBAL[_CONST_OFFSET] -- uses absolute 0 offset instead of stack frame start
* LOAD_GLOBAL -- same as store global
* CALL -- jumps to an addr popped from stack, manages new stack frame, instruction ptr and frame ptr
* RETURN -- clears stack, goes back to return address leaving only return value on stack
See directives.md for specifics

## Rationale

This allows user-defined function calling, recursion, global variable access inside functions.

PR #4534, opened 2025-12-09
FpySequencer add function calling directives

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4533  |

---
## Change Description

* Adds CALL, RETURN, STORE_GLOBAL, LOAD_GLOBAL and STORE_GLOBAL_CONST_OFFSET directives
* STORE -> STORE_LOCAL
* STORE_CONST_OFFSET -> STORE_LOCAL_CONST_OFFSET
* LOAD -> LOAD_LOCAL
* Add many overflow checks
* Add UTs for some dirs which were missing UTs

## Rationale

Allows user-defined function calling and global variable access

* AI was used to implement directives, update documentation and write unit tests.


Svc/FpySequencer/test/ut/FpySequencerTester.hpp
```diff
@@ -101,10 +101,10 @@ class FpySequencerTester : public FpySequencerGTestBase, public ::testing::Test
     void add_EXIT();
     void add_ALLOCATE(Fpy::StackSizeType size);
     void add_ALLOCATE(FpySequencer_AllocateDirective dir);
-    void add_STORE_CONST_OFFSET(Fpy::StackSizeType lvarOffset, Fpy::StackSizeType size);
-    void add_STORE_CONST_OFFSET(FpySequencer_StoreConstOffsetDirective dir);
-    void add_LOAD(Fpy::StackSizeType lvarOffset, Fpy::StackSizeType size);
-    void add_LOAD(FpySequencer_LoadDirective dir);
+    void add_STORE_LOCAL_CONST_OFFSET(I32 lvarOffset, Fpy::StackSizeType size);
```
LeStarch: Why did this move to a I32, it seems like it should still be a StackSizeType

Svc/FpySequencer/test/ut/FpySequencerTester.hpp
```diff
@@ -101,10 +101,10 @@ class FpySequencerTester : public FpySequencerGTestBase, public ::testing::Test
...
+    void add_STORE_LOCAL_CONST_OFFSET(FpySequencer_StoreLocalConstOffsetDirective dir);
+    void add_LOAD_LOCAL(I32 lvarOffset, Fpy::StackSizeType size);
```
LeStarch: Same here.

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -1075,41 +1110,60 @@ Signal FpySequencer::allocate_directiveHandler(const FpySequencer_AllocateDirect
...
         // not enough bytes to pop
         error = DirectiveError::STACK_ACCESS_OUT_OF_BOUNDS;
         return Signal::stmtResponse_failure;
     }
-    Fpy::StackSizeType stackOffset = this->m_runtime.stack.lvarOffset() + directive.get_lvarOffset();
-    // if we popped these bytes off, and put them in lvar array, would we go out of bounds
-    if (stackOffset + directive.get_size() > this->m_runtime.stack.size - directive.get_size()) {
-        // write into lvar array would go out of bounds
+    // Calculate the actual address using signed offset from stack_frame_start
+    I32 lvarOffset = directive.get_lvarOffset();
```
LeStarch: We really need to be consistent on these types.   I see `Fpy::StackSizeType`, `I32`, and now `I64` for address.

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -1075,41 +1110,60 @@ Signal FpySequencer::allocate_directiveHandler(const FpySequencer_AllocateDirect
...
+    Fpy::StackSizeType newStackSize = this->m_runtime.stack.size - directive.get_size();
+    if (static_cast<Fpy::StackSizeType>(addr) + directive.get_size() > newStackSize) {
+        // write would go out of bounds
         error = DirectiveError::STACK_ACCESS_OUT_OF_BOUNDS;
         return Signal::stmtResponse_failure;
     }
-    // i believe we can be sure the regions are not overlapping, due to the above check
-    memcpy(this->m_runtime.stack.bytes + stackOffset, this->m_runtime.stack.top() - directive.get_size(),
+    // Copy value to the local variable location
+    this->m_runtime.stack.copy(static_cast<Fpy::StackSizeType>(addr),
```
LeStarch: And now a cast back w/o overflow checking.

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -1303,24 +1355,224 @@ Signal FpySequencer::peek_directiveHandler(const FpySequencer_PeekDirective& dir
...
+    }
+    
+    // Check target is within bounds (will also be checked at execution time)
+    if (target > m_sequenceObj.get_header().get_statementCount()) {
+        error = DirectiveError::STMT_OUT_OF_BOUNDS;
+        return Signal::stmtResponse_failure;
+    }
+    
+    // Save the return address (next instruction after CALL)
+    U32 returnAddr = this->m_runtime.nextStatementIndex;
```
LeStarch, line 1426: What if a call is the last statement?  Will this work, or will we go out-of-bounds?

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -1303,24 +1355,224 @@ Signal FpySequencer::peek_directiveHandler(const FpySequencer_PeekDirective& dir
...
+    return Signal::stmtResponse_success;
+}
+
+Signal FpySequencer::return_directiveHandler(const FpySequencer_ReturnDirective& directive, DirectiveError& error) {
+    Fpy::StackSizeType returnValSize = directive.get_returnValSize();
+    Fpy::StackSizeType callArgsSize = directive.get_callArgsSize();
+    
+    // Check we have enough bytes for the return value
+    if (this->m_runtime.stack.size < returnValSize) {
+        error = DirectiveError::STACK_ACCESS_OUT_OF_BOUNDS;
```
LeStarch: This looks like an underflow.

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -1303,24 +1355,224 @@ Signal FpySequencer::peek_directiveHandler(const FpySequencer_PeekDirective& dir
...
+    // thus the memcpy won't fail
+    
+    // Save the return value if there is one
+    U8 returnValue[Fpy::MAX_STACK_SIZE] = {};
+    if (returnValSize > 0) {
+        memcpy(returnValue, this->m_runtime.stack.top() - returnValSize, returnValSize);
+    }
+    
+    // Truncate the stack to stack_frame_start (discard all local variables)
+    if (this->m_runtime.stack.currentFrameStart > this->m_runtime.stack.size) {
```
LeStarch, line 1463: How would this ever happen?  That is, should we consider an assert here?

Also, using the same return error for every access will make debugging this harder.

Svc/FpySequencer/FpySequencerDirectives.cpp
```diff
@@ -1303,24 +1355,224 @@ Signal FpySequencer::peek_directiveHandler(const FpySequencer_PeekDirective& dir
...
+    this->m_runtime.nextStatementIndex = target;
+    
+    // Push the return address to the stack
+    this->m_runtime.stack.push<U32>(returnAddr);
+    
+    // Push the current frame pointer to the stack
+    this->m_runtime.stack.push<Fpy::StackSizeType>(this->m_runtime.stack.currentFrameStart);
+    
+    // Set the new frame pointer to the current top of stack
+    this->m_runtime.stack.currentFrameStart = this->m_runtime.stack.size;
```
LeStarch, line 1438: Where do you put on the arguments?

LeStarch (commented):
First and foremost there is a lot of types confusion:

1. `StackSizeType`, `I32`, `I64`.  Can offsets and stack sizes all be represented by `StackSizeType`?  My guess is it is unsigned, but does it need to be?

2. I see a lot of copy from stack and update / copy to stack and update.  Why not make these helpers?

Overall, the types is the hardest thing.

zimri-leisher:
@LeStarch thanks for the review, but my apologies, this PR wasn't quite ready for a review yet, I just intended for it to be a draft. Hoping to finish tonight, and I will address your review comments

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: I32 is used when we want to be able to resolve negative offsets in the stack. Previously, I didn't need this because the offsets were resolved relative to 0 (stack start) and so there couldn't be anything before it. Now, there are occasionally things before the start of the stack frame that we need to get the value of--specifically, the function arguments. So the store/load directives need to be able to handle this. You're right, it's not consistent, and it does lead to the case in which there are some offsets in the stack which are not representable with the new I32 (specifically, offsets greater than i32 max). However, this is not a concern, b/c i32 max is 2 GB, and I would say if your stack is 2 GB you have other issues.

zimri-leisher: Some ideas are: we could change StackSizeType to I32, which would mean this would be consistent. However, that would cause many other instructions to have problems with invalid negative sizes.
Another idea is we could add a boolean to the Store/Load directives that specify positive or negative. This is a waste of 7 bits though.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: Good question, if nextStatementIndex is one past the last statement index, the sequence will end. See RunState around line 20. I think this is the correct behavior.

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: The arguments are put on the stack by the caller--they are already there by the time we get to the CALL directive, so we don't have to handle them.

Then we set the frame pointer to the top of the stack, so the args are at a negative offset relative to the frame pointer. This is why we need the I32 offset for store/load

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: Yes, at the moment I don't have a separate error for UNDERFLOW. I used to have it, but sometimes it was a bit ambiguous whether something was OOB or underflow. Open to adding it back but then we would want to go look at all other dirs and see if they're an underflow or OOB. If you can help me define which is which then I can do it.

Svc/FpySequencer/FpySequencerStack.cpp
```diff
@@ -123,7 +123,7 @@ void FpySequencer::Stack::pop(U8* dest, Fpy::StackSizeType destSize) {
 // leaves the source array unmodified
 // does not convert endianness
 void FpySequencer::Stack::push(U8* src, Fpy::StackSizeType srcSize) {
-    FW_ASSERT(this->size + srcSize < Fpy::MAX_STACK_SIZE, static_cast<FwAssertArgType>(this->size),
+    FW_ASSERT(this->size + srcSize <= Fpy::MAX_STACK_SIZE, static_cast<FwAssertArgType>(this->size),
```
zimri-leisher, line 126: this was a mistake, should have been = allowed

Svc/FpySequencer/test/ut/FpySequencerTester.hpp
zimri-leisher: See other comments

zimri-leisher:
@LeStarch I believe most of your feedback should be resolved. I had Claude (AI) take a look at other directives and there seemed to be other places where overflows were possible which we missed. I added unit tests for these situations and fixed them.

This update has been tested with the compiler and works, so now I'd say both PRs are ready to go

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: This could happen if the compiler messes up and pops more than it should off of the top of the stack. E.g. if it said ADD when currFrameStart==stack.size, then stack,size would go down, leaving currentFrameStart invalid. We don't check this constantly because it's not always relevant. We only check that the currFrameStart ptr is valid before we use it or modify it. It's definitely not cut and dry that this is the best way but I think it's reasonable

Svc/FpySequencer/FpySequencerDirectives.cpp
LeStarch: I think we should add an UNDERFLOW status just to help track the distinction.

LeStarch:
@zimri-leisher I think we need to pin-down the types here still. If we are going to use a named type for stack size, we should use a named type for offset.  Otherwise it is very confusing to handle. We could:

For now, I think it is easiest to define `SignedStackType` and set it to `I32` and move on.

LeStarch (commented):
I've also noticed that return addresses are in the form `U32`.

zimri-leisher:
Yes, addresses are not stack offsets or stack sizes, so they are a separate type. It's really just an index into an array of directives. I could make it an alias I guess.

> For now, I think it is easiest to define SignedStackType and set it to I32 and move on.

Okay, so a StackSizeType U32 and SignedStackSizeType I32?

Svc/FpySequencer/FpySequencerDirectives.cpp
zimri-leisher: Okay, can do.

zimri-leisher:
@LeStarch  i responded to your feedback

LeStarch (dismissed):

zimri-leisher:
@LeStarch looks like builds are failing for unrelated reasons. I pulled in devel, hopefully that solves it

zimri-leisher:
Had a shower thought that STORE_GLOBAL should really just be STORE_ABS and STORE_LOCAL should be STORE_REL. Made that change, purely a renaming change.

LeStarch (approved):

PR #4548, opened 2025-12-12
Prevent divide-by-zero in PosixFileSystem::_getFreeSpace

### LINKED ISSUE #4500: Divide-by-zero in Os::Posix::FileSystem::PosixFileSystem::_getFreeSpace
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**| OS/Posix |
---
Follow-on from #4296 which gathered static analysis findings.

## Problem Description

Os::Posix::FileSystem::PosixFileSystem::_getFreeSpace will divide by zero if block_size is zero in this code snippet:
```
    const FwSizeType block_size = static_cast<FwSizeType>(fsStat.f_frsize);
    const FwSizeType free_blocks = static_cast<FwSizeType>(fsStat.f_bfree);
    const FwSizeType total_blocks = static_cast<FwSizeType>(fsStat.f_blocks);

    // Check for overflow in multiplication
    if (free_blocks > (std::numeric_limits<FwSizeType>::max() / block_size) ||
        total_blocks > (std::numeric_limits<FwSizeType>::max() / block_size)) {
        return OVERFLOW_ERROR;
    }
```

Suggest adding a check in the code to return an error if fsStat.f_frsize is zero.  Is there a more general sanity check on what valid values fsStat could hold?

## How to Reproduce

1. Run CodeSonar using the JPL CAE infrastructure. See issue #4296.
2. Search for this specific warning:
```
DJ-caeleng% grep Division .scrub/codesonar.scrub 
codesonar004 <High> :lib/fprime/Os/Posix/FileSystem.cpp:86: Division By Zero
    	Division By Zero found in Os::Posix::FileSystem::PosixFileSystem::_getFreeSpace
```

## Expected Behavior

No such findings should be reported.

heathdutton:
|    |    |
|:---|:---|
|**_Related Issue(s)_**| Fixes #4500 |

---
## Change Description

Add a guard to check for zero `block_size` before performing division operations in `Os::Posix::FileSystem::PosixFileSystem::_getFreeSpace`. Returns `OTHER_ERROR` if the filesystem reports an invalid (zero) fragment size.

## Rationale

Static analysis (CodeSonar) identified a potential divide-by-zero vulnerability. The overflow check at lines 91-92 divides by `block_size` without first verifying it is non-zero. While `f_frsize` should never be zero on a properly functioning filesystem, this defensive guard ensures robustness against edge cases and satisfies static analysis requirements.

- The edge case can't be triggered without mocking `statvfs` to simulate some radiation-induced memory corruption (changing f_frsize from 0x1000 to 0 in a bit flip)
- CI should verify compilation with strict warnings (`-Wall -Wextra -Werror`)
- Review that `OTHER_ERROR` is the appropriate status for invalid filesystem response

### DIFF
--- a/Os/Posix/FileSystem.cpp
+++ b/Os/Posix/FileSystem.cpp
@@ -82,6 +82,11 @@ PosixFileSystem::Status PosixFileSystem::_getFreeSpace(const char* path,
     const FwSizeType free_blocks = static_cast<FwSizeType>(fsStat.f_bfree);
     const FwSizeType total_blocks = static_cast<FwSizeType>(fsStat.f_blocks);
 
+    // Guard against divide-by-zero if filesystem reports invalid block size
+    if (block_size == 0) {
+        return OTHER_ERROR;
+    }
+
     // Check for overflow in multiplication
     if (free_blocks > (std::numeric_limits<FwSizeType>::max() / block_size) ||
         total_blocks > (std::numeric_limits<FwSizeType>::max() / block_size)) {


Os/Posix/FileSystem.cpp
```diff
@@ -82,6 +82,11 @@ PosixFileSystem::Status PosixFileSystem::_getFreeSpace(const char* path,
     const FwSizeType free_blocks = static_cast<FwSizeType>(fsStat.f_bfree);
     const FwSizeType total_blocks = static_cast<FwSizeType>(fsStat.f_blocks);
 
+    // Guard against divide-by-zero if filesystem reports invalid block size
+    if (block_size == 0) {
```
djbyrne17, line 86: I'm a fan of (0 == block_size) but this is not wrong.

djbyrne17:
The POSIX statvfs() standard does not require good behavior for f_frsize, so this is a good check to maintain portability.

djbyrne17:
OTHER_ERROR seems a good choice, since any more specific error should have resulted in returning after the statvfs call.  This sanity check for a zero block_size means we are running on an unanticipated (or mis-behaving) platform.

djbyrne17 (approved):
Good to merge

LeStarch (approved):

Issue #4444, opened 2025-11-19
Add FileDispatch port to CmdSequencer


timcanham:
| | |
|:---|:---|
|**_F´ Version_**| https://github.com/nasa/fprime/releases/tag/v4.0.1a1|
|**_Affected Component_**| https://github.com/nasa/fprime/tree/devel/Svc/CmdSequencer |
---
## Feature Description

Add a port to `Svc/CmdSequencer` that that accepts an uplinked sequence binary and automatically runs it.

## Rationale

Provide for automatic execution of sequence files to minimize ground involvement after an uplink. See https://github.com/nasa/fprime/discussions/4390

zimri-leisher:
Can you scope adding this to FpySeq too? Maybe this PR could also create a CmdSequencer.fppi interface

timcanham:
Merge https://github.com/nasa/fprime/pull/4552 first.

timcanham:
Initial CI run to clear any issues. Still needs an SDD update.

Issue #4443, opened 2025-11-19
Implement FileDispatcher component


timcanham:
| | |
|:---|:---|
|**_F´ Version_**| https://github.com/nasa/fprime/releases/tag/v4.0.1a1|
|**_Affected Component_**| New Component |
---
## Feature Description

Implement a component that dispatches newly uplinked files for automatic processing without ground intervention. This will accept a user table that maps file extension to an output port.

There are two anticipated use cases so far:

1) Automatic sequence running. Based a compiled sequence file extenstion, a sequence can automatically be run.
2) Forward file bundles. File bundles will allow for a collection of files to be uplinked all at the same time, minimizing ground commands.

## Rationale

Provide for automatic handling of files after upling is done to provide some automation without ground intervention. See https://github.com/nasa/fprime/discussions/4390 and 
https://github.com/nasa/fprime/issues/4419.

PR #4552, opened 2025-12-15
Add Svc.FileDispatcher component

timcanham:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/4443  |

---
## Change Description

Adds a new Svc/FileDispatcher component for dispatching newly uplinked files

## Rationale

Provides automation for automatically handling files like sequences.

Unit tests are provided which test some entries

## Future Work

https://github.com/nasa/fprime/issues/4444 - this issue will add a port to Svc/CmdSequencer to automatically run sequences.

Git Copilot provided suggestions while writing the code.


Svc/FileDispatcher/FileDispatcher.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+FileDispatcher ::FileDispatcher(const char* const compName) : FileDispatcherComponentBase(compName) {
+    // disable entries
+    for (FwSizeType i = 0; i < Svc::FileDispatcherCfg::FileDispatchPort::MAX_FILE_DISPATCH_PORTS; i++) {
+        this->m_dispatchTable[i].enabled = false;
+    }
+}
+
+FileDispatcher ::~FileDispatcher() {
+    this->deinit();
```
LeStarch: Why was `deinit` added to the destructor?  This should be called from the topology / UT harness.

Svc/FileDispatcher/FileDispatcher.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+void FileDispatcher ::configure(FileDispatcherEntry* entries, FwSizeType numEntries) {
```
LeStarch: Typically we prefer a struct that has a fixed size array inside.

Svc/FileDispatcher/FileDispatcher.cpp
```diff
@@ -0,0 +1,84 @@
...
+// Handler implementations for typed input ports
+// ----------------------------------------------------------------------
+
+void FileDispatcher ::fileAnnounceRecv_handler(FwIndexType portNum, Fw::StringBase& file_name) {
+    // determine file extension and dispatch accordingly
+
+    // walk table to find match
+    for (FwIndexType i = 0; i < Svc::FileDispatcherCfg::FileDispatchPort::MAX_FILE_DISPATCH_PORTS; i++) {
+        if (this->m_dispatchTable[i].enabled) {
+            if (Fw::StringUtils::substring_find(file_name.toChar(), file_name.length(),
```
LeStarch: This will search for the pattern "fileExt" anywhere in the filename.   This really should do a `substring_find_last` (https://github.com/nasa/fprime/blob/1a097ade00279e7dc885c19de9117f9fa31001e8/Fw/Types/StringUtils.cpp#L77) and confirm the index is `file.length()` - `ext.length()`.

Finally, this is suffix driven, not extension driven as it is not enforcing "."

Svc/FileDispatcher/FileDispatcher.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+    //! Handler implementation for pingIn
+    //!
+    //! Ping in
+    void pingIn_handler(FwIndexType portNum,  //!< The port number
+                        U32 key               //!< Value to return to pinger
+                        ) override;
+
+    //! table of dispatch entries
+    FileDispatcherEntry m_dispatchTable[Svc::FileDispatcherCfg::FileDispatchPort::MAX_FILE_DISPATCH_PORTS];
```
LeStarch: I don't understand why this is limited to the port length.  It seems like multiple file extensions could refer to the same port, which would mean this should be "MAX_DISPATCH_EXTENSIONS", or there is exactly one file per port.....at which point, it makes most sense to me to expect the input config to be well-ordered....(as does this PR) and as such, dispatch port is unnecessary.

Svc/FileDispatcher/FileDispatcher.fpp
```diff
@@ -0,0 +1,64 @@
...
+
+        ##############################################################################
+        #### Uncomment the following examples to start customizing your component ####
+        ##############################################################################
+
+        # @ File type dispatch enabled/disabled event
+        event FileDispatchState(file_type: Svc.FileDispatcherCfg.FileDispatchPort, enabled: Fw.Enabled) severity activity high format "File dispatch {} state changed: to {}"
+
+        # @ File dispatched event
+        event FileDispatched(file_name: string size 80, file_type: Svc.FileDispatcherCfg.FileDispatchPort) severity activity high format "File {} dispatched to {}"
```
LeStarch: Use the file name size constant instead of `80`

Svc/FileDispatcher/FileDispatcher.cpp
```diff
@@ -0,0 +1,84 @@
...
+                                                this->m_dispatchTable[i].fileExt.toChar(),
+                                                this->m_dispatchTable[i].fileExt.length()) != -1) {
+                // dispatch on this port
+                this->fileDispatch_out(this->m_dispatchTable[i].port, file_name);
```
LeStarch: This doesn't work.  Right now, the configuration table is expected to be in-order because you index directly into it on line 74.   Thus, independently using `.port` for which port to dispatch does not make since.

LeStarch (changes requested):
I really think you need 2 data structures here:

1. Mapping for EXT to port
2. List of ports to disable/enable

Right now the code looks like you are trying to separate the concepts, but cannot because of the singular data structure.

Either remove "port" from the data structure, make enable/disable a loop through all table entrys matching on port type, or split them up.

Svc/FileDispatcher/FileDispatcher.cpp
timcanham: I thought it could be called from either or both. If the new pattern is to remove it, I can do that.

LeStarch: It should only be called in the destructor of the Tester.

```c++
~Tester() {
   this->m_component.deinit();
}
```

It should be called externally from the component itself.

Svc/FileDispatcher/FileDispatcher.fpp
timcanham: Done.

Svc/FileDispatcher/FileDispatcher.hpp
timcanham: Good idea. See the new implementation I just pushed. I decoupled the table size from the number of ports.

Svc/FileDispatcher/FileDispatcher.cpp
timcanham: I updated the SDD to say it is extension/suffix, and used the `find_last` function.

Svc/FileDispatcher/FileDispatcher.cpp
timcanham: Fixed!

Svc/FileDispatcher/FileDispatcher.cpp
timcanham: Fixed!

Svc/FileDispatcher/FileDispatcher.cpp
timcanham: I think this is fixed with the new design?

Svc/FileDispatcher/FileDispatcher.fpp
```diff
@@ -0,0 +1,64 @@
+module Svc {
+    @ Component to dispatch uplinked files to various services
+    active component FileDispatcher {
+
+        # @ disable dispatch of a file type
+        async command ENABLE_DISPATCH(
+                file_type: Svc.FileDispatcherCfg.FileDispatchPort @< the file type dispatch to enable/disable
```
LeStarch, line 7: Good use of the configurable enumeration pattern!

Svc/FileDispatcher/FileDispatcher.fpp
```diff
@@ -0,0 +1,64 @@
...
+        event FileDispatchState(file_type: Svc.FileDispatcherCfg.FileDispatchPort, enabled: Fw.Enabled) severity activity high format "File dispatch {} state changed: to {}"
+
+        # @ File dispatched event
+        event FileDispatched(file_name: string size FileNameStringSize, file_type: Svc.FileDispatcherCfg.FileDispatchPort) severity activity high format "File {} dispatched to {}"
+
+        ###############################################################################
+        # Standard AC Ports: Required for Channels, Events, Commands, and Parameters  #
+        ###############################################################################
+        @ Port for requesting the current time
+        time get port timeCaller
```
LeStarch, line 26: In the future, prefer standard interfaces here.  No need to change now.

Svc/FileDispatcher/FileDispatcher.cpp
```diff
@@ -0,0 +1,87 @@
...
+    FW_ASSERT(table.numEntries <= Svc::FileDispatcherCfg::FILE_DISPATCHER_MAX_TABLE_SIZE);
+    for (FwSizeType entry = 0; entry < table.numEntries; entry++) {
+        FW_ASSERT(table.entries[entry].port.isValid(),
+                  table.entries[entry].port.e);  // valid output port
+        FW_ASSERT(table.entries[entry].fileExt.length() > 0,
+                  static_cast<FwAssertArgType>(table.entries[entry].fileExt.length()));  // non-zero length
+    }
+
+    // copy to local table
+    this->m_dispatchTable = table;
```
LeStarch: I am concerned by this copy.  I am not sure it does as expected.

LeStarch (commented):
I need to think a bit on the copy, and I need to evaluate the CodeQL warnings.

Svc/FileDispatcher/FileDispatcher.cpp
```diff
@@ -0,0 +1,80 @@
...
+void FileDispatcher ::fileAnnounceRecv_handler(FwIndexType portNum, Fw::StringBase& file_name) {
+    // determine file extension and dispatch accordingly
+
+    // walk table to find match
+    for (FwSizeType i = 0; i < this->m_dispatchTable.numEntries; i++) {
+        if (!this->m_dispatchTable.entries[i].enabled) {
+            continue;
+        }
+
+        auto loc = Fw::StringUtils::substring_find_last(file_name.toChar(), file_name.length(),
```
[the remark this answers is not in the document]
LeStarch: Will be valid based on type Fw::String.

Svc/FileDispatcher/FileDispatcher.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+// ----------------------------------------------------------------------
+// Handler implementations for commands
+// ----------------------------------------------------------------------
+
+void FileDispatcher ::ENABLE_DISPATCH_cmdHandler(FwOpcodeType opCode,
+                                                 U32 cmdSeq,
+                                                 Svc::FileDispatcherCfg::FileDispatchPort file_type,
+                                                 Fw::Enabled enable) {
+    this->m_dispatchTable.entries[file_type].enabled = enable;
```
LeStarch: This should be `file_type.e` no?

Svc/FileDispatcher/FileDispatcher.cpp
LeStarch: It will work, but copies a few more bytes than strictly necessary.

Svc/FileDispatcher/FileDispatcher.cpp
```diff
@@ -0,0 +1,87 @@
...
+        auto loc = Fw::StringUtils::substring_find_last(file_name.toChar(), file_name.length(),
+                                                        this->m_dispatchTable.entries[i].fileExt.toChar(),
+                                                        this->m_dispatchTable.entries[i].fileExt.length());
+
+        if ((loc != -1) &&  // matches at all
+            (file_name.length() - this->m_dispatchTable.entries[i].fileExt.length() ==
+             static_cast<FwSizeType>(loc))  // match at end of string
+        ) {
+            // dispatch on this port
+            this->fileDispatch_out(this->m_dispatchTable.entries[i].port, file_name);
```
kevin-f-ortega: Should use `port.e` instead of just `port`.

Svc/FileDispatcher/FileDispatcher.cpp
```diff
@@ -0,0 +1,87 @@
...
+            this->log_ACTIVITY_HI_FileDispatched(file_name,
+                                                 static_cast<Svc::FileDispatcherCfg::FileDispatchPort::T>(i));
```
kevin-f-ortega: Should emit event with `port` not `i`

Svc/FileDispatcher/FileDispatcher.cpp
kevin-f-ortega: Instead of copying the entire table, in the loop, assign per element so you don't override the default `enable = false` from the constructor.

kevin-f-ortega (changes requested):
Looks good, nice work! A few comments

Svc/FileDispatcher/docs/sdd.md
```diff
@@ -0,0 +1,91 @@
...
+|---|---|---|---|
+|FPRIME-FDISP-001|File dispatcher will provide a way to dispatch files to other components|Need to have a way to supply a file name for a newly transferred file|Unit Test/System Test|
+|FPRIME-FDISP-002|File dispatcher will provide user configuration to map file types to output ports|Projects will want to customize the dispatching types|Unit Test/System Test|
+|FPRIME-FDISP-003|File dispatcher will dispatch to specified output ports based on the supplied table|Projects will want to connect the dispatch types to components for processing new files|Unit Test/System Test|
+
+## Usage Examples
+
+### Typical Usage
+
+The `Ref` example connects `Svc::FIleUplink` to `FileDispatcher` to process uplinked files. The `Ref` example configuration table has compiled sequences with a certain extension that `FileDispatcher` dispatches to `Svc::CmdSequencer` for automatic running when those files are uplinked.
```
kevin-f-ortega: Typo `FIleuplink`

LeStarch (dismissed):

Svc/FileDispatcher/FileDispatcher.cpp
```diff
@@ -74,7 +76,12 @@ void FileDispatcher ::ENABLE_DISPATCH_cmdHandler(FwOpcodeType opCode,
                                                  U32 cmdSeq,
                                                  Svc::FileDispatcherCfg::FileDispatchPort file_type,
                                                  Fw::Enabled enable) {
-    this->m_dispatchTable.entries[file_type].enabled = enable;
+
+    for (FwSizeType i = 0; i < this->m_dispatchTable.numEntries; i++) {
+        if (this->m_dispatchTable.entries[i].port == file_type) {
+            this->m_dispatchTable.entries[i].enabled = enable;
```
kevin-f-ortega: This should be `this->m_dispatchTable.entries[i].enabled = (enable == Fw::Enabled::ENABLED);`

kevin-f-ortega (commented):
Looks good, just one minor comment

Svc/FileDispatcher/FileDispatcher.cpp
LeStarch: yes, yes it should

kevin-f-ortega (approved):

Issue #4560, opened 2025-12-16
Make TimeComparison Fpp enum


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.1.0 |
|**_Affected Component_**| Fw.Time |
---
## Feature Description

Factor the Comparison enum in Fw::Time out of CPP into FPP
## Rationale

This will allow the FpySequencer to use the enum types

PR #4561, opened 2025-12-16
Switch to FPP enum for time comparisons

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4560 |

---
## Change Description

* Switch to using an Fpp enum instead of a CPP enum for TimeComparison results
## Rationale

* This allows the time comparison result to be used in Fpy

None

### DIFF
--- a/Fw/Time/Time.cpp
+++ b/Fw/Time/Time.cpp
@@ -52,29 +52,29 @@ Time& Time::operator=(const Time& other) {
 }
 
 bool Time::operator==(const Time& other) const {
-    return (Time::compare(*this, other) == EQ);
+    return (Time::compare(*this, other) == TimeComparison::EQ);
 }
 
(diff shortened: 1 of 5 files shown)


LeStarch:
Conceptually I get that this will allow the result of a time comparison to be used in Fpy as it will now be a dictionary type.

However, time comparison really wasn't intended to be used outside flight software.  It is a utility enum to make code a little cleaner to read.  Now we've taken that utility enum and wrapped it in the overhead of an code-generated class.  This seems to bring in more complexity than was intended for this type.

At the very least, we need to review this change in the CCB.

zimri-leisher:
Okay, makes sense that we'd like to review it. Actually I didn't make this a dictionary enum so I think I still have to update that.

LeStarch (approved):

Issue #4524, opened 2025-12-05
Version.cpp multiple static analysis findings


djbyrne17:
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**| Svc/Version |
---
Follow-on from #4296 which gathered static analysis findings.

## Problem Description

File `Version.cpp` has 1 of the most severe SonarQube findings in fprime.  Fix that, and opportunistically some less severe findings.

Also, more assertions would be useful, but that is not in scope of this ticket.

### At least the SonarQube "Blocker" finding

```
<High> Line 50: cpp:S5912 Use pointer or reference to avoid slicing from "ExternalString" to "StringBase".
```

### Other SonarQube findings as make sense, and as time allows

```
<High> Line 29: cpp:S3490 Use "=default" instead of the default implementation of this special member functions.

<Med> Line 115: cpp:S5817 This function should be declared "const".
<Med> Line 116: cpp:S1110 Remove these redundant parentheses.
<Med> Line 118: cpp:S1110 Remove these redundant parentheses.
<Med> Line 123: cpp:S5817 This function should be declared "const".
<Med> Line 131: cpp:S5817 This function should be declared "const".
<Med> Line 19: cpp:S3230 Do not use the constructor's initializer list for data member "m_enable". Use the in-class initializer instead.
<Med> Line 19: cpp:S3230 Do not use the constructor's initializer list for data member "m_num_custom_elements". Use the in-class initializer instead.
<Med> Line 59: cpp:S5827 Replace the redundant type with "auto".
<Med> Line 71: cpp:S1238 Pass polymorphic object "enable" by reference to const.
<Med> Line 78: cpp:S1238 Pass polymorphic object "version_type" by reference to const.
<Med> Line 82: cpp:S1110 Remove these redundant parentheses.
<Med> Line 86: cpp:S1110 Remove these redundant parentheses.
<Med> Line 90: cpp:S1110 Remove these redundant parentheses.
<Med> Line 94: cpp:S1110 Remove these redundant parentheses.
<Med> Line 98: cpp:S1110 Remove these redundant parentheses.

<Low> Line 22: cpp:S5566 Change this raw for-loop to a range for-loop or an "std::for_each".
```

### Other CodeSonar findings as make sense, and as time allows

```
<Med> Line 1: Not All Warnings Are Enabled
<Med> Line 1: Warnings Not Treated As Errors
<Med> Line 109: Unchecked Parameter Dereference
[...]

PR #4563, opened 2025-12-16
Fix object slicing and modernize types in Version.cpp

JuanPS999:
**PR Title:** `Fix object slicing and modernize types in Version.cpp`

|  |  |
| --- | --- |
| ***Related Issue(s)*** | #4524 |
| ***Has Unit Tests (y/n)*** | n |
| ***Documentation Included (y/n)*** | n |
| ***Generative AI was used in this contribution (y/n)*** | y |

---

## Change DescriptionAddresses static analysis findings in `Svc/Version/Version.cpp`:

* Fixes object slicing when retrieving the version string by using a `static_cast` to `const Fw::StringBase&`.
* Modernizes variable declaration by using `auto` to avoid redundant type specification.

## Rationale
This PR resolves a "Blocker" severity finding (`cpp:S5912`) identified by SonarQube regarding object slicing, where `ExternalString` was being sliced to `StringBase` implicitly. It also resolves a "Medium" severity finding (`cpp:S5827`) by cleaning up redundant type declarations.

## Testing/Review Recommendations* 
* **Unit Tests:** Run existing unit tests to ensure no regression: `fprime-util check`.
* **Static Analysis:** Verify that SonarQube/CodeSonar no longer reports `cpp:S5912` on line ~25 and `cpp:S5827` on line ~16 of `Svc/Version/Version.cpp`.

## Future Work
This PR addresses the most critical findings in Issue #4524. Other lower-priority findings (such as redundant parentheses and const correctness in getter methods) remain to be addressed in future contributions.

#### Generative AI (Gemini) was used to analyze the specific static analysis error messages (`cpp:S5912`) and suggest the correct C++ syntax (`static_cast`) to resolve the object slicing without altering the function signature.

### DIFF
--- a/Svc/Version/Version.cpp
+++ b/Svc/Version/Version.cpp
@@ -47,7 +47,7 @@ void Version ::getVersion_handler(FwIndexType portNum,
                                   Svc::VersionStatus& status) {
     FW_ASSERT(version_id.isValid(), version_id.e);
     U8 version_slot = VersionSlot(version_id.e);
-    version_string = this->verId_db[version_slot].get_version_value();
+    version_string = static_cast<const Fw::StringBase&>(this->verId_db[version_slot].get_version_value());
     status = this->verId_db[version_slot].get_version_status();
 }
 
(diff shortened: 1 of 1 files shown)


LeStarch:
@djbyrne17 would you take a look at this PR?

LeStarch:
Also, @djbyrne17 .... is it possible this is a false-positive.  I don't see how the compiler might be converting a reference to a value, and then back to a reference.  My loose understanding is that it shouldn't

While I agree that this should fix the problem, I still don't see how/why the problem arises with the exception that this is a place where this problem is more likely to arise in general.

LeStarch (approved):
This looks like the right change to me.  I asked the issue author to take a look too.

Issue #4411, opened 2025-11-11
Os::Test::FileTest::Tester::OpenBaseRule::action() can cause unit test to hang (infinite loop)


bevinduckett:
| | |
|:---|:---|
|**_F´ Version_**| Any commit after [2b65cc83cfdece0f6209fbff005a9cfdc739d91d ](https://github.com/nasa/fprime/commit/c7048d0b9c4f02473a947914374192f446372591) on devel branch |
|**_Affected Component_**|   Os/test/ut |
---
## Problem Description

The unbounded loop in `Os::Test::FileTest::Tester::OpenBaseRule::action()` (in `Os/test/ut/file/FileRules.cpp`) can cause a unit test to hang  if the `get_filename` function provided by a file-system unit test can't come up with a new unique name. 

```C
    // When randomly generating filenames, some seeds can result in duplicate filenames
    // Continue generating until unique, unless this is an overwrite test
    if (this->m_random && !this->m_overwrite) {
        while (state.exists(*filename)) {
            filename = state.get_filename(this->m_random);
        }
    }
```

It's on the client to update `get_filename()` so that it can provide a unique name, but `Os::Test::FileTest::Tester::OpenBaseRule::action()` should be updated to fail and report why. I've modified my version to an arbitrary limit of 100 K, after which it logs an error and breaks out of the loop. 
```C
    // When randomly generating filenames, some seeds can result in duplicate filenames
    // Continue generating until unique, unless this is an overwrite test
    U32 i = 0;
    if (this->m_random && !this->m_overwrite) {
        while (state.exists(*filename)) {
            filename = state.get_filename(this->m_random);
            i++;
            if (i > 100000){
                ASSERT_TRUE(false) << "ERROR giving up finding an unique name after " << i << " attempts";
                break;
            }
        }
    }
```

## Context / Environment

Execute `fprime-util version-check` and share the output.

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-4.18.0-553.77.1.el8_10.x86_64-x86_64-with-glibc2.28
Python version: 3.9.20
CMake version: 3.26.0
Pip version: 25.3
Pip packages:
    fprime-tools==4.0.2a1
    fprime-gds==4.0.2a9
    fprime-fpp==3.1.0a8
Project submodules:
    https://github.com/nasa/fprime.git @ v4.0.1a1-76-g2b65cc83c
```

## How to Reproduce
[...]

PR #4566, opened 2025-12-17
Prevent infinite loop in FileTest unique filename generation

heathdutton:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4411 |

---
## Change Description

Add an iteration limit to the `OpenBaseRule::action()` method in `Os/test/ut/file/FileRules.cpp`. The while loop that generates unique filenames now fails with a clear assertion after 100,000 attempts instead of potentially looping forever.

## Rationale

When `get_filename()` cannot produce unique names (e.g., limited filename space in custom file system implementations), the existing unbounded while loop causes unit tests to hang indefinitely. This makes debugging difficult since there's no indication of what went wrong.

  - All existing `PosixFileTest` and `StubFileTest` tests pass
  - The limit of 100,000 is consistent with `DEPTH_BOUND` in `Os/test/ut/queue/RulesHeaders.hpp`
  - The fix is defensive—normal test execution won't approach this limit
  
## Future Work

NA


LeStarch:
@bevinduckett would you glance at this?

LeStarch (approved):

bevinduckett:
Looks good to me

PR #4570, opened 2025-12-17
Fix/missing priority

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes minor bugs in fprime:

1. Add `aggregator` priority
2. Fix FileManager for less drops

### DIFF
--- a/Svc/FileManager/FileManager.cpp
+++ b/Svc/FileManager/FileManager.cpp
@@ -34,7 +34,8 @@ FileManager ::FileManager(const char* const compName  //!< The component name
       m_listState(IDLE),
       m_totalEntries(0),
       m_currentOpCode(0),
-      m_currentCmdSeq(0) {}
+      m_currentCmdSeq(0),
+      m_runQueued(false) {}
 
 FileManager ::~FileManager() {}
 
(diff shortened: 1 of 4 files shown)


Svc/FileManager/FileManager.cpp
```diff
@@ -204,6 +205,17 @@ void FileManager ::pingIn_handler(const FwIndexType portNum, U32 key) {
 }
 
 void FileManager ::schedIn_handler(const FwIndexType portNum, U32 context) {
+    bool isQueued = false;  // We only update when nothing is queued
+    // Check for false (no message queued) and if false, atomically make it true
+    (void)this->m_runQueued.compare_exchange_weak(isQueued, true);
```
thomas-bc, line 210: Why not use `compare_exchange_strong` ? `schedIn`'s rate might be slow enough that this could matter ?

LeStarch: In this case, `compare_exchange_weak` is more flexible/portable as it can be implemented on a larger swath of hardware.  Since the failure to compare would only result in a wait for the next RG tick (and might delay an event by 1 tick), I deemed it a suitable for the additional flexibility.

thomas-bc (approved):

PR #4579, opened 2025-12-19
Fix/weak strong

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes crash due to `_weak` failing. Thanks @thomas-bc for the solution.  Add CRC for file command.

### DIFF
--- a/Svc/FileManager/FileManager.cpp
+++ b/Svc/FileManager/FileManager.cpp
@@ -199,6 +199,26 @@ void FileManager ::ListDirectory_cmdHandler(const FwOpcodeType opCode,
     // Command response will be sent when listing completes.
 }
 
+void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, const Fw::CmdStringArg& filename) {
+    Os::File file;
+    U32 crcValue = 0;
+    this->log_ACTIVITY_HI_CalculateCrcStarted(filename);
+
+    Os::File::Status status = file.open(filename.toChar(), Os::File::OPEN_READ);
+    if (status == Os::File::OP_OK) {
+        status = file.calculateCrc(crcValue);
+    }
+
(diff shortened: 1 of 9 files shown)


Svc/FileManager/FileManager.cpp
```diff
@@ -207,7 +227,7 @@ void FileManager ::pingIn_handler(const FwIndexType portNum, U32 key) {
 void FileManager ::schedIn_handler(const FwIndexType portNum, U32 context) {
     bool isQueued = false;  // We only update when nothing is queued
     // Check for false (no message queued) and if false, atomically make it true
-    (void)this->m_runQueued.compare_exchange_weak(isQueued, true);
+    (void)this->m_runQueued.compare_exchange_strong(isQueued, true);
     if (not isQueued) {
```
bocchino: The logic here is a bit hard to follow. I think what is happening is that `isQueued` will be changed from `false` to `true` if and only if the `compare_exchange` operation fails. But in that case the `compare_exchange` operation itself should return `false`. It may be clearer to check the return value of `compare_exchange` instead of casting it to `void` and checking `isQueued`. A comment explaining the logic would also help.

From cppreference.com:
>bool compare_exchange_strong( T& expected, T desired

>Atomically compares the [object representation](https://en.cppreference.com/w/cpp/language/objects.html)(until C++20)[value representation](https://en.cppreference.com/w/cpp/language/objects.html)(since C++20) of *this with that of expected. If those are bitwise-equal, replaces the former with desired (performs read-modify-write operation). Otherwise, loads the actual value stored in *this into expected (performs load operation).

>Return value
true if the underlying atomic value was successfully changed, false otherwise.

bocchino (approved):
Looks good!

thomas-bc (approved):

PR #4592, opened 2025-12-24
Correct TmFramer size assertion math

Willmac16:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**| N  |
|**_Documentation Included (y/n)_**|  N |

---
## Change Description

The math in TmFramer's assertions was corrected to prevent runtime assertions

## Rationale

The math on the TmFramer's size assertions allowed for runtime assertions during file downlink.

```bash
Assert: "Svc/Ccsds/TmFramer/TmFramer.cpp:116" 2
FATAL 5761 handled.
Exiting with abort signal and core dump file.
```

Make sure you believe File Downlink works right at the limit of the new assertion

## Future Work

I am working on an AOS framer that can do many PDU per one Frame and many Frames per one PDU. The changes could either be ported to TM or we could move the fprime ccsds stackup over to AOS at the point that is done.

Nope

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,21 @@
 repos:
-    -   repo: https://github.com/psf/black
-        rev: 19.3b0
-        hooks:
-        -   id: black
-            name: Format Python Code (black) in Fw/Python/
-            files: '^Fw/Python/'
-        -   id: black
-            name: Format Python Code (black) in Gds/
-            files: '^Gds/'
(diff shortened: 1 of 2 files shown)


LeStarch (dismissed):
LGTM, but I'd like @thomas-bc to weigh-in too!

thomas-bc (approved):

LeStarch (approved):

LeStarch:
Thanks @Willmac16!

Issue #4604, opened 2026-01-07
Allow file creation mode parameter to be configured for Posix/File.cpp


kubiak-jpl:
| | |
|:---|:---|
|**_F´ Version_**| 4.1.1 |
|**_Affected Component_**| Os/Posix/File.cpp |
---
## Feature Description

Allow the file mode parameters in the open syscall to be configured per project, instead of hardcoded in the cpp file.

A further improvement would be to expose this parameter to the caller so that different components could set the file creation mode as needed.

## Rationale

The default values result in a file permissions of 600, which may not be appropriate for all projects. I specifically would have preferred 666 or 664 for a recent project.

thomas-bc:
https://github.com/nasa/fprime/pull/4613/changes

PR #4613, opened 2026-01-13
Expose default file create permissions through a new OsCfg.fpp config file

kubiak-jpl:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4604 |

---
## Change Description

Creates an OsCfg.fpp configuration file with a default file creation mode constant. Os/Posix/File is updated to use this constant when creating new files. This allows the F Prime developer to control permissions on files created with Os::File, at least globally for the deployment

## Rationale

This is a fix that would have been useful for a previous F Prime project, where the default 600 permissions caused operational headaches when the files were created by root, but access by a normal user

A small bit of test code was added to the reference topology to create files with this interface. All defined FILE_MODE_ bits were tested in one form or another on a Debian 13 system. The code was also compiled under Darwin and files were created with the expected permissions.

```
diff --git a/Ref/Top/RefTopology.cpp b/Ref/Top/RefTopology.cpp
index 6d18f751d..389f120f3 100644
--- a/Ref/Top/RefTopology.cpp
+++ b/Ref/Top/RefTopology.cpp
@@ -53,6 +53,10 @@ void configureTopology() {
 
     // Command sequencer needs to allocate memory to hold contents of command sequences
     cmdSeq.allocateBuffer(0, mallocator, 5 * 1024);
+
+    Os::File test_file;
+    test_file.open("test_file.txt", Os::FileInterface::OPEN_CREATE);
+    test_file.close();
 }
```

No testing was done on other operating systems. In particular, I am not sure how the VxWorks, FreeRTOS or Zephr OS layers will behave if they use the Os/Posix abstraction for File.

## Future Work

I would also like to change the default creation mode from 600 to 666. That is the default used by multiple unix tools (touch, vim). End users can drop permissions by changing the config constant or by using umask with their application.

No AI assistance used.

### DIFF
--- a/Os/Posix/File.cpp
+++ b/Os/Posix/File.cpp
@@ -17,18 +17,6 @@ namespace Os {
 namespace Posix {
 namespace File {
 
-// Sets up the default file permission as user read + user write
-// Some posix systems (e.g. Darwin) use the older S_IREAD and S_IWRITE flags while other systems (e.g. Linux) use the
-// newer S_IRUSR and S_IWUSR flags, and some don't support these flags at all. Hence, we look if flags are defined then
-// set USER_FLAGS to be the set of flags supported or 0 in the case neither is defined.
-#if defined(S_IREAD) && defined(S_IWRITE)
-#define USER_FLAGS (S_IREAD | S_IWRITE)
-#elif defined(S_IRUSR) && defined(S_IWUSR)
-#define USER_FLAGS (S_IRUSR | S_IWUSR)
-#else
-#define USER_FLAGS (0)
-#endif
-
 // O_SYNC is not defined on every system. This will set up the SYNC_FLAGS variable to be O_SYNC when defined and
 // (0) when not defined. This allows OPEN_SYNC_WRITE to fall-back to OPEN_WRITE on those systems.
 #if defined(O_SYNC)
@@ -78,6 +66,36 @@ PosixFile& PosixFile::operator=(const PosixFile& other) {
     return *this;
 }
 
+mode_t PosixFile::map_open_create_mode(const U32 create_mode) {
+    mode_t out_mode = 0;
+
+    // Some posix systems (e.g. Darwin) use the older S_IREAD and S_IWRITE flags
+    // while other systems (e.g. Linux) use the newer S_IRUSR and S_IWUSR flags.
+#if defined(S_IREAD)
+    out_mode |= (create_mode & Os::FILE_MODE_IRUSR) ? S_IREAD : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_IWUSR) ? S_IWRITE : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_IXUSR) ? S_IEXEC : 0;
+#else
+    out_mode |= (create_mode & Os::FILE_MODE_IRUSR) ? S_IRUSR : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_IWUSR) ? S_IWUSR : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_IXUSR) ? S_IXUSR : 0;
+#endif
+
+    out_mode |= (create_mode & Os::FILE_MODE_IRGRP) ? S_IRGRP : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_IWGRP) ? S_IWGRP : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_IXGRP) ? S_IXGRP : 0;
+
+    out_mode |= (create_mode & Os::FILE_MODE_IROTH) ? S_IROTH : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_IWOTH) ? S_IWOTH : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_IXOTH) ? S_IXOTH : 0;
+
+    out_mode |= (create_mode & Os::FILE_MODE_ISUID) ? S_ISUID : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_ISGID) ? S_ISGID : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_ISVTX) ? S_ISVTX : 0;
+
+    return out_mode;
+}
+
 PosixFile::Status PosixFile::open(const char* filepath,
                                   PosixFile::Mode requested_mode,
                                   PosixFile::OverwriteType overwrite) {
@@ -104,7 +122,7 @@ PosixFile::Status PosixFile::open(const char* filepath,
             FW_ASSERT(0, requested_mode);
             break;
     }
-    int descriptor = ::open(filepath, mode_flags, USER_FLAGS);
+    int descriptor = ::open(filepath, mode_flags, map_open_create_mode(Os::FILE_DEFAULT_CREATE_MODE));
     if (PosixFileHandle::INVALID_FILE_DESCRIPTOR == descriptor) {
         int errno_store = errno;
         status = Os::Posix::errno_to_file_status(errno_store);
--- a/default/config/OsCfg.fpp
+++ b/default/config/OsCfg.fpp
@@ -25,5 +25,10 @@ module Os {
     constant FILE_MODE_ISGID = 0x400
     constant FILE_MODE_ISVTX = 0x200
 
+    @ File creation mode used when calling Os::File "open" with
+    @ the OPEN_CREATE mode flag.
+    @ The FILE_DEFAULT_CREATE_MODE constant @ is a bitfield of the
+    @ above FILE_MODE_* bits that should be present in the created file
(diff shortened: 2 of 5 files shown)


Os/Posix/File.cpp
```diff
@@ -78,6 +66,36 @@
     return *this;
 }
 
+mode_t PosixFile::map_open_create_mode(const U32 create_mode) {
```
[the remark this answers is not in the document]
kubiak-jpl: Might be referring to `mode_t`? If so, that's the correct type for the open call

Os/Posix/File.cpp
```diff
@@ -104,7 +122,7 @@
             FW_ASSERT(0, requested_mode);
             break;
     }
-    int descriptor = ::open(filepath, mode_flags, USER_FLAGS);
+    int descriptor = ::open(filepath, mode_flags, map_open_create_mode(Os::FILE_DEFAULT_CREATE_MODE));
```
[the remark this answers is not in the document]
kubiak-jpl: No change from previous code

Os/Posix/File.cpp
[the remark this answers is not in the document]
kubiak-jpl: I can't think of an assertion that would provide value here

Os/Posix/File.cpp
```diff
@@ -78,6 +66,36 @@
     return *this;
 }
 
+mode_t PosixFile::map_open_create_mode(const U32 create_mode) {
+    mode_t out_mode = 0;
+
+    // Some posix systems (e.g. Darwin) use the older S_IREAD and S_IWRITE flags
+    // while other systems (e.g. Linux) use the newer S_IRUSR and S_IWUSR flags.
+#if defined(S_IREAD)
```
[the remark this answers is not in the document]
kubiak-jpl: Necessary to support Linux and MacOS (ie. BSD mode_t constants)

thomas-bc:
Note: we should clearly mark the configuration as Posix-specific

default/config/OsCfg.fpp
```diff
@@ -0,0 +1,34 @@
...
+    constant FILE_MODE_IXOTH = 0x004
+    constant FILE_MODE_IRWXO = 0x007
+
+    constant FILE_MODE_ISUID = 0x800
+    constant FILE_MODE_ISGID = 0x400
+    constant FILE_MODE_ISVTX = 0x200
+
+    @ File creation mode used when calling Os::File "open" with
+    @ the OPEN_CREATE mode flag.
+    @ The FILE_DEFAULT_CREATE_MODE constant @ is a bitfield of the
```
thomas-bc, line 30: Need to add `bitfield` to `.github/actions/spelling/expect.txt` ... odd that it isn't a recognized word 🤷‍♂️

default/config/OsCfg.fpp
```diff
@@ -0,0 +1,34 @@
+
+
+module Os {
+
+    @ File open mode permission bits
+    @ Constant values are derived from standard
```
thomas-bc, line 6: The CCB recommends that we namespace this further to make it clear it is meant to apply only to Posix. Other implementations may borrow from it if they want (e.g. VxWorks seems to have the same API?) but it should be clear that this isn't a universal config.

So something like such:
```
module Os {

  module Posix {
    [ ....]
  }
}
```

default/config/OsCfg.fpp
```diff
@@ -0,0 +1,34 @@
...
+    @ Unix values, but should not be assumed
+    @ to match
```
thomas-bc, line 8: What should not be assumed to match? I'm not sure I follow.

thomas-bc (changes requested):

default/config/OsCfg.fpp
kubiak-jpl: I disagree that this should be further namespaced.

The intention is that the `FILE_MODE_` bits can be used across different backends. I only implemented it in the Posix backend because that's the only one I'm familiar with. However, the constants could be referenced by other backends. The `Posix::File::map_open_create_mode` maps these FILE_MODE_ bits to mode_t bits for Posix, but a similar function could be defined for VxWorks or zephyr to map `FILE_DEFAULT_CREATE_MODE` to the specific permission bits for those platforms.

default/config/OsCfg.fpp
kubiak-jpl: Done

default/config/OsCfg.fpp
thomas-bc: What we want to avoid is for users to believe that this option will unambiguously happen anywhere, when that's not the case.
VxWorks and Zephyr happen to have the option to have some compliance with Posix w.r.t. file management, so I'd say it's not wrong to namespace `Os::Posix` when Posix is precisely what Zephyr/VxWorks are trying to comply with? (I might be reaching at straws haha)
I have no strong opinion, let's discuss with the team.

thomas-bc: Note that we agree that, as much as possible, OSALs should attempt to respect this config

kubiak-jpl: > Note that we agree that, as much as possible, OSALs should attempt to respect this config

If that's the case then I think that namespacing it under Os::Posix would discourage other OSALs from using the interface, and make it look like this is only intended to be a Posix thing

thomas-bc: Very true!

thomas-bc: Just a thought, but could using an HPP file instead of FPP solve some of these concerns (conditional compilation / includes or whatnot)? Also easier to just reuse the Posix constants rather than re-define them in FPP

Os/Posix/File.cpp
```diff
@@ -78,6 +66,36 @@ PosixFile& PosixFile::operator=(const PosixFile& other) {
...
+    out_mode |= (create_mode & Os::FILE_MODE_IRUSR) ? S_IREAD : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_IWUSR) ? S_IWRITE : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_IXUSR) ? S_IEXEC : 0;
+#else
+    out_mode |= (create_mode & Os::FILE_MODE_IRUSR) ? S_IRUSR : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_IWUSR) ? S_IWUSR : 0;
+    out_mode |= (create_mode & Os::FILE_MODE_IXUSR) ? S_IXUSR : 0;
+#endif
+
+    out_mode |= (create_mode & Os::FILE_MODE_IRGRP) ? S_IRGRP : 0;
```
LeStarch, line 84: Are these guaranteed to be defined in all posix implementations?

LeStarch: may need to come back later and make these conditional.   Mostly I want to know if these macros are defined on say VxWorks. @kevin-f-ortega have you tried this PR in our VxWorks + Posix stuff?

LeStarch (commented):
A few comment, but otherwise I like it!

I am fine with how the configuration is done.  Platforms outside of F Prime central repo that support posix could override this at the platform level!

default/config/OsCfg.fpp
LeStarch: I think it is fine as a FPP and in a generic namespace.  These constants happen to share names with Posix, but they represent a generic concept.  If a platform cannot support them, then at least they know they should virtualize as close as possible.

I vote this is resolved as-is.

default/config/OsCfg.fpp
LeStarch: The posix constants may change between posix systems. Posix compliant platform implementations should provide overrides if their constants do not match.

Os/Posix/File.cpp
kubiak-jpl: I'm not sure if they are defined everywhere, and it would be good to make sure the other backends that use Os/Posix still compile. Is there a VxWork project compiled in CI?

The original code did have an empty `#ifdef` statement if these values weren't defined, and mentions that might be true on some platforms, but doesn't specify which
```
// Sets up the default file permission as user read + user write
// Some posix systems (e.g. Darwin) use the older S_IREAD and S_IWRITE flags while other systems (e.g. Linux) use the
// newer S_IRUSR and S_IWUSR flags, and some don't support these flags at all. Hence, we look if flags are defined then
// set USER_FLAGS to be the set of flags supported or 0 in the case neither is defined.
#if defined(S_IREAD) && defined(S_IWRITE)
#define USER_FLAGS (S_IREAD | S_IWRITE)
#elif defined(S_IRUSR) && defined(S_IWUSR)
#define USER_FLAGS (S_IRUSR | S_IWUSR)
#else
#define USER_FLAGS (0)
#endif
```

LeStarch (approved):
We agreed that a VxWorks test would happen offline.

thomas-bc (approved):

Issue #4092, opened 2025-08-29
Move Data Product CRC calculation out of component thread context


kubiak-jpl:
| | |
|:---|:---|
|**_F´ Version_**| 4.0 |
|**_Affected Component_**| DataProducts |
---
## Feature Description

Move Data Product CRC calculation out of `dpSend`

## Rationale

While writing a record to a Data Product is generally a low overhead processing, sending the container to the DpWriter is surprisingly expensive, as the act of calling `dpSend` calls `container.updateDataHash()`. This will block the active thread while the hash is calculated, which can be an expensive task depending on the size of the data product. For data products of about 10 MB, this is enough to break my realtime loop that needs to run every 5ms.

Moving the CRC calculation to DpWriter would reduce the strain on the task which is writing to the DataProduct files and may have realtime requirements.

LeStarch:
I like this idea.  Is there a better place to put it?

kubiak-jpl:
My first thought would be DpWriter, since that thread is writing data to disk and doesn't really need to be realtime

kubiak-jpl:
For reference, pushing the `dpSend` calls to an internal thread was enough to work around the problem for what I needed

```cpp
  class SignalProcessing :
    public SignalProcessingComponentBase
  {
...
      // DpSend Thread
      std::jthread dpsend_thread_;
      std::deque<DpContainer> dpsend_q_;
      std::mutex dpsend_lock_;
      std::condition_variable dpsend_cv_;

  };

  void SignalProcessing ::
    dpsend_to_thread(
      DpContainer& c
  ) {

    std::unique_lock lk(dpsend_lock_);
    dpsend_q_.push_back(c);
    dpsend_cv_.notify_all();
  }


  void SignalProcessing ::
    dpsend_thread()
  {
    std::stop_token stop_token = dpsend_thread_.get_stop_token();
    while (true) {
      std::optional<DpContainer> container;
      {
        std::unique_lock lk(dpsend_lock_);
        dpsend_cv_.wait(lk);
        if (dpsend_q_.size() > 0) {
          container = dpsend_q_[0];
          dpsend_q_.pop_front();
        }
      }
      if (stop_token.stop_requested()) {
        break;
      }
      if (container) {
        dpSend(*container);
      }
    }
  }
```

thomas-bc:
@kubiak-jpl are we good to close this ticket?

PR #4622, opened 2026-01-15
Calculate Data Product checksum in the DpWriter component

kubiak-jpl:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4092  |

---
## Change Description

Linked with FPP PR https://github.com/nasa/fpp/pull/900

Moves the calculation of the data checksum into DpWriter and after and post processing behaviors occur

## Rationale

Moves expensive CRC calculations out of potentially realtime component contexts and into the non-realtime DpWriter context.

Calculates the CRC after post-processing on the file Buffer object, so the CRC on disk should match the file contents. This was not the case with the previous implementation

- Updated existing DpWriter unit tests to assume the incoming buffer does not have a checksum
- Added a check that the data product written by DpWriter has a valid data checksum
- Generated a data product with the Ref DpDemo component and validated the checksum

- No testing with processed data products was performed

## Future Work

None

N/A

### DIFF
--- a/Fw/Dp/test/util/DpContainerHeader.hpp
+++ b/Fw/Dp/test/util/DpContainerHeader.hpp
@@ -90,8 +90,6 @@ struct DpContainerHeader {
         checkDeserialAtOffset(deserializer, DpContainer::HEADER_HASH_OFFSET);
         // Check the header hash
         checkHeaderHash(file, line, buffer);
-        // Check the data hash
-        this->checkDataHash(file, line, buffer);
         // Move the deserialization pointer to the data offset
         DpContainerHeader::moveDeserToOffset(file, line, deserializer, DpContainer::DATA_OFFSET);
     }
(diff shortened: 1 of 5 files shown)


bocchino:
This PR is tandem with https://github.com/nasa/fpp/pull/900. Before merging it, we should produce an FPP alpha release and update this branch with the new alpha release.

bocchino:
@jwest115 I know that CI does testing to ensure that data products are generated and decoded properly. As part of this process, does it check the hashes?

jwest115:
> @jwest115 I know that CI does testing to ensure that data products are generated and decoded properly. As part of this process, does it check the hashes?

@bocchino Yes! CI runs the data product decoder tool which checks the hashes and reports an error if there is a mismatch.

bocchino (changes requested):
When running with this version of FPP

https://github.com/nasa/fpp/pull/901

I see an error in `FppTestProject/FppTest/dp`. I think the test has to be updated to conform to the new behavior. @kubiak-jpl can you investigate?

kubiak-jpl:
> When running with this version of FPP
> 
> [nasa/fpp#901](https://github.com/nasa/fpp/pull/901)
> 
> I see an error in `FppTestProject/FppTest/dp`. I think the test has to be updated to conform to the new behavior. @kubiak-jpl can you investigate?

I just pushed a commit to address this. I removed the data hash checks that were done in `Fw/Dp/test/util/DpContainerHeader.hpp` since the data hash is no longer valid when the container is pushed out the port

Fw/Dp/test/util/DpContainerHeader.hpp
bocchino:
```suggestion
    //! Check the header hash
```

bocchino:
Looks good! I will make the FPP alpha release so we can run CI with that.

bocchino:
@thomas-bc The Zephyr runs are failing for reasons that seem unrelated to this PR. Do you have any insight into what is going on here?

bocchino (dismissed):
Looks good!

kubiak-jpl:
@bocchino I re-added the utility function

bocchino (approved):
Looks good!

PR #4628, opened 2026-01-17
Suppress duplicate library warning on macos

Willmac16:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**| n |
|**_Documentation Included (y/n)_**| n |
|**_Generative AI was used in this contribution (y/n)_**| y |

---
## Change Description

Added `-no_warn_duplicate_libraries` to Darwin link options to suppress linker spam about duplicate libraries.

## Rationale

After a macOS update in the past couple of months, I've started getting linker spam about duplicate libraries when compiling the upstream fprime repo:
```
ld: warning: ignoring duplicate libraries: 'lib/Darwin/libFw_Cmd.a', 'lib/Darwin/libFw_Com.a', 'lib/Darwin/libFw_CompQueued.a', 'lib/Darwin/libFw_Fpy.a', 'lib/Darwin/libFw_Log.a', 'lib/Darwin/libFw_Prm.a', 'lib/Darwin/libFw_Tlm.a', 'lib/Darwin/libOs.a', 'lib/Darwin/libSvc_FpySequencer.a', 'lib/Darwin/libSvc_Ping.a', 'lib/Darwin/libSvc_Sched.a', 'lib/Darwin/libSvc_Seq.a', 'lib/Darwin/libdefault_config.a'
```

## Testing/Review Recommendations

Ensure compilation still works on your macOS setups

## Future Work

Potentially there is value in reducing the actual duplicate linked libraries; however,  from what I can tell, the linux standard is to ignore passing the same lib twice so not a huge deal.

## AI Usage (see [policy](https://github.com/nasa/fprime/blob/devel/AI_POLICY.md))

OpenAI Codex runing gpt-5.2-codex was used to diagnose and apply the linker option change to cmake

PR #4630, opened 2026-01-17
Add CCSDS AOS Framer

Willmac16:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Implements a CCSDS AOS Framer that supports both many packets per frame and packets than span multiple frames.

## Rationale

AOS is a newer protocol than TM (although USLP is even newer). Requiring that packets fit within one frame is overly restrictive.

Ensure code style, layout and coverage looks good. Determine if existing subset of AOS features would be sufficient for Fprime downlink.

## Future Work

An AOS Deframer will be made to accompany this, and AOS gds infrastructure will also likely be added.

N/A


Willmac16:
Only uncovered line is the (not yet implemented) EPP Idle packet generation
```
1/1 Test #62: Svc_Ccsds_AosFramer_ut_exe .......   Passed    0.09 sec

The following tests passed:
        Svc_Ccsds_AosFramer_ut_exe

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   6.33 sec
(INFO) Reading coverage data...
(INFO) Writing coverage report...
lines: 99.5% (205 out of 206)
functions: 100.0% (23 out of 23)
branches: 49.0% (423 out of 864)
```

cmake/platform/Darwin.cmake
```diff
@@ -24,3 +24,4 @@ register_fprime_config(
    BASE_CONFIG
 )
 target_compile_definitions(PlatformDarwin INTERFACE -DTGT_OS_TYPE_DARWIN)
+target_link_options(PlatformDarwin INTERFACE -Wl,-no_warn_duplicate_libraries)
```
thomas-bc, line 27: Let's revert this line which is being reviewed in https://github.com/nasa/fprime/pull/4628

.vscode/launch.json
thomas-bc, line 1: Added by mistake? It could be valuable to provide reusable VSCode settings, but we should identify which ones would be useful to devs of the core nasa/fprime. We have a CMakePresets.json already which I've found to work pretty well with the VSCode CMake extension

Svc/Ccsds/Types/Types.fpp
```diff
@@ -88,5 +88,64 @@ module Ccsds {
         constant segLengthOffset = 11
     }
 
+    # ------------------------------------------------
+    # AOS
+    # ------------------------------------------------
+    @ Describes the frame header format for a Advanced Orbiting Systems (AOS) Space Data Link (SDL) Transfer Frame
+    struct AOSHeader {
+        globalVcId: U16,         @< 2 bit Frame Version | 8 least significant bits of spacecraft ID | 6 bit virtual channel ID
```
thomas-bc, line 96: Question: why `least significant bits` for the SCID ? I haven't seen anything about that in the blue book so I wonder if I may have missed something

Svc/Ccsds/AosFramer/AosFramer.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+    // Replay Flag (4.1.2.5.2)
+    frameCountAndSignaling |= static_cast<U32>(context.get_replayFlag() << AOSHeaderSubfields::replayFlagOffset);
+
+    // Virtual Channel Frame Count Cycle Use Flag (4.1.2.5.3)
+    frameCountAndSignaling |= static_cast<U32>(1 << AOSHeaderSubfields::cycleCountFlagOffset);
+
+    // Spacecraft ID MSB (4.1.2.5.4)
+    frameCountAndSignaling |=
+        static_cast<U32>((ComCfg::SpacecraftId & 0x0300) >> (8 - AOSHeaderSubfields::spacecraftIdMsbOffset));
```
thomas-bc, line 166: So you are using the 2-bit `Reserved Spare` field for SCID MSB. That's not part of the blue book AFAIK. Are there other guidances that encourage to do that? Or what is the rationale? If we are deviating from the standard, we should clearly indicate so in the SDD.

From the standard: 
> This 2-bit field is reserved for future definition by CCSDS and shall be set to ‘00’.

default/config/ComCfg.fpp
```diff
@@ -10,13 +10,23 @@ module ComCfg {
...
+    @ Upper Bound on Fixed size of CCSDS AOS frames
+    constant AosMaxFrameFixedSize = 1536
+
     @ Aggregation buffer for ComAggregator component
     constant AggregationSize = TmFrameFixedSize - 6 - 6 - 1 - 2  # 2 header (6) + 1 idle byte + 2 trailer bytes
 
+    @ Packet Version Numbers are 3 bits with only 2 currently valid values
+    dictionary enum Pvn : U8 {
+        SPACE_PACKET_PROTOCOL         = 0x0   @< Command packet type - incoming
+        ENCAPSULATION_PACKET_PROTOCOL = 0x3   @< Telemetry packet type - outgoing
```
thomas-bc: annotations are off here

default/config/ComCfg.fpp
```diff
@@ -10,13 +10,23 @@ module ComCfg {
 
     @ Spacecraft ID (10 bits) for CCSDS Data Link layer
     dictionary constant SpacecraftId = 0x0044
-    
+
     @ Fixed size of CCSDS TM frames
     dictionary constant TmFrameFixedSize = 1024  # Needs to be at least COM_BUFFER_MAX_SIZE + (2 * SpacePacketHeaderSize) + 1
 
+    @ Upper Bound on Fixed size of CCSDS AOS frames
+    constant AosMaxFrameFixedSize = 1536
```
thomas-bc, line 18: Just out of curiosity, is there a reason this was chosen to be the default?

Svc/Ccsds/Types/Types.fpp
thomas-bc: Just saw the thing about using the reserved field for most significant bits. Let's move the discussion there. https://github.com/nasa/fprime/pull/4630#discussion_r2709983955

Svc/Ccsds/AosFramer/AosFramer.cpp
```diff
@@ -0,0 +1,422 @@
...
+            // Setup the AOS header right before we send with the context cached in outstanding
+            setup_header(currentVc.outstanding.context);
+
+            // Setup the M_PDU header w/ all zero (only continuing packets)
+            setup_m_pdu_header(currentVc.outstanding.context, true);
+        }
+
+        // Compute Trailer
+        if (this->m_fecf) {
+            compute_fecf(currentVc);
```
thomas-bc: A bit of a nit, but how would you feel about renaming that helper function to make it more clear that it's actually modifying the buffer under the hood? E.g. `compute_and_{inject/serialize/...}_fecf()` ?

Svc/Ccsds/AosFramer/docs/sdd.md
```diff
@@ -0,0 +1,38 @@
...
+## AOS Framing Support
+
+The AOS Framer and Deframer support the following subset of CCSDS AOS SDL:
+
+* Single, port context driven virtual channel
+* Space Data Link Security (SDLS) disabled
+* Frame Header Error Control Field is disabled
+* Transfer Frame Insert Zone is disabled
+* Transfer Frame Data Field is driven by the per-instance, compile time `fixed_frame_size` setting
+* This component only supports the Multiplexing Protocol Data Unit (M_PDU) service
```
thomas-bc: Can you confirm on your end, but I believe this should say Virtucal Channel Packet (VCP) Service, using Multiplexing Protocol Data Units

Svc/Ccsds/AosFramer/AosFramer.cpp
```diff
@@ -0,0 +1,422 @@
...
+        // pack_packet w/ sendNow & < 7 bytes left                   ->                           (skip) check_and_send
+        //                   \                                                                  /
+        //                    fill_with_idle_packet (spp only) -> pack_packet -> check_and_send
+        // This ensures the outer loop check_and_send doesn't try to send
+        currentVc.current_payload_offset = 0;
+        currentVc.past_first_fresh_packet = false;
+
+        // Write the completed frame
+        this->dataOut_out(0, currentVc.frame.buffer, currentVc.outstanding.context);
+    } else if (this->isConnected_comStatusOut_OutputPort(currentVc.vc_struct_index)) {
```
thomas-bc, line 252: Not applicable yet, but I wonder if this could assert unexpectedly in the future depending on how they are configured? For example if there are more VCs than ports, and you may only start seeing a specific VC late in the mission.
Not sure how you envision this would be set up in the future. But to keep in mind?

thomas-bc (commented):
I haven't been able to dive in the code itself too deep yet, as I was also discovering the AOS spec in parallel. I should be able to get to it more thoroughly in the next day or two. But in the meantime, for the sake of iterating more quickly, I figured I'd dump a first set of nits/questions/comments. 

This looks great! Love the internal pattern of tracking `AosVc`'s to extend further down the road.

.vscode/launch.json
Willmac16: Where is the CMakePresets.json at?

Svc/Ccsds/AosFramer/AosFramer.cpp
Willmac16: The 2 additional bits for the SCID is an addition in the 5th Issue of the standard. CCSDS 732.0-B-5 Page 4-3

image
My marked up copy of the AOS spec I am working from

default/config/ComCfg.fpp
Willmac16: I initialized to 1024 like TM, then realized I wrote a unit test that relied on the max value being 1024 to pass, so I switched the value to 1536 (3 * 512) to ensure I had correctly isolated that test from a users configured max frame size.

I'm happy to switch it back to 1024, or maybe we use 1472 since that (should be) the largest UDP payload that fits in a standard 1500 MTU ethernet frame

Svc/Ccsds/AosFramer/AosFramer.cpp
Willmac16: great suggestion! done

Svc/Ccsds/AosFramer/docs/sdd.md
Willmac16: Great catch!

Svc/Ccsds/AosFramer/AosFramer.cpp
Willmac16: I think in my mind more VCs than ports would represent a mission phase change, which could have its own reconfigure call into the AOS Framer.

When we do add support for multiple VCs there is going to need to be some thought into how input and output ports work. I could imagine only one port in the array being used despite multiple VCs or cases where one of the input side or the output side would have multiple ports in the array while the other operated w/ one.

That and if you mixed upstream SPP & EPP you could have multiple upstreams for a single VC.

Svc/Ccsds/AosFramer/AosFramer.cpp
thomas-bc: Oh nice! I had not caught that 5th revision. The link in the SDD is https://ccsds.org/Pubs/732x0b4.pdf which points to B4, it'd be nice if we could update that to B5

Svc/Ccsds/AosFramer/AosFramer.cpp
```diff
@@ -0,0 +1,422 @@
...
+#include "config/FppConstantsAc.hpp"
+
+namespace Svc {
+
+namespace Ccsds {
+// ----------------------------------------------------------------------
+// Component construction and destruction
+// ----------------------------------------------------------------------
+
+AosFramer ::AosFramer(const char* const compName) : AosFramerComponentBase(compName) {}
```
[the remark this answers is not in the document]
thomas-bc: Let's fix this one, either in constructor or in the HPP

.vscode/launch.json
thomas-bc: https://github.com/nasa/fprime/blob/devel/CMakePresets.json

It's also generated by `fprime-bootstrap project` now. Gives you the below panel in VScode when using the CMake extension

image

thomas-bc: There are a few VSCode config quirks to work around, notably around the Python environment that's required to be activated  when CMake is invoked, some pointers in https://github.com/nasa/fprime/discussions/2807

thomas-bc: Leaving this unresolved so you don't miss the response; please resolve once you have !

thomas-bc:
yeah those RHEL8 warnings are annoying, but we've seen that before, just need to cast the whole result of the operation... for example
https://github.com/willmac16/fprime/blob/760ca7c4554b2241906fac6ca2b7ba868ee13d9a/Svc/Ccsds/SpacePacketDeframer/test/ut/SpacePacketDeframerTester.cpp#L50-L51

default/config/ComCfg.fpp
```diff
@@ -37,14 +47,21 @@ module ComCfg {
...
+        replayFlag: bool            @< AOS bool for realtime vs replay (passed directly into the frame)
+        sendNow: bool               @< Flag to AOS Framer that the Frame this packet goes into should be sent ASAP
+
     } default {
         comQueueIndex = 0
+        pvn = Pvn.SPACE_PACKET_PROTOCOL
         apid = Apid.FW_PACKET_UNKNOWN
         sequenceCount = 0
-        vcId = 1
+        vcId = 0
```
thomas-bc: The reason I didn't default to 0 here is because I prefer to avoid making "serializing all 0s" have a meaning cause that can just be an uninitialized field (e.g. `FW_PACKET_COMMAND = 0x0000` has bitten me a few times).

Is there a reason to switch default to 0 ? If yes, we should update the GDS, otherwise revert here.

https://github.com/nasa/fprime-gds/blob/196102c7d3d0a5d80663771d3c7bfa945be0eaa4/src/fprime_gds/common/communication/ccsds/space_data_link.py#L186-L191

thomas-bc: I see that there's some relation between the index in the `AosVc m_vcs[1];` array, but I think you should be able to default `virtualChannelId = 1` ? Curious to hear your thoughts

Svc/Ccsds/AosFramer/AosFramer.cpp
Willmac16: I updated the URL and referenced version of the standard. The URL is a little less clean/obvious than the 4th version one, but its what they point to on the Blue Books page https://ccsds.org/publications/bluebooks/entry/4547/

Svc/Ccsds/AosFramer/AosFramer.cpp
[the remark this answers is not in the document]
Willmac16: hpp now defaults to true & configure is called w/ FECF on + max size constant by default

Svc/Ccsds/AosFramer/AosFramer.cpp
```diff
@@ -0,0 +1,433 @@
...
+        static_cast<U32>((ComCfg::SpacecraftId & 0x0300) >> (8 - AOSHeaderSubfields::spacecraftIdMsbOffset));
+
+    // Virtual Channel Frame Cycle Count (4.1.2.5.5)
+    frameCountAndSignaling |= static_cast<U32>((currentVc.virtualFrameCount & 0x0F000000) >> 24);
+
+    header.set_globalVcId(globalVcId);
+    header.set_frameCountAndSignaling(frameCountAndSignaling);
+
+    // Perform the modulo at serialization time makes vc cycle count easier to make optional
+    currentVc.virtualFrameCount++;  // U28 intended to wrap around (modulo 268,435,456)
```
thomas-bc: I got confused at U28 here, then noticed the note about 

> VC Frame Count Cycle effectively extends the Virtual Channel Frame Count from 24 to 28 bits.

Ah... gotta love standards

thomas-bc: jokes aside it could be good to mention it's 24 bits extended by 4 additional bits to make it unambiguous

default/config/ComCfg.fpp
```diff
@@ -37,14 +47,21 @@ module ComCfg {
     @ Type used to pass context info between components during framing/deframing
     struct FrameContext {
         comQueueIndex: FwIndexType  @< Queue Index used by the ComQueue, other components shall not modify
+        pvn: Pvn                    @< Packet Version Number (CCSDS SPP vs EPP)
         apid: Apid                  @< 11 bits APID in CCSDS
         sequenceCount: U16          @< 14 bit Sequence count - sequence count is incremented per APID
-        vcId: U8                    @< 6 bit Virtual Channel ID - used for TC and TM
+        vcId: U8                    @< 6 bit Virtual Channel ID - used for AOS
```
thomas-bc: used for AOS, TC and TM protocols

default/config/ComCfg.fpp
```diff
@@ -37,14 +47,21 @@ module ComCfg {
...
+        replayFlag: bool            @< AOS bool for realtime vs replay (passed directly into the frame)
```
thomas-bc: From the blue book, it sounds like `replayFlag` is fairly project-specific and I don't see any way we could leverage it right now - do you have a plan to use this? I'm wary of including this field in all FrameContext's without a use case for it.

default/config/ComCfg.fpp
```diff
@@ -37,14 +47,21 @@ module ComCfg {
...
+        sendNow: bool               @< Flag to AOS Framer that the Frame this packet goes into should be sent ASAP
```
thomas-bc, line 46: Could this functionality come in on a separate port instead? What component do you envision would turn that on in the framing chain?

Svc/Ccsds/AosFramer/docs/sdd.md
thomas-bc, line 1: General question; are you all planning to use AOS for a specific reason that you would be willing to share? 

I have been looking at these standards for a bit but am still out of the loop on use-cases and such. For example, why use AOS and not USLP? Is AOS better supported by other tools because of its "seniority" over USLP ?

Svc/Ccsds/AosFramer/AosFramer.cpp
thomas-bc, line 1: Has this Framer been tested end-to-end against a tool that is known to deframe AOS frames correctly?

thomas-bc (commented):
Code LGTM; just a few follow-ups and questions in some places.

Svc/Ccsds/AosFramer/AosFramer.cpp
Willmac16: Not yet. I could try to work out a YAMCS test for downlink; however, my primary need for this component is for communicating w/ payloads, so I was planning on placing my integration testing focus on comms w/ those particular payloads

default/config/ComCfg.fpp
Willmac16: Nope, don't have a use for it at this time. My initial thought was to avoid hardcoding anything that was easy enough to serialize in; however, I'm happy to remove it from ComCfg + cpp. If someone needs it, it'll still be in the Types.fpp for later

default/config/ComCfg.fpp
Willmac16: I switched it back to 1. One of my payloads needs VC ID set to 0, but no reason to change the default in ComCfg

default/config/ComCfg.fpp
Willmac16: My thought process is that some sources upstream of the ComQueue might want to request immediate sending, while others would prioritize bandwidth over latency. Since we're already conveying the APID via the context it seems reasonable to me that we'd convey sending immediacy this way too

Svc/Ccsds/AosFramer/docs/sdd.md
Willmac16: From what I understand, AOS Downlink + TC Uplink is pretty common. Or you can go USLP (In different modes I believe) both ways.

For FPrime upstream I think it makes sense to swap the ComAggregator (which I didn't know about until today) + TM Framer for the AOS Framer since it enables packets that span multiple frames. Or at least make a new AOS+TC Subtopo that does. And if FPrime gets a USLP defamer that can do variable length deframing, then it would make sense to switch over/provide USLP for both.

The big AOS upgrade--that this implementation doesn't provide--is the transfer frame insert zone, which is meant for realtime voice/video to be interspersed at a regular rate w/ standard telemetry & downlink.

We are likely not using this AOS framer for anything other than payload comms due to particulars of our radio.

cmake/platform/Darwin.cmake
```diff
@@ -24,4 +24,3 @@ register_fprime_config(
    BASE_CONFIG
 )
 target_compile_definitions(PlatformDarwin INTERFACE -DTGT_OS_TYPE_DARWIN)
-target_link_options(PlatformDarwin INTERFACE -Wl,-no_warn_duplicate_libraries)
```
thomas-bc, line 27: We should revert this deletion, maybe appeared because of the force push or something

thomas-bc:
If that last one doesn't work, I can take this to a RHEL8 machine and try to resolve the compiler warnings there

default/config/ComCfg.fpp
```diff
@@ -37,14 +47,19 @@ module ComCfg {
     @ Type used to pass context info between components during framing/deframing
     struct FrameContext {
         comQueueIndex: FwIndexType  @< Queue Index used by the ComQueue, other components shall not modify
+        pvn: Pvn                    @< Packet Version Number (CCSDS SPP vs EPP)
```
thomas-bc: It doesn't seem like this struct member is being used? Since this is configurable by the project, we should stick with what's strictly needed.

default/config/ComCfg.fpp
```diff
@@ -10,13 +10,23 @@ module ComCfg {
...
+    constant AosMaxFrameFixedSize = 1536
+
     @ Aggregation buffer for ComAggregator component
     constant AggregationSize = TmFrameFixedSize - 6 - 6 - 1 - 2  # 2 header (6) + 1 idle byte + 2 trailer bytes
 
+    @ Packet Version Numbers are 3 bits with only 2 currently valid values
+    dictionary enum Pvn : U8 {
+        SPACE_PACKET_PROTOCOL         = 0x0   @< Fully Featured CCSDS Space Packet Protocol
+        ENCAPSULATION_PACKET_PROTOCOL = 0x3   @< Bare-bones CCSDS Encapsulation Packet Protocol
+        INVALID_UNINITIALIZED         = 0x4   @< Anything equal or higher value is invalid and should not be used
```
Willmac16: EPP needs to be corrected to 7 & the subsequent invalid one to 8

Svc/Ccsds/AosFramer/docs/sdd.md
```diff
@@ -0,0 +1,38 @@
...
+
+* Single, port context driven virtual channel
+* Space Data Link Security (SDLS) disabled
+* Frame Header Error Control Field is disabled
+* Transfer Frame Insert Zone is disabled
+* Transfer Frame Data Field is driven by the per-instance, compile time `fixed_frame_size` setting
+* This component only supports the Virtual Channel Packet (VCP) service, using Multiplexing Protocol Data Units (M_PDU)
+* Operational Control Field is disabled
+* Frame Error Control Field is compile-time configurable
+* This component supports packing multiple CCSDS Recognized packets (i.e. Space Packet Protocol or Encapsulation Packet Protocol) into one AOS SDL Frame. It also supports striping a single packet across multiple frames. Extra space is padded with idle packets.
```
thomas-bc, line 21: I'm curious to know if you've looked at the Svc.ComAggregator component, which should enable the same feature. If you have seen it, are there reasons you're choosing to re-implement the accumulation here? Is it that it gives you more control about how/when to accumulate?

thomas-bc (commented):
Okay! CI is passing. Two quick questions, but other than that I think we're ready to merge.

Svc/Ccsds/AosFramer/docs/sdd.md
Willmac16: I only learned that the component existed once I'd written the code (through the review process). I haven't thoroughly reviewed the code, but I'm not convinced that packets spanning multiple frames would correctly operate or. at the very least, set the first payload header offset.

default/config/ComCfg.fpp
Willmac16: I'll need to figure out for the AosDeframer how I want the EPP vs SPP output to work (either different ports or using the context). We can remove it for now since It's not adding value.

Svc/Ccsds/AosFramer/docs/sdd.md
thomas-bc: Oh absolutely, there's no striping there. Makes sense. Thanks!

thomas-bc (approved):
LGTM! Thanks a lot @Willmac16

Willmac16:
@thomas-bc Is there anything left blocking the merge on this PR? Release Cycle?

I'm about ready to open up my AosDeframer PR and want to avoid duplicate commits btw the two

Issue #4527, opened 2025-12-05
Serializable.cpp  multiple static analysis findings


djbyrne17:
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**| Fw/Types |
---
Follow-on from #4296 which gathered static analysis findings.

## Problem Description

File `Serializable.cpp` has the only SonarQube findings of this category in fprime.  Fix that, and opportunistically some other findings.

Also, more assertions would be useful, but that is not in scope of this ticket.

### At least the unique SonarQube finding

```
<High> Line 281: cpp:S1768 result of comparison of unsigned expression < 0 is always false
```

### Other SonarQube findings as make sense, and as time allows

```
<High> Line 16: cpp:S3490 Use "=default" instead of the default implementation of this special member functions.
<High> Line 18: cpp:S3490 Use "=default" instead of the default implementation of this special member functions.
<High> Line 217: cpp:S5008 Replace this use of "void *" with a more meaningful type.
<High> Line 42: cpp:S3490 Use "=default" instead of the default implementation of this special member functions.
<High> Line 473: cpp:S5008 Replace this use of "void *" with a more meaningful type.
<High> Line 816: cpp:S5008 Replace this use of "void *" with a more meaningful type.
<High> Line 878: cpp:S5008 Replace this use of "void *" with a more meaningful type.

<Med> Line 478: cpp:S3630 Replace "reinterpret_cast" with a safer operation.
<Med> Line 496: cpp:S5827 Replace the redundant type with "auto".

<Low> Line 102: cpp:S1905 Remove this redundant cast.
<Low> Line 109: cpp:S1905 Remove this redundant cast.
<Low> Line 116: cpp:S1905 Remove this redundant cast.
<Low> Line 125: cpp:S1905 Remove this redundant cast.
<Low> Line 131: cpp:S1905 Remove this redundant cast.
<Low> Line 140: cpp:S1905 Remove this redundant cast.
<Low> Line 148: cpp:S1905 Remove this redundant cast.
<Low> Line 161: cpp:S1905 Remove this redundant cast.
<Low> Line 167: cpp:S1905 Remove this redundant cast.
<Low> Line 180: cpp:S1905 Remove this redundant cast.
<Low> Line 201: cpp:S1905 Remove this redundant cast.
<Low> Line 212: cpp:S1905 Remove this redundant cast.
<Low> Line 218: cpp:S1905 Remove this redundant cast.
[...]

PR #4634, opened 2026-01-20
Fix static analysis issues in Serializable.cpp

Sarah5567:
| | |
|:---|:---|
|**_Related Issue(s)_**| [Serializable.cpp multiple static analysis findings](https://github.com/nasa/fprime/issues/4527) |

---
## Change Description

- Fixed SonarQube finding: `<High> Line 281: cpp:S1768 result of comparison of unsigned expression < 0 is always false`
  - Removed `(size < std::numeric_limits<FwSizeStoreType>::min())` from the range check, keeping the upper-bound validation.
- Addressed several `Unchecked Parameter Dereference` findings by adding explicit assertions before dereferencing pointers, e.g.:
    - `if (length > 0) { FW_ASSERT(buff); }`
    - `FW_ASSERT(this->getBuffAddr());`
   
 The remaining High-severity findings were reviewed but did not appear to represent actual issues.

## Rationale

 - Both types are unsigned, so `min()` is 0 and `size < min()` is always false. The redundant check was removed, the upper-bound check remains.
- Assertions were added to prevent writing to a NULL buffer and to ensure the source buffer is non-null when writing more than zero bytes.

## Future Work

Review and address remaining Medium and Low severity findings.

ChatGPT was used to help identify the source of the static analysis findings.

### DIFF
--- a/Fw/Types/Serializable.cpp
+++ b/Fw/Types/Serializable.cpp
@@ -287,7 +287,7 @@ SerializeStatus LinearBufferBase::serializeFrom(const LinearBufferBase& val, End
 
 SerializeStatus LinearBufferBase::serializeSize(const FwSizeType size, Endianness mode) {
     SerializeStatus status = FW_SERIALIZE_OK;
-    if (size > std::numeric_limits<FwSizeStoreType>::max()) {
+    if ((size < std::numeric_limits<FwSizeStoreType>::min()) || (size > std::numeric_limits<FwSizeStoreType>::max())) {
         status = FW_SERIALIZE_FORMAT_ERROR;
     }
     if (status == FW_SERIALIZE_OK) {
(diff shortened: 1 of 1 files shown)


Sarah5567:
Hi @djbyrne17, @thomas-bc,
could you please take a look when you have time?
Happy to make any changes if needed.
Thanks

thomas-bc (commented):
FwSizeStoreType is meant to be configurable, so while the check might not be relevant in the default config I don't think we want to remove it altogether

Sarah5567:
@thomas-bc Got it. So the main purpose of [this issue](https://github.com/nasa/fprime/issues/4527) is no longer relevant? Should we close this PR, or keep it open for the other changes?

thomas-bc:
@Sarah5567 the purpose of #4527 is to find a resolution for these warnings, and you've proceeded exactly what I would have expected by fixing the issue! It just happens that in this case, the warning is valid but we just consider it a "false positive" or something in those lines (because of the fact that it is a configurable type), which is a perfectly valid way to disposition this warning.

So thanks a lot! And thanks for reverting the change; it's natural for reviews to go back and forth. 
Give me just some time to find out if there is a reason the other asserts weren't in in the first place, and we'll try to proceed with merging this.

LeStarch (approved):

PR #4647, opened 2026-01-24
Correct isConnected check in FpySequencer for seqStartOut

Willmac16:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Correct a isConnected check to ensure it matches the port call we are guarding

## Rationale

FSW will assert if port is unconnected otherwise

Check if anything else in FpySequencer has a mismatch

## Future Work

Can we get a func that checks then invokes if the port is connected?

N/A


LeStarch (approved):

zimri-leisher:
Oops... thanks @Willmac16 
Pretty clearly a copy-paste error.

zimri-leisher (approved):

Issue #4646, opened 2026-01-24
Os::Queue::setRegistry is declared but never defined


jsmith212:
| | |
|:---|:---|
|**_F´ Version_**| v4.1.1 |
|**_Affected Component_**| Os::Queue |
---
## Problem Description

The method `Os::Queue::setRegistry` is declared in the Queue.hpp header, but is never actually defined in Queue.cpp. This appears to be an oversight that causes a linker error when attempting to use a Queue registry.

## Context / Environment

Execute `fprime-util version-check` and share the output.

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-6.6.87.1-microsoft-standard-WSL2+-x86_64-with-glibc2.39
Python version: 3.13.4
CMake version: 3.26.0
Pip version: 25.3
Pip packages:
    fprime-tools==4.1.0
    fprime-gds==4.1.0
    fprime-fpp==3.1.0
Project submodules:
    https://github.com/nasa/fprime.git @ v4.1.1
    https://github.com/fprime-community/fprime-zephyr.git @ 5550106
```

## How to Reproduce

1. Implement a simple QueueRegistry
2. Attempt to set the registry with `Os::Queue::setRegistry(&myRegistry);`
3. Observe a linker error: `TrickQueueRegistry.cpp:(.text+0xdd): undefined reference to `Os::Queue::setRegistry(Os::QueueRegistry*)'`
4. Open Queue.cpp and note that the implementation is indeed missing.

## Expected Behavior

The deployment links successfully and the queue registry is successfully registered. 

## Workaround

This can be worked around by putting the following snippet somewhere in your deployment:

```
namespace Os {
void Queue::setRegistry(QueueRegistry* registry) {
    ScopeLock lock(Queue::getStaticMutex());
    Queue::s_queueRegistry = registry;
}
}  // namespace Os
```

jsmith212:
I took a stab at this in https://github.com/nasa/fprime/pull/4648

PR #4648, opened 2026-01-24
Implement Os::Queue::setRegistry

jsmith212:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/4646 |

---
## Change Description

Adds an implementation for `Os::Queue::setRegistry` and tests this using the Queue Tester.

## Rationale

This method was previously unimplemented, leading to linker errors if used.

1. Add basic QueueRegistry implementation
2. Attempt to register it without the fix and note the linker error
3. Attempt to register it after the fix and note the absence of the linker error

## Future Work

N/A, this should cover it.

No AI was used in the discovery, fix, or testing of this bug.

### DIFF
--- a/Os/Queue.cpp
+++ b/Os/Queue.cpp
@@ -146,4 +146,11 @@ Os::Mutex& Queue::getStaticMutex() {
     return s_mutex;
 }
 
+#if FW_QUEUE_REGISTRATION
+void Queue::setRegistry(QueueRegistry* registry) {
+    ScopeLock lock(Queue::getStaticMutex());
+    Queue::s_queueRegistry = registry;
+}
+#endif
+
 }  // namespace Os
--- a/Os/test/ut/queue/CommonTests.cpp
+++ b/Os/test/ut/queue/CommonTests.cpp
@@ -19,6 +19,10 @@ U64 Tester::QueueMessage::order_counter = 0;
 
 PriorityCompare const Tester::QueueMessageComparer::HELPER = PriorityCompare();
 
+Tester::Tester() {
+    Os::Queue::setRegistry(this);
+}
+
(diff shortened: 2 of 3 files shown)


Os/test/ut/queue/CommonTests.cpp
```diff
@@ -19,6 +19,10 @@ U64 Tester::QueueMessage::order_counter = 0;
 
 PriorityCompare const Tester::QueueMessageComparer::HELPER = PriorityCompare();
 
+Tester::Tester() {
+    Os::Queue::setRegistry(this);
```
LeStarch, line 24: Need to guard this:

```suggestion
#if FW_QUEUE_REGISTRATION
    Os::Queue::setRegistry(this);
#endif
```

Os/test/ut/queue/CommonTests.cpp
```diff
@@ -229,6 +237,7 @@ TEST(BasicRules, Create) {
     create_rule.action(tester);
     // Repetitive create
     create_rule.action(tester);
+    EXPECT_GT(tester.m_all_queues.size(), 0) << "No queues were registered.";
```
LeStarch, line 243: Same:

```suggestion
#if FW_QUEUE_REGISTRATION
 EXPECT_GT(tester.m_all_queues.size(), 0) << "No queues were registered.";
#endif
```

Os/Queue.cpp
```diff
@@ -146,4 +146,11 @@
     return s_mutex;
 }
 
+#if FW_QUEUE_REGISTRATION
```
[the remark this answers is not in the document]
jsmith212: Required to work with header declaration

LeStarch (dismissed):

LeStarch (dismissed):

LeStarch (approved):

Issue #4523, opened 2025-12-05
PolyType.cpp multiple static analysis findings


djbyrne17:
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**| Svc/Version |
---
Follow-on from #4296 which gathered static analysis findings.

## Problem Description

File `PolyType.cpp` has 1 of the most severe SonarQube findings in fprime.  Fix that, and opportunistically some less severe findings.

Also, more assertions would be useful, but that is not in scope of this ticket.

### At least the SonarQube "Blocker" finding

```
<High> Line 649: cpp:S5912 Use pointer or reference to avoid slicing from "ExternalString" to "StringBase".
```

### Other SonarQube findings as make sense, and as time allows

```
<High> Line 313: cpp:S5008 Replace this use of "void *" with a more meaningful type.
<High> Line 318: cpp:S5008 Replace this use of "void *" with a more meaningful type.
<High> Line 323: cpp:S5008 Replace this use of "void *" with a more meaningful type.
<High> Line 332: cpp:S5008 Replace this use of "void *" with a more meaningful type.
<High> Line 343: cpp:S3490 Use "=default" instead of the default implementation of this special member functions.

<Med> Line 101: cpp:S5817 This function should be declared "const".
<Med> Line 106: cpp:S5817 This function should be declared "const".
<Med> Line 111: cpp:S5817 This function should be declared "const".
<Med> Line 132: cpp:S5817 This function should be declared "const".
<Med> Line 137: cpp:S5817 This function should be declared "const".
<Med> Line 142: cpp:S5817 This function should be declared "const".
<Med> Line 159: cpp:S5817 This function should be declared "const".
<Med> Line 164: cpp:S5817 This function should be declared "const".
<Med> Line 169: cpp:S5817 This function should be declared "const".
<Med> Line  18: cpp:S5817 This function should be declared "const".
<Med> Line 189: cpp:S5817 This function should be declared "const".
<Med> Line 194: cpp:S5817 This function should be declared "const".
<Med> Line 199: cpp:S5817 This function should be declared "const".
<Med> Line 216: cpp:S5817 This function should be declared "const".
<Med> Line 221: cpp:S5817 This function should be declared "const".
<Med> Line 226: cpp:S5817 This function should be declared "const".
<Med> Line  23: cpp:S5817 This function should be declared "const".
<Med> Line 243: cpp:S5817 This function should be declared "const".
[...]

PR #4649, opened 2026-01-26
Fix PolyType Static Analysis Issues

youio:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
PolyType.cpp multiple static analysis findings #4523
n
n
n

---
## Change Description

Made various changes to address static analysis issues found in PolyType.cpp. Changes include
* Cast an ExternalString to StringBase& when assigning to a StringBase variable
* Added const modifiers to const-qualified member functions
* Changed destructor to use =default instead of {}
* Added FW_ASSERT statements to ensure the 'this' pointer is not null
* Moved data member assignments to an initializer list

Note:
Did not address the `Replace this use of "void *" with a more meaningful type.` or `Use "std::string" instead of a C-style char array.`  flags.

## Rationale

Most changes were directly suggested in the SonarQube or CodeSonar flags. Casting ExternalString to StringBase in assignment removes the object slicing blocker without affecting functionality.  The FW_ASSERT statements are to address the `Unchecked Parameter Dereference` flags and assume that those functions should never be called on a nullptr object.

The flags I did not address either did not make sense to change in the context of the code or would require large refactoring that seemed out of scope for this issue.

## Future Work

Could perhaps address the remaining flags in the future

AI was used for code understanding and debugging help.

### DIFF
--- a/Fw/Types/PolyType.cpp
+++ b/Fw/Types/PolyType.cpp
@@ -340,7 +340,7 @@ PolyType::PolyType(const PolyType& original) : Fw::Serializable() {
     this->m_val = original.m_val;
 }
 
-PolyType::~PolyType() {}
+PolyType::~PolyType() = default
 
 PolyType& PolyType::operator=(const PolyType& src) {
     this->m_dataType = src.m_dataType;
--- a/Fw/Types/PolyType.cpp
+++ b/Fw/Types/PolyType.cpp
@@ -336,10 +336,7 @@ PolyType& PolyType::operator=(void* other) {
 }
 
 PolyType::PolyType(const PolyType& original)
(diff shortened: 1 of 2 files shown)


LeStarch (dismissed):
I don't understand why there have been added a bunch of checks for `this != null`.  If static analysis is flagging these, then it is probably a false positive.

Otherwise, it looks good!

youio:
@LeStarch Okay got it thank you! I thought it was odd the 'this->' calls were flagged as an issue. I went ahead and removed the asserts and fixed some formatting issues flagged in the workflows. It dismissed your old review if you could take another look whenever you get the chance. 

Thanks!

LeStarch (approved):

PR #4669, opened 2026-01-29
Add Fw.DataProduct FPP interface

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Following the other Fw.Event, Fw.Channel etc. interfaces
Allows to have 

```python
active component MyComponent {
    import Fw.Event
    import Fw.Command
    import Fw.Channel
    import Fw.DataProduct
}
```

... and off to the races!

### DIFF
--- a/Fw/Interfaces/DataProduct.fpp
+++ b/Fw/Interfaces/DataProduct.fpp
@@ -1,0 +1,11 @@
+module Fw {
+    interface DataProduct {
+
+        @ Data product get port: allocates a data product container
+        product get port productGetOut
+
+        @ Data product send port: sends the filled data product container
+        product send port productSendOut
+
+    }
+}
--- a/Fw/Interfaces/DataProductSync.fpp
+++ b/Fw/Interfaces/DataProductSync.fpp
@@ -1,5 +1,6 @@
 module Fw {
-    interface DataProduct {
+    @ Defines ports for synchronous data product operations (get/send)
+    interface DataProductSync {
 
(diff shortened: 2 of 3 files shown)


thomas-bc:
Note, I just noticed there also are 

```
        @ Data product request port
        product request port productRequestOut

        @ Data product receive port
        async product recv port productRecvIn
```
ports. Maybe we need an async interface

kevin-f-ortega (dismissed):

Fw/Interfaces/DataProductSync.fpp
```diff
@@ -0,0 +1,11 @@
+module Fw {
+    interface DataProduct {
+
+        @ Data product get port: allocates a data product container
```
kevin-f-ortega, line 5: This is for sync ports. I think we need to add one for async ports too

Fw/Interfaces/DataProduct.fpp
```diff
@@ -0,0 +1,11 @@
+module Fw {
+    interface DataProduct {
```
LeStarch:
```suggestion
    interface SyncDataProduct
```

LeStarch (changes requested):

Fw/Interfaces/DataProduct.fpp
LeStarch: Need to change the filename too!

LeStarch (approved):

PR #4674, opened 2026-01-30
ComQueue queue ordering, drop mode, & command to reprioritize

Willmac16:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Adds the abilities to change the ordering and drop mode of each queue in the com queue at compile time.

Also adds the ability to reprioritize queues w/ a command

## Rationale

Many channels benefit from alternative queue ordering & drop mode (e.g. new telemetry is likely more important to preserve than stale telemetry). Additionally, mission demands like anomaly resolution may require the ability to reonfigure the priority of different data streams into the com queue. 

Ensure the return scheme for Queue `DROP_OLDEST` mode feels appropriate. Decide if compile time vs command-able split in features matches expected usage. Verify that testing looks sufficient.

## Future Work

N/A

Claude Code for the Web (Sonnet 4.5) was used to perform initial implementation. I went through and read, reorganized, & refactored the code

### DIFF
--- a/Svc/ComQueue/ComQueue.cpp
+++ b/Svc/ComQueue/ComQueue.cpp
@@ -182,7 +182,7 @@ void ComQueue::SET_QUEUE_PRIORITY_cmdHandler(FwOpcodeType opCode,
     for (FwIndexType prioIndex = 0; prioIndex < TOTAL_PORT_COUNT; prioIndex++) {
         // If the port based index matches, then update
         if (m_prioritizedList[prioIndex].index == queueIndex) {
-            m_prioritizedList[prioIndex].priority = static_cast<FwIndexType>(newPriority);
+            m_prioritizedList[prioIndex].priority = newPriority;
             break;  // Since we shouldn't find more than one queue at this port index
         }
     }
@@ -196,10 +196,6 @@ void ComQueue::SET_QUEUE_PRIORITY_cmdHandler(FwOpcodeType opCode,
                 QueueMetadata temp = m_prioritizedList[j];
                 m_prioritizedList[j] = m_prioritizedList[j + 1];
                 m_prioritizedList[j + 1] = temp;
-
-                // Update indices to match new positions
-                m_prioritizedList[j].index = j;
-                m_prioritizedList[j + 1].index = j + 1;
             }
         }
     }
--- a/Svc/ComQueue/ComQueue.cpp
+++ b/Svc/ComQueue/ComQueue.cpp
@@ -192,7 +192,7 @@ void ComQueue::SET_QUEUE_PRIORITY_cmdHandler(FwOpcodeType opCode,
     // Re-sort the prioritized list to maintain priority ordering
     // Using simple bubble sort since TOTAL_PORT_COUNT is typically small
     for (FwIndexType i = 0; i < TOTAL_PORT_COUNT - 1; i++) {
-        for (FwIndexType j = 0; j < TOTAL_PORT_COUNT - i - 1; j++) {
+        for (FwIndexType j = 0; (j < TOTAL_PORT_COUNT - i - 1) && (j < TOTAL_PORT_COUNT - 1); j++) {
             if (m_prioritizedList[j].priority > m_prioritizedList[j + 1].priority) {
                 // Swap metadata
                 QueueMetadata temp = m_prioritizedList[j];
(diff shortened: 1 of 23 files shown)


Svc/ComQueue/ComQueueCommands.fppi
```diff
@@ -0,0 +1,17 @@
...
+
+@ Flush all queues. This will discard all queued data removing it from eventual downlink. Buffers requiring
+@ ownership return will be returned via the bufferReturnOut port.
+async command FLUSH_ALL_QUEUES()
+
+@ Set the priority of a specific queue at runtime
+async command SET_QUEUE_PRIORITY(
+                                queueType: QueueType @< The Queue data type
+                                indexType: FwIndexType @< The index of the queue (within the supplied type) to modify
+                                newPriority: U32 @< New priority value for the queue
```
LeStarch: This uses a U32....  Is this the right type?

Svc/ComQueue/ComQueueEvents.fppi
```diff
@@ -0,0 +1,16 @@
+@ Queue overflow event
+event QueueOverflow(
+    queueType: QueueType @< The Queue data type
+    index: U32 @< index of overflowed queue
```
LeStarch: Index should be a `FwIndexType` as it must be bounded by the input port size.

Svc/ComQueue/ComQueueTelemetry.fppi
```diff
@@ -0,0 +1,5 @@
+@ Depth of queues of Fw::ComBuffer type
+telemetry comQueueDepth: ComQueueDepth id 0
+
+@ Depth of queues of Fw::Buffer type
+      telemetry buffQueueDepth: BuffQueueDepth id 1
```
LeStarch:
```suggestion
telemetry buffQueueDepth: BuffQueueDepth id 1
```

Svc/ComQueue/ComQueueCommands.fppi
LeStarch: Priority is set in the C++ as a `FwIndexType`.  We should keep this for consistency, no?

Svc/ComQueue/ComQueue.cpp
```diff
@@ -147,6 +158,59 @@ void ComQueue ::FLUSH_ALL_QUEUES_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
...
+    for (FwIndexType i = 0; i < TOTAL_PORT_COUNT - 1; i++) {
+        for (FwIndexType j = 0; j < TOTAL_PORT_COUNT - i - 1; j++) {
+            if (m_prioritizedList[j].priority > m_prioritizedList[j + 1].priority) {
+                // Swap metadata
+                QueueMetadata temp = m_prioritizedList[j];
+                m_prioritizedList[j] = m_prioritizedList[j + 1];
+                m_prioritizedList[j + 1] = temp;
+
+                // Update indices to match new positions
+                m_prioritizedList[j].index = j;
```
LeStarch: I don't think you want to do this, do you?  The index should map back to the topology index.....not the order in the prioritized list.  Or have I missed something?

Svc/ComQueue/ComQueueCommands.fppi
Willmac16: I switched all the priority instances to FwIndexType. Good Catch! Now I'm wondering if there's anywhere that ensures the priority is >= 0 for the configuration on startup

Willmac16: It looks like negative priorities won't get populated, but I don't see anything that will explicitly warn you

Svc/ComQueue/ComQueue.cpp
Willmac16: No you're totally correct. I think there was some initial confusion since the description of the QueueMetadata index field initially referred to it as the priority index not a queueIndex.

Svc/ComQueue/ComQueue.cpp
```diff
@@ -147,6 +158,55 @@ void ComQueue ::FLUSH_ALL_QUEUES_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
...
+
+    // Find our queue in the prioritized list & update the priority
+    for (FwIndexType prioIndex = 0; prioIndex < TOTAL_PORT_COUNT; prioIndex++) {
+        // If the port based index matches, then update
+        if (m_prioritizedList[prioIndex].index == queueIndex) {
+            m_prioritizedList[prioIndex].priority = newPriority;
+            break;  // Since we shouldn't find more than one queue at this port index
+        }
+    }
+
```
LeStarch, line 191: Should assert (static_assert) TOTAL_PORT_COUNT >= 1

LeStarch (dismissed):
@Willmac16 looks good to me, I would recommend a static assert that `TOTAL_PORTS >= 1` for future proofing, but it looks good to me!

LeStarch (dismissed):

LeStarch (approved):

PR #4677, opened 2026-01-31
Add option to force assertions to always abort

celskeggs:
| | |
|:---|:---|

---
## Change Description

Add an option to let projects decide whether the framework should force assertions to always abort.

If enabled, this option allows prevents code from running after an assertion trips. This enables additional compiler optimizations.

If disabled, this option allows the FATAL event handler to decide whether code should continue running after an assertion trips. This is the default and is consistent with the existing assertion behavior in F Prime. Therefore, users should not observe any change in behavior unless they make the decision to change the configuration.

## Rationale

When assertions have the possibility of returning, the compiler needs to generate code for that case. This code can in some cases be complex enough to prevent inlining of otherwise short functions when under `-Os`. This can result in increased code size and performance degradation.

The existing default behavior is important for projects that need the ability to disable abort-on-FATAL and abort-on-ASSERT at runtime. However, just because this behavior is needed for some projects, does not mean it should be required for all projects.

While the ability to provide alternative assertion hooks is valuable, it is insufficient, as it is often not possible for the compiler to predict the assertion hook that will be used at compile time. The compiler needs to be confident that `Fw::SwAssert` cannot return _at the point of each use_.

Please consider whether the newly introduced option has the right name and description.

## Future Work

None in particular.

Not used.


Fw/Types/Assert.cpp
```diff
@@ -105,15 +106,15 @@
...
-I8 defaultSwAssert(FILE_NAME_ARG file,
-                   FwSizeType lineNo,
-                   FwSizeType numArgs,
-                   FwAssertArgType arg1,
-                   FwAssertArgType arg2,
-                   FwAssertArgType arg3,
-                   FwAssertArgType arg4,
-                   FwAssertArgType arg5,
-                   FwAssertArgType arg6) {
+FW_SWASSERT_NORETURN void defaultSwAssert(FILE_NAME_ARG file,
```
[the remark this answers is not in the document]
asd585140-glitch: ضبط

Fw/Types/Assert.hpp
```diff
@@ -49,17 +49,21 @@
 #define FW_ASSERT_NO_OVERFLOW(value, T) \
     FW_ASSERT((value) <= std::numeric_limits<T>::max(), static_cast<FwAssertArgType>(value))
 
+#if FW_ASSERTIONS_ALWAYS_ABORT
```
LeStarch, line 51: Shouldn't this be removed?  I think it is already done on line 64.....so this seems irrelevant and it uses the __attribute__((noreturn)) without checking it.

Fw/Types/Assert.cpp
```diff
@@ -124,23 +125,37 @@ I8 defaultSwAssert(FILE_NAME_ARG file,
...
+#if !FW_ASSERTIONS_ALWAYS_ABORT
+    return 0;
+#endif
 }
 
 I8 SwAssert(FILE_NAME_ARG file, FwAssertArgType arg1, FwSizeType lineNo) {
-    return defaultSwAssert(file, lineNo, 1, arg1, 0, 0, 0, 0, 0);
+    defaultSwAssert(file, lineNo, 1, arg1, 0, 0, 0, 0, 0);
+#if !FW_ASSERTIONS_ALWAYS_ABORT
+    return 0;
```
LeStarch: This usage means that the `I8` argument isn't returned in all cases.  Since we are always returning 0, why not strip these returns altogether?

LeStarch (changes requested):
One minor change, one question.

Fw/Types/Assert.hpp
celskeggs: Not sure I understand... there's a big difference between `__attribute__((noreturn))` and `__attribute__((analyzer_noreturn))`

Fw/Types/Assert.cpp
celskeggs: Hmm... I tried to preserve the existing implementation. However, it may not be necessary. I will try removing these and see if it all still works.

Fw/Types/Assert.hpp
LeStarch: Sorry, miss-read the attribute. We definitely need both.  Since this is behind a #define anyway, we can get awat without that big check block for the existence of `__attribute__((noreturn))`

LeStarch (dismissed):

LeStarch (approved):
@timcanham want to take a look?

LeStarch:
Got verbal approval from @timcanham supporting this!

Issue #4360, opened 2025-10-27
Data Product Catalog Tests Fail to Clean Up Correctly


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

On failure, the Data Product Catalog will leave residual files on the disk. This should be cleaned-up.

LeStarch:
The work-around is to clean up the file by hand before running.  This should not affect CI because CI always runs in a clean state.

thomas-bc:
Fixed by #4714

PR #4714, opened 2026-02-12
Clean up DpCatalog test artifacts in destructor

heathdutton:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4360 |
|**_Has Unit Tests (y/n)_**| Y (existing tests pass) |

---
## Change Description

Adds `std::system("rm -rf ./DpTest*")` to the `DpCatalogTester` destructor, mirroring the identical cleanup already present in the constructor.

## Rationale

When a test assertion fails, the cleanup at the end of `readDps()` never runs, leaving residual `DpTest*` directories on disk. The constructor cleans up before the next run, but only if you run the tests again. Adding cleanup to the destructor ensures artifacts are removed on every exit since C++ guarantees destructor execution during stack unwinding.

All 19 existing DpCatalog unit tests pass. Verified no `DpTest*` artifacts remain on disk after the test run completes.

## Future Work

None.

Claude Code was used for codebase exploration, identifying the root cause, and drafting the one-line fix. The change was reviewed and tested manually.

### DIFF
--- a/Svc/DpCatalog/test/ut/DpCatalogTester.cpp
+++ b/Svc/DpCatalog/test/ut/DpCatalogTester.cpp
@@ -32,6 +32,7 @@ DpCatalogTester ::DpCatalogTester()
 
 DpCatalogTester ::~DpCatalogTester() {
     this->component.deinit();
+    std::system("rm -rf ./DpTest*");
 }
 
 // ----------------------------------------------------------------------


Svc/DpCatalog/test/ut/DpCatalogTester.cpp
```diff
@@ -32,6 +32,7 @@ DpCatalogTester ::DpCatalogTester()
 
 DpCatalogTester ::~DpCatalogTester() {
     this->component.deinit();
+    std::system("rm -rf ./DpTest*");
```
thomas-bc, line 35: Should we be using our OSAL instead of this platform-specific call?

LeStarch (approved):
This is consistent with our other UTs.

Issue #4737, opened 2026-02-19
Circular Buffer Tests Failure


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

From CI:

```
[STest::Random] Generated seed 543961 from system time
[==========] Running 8 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 8 tests from CircularBufferTests
[ RUN      ] CircularBufferTests.RandomCircularTests
Ran 1000 steps.
[       OK ] CircularBufferTests.RandomCircularTests (150 ms)
[ RUN      ] CircularBufferTests.BasicSerializeTest
[       OK ] CircularBufferTests.BasicSerializeTest (0 ms)
[ RUN      ] CircularBufferTests.BasicOverflowTest
/Users/runner/work/fprime/fprime/STest/STest/Rule/Rule.hpp:52: Failure
Value of: this->precondition(state)
  Actual: false
Expected: true
precondition failed applying rule serializeOverflow
[  FAILED  ] CircularBufferTests.BasicOverflowTest (6 ms)
[ RUN      ] CircularBufferTests.BasicPeekTest
[       OK ] CircularBufferTests.BasicPeekTest (0 ms)
[ RUN      ] CircularBufferTests.BasicPeekBadTest
[       OK ] CircularBufferTests.BasicPeekBadTest (0 ms)
[ RUN      ] CircularBufferTests.BasicRotateTest
[       OK ] CircularBufferTests.BasicRotateTest (0 ms)
[ RUN      ] CircularBufferTests.BasicRotateBadTest
[       OK ] CircularBufferTests.BasicRotateBadTest (0 ms)
[ RUN      ] CircularBufferTests.BoundaryCases
[       OK ] CircularBufferTests.BoundaryCases (0 ms)
[----------] 8 tests from CircularBufferTests (157 ms total)
```

## Context / Environment

CI

## How to Reproduce

Run with above seed.

## Expected Behavior

UTs pass

Gregox273:
For `SerializeOverflowRule::precondition` to return false, `CircularState::m_random_size` must be zero? Presumably we only want to use nonzero buffer sizes for this test case?

As an aside, I can't reproduce the failing test with the given seed; is there some guidance/documentation available re. how to do this?

thomas-bc:
I haven't seen this pop back up, and it appears a PR has been merged. Closing.

PR #4752, opened 2026-02-22
Modify circular buffer overflow test to handle zero-length random input

Gregox273:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4737  |

---
## Change Description

Modify basic overflow test to handle zero-length random input - if input has zero length then we don't expect a call to `Types::CircularBuffer::serialize` to return an error code, even if the circular buffer is full.

An alternative approach might be to modify `MockTypes::CircularState::generateRandomBuffer()` so that it doesn't generate zero-size random buffers, or to parameterise the function so that an allowable range of `random_size` can be passed in by the caller / per test case.

## Rationale

Fix failing test case.

### DIFF
--- a/Utils/Types/test/ut/CircularBuffer/CircularRules.cpp
+++ b/Utils/Types/test/ut/CircularBuffer/CircularRules.cpp
@@ -42,12 +42,13 @@ void SerializeOkRule::action(MockTypes::CircularState& state) {
 SerializeOverflowRule::SerializeOverflowRule(const char* const name) : STest::Rule<MockTypes::CircularState>(name) {}
 
 bool SerializeOverflowRule::precondition(const MockTypes::CircularState& state) {
-    return state.getRemainingSize() < state.getRandomSize();
+    return state.getRemainingSize() <= state.getRandomSize();
 }
 
 void SerializeOverflowRule::action(MockTypes::CircularState& state) {
     Fw::SerializeStatus status = state.getTestBuffer().serialize(state.getBuffer(), state.getRandomSize());
-    ASSERT_EQ(status, Fw::FW_SERIALIZE_NO_ROOM_LEFT);
+    Fw::SerializeStatus expected_status = (state.getRandomSize() == 0) ? (Fw::FW_SERIALIZE_OK) : (Fw::FW_SERIALIZE_NO_ROOM_LEFT);
+    ASSERT_EQ(status, expected_status);
 }
 
 PeekOkRule::PeekOkRule(const char* const name) : STest::Rule<MockTypes::CircularState>(name) {}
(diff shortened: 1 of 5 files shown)


Utils/Types/test/ut/CircularBuffer/CircularRules.cpp
```diff
@@ -42,12 +42,13 @@ void SerializeOkRule::action(MockTypes::CircularState& state) {
 SerializeOverflowRule::SerializeOverflowRule(const char* const name) : STest::Rule<MockTypes::CircularState>(name) {}
 
 bool SerializeOverflowRule::precondition(const MockTypes::CircularState& state) {
-    return state.getRemainingSize() < state.getRandomSize();
+    return state.getRemainingSize() <= state.getRandomSize();
```
LeStarch: The overflow rule is about testing specifically the overflow case.  I think in this case, we should have the precondition:

```suggestion
    return (state.getRemainingSize() < state.getRandomSize()) && (state.getRandomSize() != 0);
```

As this would just prevent the rule when not overflowing.

LeStarch (changes requested):

Utils/Types/test/ut/CircularBuffer/CircularRules.cpp
LeStarch: We should also check that `state.getRandomSize()` is idempotent across the access in `precondition` and `apply`.  Otherwise this does not work.

Gregox273: OK, yes, makes sense. This doesn't address the manual `BasicOverflowTest` failing though since the precondition ASSERT fails if `state.getRandomSize()` returns `0`. Would it make sense to parameterize `Types::RandomizeRule` so that bounds on buffer size can be (optionally) specified in the constructor? That way we can place a lower limit of `1` on the random buffer size used in `BasicOverflowTest`.

I don't believe `CircularState::m_random_size` is mutated after it gets set in `randomGo.apply()` so the precondition check should be valid.

LeStarch: Based on the description, yes, I think this would make sense. Want to take a pass at it so I can see what it looks like in practice?

If it is possible, use default parameters so that we don't have lots of other code churn.

Gregox273: Sure, I'll have a pass at it

Gregox273: I think this addresses the failing `BasicOverflowTest` by placing a lower bound of 1 on the random buffer size. I've updated the precondition to match your suggestion so that the overflow test doesn't get automatically run with a random buffer size of zero in `RandomCircularTests`.

LeStarch (approved):

thomas-bc (approved):

PR #4764, opened 2026-02-24
Use GH Actions inputs through environment variables

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Follow best practice. This prevents potential problems when the input strings are user-defined and can potentially lead to unexpected behavior. Using env vars, the inputs are strictly interpreted as strings.

## Rationale

Security best practice

## AI Usage

Fully authored by AI with some handholding. Reviewed myself before pushing the PR.

### DIFF
--- a/.github/workflows/ext-cookiecutters-test.yml
+++ b/.github/workflows/ext-cookiecutters-test.yml
@@ -41,6 +41,9 @@ jobs:
   Validate:
     runs-on: ubuntu-latest
     needs: [ get-tools-branch, get-bootstrap-branch ]
+    env:
+      BOOTSTRAP_BRANCH: ${{ needs.get-bootstrap-branch.outputs.target-branch }}
+      TOOLS_BRANCH: ${{ needs.get-tools-branch.outputs.target-branch }}
     steps:
       # Checkout only the bootstrap.expect file, since the full F´ repo will be 
       # checked out as part of the fprime-bootstrap process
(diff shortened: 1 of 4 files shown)


LeStarch (commented):
Why is this necessary/what problem does it solve?  Can you elaborate more?

thomas-bc:
cc @bitWarrior 

This is a security best practice documented by GitHub: [Use an intermediate environment variable](https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable)

These changes fix a potential issue where user-controllable inputs (like `inputs.target_repository`) were directly interpolated into bash scripts. These aren't untrusted inputs as they are source controlled in our own repo; so the risk is quite low. But it doesn't hurt to follow best practice.

Most changes are to direct inputs, some are indirect inputs (output of another Action), and a few are purely stylistic to match with the other inputs (e.g. such as `VERBOSE_FLAG`)

thomas-bc:
Actually, because we intend for some of these workflows to be reusable by other projects (and we don't control where these projects use inputs from), this is a very legitimate use of this best practice imo.

RinZ27:
Reviewed the diff and the approach is solid. Mapping inputs to environment variables is the right way to handle potential shell injection in these workflows. I like that you caught all the instances across the different YAML files.

LeStarch (approved):

PR #4765, opened 2026-02-24
Topology connection unit tests

Kronos3:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fpp/issues/840 |

---
## Change Description

Adds GTest-tests for checking connections inside a topology. The purpose of this is to test the topology code-generation in preparation for "direct/optimized" topology port connections.

The idea here is to use two components: Sender/Receiver. The sender will drive the test cases. It'll send out a message to the receiver. The receiver will serialize the message and reply back to the sender on a single port.

We are utilizing the TypedPorts interface (split up into multiple sub-interfaces) to test type serialization/deserialization/abi across the port boundary.


default/config/FpConfig.h
```diff
@@ -21,6 +21,11 @@ extern "C" {
 // Configuration switches
 // ----------------------------------------------------------------------
 
+// Direct port calls (TODO)
```
LeStarch, line 24: Why is this marked as (TODO) should it not be done or removed?

Kronos3: We need this option in the code generation config (even though it's not merged in yet). I think the `TODO` here is more of a `TODO` for the description of this option.

Kronos3:
@LeStarch This should be good to go (unless you want me to remove the above TODO). The `-DFW_DIRECT_PORT_CALLS` currently doesn't do anything with this version of FPP though it is needed for development of the direct port call codegen.

LeStarch (approved):

Issue #4767, opened 2026-02-25
S_ISVTX issue in File.cpp


kevin-f-ortega:
| | |
|:---|:---|
|**_F´ Version_**| `4.1.1` |
|**_Affected Component_**|  Os/Posix/File.cpp |
---
## Problem Description

`S_ISVTX` is part of the optional XSI -- which is an optional interface in POSIX. Some POSIX build systems don't use XSI, so this file fails to compile.

## Context / Environment

Execute `fprime-util version-check` and share the output.

```
prime-util version-check
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-4.18.0-553.40.1.el8_10.x86_64-x86_64-with-glibc2.28
Python version: 3.13.3
CMake version: 3.26.0
Pip version: 26.0.1
Pip packages:
    fprime-tools==4.1.0
    fprime-gds==4.1.1a4
    fprime-fpp==3.1.1a4
Project submodules:
    https://github.com/fprime-community/fprime-vxworks.git @ v4.1.1
```

## How to Reproduce

1. Check out fprime-vxworks-reference
2. Compile reference for the POSIX version

## Expected Behavior

To be able to compile

PR #4768, opened 2026-02-25
S_ISVTX is part of XSI, which is an optional set of interfaces and ex…

kevin-f-ortega:
…tensions on top of the POSIX standard. Some systems do not build with the optional XSI so this file fails to compile in those cases.

| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/4767 |

---
## Change Description

Use a build macro to include the line of code iff the definition exists

## Rationale

Some POSIX build systems don't use the optional XSI interface

CI

## Future Work

None

N/A


kevin-f-ortega:
I ran this update with my vxworks CI system. It passed

LeStarch (approved):

PR #4779, opened 2026-02-27
CCSDS AOS Deframer

Willmac16:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|  y |
|**_Documentation Included (y/n)_**| n  |

---
## Change Description

Add a CCSDS AOS SDL version 5 defamer that supports:
* Space Packet Protocol
* Encapsulation Packet Protocol
* FECF on or off
* 1 virtual channel--all the internal code should be ready for multiple just needs agreement on config & output port setup (i.e. vc per port array index)

## Rationale

You need an AOS Deframer to rx AOS Frames from payloads/other craft.
AOS lets you do fixed frame width w/ variable packet size (including packets that span multiple frames)

Take a look at the processing I'm doing on SPP & EPP packets and what I'm emitting in the context for them. I could imagine a later rev of this component sharing w/ SPP defamer more logic/doing its job for it.

Other facet on the SPP & EPP processing is how light my validation is. I'm only checking that they contain the right PVN & grabbing size + idle. If we moved the full SPP parsing into this/called into a lib then you could maybe avoid throwing away packets by ID-ing a problem earlier (although dropping a frame upon packet issues is probably the only safe thing to do).

A double check on my read of the EPP spec would be beneficial since I caught a few bugs that lingered all the way into testing.

## Future Work

I am likely done w/ AOS for near term; however, future work could be:
* Porting downlink stack to AOS
* Using AOS Framer to bootstrap USLP
* Making a proper EPP Framer & Deframer
* Consolidating logic into a lib btw AOS, SPP, & EPP deframers
* Writing an AOS Frame Detector if using AOS over a byte stream like tcp vs a datagram protocol like UDP

* Claude Code for Web did the first pass
* The Claude Code GitHub extension addressed feedback on 3-4 commits
* Over the past couple days, while I worked in VSCode I would use ChatGPT Codex & Claude Code CLI for minor code movement/event creation and test tweaks. I've reviewed (and quite often heavily refactored) all of the flight code to match my stylistic and structural preferences

Overall, I'm not convinced the first 2 steps really provided that much of a speedup since much of the EPP structure was completely wrong, and the approach to spanning packets had significant code duplication. Maybe a minor psychological speedup simply because I had code to review to bootstrap rather than a blank slate.

The minor test writing & event creation usage felt like a speedup, but better editor skills/snippets might be able to handily beat that.


Willmac16:
This PR should look less scary once #4630 merges

Svc/Ccsds/AosDeframer/AosDeframer.hpp
```diff
@@ -0,0 +1,193 @@
...
+    struct AosDeframerVc {
+        U8 vcStructIndex = 0xFF;                                     //!< Index into VC array for this vc struct
+        U8 virtualChannelId = 0;                                     //!< VCID for this virtual channel
+        U8 pvnMask = PvnBitfield::SPP_MASK | PvnBitfield::EPP_MASK;  //!< Bitmask of enabled PVNs
+
+        // Telemetry counters (per-VC)
+        U32 framesProcessed = 0;   //!< Total frames received on this VC
+        U32 packetsExtracted = 0;  //!< Total packets extracted from this VC
+        U32 vcFrameCount = 0;      //!< Last received virtual channel frame count from header
+
```
thomas-bc: Should some of these be `FwSizeType` - or is there a reason to use strictly 32 bits?

Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -0,0 +1,530 @@
...
+    // Extract VC Frame Count Cycle if in use (Section 4.1.2.5.3)
+    if ((header.get_frameCountAndSignaling() & AOSHeaderSubfields::cycleCountFlagMask) != 0) {
+        const U8 vcFrameCountCycle = header.get_frameCountAndSignaling() & AOSHeaderSubfields::vcFrameCountCycleMask;
+        // Extend the 24-bit frame count with the 4-bit cycle count
+        vcFrameCount |= static_cast<U32>(vcFrameCountCycle) << 24;
+    }
+
+    // Gap detect after the first accepted frame on a VC
+    if (vc->framesProcessed > 0U) {
+        const U32 expectedVcFrameCount = vc->vcFrameCount + 1U;
```
thomas-bc, line 236: This should wrap around at 24 bits, no?

Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -0,0 +1,530 @@
...
+    }
+
+    // Set the default context only if we haven't for this packet already
+    // Otherwise our PVN tracker gets overwritten
+    if (!vc->spanningPacket.buffer.isValid()) {
+        vc->spanningPacket.context = packetContext;
+    }
+
+    // Update telemetry
+    this->tlmWrite_FramesProcessed(++vc->framesProcessed);
```
thomas-bc: Should this be moved to after `extractPackets` in the event that packet extraction asserts? Or is it voluntarily prior to it?

Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -0,0 +1,530 @@
...
+        }
+        // If no spanning packet active, this continuation data cannot be used
+        return;
+    }
+
+    // There is continuation data before the first packet header
+    if (firstHeaderPointer > 0 && vc.spanningPacket.bytesReceived > 0) {
+        (void)this->appendToSpanningPacket(vc, dataZone, static_cast<FwSizeType>(firstHeaderPointer));
+        // We must be done w/ the prior packet since we have a FHP
+        this->abandonSpanningPacket(vc);
```
thomas-bc, line 407: It may be useful to log an event or something here in case we're abandoning something that hasn't been emitted - could this done by checking if `vc.spanningPacket.buffer` has been reset?

thomas-bc: Nevermind I just saw this is already done within `abandonSpanningPacket` which makes sense

Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -0,0 +1,530 @@
...
+        if (vcFrameCount != expectedVcFrameCount) {
+            this->log_WARNING_HI_VcFrameCountGap(vcId, vcFrameCount, expectedVcFrameCount);
+            this->notifyErrorIfConnected(Ccsds::FrameError::AOS_VC_FRAME_COUNT_GAP);
+            // Other errors will implicitly drop their spanning packet once we finally lock back onto a valid frame
+            this->abandonSpanningPacket(*vc);
+        }
+    }
+
+    // Store VC frame count in the VC struct for reference (e.g., gap detection)
+    this->tlmWrite_LatestVcFrameCount(vc->vcFrameCount = vcFrameCount);
```
thomas-bc: Let's split this in two lines for clarity

Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -0,0 +1,530 @@
...
+        this->log_WARNING_HI_InvalidFrameLength(data.getSize(), m_fixedFrameSize);
+        this->notifyErrorIfConnected(Ccsds::FrameError::AOS_INVALID_LENGTH);
+        this->dataReturnOut_out(0, data, context);
+        return;
+    }
+
+    // Validate FECF if enabled (Section 4.1.6)
+    // FrameDetector + FrameAccumulator or Lower Protocol Layer should enforce whole AOS Frames
+    if (m_fecfEnabled && !this->validateFecf(data)) {
+        this->dataReturnOut_out(0, data, context);
```
thomas-bc, line 107: abandon current spanning packet if failed to validate Fecf ?

Svc/Ccsds/AosDeframer/AosDeframer.fpp
```diff
@@ -0,0 +1,51 @@
...
+    @ - Frame Error Control Field (FECF) per Section 4.1.6
+    @ - Space Packet Protocol (SPP) extraction per CCSDS 133.0-B-2
+    @ - Encapsulation Packet Protocol (EPP) extraction per CCSDS 133.1-B-3
+    passive component AosDeframer {
+
+        constant NumVcs = 1
+
+        # TODO: Multi VC | figure out storage and telemetry round up of per VC stat counters
+        type VCxU32 = U32
+        type VCxU8 = U8
```
thomas-bc: Why define these aliases?

Svc/Ccsds/AosDeframer/AosDeframer.cpp
thomas-bc: I'm noticing the value needs to be incremented prior, but I still wonder whether it'd be more accurate to write the tlm afterwards? idk

Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -0,0 +1,530 @@
...
+    const FwSizeType dataZoneStart = AOSHeader::SERIALIZED_SIZE + M_PDUHeader::SERIALIZED_SIZE;
+    const FwSizeType dataZoneEnd = m_fixedFrameSize - (m_fecfEnabled ? AOSTrailer::SERIALIZED_SIZE : 0);
+    const FwSizeType dataZoneSize = dataZoneEnd - dataZoneStart;
+    U8* dataZone = data.getData() + dataZoneStart;
+
+    // Handle special First Header Pointer values (Section 4.1.4.2.2.4)
+    if (firstHeaderPointer == M_PDUSubfields::FHP_IDLE_DATA_ONLY) {
+        // Frame contains only idle data - abandon any in-progress spanning packet
+        this->log_ACTIVITY_LO_IdleFrame(vc.virtualChannelId);
+        this->abandonSpanningPacket(vc);
```
thomas-bc: I get that it wouldn't make much sense from a ground side to end up in that scenario, but is this in the standard that an Only Idle Data frame means we need to drop any currently spanning packet? I have not found a reference of that.

Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -0,0 +1,530 @@
...
+
+    // Seek amount
+    FwSizeType seekForward = 0;
+
+    if (!vc.spanningPacket.buffer.isValid()) {
+        // Fill the tmp header buff w/ what we got
+        const FwSizeType headerCap = AosDeframerVc::SpanningPacketState::HEADER_BUF_SIZE;
+        const FwSizeType toHeader = FW_MIN(size, headerCap - vc.spanningPacket.bytesReceived);
+        if (toHeader > 0) {
+            ::memcpy(vc.spanningPacket.headerBuf + vc.spanningPacket.bytesReceived, data, toHeader);
```
thomas-bc, line 302: The logic is a bit complex to follow here, could we add a few comments? I'm also not entirely certain that there isn't a very niche edge-case in case FwSizeType isn't large, could there be a scenario where:
1. you've accumulated a very large spanning packet
2. `headerCap - vc.spanningPacket.bytesReceived` ends up wrapping around to a value lower than received `size`
3. I think you'd hit an invalid `memcpy()` below?

Probably at least should assert both `toHeader <= 8` and `vc.spanningPacket.bytesReceived < 8` here?

Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -0,0 +1,530 @@
...
+    // Determine packet type from PVN (upper 3 bits of first byte)
+    U8 pvn = getPacketVersion(packetStart[0]);
+    // Default to invalid, override if valid (non-idle) packet
+    vc.spanningPacket.context.set_pvn(ComCfg::Pvn::INVALID_UNINITIALIZED);
+
+    // Size the Packet (so we can alloc a buffer)
+    switch (ComCfg::Pvn pvnEnum = static_cast<ComCfg::Pvn::T>(pvn)) {
+        case ComCfg::Pvn::SPACE_PACKET_PROTOCOL:
+        // Intentionally fallthrough since logic is more condensed this way
+        case ComCfg::Pvn::ENCAPSULATION_PACKET_PROTOCOL:
```
thomas-bc, line 458: Maybe it would be easier to just remove the switch-case and handle this with a guarding if(invalidPvn) { return } ? This feels a bit odd.

Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -0,0 +1,530 @@
...
+
+    if (status != Fw::FW_SERIALIZE_OK) {
+        return 0;  // Incomplete header - spans to next frame
+    }
+
+    // Per CCSDS 133.0-B-2 Section 4.1.3.5.2, packet data length = (actual length - 1)
+    FwSizeType totalPacketSize = SpacePacketHeader::SERIALIZED_SIZE + header.get_packetDataLength() + 1;
+
+    // TODO: Unify Deframers | bring the whole spp processing into this component
+    // since we're only missing seq count logic?
```
thomas-bc, line 482: ### Future work 

I agree. I've also had this feeling for a while that maybe there's a broader issue, it could be that a cleaner architecture would be to **not** have SpacePacket Framer/Deframer in F Prime, and instead have helper mechanisms (e.g. helper classes? helper component?) that enables components to work _with_ Space Packets. This would fit in the idea that components could benefit from being able to "speak" SpacePacket (e.g. have a SpacePacketHub).

After all, they are different layers of the protocol stack, so it does sound like they shouldn't be handled at the same layer of our Comms component stack (`Svc.Framer` interface for both protocols)

Thoughts?

thomas-bc: Another solution that's kind of in the middle of both worlds is that the SpacePacketFramer/Deframer should offer interfaces to inspect/generate spacepackets, without passing them out directly to the comms stack.

Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -0,0 +1,530 @@
...
+        // Load the header into the dynamic buffer
+        ::memcpy(vc.spanningPacket.buffer.getData(), vc.spanningPacket.headerBuf, vc.spanningPacket.bytesReceived);
+    }
+
+    // Already have the dynamic buffer, so fill away
+    const FwSizeType spaceLeft = vc.spanningPacket.buffer.getSize() - vc.spanningPacket.bytesReceived;
+    // Copy what we got
+    const FwSizeType toBody = FW_MIN(size, spaceLeft);
+    if (toBody > 0) {
+        ::memcpy(vc.spanningPacket.buffer.getData() + vc.spanningPacket.bytesReceived, data, toBody);
```
thomas-bc, line 351: also add asserts on the size/pointers being copied here

Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -0,0 +1,530 @@
...
+            const FwSizeType remainingLength = seekForward + remainingBody;
+            if (remainingLength > size) {
+                return 0;
+            } else {
+                return remainingLength;
+            }
+        }
+
+        // Load the header into the dynamic buffer
+        ::memcpy(vc.spanningPacket.buffer.getData(), vc.spanningPacket.headerBuf, vc.spanningPacket.bytesReceived);
```
thomas-bc, line 340: assert bytesReceived <= HEADER_BUF_SIZE ?

thomas-bc (changes requested):
Looks good! A few notes and questions. I'm also overall seeing very few asserts in the processing parts of the code, it'd be good if we add some.

Svc/Ccsds/AosDeframer/AosDeframer.cpp
Willmac16: 24, base & 28 w/ the cyclic extension. right architecture in my mind is to compare whatever bits are enabled in the source (e.g. match to 24 or 28)

Svc/Ccsds/AosDeframer/AosDeframer.cpp
Willmac16: I'm a pretty firm believer in making the assignment and telemetry of values inseparable if possible. If this really goes against F' house style I can be convinced to change it, but my preference is to write to member var + to tlm stream in one motion.

Svc/Ccsds/AosDeframer/AosDeframer.fpp
Willmac16: These can be cut from this rev of it. The thought process was to make array types for multi-VC support. The only real advantage of keeping these is that they call out what is on a per vc basis (like how the vc struct separates global from per vc state).

Svc/Ccsds/AosDeframer/AosDeframer.cpp
Willmac16: We wait to drop the packet until we find out that we've skipped a vc frame count (and therefore are missing the rest of the packet) on a valid FECF frame

Svc/Ccsds/AosDeframer/AosDeframer.cpp
Willmac16: In my mind packet extraction should never assert. Bytes on the wire are untrusted input so any protocol violations should result in events not asserts.

I'll take a look at moving this elsewhere

Svc/Ccsds/AosDeframer/AosDeframer.cpp
Willmac16: If you have more data for your spanning packet, then  you'd shove the rest of your packet into the frame. Only case where I'd really understand idle in the middle is if the whole frame ends up idle (because continuations must be at the start of a frame). But I don't think you'd utilize the packet service for that, that feels like an AOS frame level/virtual channel level idle.

Svc/Ccsds/AosDeframer/AosDeframer.cpp
Willmac16: yeah, the switch case made a whole lot more sense when I forgot that I called 2 separate functions for epp and spp

Svc/Ccsds/AosDeframer/AosDeframer.cpp
Willmac16: Either of those seems quite reasonable to me. I do like the outcome where Fprime natively supports directing space packets to components & can emit them from components as well

Svc/Ccsds/AosDeframer/AosDeframer.cpp
Willmac16: replaced w/ an early return if PVN is not enabled & now the switch just calls epp vs spp specifics

Svc/Ccsds/AosDeframer/AosDeframer.cpp
thomas-bc: I'm not strongly opinionated on this, but as far as I understand it, this goes against the JPL coding standards, which we try to abide by.

https://yurichev.com/mirrors/C/JPL_Coding_Standard_C.pdf

> Rule 24
There should be no more than one statement or variable declaration per
line. A single exception is the C for-loop, where the three controlling
expressions (initialization, loop bound, and increment) can be placed on a
single line. 

There are cases where this doesn't really apply to C++, but I think this usage here is an example of things that this rule targets.

Willmac16: I'll split it up then. Thanks for finding the standard

Svc/Ccsds/AosDeframer/AosDeframer.fpp
Willmac16: removed

Svc/Ccsds/AosDeframer/AosDeframer.cpp
Willmac16: added those asserts & more more comments

Willmac16:
I believe that should address all of your comments

Svc/Ccsds/AosDeframer/AosDeframer.cpp
Willmac16: image
I took another look at CCSDS 732.0-B-5 Page 4-13 and  looks like idle packet tx in the middle of a packet that spans multiple frames is fully valid. I'll go back through and clean this logic up.

I'll also make sure this deframer properly handles idle packets that span (i believe currently works by invalidating the current spanning packet).

thomas-bc: Huh - interesting that it's called out. Thanks for finding it!

Willmac16: Not sure what to do other than drop both if you get a partial idle in the middle of a partial packet. Otherwise if the idle doesn’t span, should be able to keep the partial going

Willmac16: Thought I wrote up a follow up, but looks like I never sent it. I was confusing OID (only idle data) frames w/ idle packets. I believe my initial reply does correctly summarize behavior for Idle packets; however, for an OID frame we now no longer drop any in progress spanning packets.

thomas-bc:
@Willmac16  `Svc_Ccsds_AosDeframer_ut_exe` is failing, are you able to take a look?

Willmac16:
Yeah I’ll try to resolve tmrwSent from my iPhoneOn Apr 30, 2026, at 20:30, Thomas Boyer-Chammard ***@***.***> wrote:﻿thomas-bc left a comment (nasa/fprime#4779)
@Willmac16  Svc_Ccsds_AosDeframer_ut_exe is failing, are you able to take a look?

Willmac16:
Looks like the macOS/regular linux fails are bc I changed that Only Idle Data abandon behavior but failed to correct the UT. Patch going up shortly for that

And then RHEL8 casting errors, my old friend.

Willmac16:
Can you fight the RHEL8 error for me? Unless I can quickly run the tests against it, it's very hard to iterate and converge on what exactly its mad about

thomas-bc:
Here is what Opus 4.7 helped me find:

> The error isn't really about the shift operand or the inner static_cast<U16>(...) — it's the compound assignment itself. U16 |= U16 expands to U16 = U16 | U16, and per C++ integer-promotion rules the bitwise OR of two unsigned short operands is computed as int, so the implicit narrowing on the assignment back (int → U16) is what -Wconversion flags. GCC unhelpfully reports the location at the closing ) of the inner static_cast, which makes it look like the cast is the problem when it's actually one level up.

So the way to appease GCC (not specific to RHEL8!! it just looks like the rest of our CI is using clang? _EDIT: No, RHEL CI is using an old version of GCC and that's the issue - the Ubuntu CI is using a more recent version of GCC that doesn't have the issue_) is to not do the U16 |= U16, and do everything in one operation. Or expand into `scid = static_cast<U16>(scId | [expr])` but it became even less readable imo. I'm happy with that. Thoughts?

Willmac16:
Whatever makes it pass works w/ me! Do we really use clang for Ubuntu?Sent from my iPhoneOn May 6, 2026, at 17:29, Thomas Boyer-Chammard ***@***.***> wrote:﻿thomas-bc left a comment (nasa/fprime#4779)
Here is what Opus 4.7 helped me find:

The error isn't really about the shift operand or the inner static_cast(...) — it's the compound assignment itself. U16 |= U16 expands to U16 = U16 | U16, and per C++ integer-promotion rules the bitwise OR of two unsigned short operands is computed as int, so the implicit narrowing on the assignment back (int → U16) is what -Wconversion flags. GCC unhelpfully reports the location at the closing ) of the inner static_cast, which makes it look like the cast is the problem when it's actually one level up.

So the way to appease GCC (not specific to RHEL8!! it just looks like the rest of our CI is using clang?) is to not do the U16 |= U16, and do everything in one operation. I'm happy with that. Thoughts?

thomas-bc:
@Willmac16 ah you're right. Then I think it may be that RHEL is using an older version of GCC (`GNU 8.5.0` per CI logs) that has this quirk, where ubuntu's version (`GNU 11.4.0` per CI logs) is smarter about it.

thomas-bc:
@Willmac16 ok I think we're almost ready to merge this. I opened a PR on your branch for a few security-related items: https://github.com/Willmac16/fprime/pull/8 - let me know. 

These are the kind of issues that the community has been reporting to us in our comms stack, and we've been spending a lot of time fixing. So I figured it'd be worth fixing them from the get-go here.

Once those changes are accepted into your branch I am ready to merge this.

Willmac16:
Totally agree with fixing things before pre-merge; thanks for taking care of them!

Willmac16:
I think we might need to abandon any spanning packets upon an out of bounds first header pointer. I’ll double check if that’s already happening, but I think you can’t assume that any state should continue through an invalid frame. Otherwise we’ll get some Franken-packetSent from my iPhoneOn May 6, 2026, at 20:06, Thomas Boyer-Chammard ***@***.***> wrote:﻿thomas-bc left a comment (nasa/fprime#4779)
@Willmac16 ok I think we're almost ready to merge this. I opened a PR on your branch for a few security-related items: Willmac16#8 - let me know.
These are the kind of issues that the community has been reporting to us in our comms stack, and we've been spending a lot of time fixing. So I figured it'd be worth fixing them from the get-go here.
Once those changes are accepted into your branch I am ready to merge this.

Willmac16:
Is this ready to merge? finally

thomas-bc (approved):

Issue #4512, opened 2025-12-05
Utils/Hash multiple static analysis findings


djbyrne17:
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**|  Utils/Hash |
---
## Problem Description

Utils/Hash has 54 SonarQube findings of varying severity, and 14 codesonar findings worth looking at.

Also, more assertions would be useful, but that is not in scope of this ticket.

### At least the four SonarQube "Blocker" severity findings should be fixed:
```
S1669  Replace "final" with another name to avoid conflict with the keyword "final".
<High> Utils/Hash/Hash.hpp:87: cpp:S1669
<High> Utils/Hash/Hash.hpp:93: cpp:S1669
<High> Utils/Hash/libcrc/CRC32.cpp:67: cpp:S1669
<High> Utils/Hash/libcrc/CRC32.cpp:77: cpp:S1669
```

### And at least the two CodeSonar "High" severity findings should be fixed:
```
<High> Utils/Hash/libcrc/CRC32.cpp:61: Uninitialized Variable Uninitialized Variable found in Utils::Hash::update
<High> Utils/Hash/HashBufferCommon.cpp:22: Virtual Call in Constructor Virtual Call in Constructor found in Utils::HashBuffer::HashBuffer
```

### The other SonarQube findings should also be reduced if not all eliminated
They are:
```
<High> Utils/Hash/HashBuffer.hpp:26: cpp:S3624 Customize this class' destructor to participate in resource management.
<High> Utils/Hash/HashBufferCommon.cpp:10: cpp:S3490 Use "=default" instead of the default implementation of this special member functions.
<High> Utils/Hash/HashBufferCommon.cpp:17: cpp:S3490 Use "=default" instead of the default implementation of this special member functions.
<High> Utils/Hash/libcrc/CRC32.cpp:27: cpp:S3490 Use "=default" instead of the default implementation of this special member functions.
<High> Utils/Hash/libcrc/CRC32.cpp:32: cpp:S5008 Replace this use of "void *" with a more meaningful type.
<High> Utils/Hash/libcrc/CRC32.cpp:56: cpp:S5008 Replace this use of "void *" with a more meaningful type.
<High> Utils/Hash/libcrc/lib_crc.h:57: cpp:S5028 Replace this macro by "const", "constexpr" or an "enum".
[...]

PR #4799, opened 2026-03-03
Resolve Utils/Hash static analysis findings

vsoulgard:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4512 |

---
## Change Description

All severity findings have been resolved. Likely all other findings are resolved as well, excluding false positives.

### Details:
- `Utils::Hash::final` renamed to `Utils::Hash::finalize` to resolve the conflict with `final` keyword.
- `Utils::HashBuffer::operator==` refactored as a friend function.

## AI Usage

AI was used for research. All code implementation was done by me.


thomas-bc:

```bash
```

vsoulgard:
Thank you very much for the guidance @thomas-bc! Fixed.

Utils/Hash/libcrc/lib_crc.c
thomas-bc, line 1: This file is not authored by F´, let's revert the changes there.

Utils/Hash/HashBuffer.hpp
```diff
@@ -49,21 +49,21 @@ class HashBuffer : public Fw::SerializeBufferBase {
 
     //! Compare two hash buffers for equality
     //!
-    bool operator==(const HashBuffer& other) const;
+    friend bool operator==(const HashBuffer& lhs, const HashBuffer& rhs);
```
thomas-bc: @bocchino do you have thoughts on the operator change here? It seems good to me, but I haven't seen that used anywhere in the framework so maybe there is a reason?

thomas-bc (changes requested):

Utils/Hash/HashBuffer.hpp
bocchino: From an internet search, it looks like the main reason to use a non-member `==` operator is to preserve symmetry of implicit type conversions. However, we try to avoid implicit type conversions, and the member `==` operator is better for encapsulation. What is motivating this change?

vsoulgard: This change was motivated by these SonarQube findings, mentioned in #4512:

<Med> Utils/Hash/HashBuffer.hpp:52: cpp:S2807 Make this member overloaded operator a hidden friend.
<Med> Utils/Hash/HashBuffer.hpp:56: cpp:S2807 Make this member overloaded operator a hidden friend.

thomas-bc: Ok. Static analysis has flagged this. Here is the SonarQube rule: https://next.sonarqube.com/sonarqube/coding_rules?open=cpp%3AS2807&rule_key=cpp%3AS2807

Given what you just said, I am inclined to revert this change and dispose of static analysis as a Won't fix.

Utils/Hash/HashBufferCommon.cpp
```diff
@@ -7,17 +7,18 @@
...
     FW_ASSERT(Fw::FW_SERIALIZE_OK == stat, static_cast<FwAssertArgType>(stat));
 }
 
-HashBuffer::~HashBuffer() {}
+HashBuffer::~HashBuffer() = default;
 
 HashBuffer::HashBuffer(const HashBuffer& other) : Fw::SerializeBufferBase() {
-    Fw::SerializeStatus stat = Fw::SerializeBufferBase::setBuff(other.m_bufferData, other.getSize());
+    Fw::SerializeStatus stat =
+        Fw::SerializeBufferBase::setBuff(other.m_bufferData, other.Fw::SerializeBufferBase::getSize());
```
thomas-bc: Just noticing this `other.Fw::SerializeBufferBase::getSize()` - is this needed? Doesn't look like it should be? It's quite unusual looking

vsoulgard: It should fix this CodeSonar finding, but I'm not 100% sure:
<High> Utils/Hash/HashBufferCommon.cpp:22: Virtual Call in Constructor Virtual Call in Constructor found in Utils::HashBuffer::HashBuffer

We call explicitly `Fw::SerializeBufferBase::getSize()`, so this is not virtual call anymore.

vsoulgard: Nevermind. Probably this is a false positive from CodeSonar, the virtual call is performed on the other object which is fully constructed, not on this. I should have noticed this. I'll revert this change.

thomas-bc (approved):

thomas-bc:
Breaking Change Notes:

### `Utils::Hash` finalize method

The `Utils::Hash::final()` method has been renamed `Utils::Hash::finalize()` to deconflict with the C++ `final` keyword. No functional change. Users of the `Utils::Hash` class must update their usage of that method to the new name.

Issue #4745, opened 2026-02-20
Update Core components to use FPP constants for string arguments


timcanham:
| | |
|:---|:---|
|**_F´ Version_**| 4.x|
|**_Affected Component_**| various |
---
## Feature Description

Change the string length arguments for common components to match the FPP configuration values.

For instance, FPP configuration says command strings are 40 characters:

https://github.com/nasa/fprime/blob/devel/default/config/FpConstants.fpp#L29

`Svc/FileManager` has the following string argument length (256):

https://github.com/nasa/fprime/blob/devel/Svc/FileManager/Commands.fppi#L29

Could we do:

```
async command ShellCommand(
                            $command: string size FW_CMD_STRING_MAX_SIZE @< The shell command string
                            logFileName: string size FileNameStringSize @< The name of the log file
                          ) \
  opcode 0x04
```

## Rationale

Users can't change the string length in F Prime core repo components, so it causes a mismatch and misunderstandings about what the argument length really can be.

timcanham:
@bocchino Why wouldn't this work?

```
/home/tcanham/source/fprime/Svc/FileManager/Commands.fppi:29.51
                            $command: string size FW_CMD_STRING_MAX_SIZE @< The shell command string
                                                  ^
  included at /home/tcanham/source/fprime/Svc/FileManager/FileManager.fpp:51.13
error: symbol FW_CMD_STRING_MAX_SIZE is not defined
note: looking for a constant here
```

bocchino:
That seems like it should work. Did you try regenerating and rebuilding? Sometimes the build system doesn't regenerate the dependency set properly.

timcanham:
@LeStarch @thomas-bc I see this used a lot for event string arguments:

```
    @ File written
    event FileWritten(
                       bytes: U32 @< The number of bytes written
                       file: string size FileNameStringSize @< The file name
                     ) \
      severity activity low \
      format "Wrote {} bytes to file {}"
```

Would it be better to replace the `FileNameStringSize` with `FW_LOG_STRING_MAX_SIZE`? `FileNameStringSize` could be bigger.

bocchino:
> Would it be better to replace the FileNameStringSize with FW_LOG_STRING_MAX_SIZE? FileNameStringSize could be bigger.

[...]

PR #4804, opened 2026-03-04
Update string size arguments for Svc Components to match configuration values in FPP

timcanham:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/4745  |

---
## Change Description

Update some string length arguments to tie them to FPP  constants

## Rationale

They are currently set independently from the config, so sometimes they are way to small or way too large compared to max values in the config FPP. Users can't change them, so they are stuck with the values.

Re-run unit tests.

## Future Work

None

No usage


LeStarch:
@timcanham are you going to move this out of draft, or should I close it?

PR #4805, opened 2026-03-04
Topology Ports in FPP

Kronos3:
Testing for https://github.com/nasa/fpp/pull/940


bocchino (approved):
Looks good!

Ref/Top/RefPackets.fppi
```diff
@@ -9,10 +9,10 @@ telemetry packets RefPackets {
...
     FileHandling.fileManager.CommandsExecuted
 
-    Ref.cmdSeq.CS_LoadCommands
-    Ref.cmdSeq.CS_CancelCommands
-    Ref.cmdSeq.CS_CommandsExecuted
-    Ref.cmdSeq.CS_SequencesCompleted
+    cmdSeq.CS_LoadCommands
+    cmdSeq.CS_CancelCommands
+    cmdSeq.CS_CommandsExecuted
+    cmdSeq.CS_SequencesCompleted
```
Kronos3, line 15: Heads up @thomas-bc @LeStarch.

Behavior of component instance symbol resolution with this FPP update changes a bit. `Ref` here picks up the Ref topology since topologies & component instances are now treated as "interface instances" and names resolve as one.

Kronos3: The zephr build failure here is a real build failure because of this change. We'll need to update the reference deployment with this corresponding change in the telemetry packets.

thomas-bc: @Kronos3 thanks for the heads up. Does the FPP spec specify how instance symbol resolution is done? 
That, and how topologies/instances are namespaced in C++ (`::` vs `_` is sometimes used in my experience) is not entirely clear to me. 
Do we have docs about that somewhere? That would help me (and likely other people) a lot.

bocchino: @thomas-bc @LeStarch Can we revise the Zephyr CI so it builds but does not attempt to run? And/or make the build and run separate checks? That way we can get the benefit of CI on the build. Right now a build failure can be masked by the "it always fails" attempt to run.

thomas-bc: For the needed Zephyr reference update, are you familiar with the `pr-xxxx` branch process?
https://github.com/nasa/fprime/blob/devel/CONTRIBUTING.md#automated-checks-on-reference-repositories
Let me know if you don't have permissions to create a branch on that repo.

bocchino: > Do we have docs about that somewhere? 

Yes, this is all extensively described in the User's Guide and in the spec. In particular, every place where we use `_` instead of `::` in the code gen (and why) is called out in the User's Guide.

thomas-bc: @Kronos3 I sent you an invite for permissions on https://github.com/fprime-community

thomas-bc: @bocchino Ok thank you I will go read that ! :D 

> Can we revise the Zephyr CI so it builds but does not attempt to run? 

Can you clarify? Currently there's two sequential jobs, first one builds (no run), and second one pulls the build artifacts and runs it with integration tests.

bocchino: I think the issue that we are seeing here is that if you have this
```
module M {
  constant a = 0
  module M {
    constant b = a
  }
}
```
The outer `M` is hidden in the inner scope, so `constant b = M.a` would not work; but `a` is visible in the inner scope.

In general, I've come to believe that shadowing/hiding symbols is confusing. FPP allows it because other languages allow it. But maybe we should avoid it as a matter of style. In particular, maybe putting the `Ref` topology in the `Ref` namespace is not ideal, and we should consider revising one of the names to something else -- for example, we could have `Ref.Topology`, as in the subtopology setup.

bocchino: > Currently there's two sequential jobs, first one builds (no run), and second one pulls the build artifacts and runs it with integration tests.

But they both produce the same red X in the top-level CI output. So a run failure can mask a build failure unless one goes in and checks the details. If the runs fail a lot, there's a danger of mistaking a build failure as a run failure.

bocchino: > FPP allows it because other languages allow it.

Note that in the C++ compilation we have moved to using `-Wshadow -Werror` which disallows shadowing.

bocchino: > Note that in the C++ compilation we have moved to using -Wshadow -Werror which disallows shadowing.

Actually it just disallows shadowing of variables. It still allows hiding of namespaces, which can cause confusion.

thomas-bc: Absolutely agree that `Ref.Ref` is a confusing name and we should switch one.

Kronos3:
@thomas-bc I believe this is ready to merge. We'd like to get the FPP change merged in ASAP since there are a couple of features that will need to untangle merge changes since it's fairly deep cuttting.

thomas-bc (approved):

thomas-bc:
🎉 🥳 🎉 🎉 🥳 🎉 🎉 🥳 🎉 🎉 🥳 🎉

PR #4810, opened 2026-03-04
FileWorker

### LINKED ISSUE #4795: Deliver FileWorker component
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

`Svc.FileWorker` component is an active F' component used for writing and reading files on the filesystem. Other components needing to read or write files may use this component to perform large/slow file operations without missing their deadlines.

rjtsao:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/4795 |

---
## Change Description

Adding FileWorker common component, converting from old XML format

Pending integrated test with FSW image manager component (work in progress)


Svc/FileWorker/docs/sdd.md
```diff
@@ -0,0 +1,102 @@
...
+|---|---|
+| IDLE | FileWorker is ready to accept requests for transfer |
+| READING | In the read process |
+| WRITING | In the write process |
+
+
+## Ports
+| Kind | Port Name | Port Type | Usage |
+|------|------|-----------|-------|
+|guarded input|`writeIn`|`Svc.FileReadWrite`|Initiates a file write|
```
LeStarch: Port type is invalid here.  Should be `async` according to the FPP

Svc/FileWorker/docs/sdd.md
```diff
@@ -0,0 +1,102 @@
+# Svc::FileWorker
+
+Svc::FileWorker is an active F' component used for writing and reading files on the filesystem. Other components needing to read or write files may use this component to perform large/slow file operations without missing their deadlines.
+
+## Requirements
+| Name | Description | Validation |
```
thomas-bc, line 7:
```suggestion

| Name | Description | Validation |
```
Our Markdown rendering engine doesn't like tables that don't have newlines above/below. Can we also fix in the other places?

Svc/FileWorker/FileWorker.cpp
```diff
@@ -0,0 +1,552 @@
...
+    this->verifyDoneOut_out(0, workerStat, static_cast<U32>(fileSize));
+}
+
+void FileWorker ::writeIn_handler(FwIndexType portNum, const Fw::StringBase& path, Fw::Buffer& buffer, U64 offset) {
+    
+    this->m_lock.lock();
+    this->m_abort = true;
+    this->m_lock.unLock();
+
+    char fileName[FppConstant_MAX_STRING_BYTES::MAX_STRING_BYTES];
```
thomas-bc: Using the full `FppConstant_MAX_STRING_BYTES::MAX_STRING_BYTES` shouldn't be needed FYI, we usually just use `<Namespace>::CONSTANT_NAME`. So in your case it seems that `Svc::MAX_STRING_BYTES` should work. We should refine that name/namespace since it's global to the framework.

Svc/FileWorker/FileWorker.cpp
```diff
@@ -0,0 +1,552 @@
...
+        Os::File::Status readStat = file.read(hashValuePtr, hashSize);
+        if(readStat != Os::File::OP_OK) {
+            Fw::LogStringArg s(hashFileName);
+            this->log_WARNING_HI_WriteValidationReadError(s, readStat);
+            return false;
+        }
+        Utils::HashBuffer tmp(hashValuePtr, hashSize);
+        hash.setHashValue(tmp);
+        hash.update(data, size);
+        hash.final(hashBuffer);
```
thomas-bc: This function has been renamed

```suggestion
        hash.finalize(hashBuffer);
```

thomas-bc: @rjtsao looks like this has been reverted

Svc/FileWorker/docs/sdd.md
```diff
@@ -0,0 +1,101 @@
...
+|---|---|
+| IDLE | FileWorker is ready to accept requests for transfer |
+| READING | In the read process |
+| WRITING | In the write process |
+
+## Ports
+
+| Kind | Port Name | Port Type | Usage |
+|------|------|-----------|-------|
+|guarded input|`writeIn`|`Svc.FileReadWrite`|Initiates a file write|
```
LeStarch: This should be a async input:

```suggestion
|async input|`writeIn`|`Svc.FileReadWrite`|Initiates a file write|
```

Svc/Ports/FilePorts/FileWorkerPorts.fpp
```diff
@@ -0,0 +1,29 @@
+module Svc {
+
+    constant MAX_STRING_BYTES = 120
```
LeStarch: We have a standard file string name length that should be used here.

```suggestion
```

Svc/Ports/FilePorts/FileWorkerPorts.fpp
```diff
@@ -0,0 +1,29 @@
...
+
+    port FileRead(
+        path: string size MAX_STRING_BYTES
```
LeStarch:
```suggestion
        path: string size FileNameStringSize
```

Svc/Ports/FilePorts/FileWorkerPorts.fpp
```diff
@@ -0,0 +1,29 @@
...
+        ref buffer: Fw.Buffer
+    )
+
+    port FileWrite(
+        path: string size MAX_STRING_BYTES
```
LeStarch:
```suggestion
        path: string size FileNameStringSize
```

Svc/Ports/FilePorts/FileWorkerPorts.fpp
```diff
@@ -0,0 +1,29 @@
...
+    port SignalDone(
+        status: U32
+        arg: U32
+    )
+
+    port CancelStatus(
+    ) -> I8
+
+    port VerifyStatus(
+        path: string size MAX_STRING_BYTES
```
LeStarch:
```suggestion
        path: string size FileNameStringSize
```

Svc/FileWorker/FileWorker.fpp
```diff
@@ -0,0 +1,215 @@
+module Svc {
+
+    active component FileWorker {
+
+        # ----------------------------------------------------------------------
+        # Ports
+        # ----------------------------------------------------------------------
+
+        async input port writeIn: Svc.FileWrite
```
LeStarch, line 10: Minor: it would be nice to have these annotated

Svc/Ports/FilePorts/FileWorkerPorts.fpp
```diff
@@ -0,0 +1,29 @@
...
+        offsetBytes: U64
+    )
+
+    port SignalDone(
+        status: U32
+        arg: U32
+    )
+
+    port CancelStatus(
+    ) -> I8
```
LeStarch: This has a return status that is always "DONE".  What is the purpose when there is no differentiation of status?

```suggestion
    )
```

Svc/FileWorker/FileWorker.cpp
```diff
@@ -0,0 +1,544 @@
...
+    this->m_append = append;
+}
+
+FileWorker ::~FileWorker() {}
+
+// ----------------------------------------------------------------------
+// Handler implementations for typed input ports
+// ----------------------------------------------------------------------
+
+I8 FileWorker ::cancelIn_handler(FwIndexType portNum) {
```
LeStarch: I think we need to remove the status.

```suggestion
void  FileWorker ::cancelIn_handler(FwIndexType portNum) {
```

Svc/FileWorker/FileWorker.cpp
```diff
@@ -0,0 +1,544 @@
...
+    // Cancel for reading
+    this->m_lock.lock();
+    this->m_abort = true;
+    this->m_lock.unLock();
+    return DONE;
```
LeStarch:
```suggestion
```

Svc/FileWorker/FileWorker.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+    void configure(U64 chunkSize, bool append);
+
+  private:
+    static constexpr U64 BLOCK_SIZE_BYTES = 4096;
+    static constexpr U32 TIMEOUT_MS = 1000000;
+    static constexpr U32 MAX_LOOP_ITERATIONS = 32;
+
+    Svc::FileWorkerState m_state;
+    Os::Mutex m_lock;
```
LeStarch: Minor: we could remove this lock by making the `m_abort` into a `std::atomic`.

Svc/FileWorker/FileWorker.cpp
```diff
@@ -0,0 +1,544 @@
...
+bool FileWorker ::writeBufferToFile(Fw::Buffer& buffer, const char* fileName, U64 offset) {
+    FW_ASSERT(buffer.getData() != nullptr);
+    FW_ASSERT(fileName != nullptr);
+    Fw::LogStringArg logStringArg(fileName);
+
+    // Open file
+    Os::File file;
+    Os::File::Status stat = Os::File::OP_OK;
+
+    if (!this->m_append) {
```
LeStarch: The append boolean is confusing.  Shouldn't this be an argument to the write command?  Seems like some writes should be appends, and others plain writes.

Svc/FileWorker/FileWorker.cpp
```diff
@@ -0,0 +1,544 @@
...
+    }
+
+    this->verifyDoneOut_out(0, workerStat, static_cast<U32>(fileSize));
+}
+
+void FileWorker ::writeIn_handler(FwIndexType portNum,
+                                  const Fw::StringBase& path,
+                                  Fw::Buffer& buffer,
+                                  U64 offsetBytes) {
+    this->m_lock.lock();
```
LeStarch: This block doesn't make sense.  It will force an abort if busy and then immediately exit as "not in idle".

Svc/FileWorker/FileWorker.cpp
```diff
@@ -0,0 +1,544 @@
...
+            // Only check timeout if > 0
+            now = this->getTime();
+            diff = Fw::Time::sub(now, start);
+            elapsed = diff.getSeconds() * static_cast<U64>(1000000) + diff.getUSeconds();
+            if (elapsed >= timeout) {
+                return FW_READ_TIMEOUT;
+            }
+        }
+
+        bytesRead += static_cast<U32>(readAmt);
```
LeStarch: Use of U32 here is odd.  Everywhere else sizes are `U64`.

Svc/FileWorker/FileWorker.cpp
```diff
@@ -0,0 +1,544 @@
...
+    Os::FileSystem::Status fsStat;
+    FwSizeType fileSize = 0;
+    FileWorkerStatus workerStat = FW_STATUS_DONE;
+
+    FW_ASSERT(fileNamePtr != nullptr);
+
+    // Get checksum
+    crcStat = Utils::verify_checksum(fileNamePtr, crcFromFile, crcCalculated);
+
+    switch (crcStat) {
```
LeStarch: Minor: this block seems to be present 3x in the file.  It is a good candidate for the helper.

Svc/FileWorker/FileWorker.cpp
```diff
@@ -0,0 +1,544 @@
...
+
+    // New write request overrides any leftover abort state
+    this->m_lock.lock();
+    this->m_abort = false;
+    this->m_lock.unLock();
+
+    // Save filename
+    // NB: may count null terminator due to FPRIME/fprime-sw#57, but should still be less than MAX_STRING_BYTES in any
+    // case
+    length = static_cast<U32>(strnlen(path.toChar(), Svc::MAX_STRING_BYTES));
```
LeStarch: This static cast to U32 seems wrong.  Shouldn't it be U64?

Also, we should probably use Fw::StringUtil::string_length() instead of `strnlen`.

Svc/FileWorker/FileWorker.cpp
```diff
@@ -0,0 +1,544 @@
...
+    FW_ASSERT(length < Svc::MAX_STRING_BYTES && length < sizeof(fileName), static_cast<FwAssertArgType>(length));
+
+    (void)strncpy(fileName, path.toChar(), sizeof(fileName));
```
LeStarch: Same with this.

Svc/FileWorker/FileWorker.cpp
```diff
@@ -0,0 +1,544 @@
...
+
+    // Open file
+    fileStat = f.open(fileName, Os::File::OPEN_READ);
+    if (fileStat != Os::File::OP_OK) {
+        this->log_WARNING_HI_OpenFileError(fileNameStr, fileStat);
+        return FW_STATUS_FAILED_TO_OPEN;
+    }
+
+    // Get buffer data and size
+    readSize = static_cast<U32>(buffer.getSize());
```
LeStarch: Same with this cast

LeStarch (changes requested):
Must fix: there are a lot of `static_cast<U32>()` for managing sizes, but where the selected size elsewhere are `U64`.  We must cast to `U64`.

Minor:  `U64` is used for sizes.  In modern F Prime we prefer `FwSizeType`.

LeStarch (approved):

PR #4811, opened 2026-03-04
Fix RawTime serializability concerns

### LINKED ISSUE #4351: Os::RawTime and Serializability
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

`Os::RawTime` implements serializable.  However, `Os::RawTime` does not know its own type, rather, it uses a delegate object of some `Os::RawTimeInterface` to determine the structure of the raw time object.

This was done in order to allow raw time to be implemented by different backing implementations based on the platform.

However, since the structure is not known, serialization, equivalence, etc is invalid because it depends on some unknown type that may, or may not, be the same unknown type as the receiver.

To fix this we should do one of the following:
1. Remove serializable, equivalence, etc from raw time --or--
2. Implement some system to track RawTime's unknown type and prevent copying in this case (may involve rtti) --or--
3. Have Os::RawTime convert to Fw::Time as part of its implementation and deal in Fw::Time

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix #4351

### DIFF
--- a/Fw/Types/Serializable.hpp
+++ b/Fw/Types/Serializable.hpp
@@ -69,10 +69,6 @@ class Serializable {
     //! \return SerializeStatus indicating the result of the operation
     virtual SerializeStatus deserializeFrom(SerialBufferBase& buffer, Endianness mode = Endianness::BIG) = 0;
 
-    //! TODO: this operator should be deleted, this must be done after RawTime is modified though
-    // as it currently depends on this being defined
-    Serializable& operator=(const Serializable& src) = default;
-
     // ----------------------------------------------------------------------
     // Legacy methods for backward compatibility
     // ----------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


thomas-bc:
Couple notes after investigation:

https://github.com/nasa/fprime/blob/ffa562994669cbe038a483fc244b5874e0acfde1/Os/RawTime.cpp#L26

The assignment (`this->m_delegate = <...>`) is actually a no-op. The handles are copied inside the `RawTimeInterface::getDelegate(m_handle_storage, &other.m_delegate)` call. That's why this all works even though the assignment operator is inherited from Serializable.

LeStarch (approved):

thomas-bc:
LOL to the commit message that went into devel

Issue #4757, opened 2026-02-24
TlmPacketizer: Parameterize Configuration of MIN/MAX Values


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Parameterize the MIN/MAX table with a configurable default.

## Rationale

The MIN/MAX tables should be set on start-up and would ideally be a non-volatile.  The configurable default would provide a fallback to project static configuration.

PR #4813, opened 2026-03-05
Configurable Default for TlmPacketizer MIN/MAX Tables

Sarah5567:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4757 |

---
## Change Description

Parameterizes the MIN/MAX group config tables in `TlmPacketizer` with two new parameters: `DefaultMinDelta` and `DefaultMaxDelta`. Values are loaded from non-volatile storage on startup and can be updated at runtime via SET commands. Telemetry readback channels are included for both.

## Rationale

MIN/MAX tables need to be set on startup from non-volatile storage, with a static default as fallback.

Three unit tests added: default load, stored value load, and runtime SET update. All tests pass.

### DIFF
--- a/Svc/TlmPacketizer/TlmPacketizer.cpp
+++ b/Svc/TlmPacketizer/TlmPacketizer.cpp
@@ -628,15 +628,15 @@ FwIndexType TlmPacketizer::sectionGroupToPort(const FwIndexType section, const F
 }
 
 void TlmPacketizer::parametersLoaded() {
-    parameterUpdated(PARAMID_DEFAULTMINDELTA);
-    parameterUpdated(PARAMID_DEFAULTMAXDELTA);
+    parameterUpdated(PARAMID_DEFAULT_MIN_DELTA);
+    parameterUpdated(PARAMID_DEFAULT_MAX_DELTA);
 }
 
 void TlmPacketizer::parameterUpdated(FwPrmIdType id) {
     Fw::ParamValid valid;
     switch (id) {
-        case PARAMID_DEFAULTMINDELTA: {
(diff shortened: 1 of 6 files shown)


thomas-bc:
@Lex-ari if you are able to review on your end it would be greatly appreciated!

Lex-ari (changes requested):
I think the SDD should probably be updated  to describe the usage of the new parameters, since they effectively override MIN/MAX on all sections and groups if the parameter gets updated / loaded.

This is really cool! Now there's two ways of configuring MIN/MAX on start up, via Parameters or Sequences!

LeStarch:
I was expecting the whole table for these parameters rather than just a single cross-table version. @Lex-ari thoughts?

Lex-ari:
@LeStarch Having a table / parameter array for every group in each section is the ideal change, which allows the user to use both parameters and set cmd configuration together. This seems like a first step towards a parameterized default, and the upgrade would be turning this single use parameter into a table so that each group is configured independently. 

Although the desc for this PR describe using tables, which should be included in this change but seems like it's a generic U32 for the whole config @Sarah5567

Sarah5567:
@LeStarch @Lex-ari you're right, the current implementation uses a single value for min/max. I can refactor the PR to introduce a table instead.
Thanks for the review!

LeStarch:
@Sarah5567 I already have a partial implementation of the parameter table.  Mind if I close that one out?  Would you mind helping review that work?

Sarah5567:
No problem, feel free to close it. I'd be happy to take a look and review.

Issue #4815, opened 2026-03-05
Reader-Writer Lock


kevin-f-ortega:
| | |
|:---|:---|
|**_F´ Version_**|4.1.1 |
|**_Affected Component_**| Os::Mutex |
---
## Feature Description

Add a reader-writer mutex that allows multiple readers and a single writer. This mutex should be allowed to be used alongside the normal Os::Mutex

## Rationale

Sometimes multiple readers need to access a shared data

kevin-f-ortega:
Upon further study, this is not a feature we want to maintain in F Prime core. If there is a need for a project to use rw-locks, then they can implement it for their platform. If there is a high-enough need from multiple project for this feature, then we can reconsider.

More info here https://github.com/nasa/fprime/pull/4819#issuecomment-4210307175

PR #4819, opened 2026-03-09
Implement Reader-Writer Mutex (Os::RwMutex)

vsoulgard:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4815 |

---
## Change Description

Implement reader-writer mutex (`Os::RwMutex`), enabling shared read access with exclusive write access. The implementation follows the existing delegation pattern used by `Os::Mutex`:
- `Os::RwMutexInterface`: Abstract interface inheriting from `Os::MutexInterface`, adding `takeRead()`/`releaseRead()` methods for shared (read) locking alongside inherited `take()`/`release()` for exclusive (write) locking.
- `Os::RwMutex`: RAII wrapper class that delegates to a platform-specific implementation via the MutexHandleStorage pattern.
- `Os::Posix::RwMutex`: POSIX implementation using `pthread_rwlock_t` as the underlying handle for POSIX-compliant platforms.
- `Os::Stub::RwMutex`: Stub implementation using C++ atomic operations (std::atomic) to simulate reader-writer semantics. Uses a signed atomic state variable to track reader count (>0), unlocked (0), or writer-locked (-1)
- `Os::ReadLock` / `Os::WriteLock`: RAII guard classes for automatic lock management within scope, analogous to `Os::ScopeLock`.

### Details:
- Writer-preference policy: The implementation prioritizes waiting writers over new readers to minimize writer starvation. When a writer requests the lock, subsequent readers block until the writer completes.
- While writer-preference prevents writer starvation, recursive read locks by the same thread while write lock is waiting can cause deadlock.
- Unlike `Os::Stub::Mutex`, `Os::Stub::RwMutex` blocks via spin-lock waiting rather than returning `ERROR_BUSY`. This allows to simulate blocking synchronization.
- In contrast to `Os::Posix::Mutex`, the `Os::Posix::RwMutex` implementation lacks support for error-checking mode (`PTHREAD_MUTEX_ERRORCHECK`) and priority inheritance (`PTHREAD_PRIO_INHERIT`) due to POSIX `pthread_rwlock_t` constraints.

#### Usage example:
```cpp
Os::RwMutex rw_mutex;

// Shared access
void readData() {
    Os::ReadLock lock(rw_mutex);
    // ... read shared data ...
}

// Exclusive access
void updateData() {
    Os::WriteLock lock(rw_mutex);
    // ... modify shared data ...
}
```

## Rationale

`Os::RwMutex` could be more efficient in scenarios with frequent reading and infrequent writing than a regular mutex.

Two tests are intentionally skipped with `GTEST_SKIP()`:
```cpp
TEST_F(FunctionalityTester, PosixDeleteWriteLockedRwMutex) {
    GTEST_SKIP() << "Skipped: pthread_rwlock_destroy() does not reliably " \
        "return EBUSY for locked rwlock (POSIX undefined behavior). ";
}

TEST_F(FunctionalityTester, PosixDeleteReadLockedRwMutex) {
    GTEST_SKIP() << "Skipped: pthread_rwlock_destroy() does not reliably " \
        "return EBUSY for locked rwlock (POSIX undefined behavior). ";
}
```
Rationale for skipping: Per the POSIX specification ([pthread_rwlock_destroy](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_rwlock_destroy.html)), calling `pthread_rwlock_destroy()` on a locked reader-writer mutex results in undefined behavior. The standard does not require `pthread_rwlock_destroy()` to return any specific error code for this condition.

## Future Work

Adding docs might be a good idea.

## AI Usage

AI (QWEN) was used for the initial implementation of StubRwMutex, unit-tests and doxygen comments. All AI generated content was reviewed, modified, and validated for correctness.


thomas-bc:
This issue has yet to go through the CCB - we'll get back to you once it does. Thanks!

kevin-f-ortega:
Thank you, @vsoulgard ! This is the exact implementation I was thinking when I initially created https://github.com/nasa/fprime/issues/4815. However, discussing with other F Prime developers, we decided to update the current [Mutex interface](https://github.com/nasa/fprime/blob/devel/Os/Mutex.hpp) and implementation to handle multi-read single-write locks instead of having 2 different mutex implementations. I appreciate your efforts for addressing https://github.com/nasa/fprime/issues/4815

vsoulgard:
No problem, @kevin-f-ortega. But how you want to implement this? Since rwlock and mutex are different handles in POSIX, I cannot see them sharing one interface.

kevin-f-ortega:
@vsoulgard, actually, deeper evaluation and discussion with the team led us to understand that we do need this different interface. Main reason is Os::ConditionalVariable needs a regular Mutex and not a Rw-lock. 

So I'm reopening this PR

kevin-f-ortega:
Thank you for this contribution. I can see you put significant effort into following F' Prime's patterns. The delegation design, test framework, and documentation structure are well done.

Upon further investigation, read-write lock is not functionality that we want to maintain inside of core F Prime. Some reasons:

1. Avoiding writer starvation can be implementation dependent.  See here https://pubs.opengroup.org/onlinepubs/009696899/functions/pthread_rwlock_rdlock.html under DESCRIPTION. I saw you used `pthread_rwlockattr_setkind_np` to avoid writer starvation, but that function is part of the GNU extension. That makes this implementation not portable.
2. The posix read-write lock does not handle priority inheritance, which leads to priority inversion. See here https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_rwlock_wrlock.html under APPLICATION USAGE.
3. There is not a pressing need to have RWLock

Thank you for your time and effort. I look forward to a future contribution from you.

PR #4821, opened 2026-03-09
Fixed corner case in shadow write when writing 0 bytes in append mode

kevin-f-ortega:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/fprime-community/fprime-baremetal/issues/18 |
|**_Documentation Included (y/n)_**|  n|

---
## Change Description

Update shadow-write function to update current position when in APPEND mode

## Rationale

A test that opens a file in APPEND mode and tries to write 0 bytes to a file would result in the file position changing to the end of file. The shadow class is missing this feature, which resulted in some random-based unit tests failing.

Run CI

## Future Work

N/A

N/A

### DIFF
--- a/Os/test/ut/file/FileRules.cpp
+++ b/Os/test/ut/file/FileRules.cpp
@@ -53,6 +53,12 @@ void Os::Test::FileTest::Tester::shadow_write(const std::vector<U8>& write_data)
     if (write_data.data() != nullptr) {
         status = m_shadow.write(write_data.data(), size, Os::File::WaitType::WAIT);
     }
+    // If in APPEND mode, need to set position to the end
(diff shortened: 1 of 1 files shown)


thomas-bc (approved):

thomas-bc:
Noting down here in case it comes back: saw a new failure on Os::File that I had not seen before since merging this:

https://github.com/nasa/fprime/actions/runs/22924426388/job/66531178040?pr=4828

kevin-f-ortega:
> Noting down here in case it comes back: saw a new failure on Os::File that I had not seen before since merging this:
> 
> https://github.com/nasa/fprime/actions/runs/22924426388/job/66531178040?pr=4828

Do you know what the UT seed was that caused that error?

Issue #4758, opened 2026-02-24
Increase Copying Efficiency and Remove Double-Buffering


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

We should increase the copying efficiency and remove the double buffering scheme. This will improve both CPU and memory efficiency.

This can be done by pre-computing the "ready to send" value and then copy into a single stack com buffer, rather than a full copy of the table.

## Rationale

Locking over a full packet set copy is more CPU intensive then required. Storing a complete list of copied buffers increases memory by 2x.

PR #4822, opened 2026-03-10
TlmPacketizer: Remove Double-Buffering

Sarah5567:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4758 |

---
## Change Description

Replace the double-buffering in TlmPacketizer::Run_handler with a per-packet stack snapshot. Instead of copying the entire fill buffer table into a separate send buffer table, each packet is snapshotted individually (buffer, time, level, updated flag) under a short lock before iterating over sections.

## Rationale

The previous double-buffer approach stored a full mirror of all packet buffers, doubling memory usage. It also locked over a full table copy which is more CPU intensive than needed. The new approach snapshots one packet at a time on the stack, keeping the lock duration minimal and eliminating the extra buffer table entirely.

### DIFF
--- a/Svc/TlmPacketizer/TlmPacketizer.cpp
+++ b/Svc/TlmPacketizer/TlmPacketizer.cpp
@@ -47,7 +47,6 @@ TlmPacketizer ::TlmPacketizer(const char* const compName)
     // clear packet buffers
     for (FwChanIdType buffer = 0; buffer < MAX_PACKETIZER_PACKETS; buffer++) {
         this->m_fillBuffers[buffer].updated = false;
-        this->m_sendBuffers[buffer].updated = false;
     }
 
     // enable sections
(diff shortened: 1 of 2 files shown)


Sarah5567:
Hey, @LeStarch,
did you mean that it should copy lazily, only when a section actually needs to send?
I looked into it, but it seems problematic: if section 0 skips sending and section 1 decides to send later, the buffer may have been updated in between, The sections would then be working with different versions of the same packet, making the updated flag ambiguous.
I think we need to snapshot the packet once before iterating over sections, so all sections work from a consistent version. Is that what you had in mind?

LeStarch (approved):
This fixes the double-buffering part.  We still need to address the fact that every packet is copied in this model, not just the updated ones.  This can be done in a follow-on PR.

LeStarch:
@Sarah5567 I am going to merge this, because it is a step in the right direction.  To fix the copy-every-buffer-problem, we would need to effectively do this:

1. loop through all sections/packets and calculate a "need-to-send" flag.
2. loop again and copy the buffer/update time send

Although I haven't looked too deeply at the specifics.

Issue #4494, opened 2025-12-03
`teardown` Is Non-Pure Virtual


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

`teardown` in `Os::QueueInterface` is non-pure-virtual.  Thus `Os::QueueInterface` is not pure.

PR #4824, opened 2026-03-10
Make `Os::QueueInterface::teardown()` pure virtual

vsoulgard:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4494 |

---
## Change Description

Make `Os::QueueInterface::teardown()` pure virtual and implement it in derived classes.

### DIFF
--- a/Os/Queue.hpp
+++ b/Os/Queue.hpp
@@ -83,8 +83,6 @@ class QueueInterface {
     //!
     //! Allow for queues to deallocate resources as part of system shutdown. This delegates to the underlying queue
     //! implementation.
-    //!
-    //! Note: the default implementation does nothing.
     virtual void teardown() = 0;
 
     //! \brief send a message into the queue
--- a/Os/Queue.hpp
+++ b/Os/Queue.hpp
@@ -85,7 +85,7 @@ class QueueInterface {
     //! implementation.
     //!
     //! Note: the default implementation does nothing.
-    virtual void teardown() {}
+    virtual void teardown() = 0;
 
     //! \brief send a message into the queue
     //!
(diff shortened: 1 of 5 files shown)


thomas-bc (approved):

thomas-bc (changes requested):
From CI:
```
/Users/runner/work/fprime/fprime/Os/Queue.hpp:86:18: note: unimplemented pure virtual method 'teardown' in 'StubQueue'
   86 |     virtual void teardown() = 0;
```
Looks like `Os/Stub` needs to implement `teardown`

thomas-bc:
And by the way, thank you for all your recent contributions!

thomas-bc:

```bash
```

vsoulgard:
Thanks again @thomas-bc!

LeStarch:
The changes look good, but what is the rationale for doing this?  The reason this was implemented as a virtual function with an empty implementation was that most `Os::Queue` implementations don't need to teardown.  Thus I wonder why this was changed?

vsoulgard:
@LeStarch, I assumed the scope of the issue was to make the function pure to match the interface pattern.

vsoulgard:
> I assumed the scope of the issue was to make the function pure to match the interface pattern.

Otherwise, why would interface ever have a function that isn't intended to be implemented?

LeStarch:
I am ok with this, but want to have it marked as a breaking change.

LeStarch (approved):

thomas-bc (approved):
We'll need to update https://github.com/fprime-community/fprime-vxworks/tree/devel/VxWorks/Os

Issue #4246, opened 2025-10-01
Develop Pass-Through Router Component


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Develop a router component that is a Fw.Buffer direct pass through for single custom route use cases.

DJKessler:
This feature description is pretty sparse. Is there a reason that the [Svc::BufferRepeater](https://github.com/DJKessler/fprime/blob/devel/Svc/BufferRepeater/docs/sdd.md) can't be used for the "single custom route use cases" mentioned in the description?

thomas-bc:
@DJKessler the goal is to have a component that can slot in the `Svc.Router` interface https://github.com/nasa/fprime/blob/devel/Svc/Interfaces/Router.fpp

This will be useful for example for the GenericHub architecture:
```cpp
ComDriver -> Deframer -> PassThroughRouter -> GenericHub // (single pipeline, no actual "routing", everything goes to GenericHub)
```

PR #4825, opened 2026-03-10
Add PassThroughRouter component

DJKessler:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4246  |

---
## Change Description

Implemented a Pass-Through Router Component


LeStarch (dismissed):

Svc/PassThroughRouter/PassThroughRouter.cpp
```diff
@@ -0,0 +1,49 @@
...
+void PassThroughRouter::allPacketsReturnIn_handler(FwIndexType portNum, Fw::Buffer& fwBuffer) {
+    this->bufferDeallocate_out(0, fwBuffer);
+}
+
+void PassThroughRouter::dataIn_handler(FwIndexType portNum,
+                                       Fw::Buffer& packetBuffer,
+                                       const ComCfg::FrameContext& context) {
+    if (this->isConnected_allPacketsOut_OutputPort(0)) {
+        // Copy buffer into a new allocated buffer. This lets us return the original buffer with dataReturnOut,
+        auto packetBufferCopy = this->bufferAllocate_out(0, packetBuffer.getSize());
```
thomas-bc: Are the allocations+copies needed? Could this not just be passing ownership of the Fw::Buffer back-and-forth?

DJKessler: I honestly don't know if the allocate and copy is strictly needed here. But that is how the `Svc::FprimeRouter` is implemented ([see HERE](https://github.com/nasa/fprime/blob/1fb1c71005c795cebd2ccc9452c7b0f1698f8d3b/Svc/FprimeRouter/FprimeRouter.cpp#L53)).

bocchino: For a pass-through router, I think we should not be doing any allocation. The allocation seems to cause unnecessary complexity and overhead. I'm also not sure why `Svc::FprimeRouter` is doing allocation and copying here. Since the data is unchanged, it seems simpler to forward the data as is.

thomas-bc: After in-person discussion: allocation was done in FprimeRouter for other reasons. Here, we should be ok to simply pass things through, no allocation, and return a zeroed out context.

bocchino: After discussion, we determined that the F Prime router does an allocation and copy so that it can immediately return the original buffer with the incoming context. That way it doesn't have to remember which context goes with which buffer. However, in the case of the pass-through router, there is no meaningful routing context, so we shouldn't have to do this. We can just return the default context value each time.

thomas-bc:

```bash
```

Svc/PassThroughRouter/PassThroughRouter.cpp
```diff
@@ -0,0 +1,36 @@
...
+    ComCfg::FrameContext context = {}; // default context used to satisfy dataReturnOut port interface
+    this->dataReturnOut_out(0, fwBuffer, context);
+}
+
+void PassThroughRouter::dataIn_handler(FwIndexType portNum,
+                                       Fw::Buffer& packetBuffer,
+                                       const ComCfg::FrameContext& context) {
+    if (this->isConnected_allPacketsOut_OutputPort(0)) {
+        this->allPacketsOut_out(0, packetBuffer);
+    }
```
LeStarch:
```suggestion
    } else {
        this->dataReturnOut_out(0, fwBuffer, context);  // Deallocate when port is not connected
    }
```

LeStarch: If it were me, I would just call the output port, and ASSERT if it isn't connected as I don't see a likely use case where a usere would want to have the port disconnected.

However, if we are to check if the port is connected, then returning the buffer in the case of the port not being connected is important.

I'll be happy either way.

thomas-bc (dismissed):

LeStarch (approved):

thomas-bc (approved):

Issue #4755, opened 2026-02-23
Further Improvements to TlmPacketizer


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

We would like to see the following improvements to TlmPacketizer:

1. Make parameter of MIN/MAX properties
2. Add configurable default parameter
3. Increase copying efficiency
4. Add port to set group/section config
5. Remove double-buffer scheme
6. Switch "omit list" to a pointer allow null to disable the omit lookup.  Thus removing this potential large structure.
7. Add configuration section and description to SDD.

PR #4830, opened 2026-03-10
Making TlmPacketizer configuration tables driven from parameter

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4757, #4813 |

---
## Change Description

This uses an external parameter to update the default configuration of the rate limit / min-max tables in F Prime. This allows for both parameter-driven defaults (start state) as well as commands to control individual rows of the table.

## Rationale

See: #4757.

**Note:** configuration was moved from `default/config` to `Svc/TlmPacketizer/TlmPacketizerConfig` because the configuration system is not properly set up and thus could not accommodate a config that built on `Fw.Types`. 

This should be run on a deployment including telemetry packetizer.

## Future Work

See: #4755

Autocomplete.


LeStarch:
Need UTs.

LeStarch:
@Sarah5567, @Lex-ari here was the implementation of packetizer parameters.

Svc/TlmPacketizer/TlmPacketizer.cpp
```diff
@@ -67,11 +67,7 @@ void TlmPacketizer::setPacketList(const TlmPacketizerPacketList& packetList,
                                   const FwChanIdType startLevel,
                                   const TlmPacketizer_GroupConfig& defaultGroupConfig) {
```
Lex-ari: Will this need to be removed and insert a paramGet_SECTION_CONFIGS(m_groupConfigs) here?

Lex-ari (changes requested):

Svc/TlmPacketizer/TlmPacketizer.cpp
Sarah5567: Instead of removing the argument, we could add an overload that calls the existing function with paramGet_SECTION_CONFIGS(m_groupConfigs). This keeps backward compatibility and still allows callers who want to pass the config manually to do so.

LeStarch: I chose to remove that.  Since that was added after the last release, I figured we could remove it without needed to preserve backwards compatibility.

LeStarch:
@Lex-ari and @Sarah5567 ready for re-review (UTs work now)!

Svc/TlmPacketizer/TlmPacketizer.fpp
```diff
@@ -126,7 +126,9 @@ module Svc {
                                       ) \
       opcode 5
 
-    
+    external param SECTION_ENABLED: SectionEnabled default TELEMETRY_SECTION_ENABLED_DEFAULTS
```
timcanham, line 129: We still have the problem: How do we sequence this with the existing sequencers?

LeStarch: You use static config at this time, which is exactly how the old version worked, so no regression.

timcanham: If we wish to use the full run-time configurability beyond the current static config, we will have to provide sequencer support.

timcanham (approved):

Svc/TlmPacketizer/TlmPacketizer.cpp
```diff
@@ -49,14 +47,9 @@ TlmPacketizer ::~TlmPacketizer() {}
...
                                   const Svc::TlmPacketizerPacket& ignoreList,
-                                  const FwChanIdType startLevel,
-                                  const TlmPacketizer_GroupConfig& defaultGroupConfig) {
+                                  const FwChanIdType startLevel) {
     FW_ASSERT(packetList.list);
-    // Ignore list may be nullptr as long as numEntries is 0. Providing an ignore list with numEntries 0 disables
-    // functionality for two reasons:
-    //     1. There are no ignored channels as configured by FPP.
-    //     2. Ignore functionality is intentionally disabled by project where nullptr was intentionally supplied.
-    FW_ASSERT(ignoreList.list || ignoreList.numEntries == 0);
```
thomas-bc, line 59: this should be re-added?

Svc/TlmPacketizer/TlmPacketizer.fpp
thomas-bc, line 129: add annotaiton

Svc/TlmPacketizer/TlmPacketizer.fpp
```diff
@@ -11,17 +11,11 @@ module Svc {
...
       min: U32                  @< Minimum Sched Ticks when in ON_CHANGE_MIN
       max: U32                  @< Maximum Sched Ticks when in EVERY_MAX
-    } default {
-      enabled = Fw.Enabled.ENABLED
-      forceEnabled = Fw.Enabled.DISABLED
-      rateLogic =  RateLogic.ON_CHANGE_MIN
-      min = 0
-      max = 0
-    }
+    } default DEFAULT_GROUP_CONFIG
```
thomas-bc: remove default from the type def ?

thomas-bc (approved):

Lex-ari (approved):

PR #4831, opened 2026-03-10
Add deinit function for passive components

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Add `deinit` to passive components

## Rationale

Passives have `init` and thus should have `deinit`.

None.

## Future Work

None.

Autocomplete.

### DIFF
--- a/Fw/Comp/PassiveComponentBase.cpp
+++ b/Fw/Comp/PassiveComponentBase.cpp
@@ -32,6 +32,8 @@ void PassiveComponentBase::toString(char* buffer, FwSizeType size) {
 
 PassiveComponentBase::~PassiveComponentBase() {}
 
+void deinit() {}
+
 void PassiveComponentBase::init(FwEnumStoreType instance) {
     ObjBase::init();
     this->m_instance = instance;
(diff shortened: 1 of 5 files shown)


LeStarch:
Needed for nasa/fpp#896

Fw/Comp/PassiveComponentBase.cpp
```diff
@@ -32,6 +32,8 @@ void PassiveComponentBase::toString(char* buffer, FwSizeType size) {
 
 PassiveComponentBase::~PassiveComponentBase() {}
 
+void deinit() {}
```
bocchino: Shouldn't this function be virtual, so implementations can override it? Otherwise I don't see any purpose to having it. It's a function that does nothing and will never do anything.

LeStarch: It should.  This was done to minimize variance to `init` which improperly uses shadowing....which is not great.  I have made `deinit` virtual.

bocchino:
@LeStarch As part of this PR, can we rename this phase to `deinitComponents`?

https://github.com/nasa/fprime/blob/5f1c648a51a6106d5a3e0ba17c25d4a57a6322a9/Fpp/ToCpp.fpp#L18

It will be more consistent with the other phases, e.g., `initComponents`. I will make the corresponding change on the FPP side here: https://github.com/nasa/fpp/pull/938.

LeStarch:
@bocchino done!

thomas-bc (approved):

bocchino (approved):
Looks good!

Issue #4641, opened 2026-01-21
Update hardcoded constants to use sizeof


jwest115:
This capability will be enabled by https://github.com/nasa/fpp/issues/725
See default/config/FpConfig.fpp
Example:
```
@ The id type.
type FwIdType = U32
constant SIZE_OF_FwIdType = 4  @< Size of FwIdType in bytes
```
Would be `constant SIZE_OF_FwIdType = sizeof(FwIdType)`

thomas-bc:
I would be in favor of removing `constant SIZE_OF_FwIdType = sizeof(FwIdType)` altogether and simply using `sizeof(FwIdType)` when needed

PR #4859, opened 2026-03-17
Update FpConstants to use sizeof, add sizeof tests to FppTest

jwest115:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/4641 |

---
## Change Description

Updated FpConstants.fpp to use sizeof rather than using the hardcoded sizes
Added sizeof tests to FppTest

Ran FppTests and they all passed. Also manually confirmed that sizeof calculations match expectations.


FppTestProject/FppTest/sizeof/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "FppTest/sizeof/SSerializableAc.hpp"
+#include "FppTest/sizeof/S2SerializableAc.hpp"
+#include "FppTest/sizeof/StringAliasAliasAc.hpp"
+#include "FppTest/sizeof/StringDefaultAliasAliasAc.hpp"
+#include "FppTest/sizeof/U64AliasAliasAc.hpp"
+#include "FppTest/sizeof/EnumAliasAliasAc.hpp"
+#include <gtest/gtest.h>
+
+TEST(SizeofTest, Primitives) {
+    ASSERT_TRUE(SizeofU8 == sizeof(U8));
```
bocchino: Instead of ASSERT_TRUE here, can we use ASSERT_EQ? It gives a more informative error message.

FppTestProject/FppTest/sizeof/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+    ASSERT_TRUE(SizeofF32 == sizeof(F32));
+    ASSERT_TRUE(SizeofF64 == sizeof(F64));
+    ASSERT_TRUE(SizeofBool == sizeof(bool));
+}
+
+TEST(SizeofTest, AliasType) {
+    ASSERT_TRUE(SizeofStringAlias == 102);
+    ASSERT_TRUE(SizeofStringDefaultAlias == 258);
+    ASSERT_TRUE(SizeofU64Alias == sizeof(U64Alias));
+    ASSERT_TRUE(SizeofEnumAlias == EnumAlias::SERIALIZED_SIZE && SizeofEnumAlias == 2);
```
bocchino: Instead of `ASSERT_TRUE(A == B && C == D)` let's use
```
ASSERT_EQ(A, B)
ASSERT_EQ(C, D)
```

FppTestProject/FppTest/sizeof/main.cpp
jwest115: Done!

FppTestProject/FppTest/sizeof/main.cpp
jwest115: Done!

FppTestProject/FppTest/sizeof/main.cpp
```diff
@@ -0,0 +1,56 @@
...
+    ASSERT_EQ(SizeofArray2, 48);
+}
+
+TEST(SizeofTest, Enum) {
+    ASSERT_EQ(SizeofEnum, E::SERIALIZED_SIZE);
+    ASSERT_EQ(SizeofEnum, 2);
+}
+
+TEST(SizeofTest, Struct) {
+    ASSERT_EQ(SizeofStruct, S::SERIALIZED_SIZE);
```
thomas-bc: Very much of a nit, but is there a reason to keep the test element names so short?

`struct S` -> `struct TestStruct1`
`struct S2` -> `struct TestStruct2`

It would have the benefits:
- a lot more searchable if someone needs to look at this code (can easily go from FPP <-> CPP)
- doesn't need that odd `SSerializable` spellcheck

jwest115: Good suggestion! Will make that change!

LeStarch (approved):

PR #4865, opened 2026-03-17
refactor: replace FW_MIN/FW_MAX macros with std::min/std::max

### LINKED ISSUE #4826: Replace `FW_MIN` and `FW_MAX` macros with `std::min` and `std::max`
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Is there any reason to use `FW_MIN` and `FW_MAX` which expands into ternary operator over `std::min` and `std::max`?

## Rationale

`std::min` and `std::max` are type-safe and more common practice.

dcpagotto:
Hey, this PR replaces the FW_MIN and FW_MAX macros with std::min and std::max from the standard library in all cpp files.

As discussed in #4826, the macros expand to ternary operators wich isnt type safe. Using std::min/std::max is the more common c++ practice and provides better type checking at compile time.

Changes:
  - added #include <algorithm> where needed
  - replaced FW_MIN with std::min
  - replaced FW_MAX with std::max
  - used std::min with initializer list for nested min calls (cleaner imo)

The macros in BasicTypes.h are kept for backwards compatability with any C code that might include the headers.

Only touched .cpp files to be safe, as LeStarch mentioned in the issue that .h files might be used in C context.

Let me know if theres anything that needs adjusting.

Closes #4826

### DIFF
--- a/Utils/TokenBucket.cpp
+++ b/Utils/TokenBucket.cpp
@@ -12,6 +12,7 @@
 // ======================================================================
 
 #include <Utils/TokenBucket.hpp>
+#include <algorithm>
 
 namespace Utils {
 
@@ -72,7 +73,8 @@ bool TokenBucket ::trigger(const Fw::Time time) {
         Fw::Time nextTime = Fw::Time::add(this->m_time, replenishInterval);
         while (this->m_tokens < this->m_maxTokens && nextTime <= time) {
             // replenish by replenish rate, or up to maxTokens
-            this->m_tokens += FW_MIN(this->m_replenishRate, this->m_maxTokens - this->m_tokens);
+            // replenish tokens up to max capacity
+            this->m_tokens += std::min(this->m_replenishRate, this->m_maxTokens - this->m_tokens);
             this->m_time = nextTime;
             nextTime = Fw::Time::add(this->m_time, replenishInterval);
         }
--- a/Utils/TokenBucket.cpp
+++ b/Utils/TokenBucket.cpp
@@ -73,7 +73,6 @@ bool TokenBucket ::trigger(const Fw::Time time) {
         Fw::Time nextTime = Fw::Time::add(this->m_time, replenishInterval);
         while (this->m_tokens < this->m_maxTokens && nextTime <= time) {
             // replenish by replenish rate, or up to maxTokens
-            // replenish tokens up to max capacity
             this->m_tokens += std::min(this->m_replenishRate, this->m_maxTokens - this->m_tokens);
             this->m_time = nextTime;
             nextTime = Fw::Time::add(this->m_time, replenishInterval);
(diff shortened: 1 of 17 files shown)


Utils/TokenBucket.cpp
```diff
@@ -72,7 +73,8 @@ bool TokenBucket ::trigger(const Fw::Time time) {
         Fw::Time nextTime = Fw::Time::add(this->m_time, replenishInterval);
         while (this->m_tokens < this->m_maxTokens && nextTime <= time) {
             // replenish by replenish rate, or up to maxTokens
-            this->m_tokens += FW_MIN(this->m_replenishRate, this->m_maxTokens - this->m_tokens);
+            // replenish tokens up to max capacity
```
LeStarch: Superfluous comment which is less valid than line 75

```suggestion
```

LeStarch (changes requested):

LeStarch:
@dcpagotto looks like this PR breaks the build?  Can you see the errors, or do I need to copy them over?  Here is one:

```
FileSystem.cpp:(.text+0xe23): undefined reference to `Os::FileSystemInterface::FILE_SYSTEM_FILE_CHUNK_SIZE'
```

dcpagotto:
fixed both issues:

1. added static_cast for FILE_SYSTEM_FILE_CHUNK_SIZE to resolve the linker error. the constexpr needs explicit type when used with std::min (odr-use issue)

2. removed the extra comment in TokenBucket

should be good now, let me know if theres anything else

dcpagotto:
found another issue in StubQueueTests.cpp. the std::min calls had mismatched types (FwSizeType is long unsigned int, U32 is unsigned int). added static_cast to fix.

should build now

dcpagotto:
Hey, nice refactor! took a look at the changes and have a few suggestions

The static_cast on FILE_SYSTEM_FILE_CHUNK_SIZE in FileSystem.cpp works fine but its not super obvious why theres a cast when the types already match. added a comment explaining the ODR linker issue so the next person doesnt try to remove it and break the build again lol

also removed the "track high water mark" comment in PriorityQueue.cpp, the variable is literally called m_highMark and its using std::max so the code speaks for itself there. LeStarch already flagged a similar superflous comment earlier so just keeping things consistent

and added a deprecation notice on FW_MIN/FW_MAX in BasicTypes.h since were moving to std::min/std::max in cpp files, makes sense to mark the old macros so nobody keeps using them. kept them around for C compatiblity tho

two things that might be worth looking into later:
1) the ODR workaround with static_cast is pragmatic but idealy the constexpr should get a proper out of class definition or inline constexpr if c++17 is on the table
2) the CodeQL warnings on ConstStringBase.cpp about uncheked maxLength are pre existing, the macro was just hiding them from the analyzer. might be worth tracking separatly

let me know what you think

LeStarch:

```bash
```

thomas-bc (approved):

LeStarch (approved):

PR #4872, opened 2026-03-18
Adding sscanf replacement implementation, and tests

### LINKED ISSUE #4040: Add `sscanf` Implementation
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

String format (sprintf) is provided as an implementation.  `printf` is handled via the logging system.  `sscanf` is not supplied in any overridable way.

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4040  |

---
## Change Description

Adds sscanf implementation.

## Rationale

Allow projects to replace it if necessary.

Autocomplete and UT generation.

### DIFF
--- a/Fw/Types/scan.hpp
+++ b/Fw/Types/scan.hpp
@@ -1,0 +1,66 @@
+// ======================================================================
+// \title  scan.hpp
+// \author mstarch
+// \brief  hpp file for c-string scan function (sscanf equivalent)
+//
+// \copyright
+// Copyright (C) 2026 California Institute of Technology.
+// ALL RIGHTS RESERVED.  United States Government Sponsorship
+// acknowledged.
+// ======================================================================
+#ifndef FW_TYPES_SCAN_HPP_
+#define FW_TYPES_SCAN_HPP_
+#include <Fw/FPrimeBasicTypes.hpp>
+#include <cstdarg>
+namespace Fw {
+
+//! \brief status of string scan calls
+enum class ScanStatus {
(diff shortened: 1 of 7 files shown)


thomas-bc (dismissed):

Fw/Types/scan.hpp
```diff
@@ -0,0 +1,70 @@
...
+//!   OTHER_ERROR: another error occurred in an underlying function call
+//! Otherwise SUCCESS is returned.
+//!
+//! \param count: [output] number of filled fields
+//! \param source: source string to scan
+//! \param maximumSize: size of the buffer represented by source
+//! \param formatString: format string to fill
+//! \param ...: variable arguments inputs
+//! \return: SUCCESS on successful formatting, SIZE_OVERFLOW on overflow, and something else on any error
+ScanStatus stringScan(FwSizeType& count, char* source, const FwSizeType maximumSize, const char* formatString, ...);
```
vsoulgard: I think `source` should be `const char*` and `maximumSize` should be just `FwSizeType` withount `const` since we pass it by value

thomas-bc: agreed - good catch

LeStarch: Good call.  I have made that change!

thomas-bc (approved):

LeStarch:
The breaking change here is that each platform will need to select an implement of scanf.  Platforms are advised to use the standard implementation.

PR #4888, opened 2026-03-19
Replace command sequence table with map

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Part 2, replace the sequence table with a map.

## Rationale

Complexity reduction.

Co-Pilot for change, Man for UTs.

### DIFF
--- a/Svc/CmdDispatcher/CommandDispatcherImpl.hpp
+++ b/Svc/CmdDispatcher/CommandDispatcherImpl.hpp
@@ -150,26 +150,20 @@ class CommandDispatcherImpl final : public CommandDispatcherComponentBase {
     //! array with an O(log n) red-black tree lookup.
     Fw::RedBlackTreeMap<FwOpcodeType, FwIndexType, CMD_DISPATCHER_DISPATCH_TABLE_SIZE> m_entryTable;
 
-    //! \struct SequenceTracker
-    //! \brief table used to store opcode that are being executed
+    //! \struct SequenceTrackerEntry
+    //! \brief data tracked for commands that are currently executing
     //!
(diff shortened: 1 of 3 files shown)


LeStarch:
Note from @bocchino this one would be better served with an array map given it is typically small with many insertions and deletions.

Svc/CmdDispatcher/CommandDispatcherImpl.hpp
```diff
@@ -150,26 +150,20 @@ class CommandDispatcherImpl final : public CommandDispatcherComponentBase {
...
+    //! command that expects a completion callback.
+
+    struct SequenceTrackerEntry {
+        FwOpcodeType opCode;     //!< opcode being tracked
+        U32 context;             //!< context passed by user
+        FwIndexType callerPort;  //!< port command source port
+    };
+
+    //! \brief map from command sequence number to pending command state
+    Fw::RedBlackTreeMap<U32, SequenceTrackerEntry, CMD_DISPATCHER_SEQUENCER_TABLE_SIZE> m_sequenceTracker;
```
thomas-bc: Should this be an ArrayMap? I would expect this structure to get lots insert/remove operations, but not to grow too large in size?

thomas-bc (commented):
LGTM - just one note about which type of Map to use.

Svc/CmdDispatcher/CommandDispatcherImpl.hpp
LeStarch: Yes, see above comment.

thomas-bc (approved):

Issue #4354, opened 2025-10-24
Remove Old Buffer Patterns


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

Refactor components for new pass-out-pass-back pattern.  We did Com's but we need more.

PR #4891, opened 2026-03-23
refactor: migrate components to pass-out-pass-back pattern

Sara-Cohen-dev:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4354  |

---
## Change Description

Refactored the main F´ components (BufferLogger, ComLogger, TlmChan) to use the pass-out-pass-back pattern instead of the old buffer pattern (Fw.Com):

BufferLogger: Removed duplicate buffer ports, consolidated handlers, updated test code.

ComLogger: Migrated to Fw::Buffer instead of Fw::ComBuffer, added output port for buffer return, updated handlers and tests.

TlmChan: Replaced PktSend with bufferSendOut and bufferSendInReturn, updated handlers, connected new ports in tests.

All existing unit tests pass, ensuring correctness of the new pattern implementation.
## Rationale

This change eliminates redundant buffer handling, prevents memory leaks, and ensures that Buffers are returned after use, 

following FPRIME’s real-time safety patterns.

Verify that all unit tests pass (Svc/BufferLogger, Svc/ComLogger, Svc/TlmChan).

Confirm that buffer ports correctly send and return Buffers according to the new pattern.

Check that no duplicate handlers or ports remain.


thomas-bc:
This work is not what was requested by #4354 . It was about the Return-to-Sender pattern (https://fprime.jpl.nasa.gov/devel/docs/user-manual/framework/memory-management/buffer-pool/#deallocation-strategy). We need to identify if there is still some core Svc components that do not respect it (and deallocate in a different component that allocated).
As it stands this PR changes more things than we're ready to deal with. Closing.

Issue #4898, opened 2026-03-25
Memory Leak in PriorityQueue.cpp - Newly Generated Unittests fail.


jtcass01:
| | |
|:---|:---|
|**_F´ Version_**| v4.1.1 |
|**_Affected Component_**|  Os::Generic::PriorityQueue |
---
## Problem Description

When generating new unittests using --impl, the tests fail without changes due to a memory leak in PriorityQueue.cpp.  I have fixed this memory leak internally by adding to the destructor 

```cpp
this->teardownInternal(); 
```

From looking at your commit history, it appears this problem was injected into your code base in early December of last year.

## Context / Environment

Execute `fprime-util version-check` and share the output. From looking at your 

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-6.8.0-90-generic-x86_64-with-glibc2.35
Python version: 3.10.12
CMake version: 3.26.0
Pip version: 22.0.2
Pip packages:
    fprime-tools==4.1.0
    fprime-gds==4.1.0
    fprime-fpp==3.1.0
Project submodules:
    https://github.com/nasa/fprime.git @ v4.1.1
```

## How to Reproduce

1. Create a new component using fprime-util new --component
2. Generate unittest implements
3. Modify CMakeLists.txt to include new unittests
4. Read test log, see that there is a memory leak.

LeStarch:
This has been fixed by adding `component.deinit()` to the autocoded Tester classes.  For pre-release-v4.2.0 code, users can add that line by hand.

PR #4899, opened 2026-03-25
Fix memory leak in priority queue (#4898)

jtcass01:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

teardownInternal() into PriorityQueue::~PriorityQueue() in Os/Generic/PriorityQueue.cpp. The destructor was previously empty despite teardownInternal() already existing with complete and correct deallocation logic.

## Rationale

teardownInternal() correctly deallocates all 4 of these allocations and already null-guards against double-free. However, it was never called from the destructor, meaning any active or queued component that goes out of scope leaks all 4 allocations. This is detectable by LeakSanitizer in unit tests as a consistent 5640 byte leak in 4 allocations originating from QueuedComponentBase::createQueue().

Run fprime-util check on any active or queued component unit test with AddressSanitizer enabled (default in F Prime UT builds).

## Future Work

None. The fix is self-contained. teardownInternal() was already written correctly — this change only wires it to the destructor.

### DIFF
--- a/Os/Generic/PriorityQueue.cpp
+++ b/Os/Generic/PriorityQueue.cpp
@@ -39,7 +39,9 @@ void PriorityQueueHandle ::load_data(FwSizeType index, U8* destination, FwSizeTy
     (void)::memcpy(destination, this->m_data + offset, static_cast<size_t>(size));
 }
 
-PriorityQueue::~PriorityQueue() {}
+PriorityQueue::~PriorityQueue() {
+    this->teardownInternal();
+}
 
 QueueInterface::Status PriorityQueue::create(FwEnumStoreType id,
                                              const Fw::ConstStringBase& name,


LeStarch:
When I originally wrote this code, I intentionally broke-out teardownInternal for use in the destructor.  I don't remember the reason why it was not called in the destructor.  CI may tell us.  I'll look through the code too.

LeStarch:
Here is the reason:

```
Fw::MemAllocatorRegistry::getInstance().getAnAllocator(
            Fw::MemoryAllocation::MemoryAllocatorType::OS_GENERIC_PRIORITY_QUEUE);
```

Since the `MemAllocatorRegistry`  is used and supplied not by the constructor but by the init method, it isn't guaranteed to exist before construction of these objects.   Therefore it may be destructed before the the destruction call of this object leading to a use of a destructed object.

This is the problem with not using RAII is that you need to intentionally call deallocation before destruction because the construction ordering of static objects is not guaranteed.

The correct fix is to call `~deinit()` in all the UTs' destructors thus forcing a proper shutdown.

LeStarch:
I am going to close this PR given that we've shown how do correct this leak properly.

Issue #4903, opened 2026-03-25
DpWriter reprocess container header after proc steps


kubiak-jpl:
| | |
|:---|:---|
|**_F´ Version_**| 4.1.0 |
|**_Affected Component_**| DpWriter |
---
## Feature Description

Reprocess the and serialize the container structure after calls to `procBufferSendOut_out` in DpWriter

## Rationale

The DpWriter `procBufferSendOut` can be used to post-process data product containers. Such as compressing the data section prior to downlink. A processing component can shrink and manipulate the Fw::Buffer object representing the container but DpWriter does not reprocessing the Fw::Buffer object to see if any changes were made to the container size.

PR #4904, opened 2026-03-25
DpWriter shrink container buffer after processing

kubiak-jpl:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/4903 |

---
## Change Description

Updates DpWriter to shrink and re-serialize the container header if a processing step shrinks the container buffer.

## Rationale

Necessary to develop data product compression processing components.

Testing with an external deployment to compress camera images into a PNG

Added unit tests to test the new behavior

## Future Work

None.

N/A


LeStarch (dismissed):

LeStarch:
Looks ok to me, but I could really use @bocchino or @timcanham to review this.

Svc/DpWriter/DpWriter.cpp
```diff
@@ -169,6 +169,11 @@ void DpWriter::performProcessing(const Fw::DpContainer& container) {
             this->procBufferSendOut_out(portNum, buffer);
         }
     }
+
+    // Update container buffer size
+    container.getBuffer().setSize(buffer.getSize());
```
bocchino: This logic looks right. However:
1. Before doing the subtraction, we need to assert that `buffer.getSize() >= Fw::DpContainer::MIN_PACKET_SIZE`.
2. I think `DpContainer` should encapsulate this operation as a member function `setBufferSize`. Inlining the code here feels like tinkering with the internals of the container.

bocchino: Also, maybe the call to `setBufferSize` should occur in the callee processing function. What if the processing is a byte swap or some other operation that doesn't require changing the size?

kubiak-jpl: Would you prefer that the callee is responsible for updating the full container header? The DpWriter would then recalculate the DpContainer structure from the Fw::Buffer in receives back

bocchino: I discussed this with @timcanham. Yes, we think this is a good idea. Can we say the following:
1. If DpWriter sends a container to any processor, then its local copy of the DpContainer object becomes invalid and has to be recomputed. The header hash also has to be recomputed in this case.
2. A data processor can modify the serialized data and/or the serialized header, but it must keep the header consistent with the data. In particular, if the data size changes, then it needs to update the data size field in the header. Let's say this in the SDD for DpWriter.

bocchino: Also, the SDD has to be updated to reflect the new design.

kubiak-jpl: Sounds good. I'll make these changes

kubiak-jpl: Does it make sense to compute and update the data hash inside the processing component as well? I don't think so but if so the checksum could be used as an additional check that the data product returned from processing is valid. If so it could lead to redundant calculations of the data hash, one for each processing step.

In a similar vain, should DpWrite check/assert that the checksum on the header of the returned DpContainer is valid? I'm thinking that DpWriter should  EVR if the returned header is invalid, but still re-compute the header checksum and write the container data to disk.

bocchino: I think it's OK to have the processing component update only the header and data, and have the DpWriter compute the hashes.

kubiak-jpl: I've made the requested changes. Feel free to take a look. I'm going to re-run a test of this against a processing component in another repo and make sure everything still works there

Update: My image compression components still works if I update the header size so I'm okay to merge this again. I'll mark it read

Svc/DpWriter/DpWriter.cpp
```diff
@@ -169,6 +169,24 @@ void DpWriter::performProcessing(const Fw::DpContainer& container) {
             this->procBufferSendOut_out(portNum, buffer);
         }
     }
+
+    // Updated DpContainer object state with the returned value in the
```
bocchino: Can we check whether there was any call to `procBufferSendOut_out` and update the DpContainer only if there was a call? That way if there's no processing, we don't have to recompute the state.

Svc/DpWriter/DpWriter.cpp
```diff
@@ -169,6 +169,24 @@ void DpWriter::performProcessing(const Fw::DpContainer& container) {
...
+    // container buffer
+    Fw::SerializeStatus stat = container.deserializeHeader();
+    FW_ASSERT(stat == Fw::FW_SERIALIZE_OK, stat);
+
+    // Check that the buffer size is compatible with the data size in
+    // the container header
+    FW_ASSERT(container.getDataSize() <= buffer.getSize(),
+              // Note: May perform a 64 to 32 bit conversion
```
bocchino: This comment seems too specific. It seems true only if size is set to 64 bits and FwAssertArg is set to 32 bits. But isn't that a special cases? In general this can be an _n_ bit to _m_ bit conversion.

Svc/DpWriter/DpWriter.cpp
bocchino: The changes look good!

bocchino:
This PR is looking good! I just had two small comments on the revised code.

Svc/DpWriter/DpWriter.cpp
kubiak-jpl: Maybe this a nit, but the cast here feels off to me to begin with. I added it because clang was complaining about the 64->32 bit conversion with the default config.

I need this cast because FwSizeType is defined as U64 bud FwAssertArgType is an I32. In this case there is no general safe way to assert the full value of an FwSizeType. Is there an implicit assumption the FwAssertArgType is the size of the maximum value type?

I can remove the comment though since it doesn't any general value.

bocchino: Let's remove the comment here but refer the question to the F Prime team. We have been consistently putting in these casts to avoid compiler warnings. However, as you say, there's a danger that some information could be lost. A better practice might be to require that FwAssertArgType is large enough to hold any value of any other framework type. So for example, if FwSizeType is U64, then FwAssertArgType should be U64.

bocchino: @LeStarch points out that in some cases, even with a big enough FwAssertArgType, the casts are required for losses type conversions, e.g., signed to unsigned.

bocchino: So for now maybe the position is that we keep the casts, but we recommend setting FwAssertArgType large enough to handle any value of any framework type. Note that we are not currently doing in this in the default F Prime config, since FwSizeType is larger than FwAssertArgType.

kubiak-jpl: Right. I think the largest signed or unsigned type would work best, depending on whether you'd rather see negative values or not. Not sure if the c++ standard marks one of those transitions as undefined or not.

Variadic templates can probably solve this problem well too.

kubiak-jpl:
@bocchino I pushed an update to address your two comments

bocchino (dismissed):
Looks good!

thomas-bc:
CI can be fixed by:

1. Fixing formatting
2. Merging `devel` into this PR branch (`devel` now contains the spelling fixes)

kubiak-jpl:
@thomas-bc Done

bocchino (approved):

Issue #4816, opened 2026-03-06
Add topology ports to core subtopologies


thomas-bc:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

https://github.com/nasa/fprime/pull/4805 introduced the concept of topology ports. A topology can now expose ports, allowing a subtopology to be treated as more of a blackbox (instead of having to connect to ports of components within the subtopology).

We should add ports to the core Svc/Subtopologies ! And using them would look like:

```diff
# In Ref/topoology.fpp
-      ComCcsds.fprimeRouter.commandOut  -> CdhCore.cmdDisp.seqCmdBuff
-      CdhCore.cmdDisp.seqCmdStatus           -> ComCcsds.fprimeRouter.cmdResponseIn
-      cmdSeq.comCmdOut                      -> CdhCore.cmdDisp.seqCmdBuff
-      CdhCore.cmdDisp.seqCmdStatus -> cmdSeq.cmdResponseIn
+      ComCcsds.commandOut   -> CdhCore.seqCmdBuff
+      CdhCore.seqCmdStatus  -> ComCcsds.cmdResponseIn
+      cmdSeq.comCmdOut      -> CdhCore.seqCmdBuff
+      CdhCore.seqCmdStatus  -> cmdSeq.cmdResponseIn
```

PR #4907, opened 2026-03-26
Add topology ports to subtopologies

Sarah5567:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4816 |

---
## Change Description

Add topology ports to all core Svc/Subtopologies, and update the relevant usages to use them instead of directly referencing internal component ports.

## Rationale

Adopts the topology ports feature from #4805 in the core subtopologies, enabling them to be used as black boxes.

GitHub Copilot CLI

### DIFF
--- a/Ref/Top/topology.fpp
+++ b/Ref/Top/topology.fpp
@@ -164,8 +164,8 @@ module Ref {
       ComCcsds.Subtopology.bufferReturnOut[ComCcsds.Ports_ComBufferQueue.FILE] -> FileHandling.Subtopology.fileDownlinkBufferReturn
 
       # Router <-> FileUplink
-      ComCcsds.Subtopology.fileOut                         -> FileHandling.Subtopology.fileUplinkBufferSendIn
-      FileHandling.Subtopology.fileUplinkBufferSendOut     -> ComCcsds.Subtopology.fileBufferReturnIn
+      ComCcsds.Subtopology.fileUplinkOut                    -> FileHandling.Subtopology.fileUplinkBufferSendIn
+      FileHandling.Subtopology.fileUplinkBufferSendOut     -> ComCcsds.Subtopology.fileUplinkReturnIn
     }
 
     connections FileHandling_DataProducts {
--- a/Ref/Top/topology.fpp
+++ b/Ref/Top/topology.fpp
@@ -83,12 +83,12 @@ module Ref {
       rateGroupDriverComp.CycleOut[Ports_RateGroups.rateGroup1] -> rateGroup1Comp.CycleIn
       rateGroup1Comp.RateGroupMemberOut[0] -> SG1.schedIn
       rateGroup1Comp.RateGroupMemberOut[1] -> SG2.schedIn
-      rateGroup1Comp.RateGroupMemberOut[2] -> CdhCore.tlmSend.Run
-      rateGroup1Comp.RateGroupMemberOut[3] -> FileHandling.fileDownlink.Run
+      rateGroup1Comp.RateGroupMemberOut[2] -> CdhCore.tlmSendRun
+      rateGroup1Comp.RateGroupMemberOut[3] -> FileHandling.fileDownlinkRun
       rateGroup1Comp.RateGroupMemberOut[4] -> systemResources.run
-      rateGroup1Comp.RateGroupMemberOut[5] -> ComCcsds.comQueue.run
-      rateGroup1Comp.RateGroupMemberOut[6] -> CdhCore.cmdDisp.run
-      rateGroup1Comp.RateGroupMemberOut[7] -> ComCcsds.aggregator.timeout
+      rateGroup1Comp.RateGroupMemberOut[5] -> ComCcsds.comQueueRun
+      rateGroup1Comp.RateGroupMemberOut[6] -> CdhCore.cmdDispRun
+      rateGroup1Comp.RateGroupMemberOut[7] -> ComCcsds.aggregatorTimeout
 
       # Rate group 2
       rateGroupDriverComp.CycleOut[Ports_RateGroups.rateGroup2] -> rateGroup2Comp.CycleIn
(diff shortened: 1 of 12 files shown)


Ref/Top/topology.fpp
```diff
@@ -131,47 +131,46 @@ module Ref {
...
+      # Events and telemetry to comQueue
+      CdhCore.Subtopology.eventsPktSend  -> ComCcsds.Subtopology.comPacketQueueIn[ComCcsds.Ports_ComPacketQueue.EVENTS]
+      CdhCore.Subtopology.tlmSendPktSend -> ComCcsds.Subtopology.comPacketQueueIn[ComCcsds.Ports_ComPacketQueue.TELEMETRY]
 
       # Router <-> CmdDispatcher
-      ComCcsds.fprimeRouter.commandOut  -> CdhCore.cmdDisp.seqCmdBuff
-      CdhCore.cmdDisp.seqCmdStatus     -> ComCcsds.fprimeRouter.cmdResponseIn
-      cmdSeq.comCmdOut -> CdhCore.cmdDisp.seqCmdBuff
-      CdhCore.cmdDisp.seqCmdStatus -> cmdSeq.cmdResponseIn
+      ComCcsds.Subtopology.commandOut        -> CdhCore.Subtopology.seqCmdBuff
```
Sarah5567, line 155: @Kronos3 Is there any way to access the port directly (like mentioned in the issue: ComCcsds.commandOut -> CdhCore.seqCmdBuff)?
I wasn't able to find documentation on this, currently it requires the full path through module -> topology -> port.

Kronos3: This is correct. Ports are on the topology not the module, that comment was wrong. We'll probably end up renaming that subtopology to `ComCcsds` but there is an additional feature we'll need before that becomes feasible. At the moment we have module `ComCcsds` with a `Subtopology` topology inside.

Svc/Subtopologies/CdhCore/CdhCore.fpp
```diff
@@ -66,6 +66,17 @@ module CdhCore {
...
+        # Topology ports
+        port seqCmdBuff   = cmdDisp.seqCmdBuff
+        port seqCmdStatus = cmdDisp.seqCmdStatus
+
+        port eventsPktSend  = events.PktSend
+        port tlmSendPktSend = tlmSend.PktSend
+
+        port tlmSendRun = tlmSend.Run
+        port cmdDispRun = cmdDisp.run
+        port healthRun  = $health.Run
```
Kronos3, line 93: Could you add annotations to each of these ports to describe what they do?

Ref/Top/topology.fpp
```diff
@@ -14,12 +14,12 @@ module Ref {
...
+    # Subtopology instances
     # ----------------------------------------------------------------------
-    import CdhCore.Subtopology
-    import ComCcsds.Subtopology
-    import FileHandling.Subtopology
-    import DataProducts.Subtopology
+    instance CdhCore.Subtopology
+    instance ComCcsds.Subtopology
+    instance FileHandling.Subtopology
+    instance DataProducts.Subtopology
```
Kronos3, line 22: Lets keep these as `import` rather than `instance`. They are technically equivalent now but it is more clear to users that these are subtopologies when we do `import`.

Svc/Subtopologies/ComFprime/ComFprime.fpp
```diff
@@ -164,6 +164,32 @@ module ComFprime {
             comStub.dataOut -> ComFprime.frameAccumulator.dataIn
             ComFprime.frameAccumulator.dataReturnOut -> comStub.dataReturnIn
         }
+
+        # Topology ports
+        # Command routing
+        port commandOut         = fprimeRouter.commandOut
+        port cmdResponseIn      = fprimeRouter.cmdResponseIn
+        port fileOut            = fprimeRouter.fileOut
```
Kronos3: `fileOut` sounds like downlink but it's actually uplink? Can you make this name more descriptive

Svc/Subtopologies/ComFprime/ComFprime.fpp
```diff
@@ -164,6 +164,32 @@ module ComFprime {
...
+        port fileBufferReturnIn = fprimeRouter.fileBufferReturnIn
```
Kronos3: Also rename this to describe where the data is flowing

Kronos3 (changes requested):
Thanks for doing this! This PR is all technically correct but I added some comments on naming and docs. I think topology ports are a good opportunity to give port names that describe their purpose at a _system_ level rather than a _component_ level.

Ref/Top/topology.fpp
bocchino: In the FPP User's Guide, we suggest using `instance` to indicate that the subtopology is a pure port interface (i.e., there are no connections directly to its internals). Is that going on here?

bocchino:
This PR is very cool!

bocchino:
It's good to see the topology ports in action.

Ref/Top/topology.fpp
Kronos3: It's used in the "blackbox"/pure port interface form here so maybe this is fine. I forgot we suggested `instance` here in the user's guide.

Svc/Subtopologies/FileHandling/FileHandling.fpp
```diff
@@ -47,5 +47,20 @@ module FileHandling {
         instance fileManager
         instance prmDb
 
+        # Pattern graph specifiers
+        param connections instance prmDb
```
bocchino: Why do we need to add this one pattern specifier here?

Sarah5567: You're right, it was added by accident.

Sarah5567:
@Kronos3 @bocchino Thanks for the comments!
I’ve addressed everything, it’s ready for re-review.

LeStarch:
@Kronos3 @bocchino I confirmed the changes were made.  Would one of you do one last review?

Kronos3 (approved):
Looks great!

bocchino (approved):
Looks great! My comment was addressed.

PR #4922, opened 2026-03-31
Serial port topology testing

Kronos3:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fpp/pull/981 |

---
## Change Description

This PR implements three types of topology connection tests:

1. Typed to serial (for all our typed ports that don't have a return value)
2. Serial to typed (for all our typed ports that don't have a return value)
3. Serial to serial (for a single type of port data)

You can read more about how the tests work in the [README](https://github.com/Kronos3/fprime/blob/6cfa049f71b1439bf09bfe139103f8ad4c4faab2/FppTestProject/FppTest/topology/main/README.md) since it's a bit hard to follow the connection trail

## Rationale

We need comprehensive topology-level testing to validate no-regressions when we bring in direct port calling

## Future Work

No

I manually wrote ASYNC tests and had Claude write GUARDED and SYNC based on that

### DIFF
--- a/FppTestProject/FppTest/interfaces/typed_ports.fpp
+++ b/FppTestProject/FppTest/interfaces/typed_ports.fpp
@@ -4,37 +4,37 @@ module FppTest {
         # Typed input ports with no return type
         # ----------------------------------------------------------------------
 
-        sync input port noArgsSync: [2] NoArgs
+        sync input port noArgsSync: [3] NoArgs
 
-        sync input port primitiveArgsSync: [2] PrimitiveArgs
+        sync input port primitiveArgsSync: [3] PrimitiveArgs
 
-        sync input port stringArgsSync: [2] StringArgs
+        sync input port stringArgsSync: [3] StringArgs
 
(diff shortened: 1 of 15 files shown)


bocchino:
Looks good! How is this passing CI, since the direct serial port connections are not implemented in this version? Are the direct port tests disabled for some or all of the test cases?

Kronos3:
@bocchino Because we are not running direct port calls in CI. This is running off of the FPP in the `requirements.txt`. To enable direct port calls you need to generate with `fprime-util generate -f -DFPRIME_ENABLE_DIRECT_PORT_CALLS=ON`.

We'll need to update CI and some CMake stuff to properly hook up direct port calls into the rest of the framework but this should be enough to test the codegen

bocchino:
> This is running off of the FPP in the requirements.txt. To enable direct port calls you need to generate with fprime-util generate -f -DFPRIME_ENABLE_DIRECT_PORT_CALLS=ON.

Thanks, that makes sense!

bocchino (approved):
Looks good!

PR #4938, opened 2026-04-01
Optimize memory management

i-am-malicious:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---


LeStarch:
@i-am-malicious we cannot accept this PR.  First, as pointed out by @copilot, the memory management is wrong and results in a leak. Second, it is not allow. Third, why?

Issue #4716, opened 2026-02-13
Type Errors in ParamDB masked by U32


SterlingPeet:
| | |
|:---|:---|
|**_F´ Version_**| 3.6.X |
|**_Affected Component_**|  Svc::ParamDB |
---
## Problem Description

Using non-default type sizes causes a mismatch and the database read and write formats.  For a deployment using primarily 16 bit types, this is unusable.

I have a correction available for 3.6 branch.

SterlingPeet:
This is one of the offending lines, I'll set up a PR with my full changeset for this problem.

https://github.com/nasa/fprime/blob/7538242e9a4e7b79e56889d818cb59e4deef8d94/Svc/PrmDb/PrmDbImpl.cpp#L299

PR #4954, opened 2026-04-04
PrmDb: use canonical serialized size for record size

hidirektor:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4716 |

---
## Change Description

Updates `Svc/PrmDb` to use the canonical F Prime serialized size format for the file record size field.

- Replaces raw `U32` record-size serialization/deserialization with `serializeSize(...)` / `deserializeSize(...)`
- Aligns record-size read/write checks with `sizeof(FwSizeStoreType)`
- Adds a new `FW_HAS_16_BIT == 1` regression test that:
  - writes a 16-bit parameter value,
  - saves the parameter file,
  - verifies the on-disk layout by deserializing the saved bytes,
  - reloads the file and confirms round-trip behavior

## Rationale

This fixes the parameter database file format so it is not tied to host `U32` assumptions and is more robust across configurations with different type sizes.

The added regression test protects against future format regressions, especially in non-default build configurations such as 16-bit payloads.

- Review the record-size serialization path in `Svc/PrmDb/PrmDbImpl.cpp`
- Review the new 16-bit regression coverage in `Svc/PrmDb/test/ut/PrmDbTester.cpp`
- Confirm the read/write error expectations now use `FwSizeStoreType`-based sizing
- Validate that the round-trip test still passes in `FW_HAS_16_BIT == 1` builds

## Future Work

- Add broader coverage for additional non-default size/type configurations if needed
- Consider adding a dedicated file-format compatibility test for other payload widths

AI was used for review support.

### DIFF
--- a/Svc/PrmDb/PrmDbImpl.cpp
+++ b/Svc/PrmDb/PrmDbImpl.cpp
@@ -200,12 +200,12 @@ void PrmDbImpl::PRM_SAVE_FILE_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
         // add delimiter to CRC
         crc = this->computeCrc(crc, &delim, sizeof(delim));
 
-        // serialize record size = id field + data
-        U32 recordSize = static_cast<U32>(sizeof(FwPrmIdType) + entry.getValue().getSize());
+        // serialize record size = id field + data using the canonical F Prime size format
+        FwSizeType recordSize = static_cast<FwSizeType>(sizeof(FwPrmIdType) + entry.getValue().getSize());
 
         // reset buffer
         buff.resetSer();
-        Fw::SerializeStatus serStat = buff.serializeFrom(recordSize);
+        Fw::SerializeStatus serStat = buff.serializeSize(recordSize);
         // should always work
         FW_ASSERT(Fw::FW_SERIALIZE_OK == serStat, static_cast<FwAssertArgType>(serStat));
 
@@ -218,7 +218,7 @@ void PrmDbImpl::PRM_SAVE_FILE_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
             this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
             return;
         }
-        if (writeSize != sizeof(recordSize)) {
+        if (writeSize != static_cast<FwSizeType>(sizeof(FwSizeStoreType))) {
             this->unLock();
             this->log_WARNING_HI_PrmFileWriteError(PrmWriteError::RECORD_SIZE_SIZE, static_cast<I32>(numRecords),
                                                    static_cast<I32>(writeSize));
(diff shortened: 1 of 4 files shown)


Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -200,12 +200,12 @@ void PrmDbImpl::PRM_SAVE_FILE_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
...
 
-        // serialize record size = id field + data
-        U32 recordSize = static_cast<U32>(sizeof(FwPrmIdType) + entry.getValue().getSize());
+        // serialize record size = id field + data using the canonical F Prime size format
+        FwSizeType recordSize = static_cast<FwSizeType>(sizeof(FwPrmIdType) + entry.getValue().getSize());
 
         // reset buffer
         buff.resetSer();
-        Fw::SerializeStatus serStat = buff.serializeFrom(recordSize);
+        Fw::SerializeStatus serStat = buff.serializeSize(recordSize);
```
LeStarch, line 213: Should add a comment that this is actually serializing as `FwSizeStoreType`

Svc/PrmDb/test/ut/PrmDbTester.cpp
```diff
@@ -371,11 +463,11 @@ void PrmDbTester::runFileReadError() {
                 break;
             case 2:
                 ASSERT_EVENTS_PrmFileReadError_SIZE(1);
-                ASSERT_EVENTS_PrmFileReadError(0, PrmReadError::RECORD_SIZE_SIZE, 0, sizeof(U32) + 1);
+                ASSERT_EVENTS_PrmFileReadError(0, PrmReadError::RECORD_SIZE_SIZE, 0, sizeof(FwSizeStoreType) + 1);
                 break;
             case 3:
                 ASSERT_EVENTS_PrmFileReadError_SIZE(1);
-                ASSERT_EVENTS_PrmFileReadError(0, PrmReadError::PARAMETER_ID_SIZE, 0, sizeof(FwPrmIdType) + 1);
+                ASSERT_EVENTS_PrmFileReadError(0, PrmReadError::RECORD_SIZE_SIZE, 0, sizeof(FwSizeStoreType) + 1);
```
LeStarch, line 469: Why did this change?  I believe this loops through each record. Size was in case 2, so my expectation that case 3 remain PARAMETER_ID

LeStarch (changes requested):

LeStarch:
#

```bash
```

Svc/PrmDb/PrmDbImpl.cpp
hidirektor: Comment added especially for that part :))

Svc/PrmDb/test/ut/PrmDbTester.cpp
hidirektor: That may seem like an odd change at first, but this test fails by read() call index (m_waits), not by field name. With the size-format update, case 3 now hits a record-size read at the next record boundary, so PrmReadError::RECORD_SIZE_SIZE is the correct expectation

thomas-bc:
@hidirektor the code doesn't compile. See CI errors:

```
/__w/fprime/fprime/Svc/PrmDb/PrmDbImpl.cpp:213:29: error: declaration of ‘serStat’ shadows a previous local [-Werror=shadow]
         Fw::SerializeStatus serStat = buff.serializeSize(recordSize);
                             ^~~~~~~
/__w/fprime/fprime/Svc/PrmDb/PrmDbImpl.cpp:166:25: note: shadowed declaration is here
     Fw::SerializeStatus serStat = buff.serializeFrom(crc);
                         ^~~~~~~
```

Are you able to fix them?

LeStarch:
This has not been updated in 2 weeks. Closing.

Issue #4931, opened 2026-03-31
PrmDb CRC Storage is Endianness-Specific


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

The CRC written by PrmDb casts the U32 value to a U8* for writing.  This means the stored value is determined by the endinanness of the processor, not a generic. 

Fix is to use a serial buffer in F Prime to serialize the U32 then write it out.

PR #4956, opened 2026-04-06
Issue 4931 prm db crc endian fix

odacindy-fprime:
| | |
|:---|:---|
|**_Related Issue(s)_**|  https://github.com/nasa/fprime/issues/4931|
|**_Documentation Included (y/n)_**|  y|

---
## Change Description

Fix the PrmDb CRC to use a serial buffer in F Prime to serialize the CRC for writing/reading to/from the param file.

## Rationale

Without this serialization,  the stored CRC value is determined by the endian-ness of the processor, not a generic.

I just fixed and re-ran the existing unit test.

I used AI to confirm my approach.  I also made changes to it to better fit my programming style and fixed an error in it since it was trying to use a non-existent (or perhaps old) method name.

### DIFF
--- a/Svc/PrmDb/PrmDbImpl.cpp
+++ b/Svc/PrmDb/PrmDbImpl.cpp
@@ -307,11 +307,8 @@ void PrmDbImpl::PRM_SAVE_FILE_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
     buff.resetSer();
     serStat = buff.serializeFrom(crc);
 
-    FW_ASSERT(Fw::FW_SERIALIZE_OK == serStat, static_cast<FwAssertArgType>(serStat));  
+    FW_ASSERT(Fw::FW_SERIALIZE_OK == serStat, static_cast<FwAssertArgType>(serStat));
     writeSize = static_cast<FwSizeType>(buff.getSize());
-
-    printf("DEBUG: file: %s, line: %d CINDY PRM_SAVE_FILE_cmdHandler: crc=0x%08x, writeSize=%llu\n ", 
-                __FILE__, __LINE__,  crc, static_cast<unsigned long long>(writeSize));
 
     stat = paramFile.write(buff.getBuffAddr(), writeSize, Os::File::WaitType::WAIT);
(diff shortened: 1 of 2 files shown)


Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -161,8 +161,11 @@ void PrmDbImpl::PRM_SAVE_FILE_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
 
     // write placeholder for the CRC
     U32 crc = 0xFFFFFFFF;
-    FwSizeType writeSize = static_cast<FwSizeType>(sizeof(crc));
-    stat = paramFile.write(reinterpret_cast<const U8*>(&crc), writeSize, Os::File::WaitType::WAIT);
+    buff.resetSer();
+    Fw::SerializeStatus serStat = buff.serializeFrom(crc);
```
LeStarch, line 166: I'd add a static_assert that the storage of buff >= sizeof(U32).

I'd do it right on line 37, static_assert sizeof(m_buff) >= sizeof(U32)

Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -406,10 +414,12 @@ PrmDbImpl::PrmLoadStatus PrmDbImpl::readParamFileImpl(const Fw::StringBase& file
     }
     //===========================================================================
     // read CRC from beginning of file
+    WorkingBuffer buff;
     U32 fileCrc;
     FwSizeType readSize = static_cast<FwSizeType>(sizeof(fileCrc));
```
LeStarch, line 420: For this code you **absolutely** need the static_assert or you risk overflowing WorkingBuffer.

LeStarch (changes requested):

Svc/PrmDb/PrmDbImpl.cpp
odacindy-fprime: addressed in git commit f7545461d68b44fe5fd0550f5c5cb0611469013a

LeStarch (approved):

Issue #4942, opened 2026-04-02
Add argument buffer to SeqDisp.RUN and FpySequencer.RUN


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.2 |
|**_Affected Component_**| SeqDispatcher, FpySequencer |
---
## Feature Description

In order to support argument passing to sequences, the RUN commands for both the SeqDispatcher and FpySequencer will need to be updated to have an "opaque" argument buffer, which should probably just be an array of U8s of a configurable compile time constant size.

However, this would be annoying for GDS users who don't want to pass arguments to sequences, because it would fill their commanding page with a bunch of fields for the individual bytes. So this change might need to come with either GDS updates to make this less of a problem (e.g. filling in defaults? or having collapsible sections under args? or having U8 array types be input in a manner other than one byte at a time (this could be handy)), or it could come with perhaps an FPP change which allows optional command arguments with defaults.

The less ideal solution would be to add a second command, like RUN_ARGS, to both SeqDisp and FpySequencer. This is not great because it's a bit confusing, and at the moment I don't see any "real", "user driven" reason why they should be separate commands. I could be convinced otherwise though if anyone has thoughts.

## Rationale

This is needed for supporting passing arguments to sequences (to FpySequencer now, to WasmSequencer in the future).

Lex-ari:
I'm looking into working on this change, but I'd like feedback from the community aswell!

One of our ideas was that the RUN_ARGS command is essentially a way for other tools to interact with the fprime-gds to send in their own serialized arguments into the CmdDisp. It's a command that shouldn't be accessed directly by the user, but rather for external tools (such as a supposed fpy-cli tool) to pass in data through the CmdDisp to an FpySequencer.

Another was to add a sequencing page to GDS, but we'll need to resolve creating a sequence dictionary with arguments, as well as handle adding new sequences. However, this still requires changes to allow arguments to pass in via a buffer or array.

Issue #4662, opened 2026-01-27
FpySequencer sequence args


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

A quick description of the requested feature.

## Rationale

A rationale for why the feature should be implemented in F´.

zimri-leisher:
Option A: -- likely choosing this
```
param x: U8
# some lines of code here
# probably disallow non parameter statements in between
param y: U8
```

Option B:
```
# everything has to be wrapped in a main function
def main(x: U8, y: U8):
    pass
```

Option C:
```
gvar: U8 = 1
def func_1(arg1):
   gvar = 123
def func_1(arg1):
   pass
```

Two ways of calling it:
# 1: from fpy
func(123, 123)
# 2: from GDS
FpySeq.RUN(file, varargs)



Fpy sequence calling:
Option A:
```
import some_sequence_name

some_sequence_name(1,2,3)
```

Option B:
```
call("sequence_file_name", 1, 2, 3)
```


Going to need sequence dictionary, containing generated info about sequence arguments/other things needed

Notes:
* special keyword called "redline" (abort criteria)--put as much meta info in the dictionary
* Lots of analysis is needed--control flow of sequences will need to be displayed, etc

Issue #4181, opened 2025-09-22
Fpy sequence args


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.1.0 |
|**_Affected Component_**| fpyc, FpySequencer |
---
## Feature Description

Add the ability to pass in arbitrary bytes to a sequence at runtime. Retain some type information about these bytes

zimri-leisher:
How should sequence arguments be denoted in the sequence?
How should they be stored in the file?
How should the GDS read in these arguments?
Where will the argument type checking happen?

zimri-leisher:
argument types and number must be in the sequence itself

1 idea:
* entrypoint sequence with a "main" function?
* link all the sequences together

2 idea:
* "sequence dictionary" json containing the arg types for each compiled sequence
* build-artifacts/Linux.../RefTopologySeqDict.json
* so some command that scans a folder of sequences to get a sequence dictionary?

3 idea: (nonstarter)
* no type checking on the ground for sequence args
* you would only know if the args were wrong type if the sequence fails at runtime on the spacecraft
* could have some more stringent checks in the sequence



okay going back to syntax:
```
idea 1: (seems likeliest)

def main(arg: U32):
    ...


def something_else():
   ...


idea 2:
abc: U32
test: bool

<stmt>
<stmt>

```

also consider importing sequences ("inlining"?)

zimri-leisher:
* take asm, annotate it with args, build a tool that will strip the annotations and gen a seq dictionary, and write a tool to dispatch a run cmd that processes those args and dispatches correct cmd, and update the fprime-gds to be able to supply args to sequences

PR #4962, opened 2026-04-07
Add args to CmdSeqIn

Lex-ari:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4942 #4662 #4181 [fpy#27](https://github.com/fprime-community/fpy/pull/27) [fpy#30](https://github.com/fprime-community/fpy/pull/30) |

---
## Change Description

Add a U8 array to CmdSeqIn to allow pass through arguments from SeqDisp to CmdSequencer / FpySequencer / etc

## Rationale

This is a feature to pass through arguments from the SequenceDispatcher to Sequencers (current and future) that support it. 

Ensure SequenceDispatch functionality is working, both on fprime and fpy side.
These changes should not break / altar core functionality.

## Future Work

This is a chain of pull requests to support arguments in sequences. Future work includes:
- Creating a tool to pass arguments to the GDS to start a sequence with arguments
- Fpy#30 Future Work: Replacing arguments onto the Fpy stack
- Discussion on call syntax for sequence with arguments within sequences
- Documentation

Claude Code v2.1.92 sonnet used to autocomplete and find instances where CmdSeqIn changes were required.
FPP coding was done by hand and claude was used to autocomplete C++ calls.
All lines of code were human reviewed.

@LeStarch @zimri-leisher


default/config/AcConstants.fpp
```diff
@@ -59,3 +59,5 @@ constant FwAssertTextSize = 256
 @ the constants FW_ASSERT_TEXT_SIZE and FW_LOG_STRING_MAX_SIZE, set
 @ in FpConfig.h.
 constant AssertFatalAdapterEventFileSize = FileNameStringSize
+
```
LeStarch, line 62: Annotate this with an explanation on how to set it.

Svc/CmdSequencer/CmdSequencerImpl.cpp
```diff
@@ -322,7 +329,9 @@ void CmdSequencerComponentImpl ::CS_START_cmdHandler(FwOpcodeType opcode, U32 cm
     this->performCmd_Step();
     this->log_ACTIVITY_HI_CS_CmdStarted(this->m_sequence->getLogFileName());
     if (this->isConnected_seqStartOut_OutputPort(0)) {
-        this->seqStartOut_out(0, this->m_sequence->getStringFileName());
+        // Create empty SeqArgs as placeholder
+        Svc::SeqArgs emptyArgs;
```
LeStarch: Here you did not use the parameterized constructor.  Probably want consistency.

Svc/SeqDispatcher/SeqDispatcher.cpp
LeStarch: I would have expected this component to be updated.  It has all the information to plumb those arguments through, no?

LeStarch (changes requested):

Svc/Seq/Seq.fpp
```diff
@@ -1,8 +1,13 @@
 module Svc {
+  struct SeqArgs {
```
zimri-leisher, line 2: I would name the buffer "buffer" or something. Otherwise it's going to look awkward going like `args.get_args()` all the time, kinda confusing. Also please initialize it to zero like `default { $size = 0, buffer = 0 }` (yes, this zero initialization works)

zimri-leisher:
So is modifying the RUN going to be a separate PR? Why not just do it in this PR, I think they are related. New RUN_ARGS cmd for SeqDispatcher and update existing RUN cmd for FpySeq

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -358,7 +358,8 @@ void FpySequencer::cmdResponseIn_handler(FwIndexType portNum,             //!< T
 }
 
 //! Handler for input port seqRunIn
-void FpySequencer::seqRunIn_handler(FwIndexType portNum, const Fw::StringBase& filename) {
+void FpySequencer::seqRunIn_handler(FwIndexType portNum, const Fw::StringBase& filename, const Svc::SeqArgs& args) {
+    (void)args;  // Suppress unused parameter warning
```
zimri-leisher: Shouldn't we implement this now? The arg buf should be pushed to the stack.
Important: you're going to have to add the SeqArgs struct to the SequenceExecutionArgs struct, and pass it in on line 370 here. You should probably add a new SM action called "pushArgsToStack" or something, which happens when cmd_RUN is received, and all it does is take the SeqArgs out of the SequenceExecutionArgs struct, and pushes them to the stack. Make sure it happens _after_ resetRuntime

Svc/SeqDispatcher/SeqDispatcher.cpp
zimri-leisher: agreed, the buf should be passed in here

zimri-leisher (changes requested):

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -53,6 +64,8 @@ void FpySequencer::RUN_cmdHandler(FwOpcodeType opCode,               //!< The op
         this->m_savedCmdSeq = cmdSeq;
     }
 
+    // Store args for pushArgsToStack action
+    this->m_pendingSeqArgs = buffer;
```
zimri-leisher: There is a cleaner way of passing around the pending sequence arguments without adding new state.

I described it in an earlier comment: 
> Important: you're going to have to add the SeqArgs struct to the SequenceExecutionArgs struct, and pass it in on line 370 here. You should probably add a new SM action called "pushArgsToStack" or something, which happens when cmd_RUN is received,

Svc/FpySequencer/FpySequencerStateMachine.fppi
zimri-leisher: pushArgsToStack should go here, and it should take an argument, of type SequenceExecutionArgs, and the arg buf should be added to that struct

just copy what's already done with setSequenceFilePath or setSequenceBlockState.

default/config/AcConstants.fpp
```diff
@@ -61,4 +61,7 @@ constant FwAssertTextSize = 256
 constant AssertFatalAdapterEventFileSize = FileNameStringSize
 
 @ The maximum size in bytes for passing sequence arguments through CmdSeqIn ports
-constant SequenceArgumentsMaxSize = 512
\ No newline at end of file
+@ Note: This must fit within FW_CMD_ARG_BUFFER_MAX_SIZE along with other command arguments
+@ Total serialized size: fileName (~44 bytes) + Fw::Wait (4 bytes) + SeqArgs (8 + buffer_size)
+@ With FW_CMD_ARG_BUFFER_MAX_SIZE ~= 500, we limit this to 400 bytes
```
zimri-leisher: FW_CMD_ARG_BUFFER_MAX_SIZE is now an FPP constant. Just use it directly to calculate the value here. That way it will change if the cmd arg buf size changes.

Svc/FpySequencer/FpySequencerCommands.fppi
```diff
@@ -6,25 +6,32 @@ async command RUN(
...
     fileName: string size FileNameStringSize @< The name of the sequence file
 ) \
-    opcode 1 priority 7 assert
+    opcode 2 priority 7 assert
 
 @ Must be called after VALIDATE. Runs the sequence that was validated.
 # prio: lower than sm sig and CANCEL
 async command RUN_VALIDATED(
     $block: BlockState @< Return command status when complete or not
 ) \
```
zimri-leisher, line 28: This command needs an argument buf too. And that would mean that we'd need a RUN_VALIDATED_ARGS command. However, I think it would be okay to skip this for now, because this command is not used frequently. 

Please make a github issue to track adding the args buf to this command when we fix the GDS.

Svc/FpySequencer/FpySequencerStateMachine.cpp
```diff
@@ -255,6 +255,34 @@ void FpySequencer::Svc_FpySequencer_SequencerStateMachine_action_incrementSequen
...
+    const Svc::SeqArgs& args = this->m_pendingSeqArgs;
+
+    // Early return if no arguments provided
+    if (args.get_size() == 0) {
+        return;
+    }
+
+    Fpy::StackSizeType availableSpace = Fpy::MAX_STACK_SIZE - this->m_runtime.stack.size;
+
+    if (args.get_size() > availableSpace) {
```
zimri-leisher: The way this PR is currently implemented, sendSignal_result_failure does not do what you expect it does here. It actually would do nothing, and the FSW would try to run the sequence. Could be quite bad. This is because, when this action is called, you're in the RUN state. Homework for you: make sure you know which signals do what in the RUN state. Many signals won't do anything.

Once you make the modification I mentioned regarding the SequenceExecutionArgs, this function will just set the m_pendingSeqArgs blindly. See action_setSequenceFilePath.

Then, the size check will go in the FpySequencer::validate method. You can just return Fw::Success::FAILURE.

Please also add a unit test for this case.

zimri-leisher (changes requested):

LeStarch:
Port buffers approved.

Svc/FpySequencer/FpySequencer.cpp
Lex-ari: Discussion: Will still need this->m_pendingSeqArgs to pass in arguments after the runtime is reset. Separated new action: setSequenceArguments and pushArgsToStack

Svc/FpySequencer/FpySequencerStateMachine.fppi
Lex-ari: Discussed: Cannot put pushArgsToStack due to resetRuntime called after cmd_VALIDATE and cmd_RUN. Instead put setSequenceArguments on cmd_RUN / VALIDATE and pushArgsToStack stays as an on-entry action of state RUNNING

Svc/FpySequencer/FpySequencerCommands.fppi
Lex-ari: Discussed: Not going to support a RUN_VALIDATED_ARGS. This is to prevent the user from inputting arguments twice from VALIDATE_ARGS and RUN_VALIDATED_ARGS.

Users will have to either supply arguments on a RUN_ARGS cmd (where validation and run occurs) or supply them on a VALIDATE command (and then a subsequent RUN command).

default/config/AcConstants.fpp
```diff
@@ -59,3 +59,8 @@ constant FwAssertTextSize = 256
 @ the constants FW_ASSERT_TEXT_SIZE and FW_LOG_STRING_MAX_SIZE, set
 @ in FpConfig.h.
 constant AssertFatalAdapterEventFileSize = FileNameStringSize
+
+@ The maximum size in bytes for passing sequence arguments through CmdSeqIn ports
+@ Note: This must fit within FW_CMD_ARG_BUFFER_MAX_SIZE along with cmd arguments using Svc::SeqArgs
+@ Total serialized size: fileName (~200 bytes) + Fw::Wait (4 bytes) + SeqArgs (8 + buffer_size)
+constant SequenceArgumentsMaxSize = FW_CMD_ARG_BUFFER_MAX_SIZE - FileNameStringSize - 8 - 8
```
zimri-leisher: pls explain the -8 -8 in the comment. why does it say 4 bytes for Fw::Wait when ur doing -8 here?

zimri-leisher (approved):
Spectacular work! Let's merge it.

default/config/AcConstants.fpp
Lex-ari: i can't do math

Svc/FpySequencer/FpySequencerStateMachine.cpp
Lex-ari: Changed state machine action logic
Added UT to cover incorrect argument sizing in validation

default/config/AcConstants.fpp
Lex-ari: Made FW_CMD_ARG_BUFFER_MAX_SIZE use sizeof() functions for its dependencies

zimri-leisher (approved):

default/config/AcConstants.fpp
```diff
@@ -49,7 +49,7 @@ constant DpManagerNumPorts = 5
 constant DpWriterNumProcPorts = 5
 
 @ The size of a file name string
-constant FileNameStringSize = 200
+constant FileNameStringSize = 240
```
LeStarch, line 52: Why was this change made, seems irrelevant.

LeStarch (dismissed):
Looks good to me.

default/config/AcConstants.fpp
LeStarch: This was done because it replaced hardcoded 240.

LeStarch (approved):

Issue #4521, opened 2025-12-05
DpCatalog.cpp multiple static analysis findings


djbyrne17:
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**| Svc/DpCatalog |
---

Follow-on from #4296 which gathered static analysis findings.

## Problem Description

File `DpCatalog.cpp` has 1 of the most severe SonarQube findings in fprime.  Fix that, and opportunistically some less severe findings.

### At least the SonarQube "Blocker" finding
Line 399: Access of the field 'm_directories' at index 2, while it holds only 2 'class Fw::FileNameString' elements

### Other SonarQube findings as make sense, and as time allows
```
<High> Line 385: cpp:S3776 Refactor this function to reduce its Cognitive Complexity from 36 to the 25 allowed.
<High> Line  47: cpp:S3490 Use "=default" instead of the default implementation of this special member functions.
<High> Line 560: cpp:S134 Refactor this code to not nest more than 3 if|for|do|while|switch statements.
<High> Line 635: cpp:S1186 Add a nested comment explaining why this method is empty, or complete the implementation.
<High> Line 673: cpp:S3776 Refactor this function to reduce its Cognitive Complexity from 28 to the 25 allowed.
<High> Line 695: cpp:S134 Refactor this code to not nest more than 3 if|for|do|while|switch statements.
<High> Line 699: cpp:S134 Refactor this code to not nest more than 3 if|for|do|while|switch statements.
<High> Line 711: cpp:S134 Refactor this code to not nest more than 3 if|for|do|while|switch statements.
<High> Line 718: cpp:S134 Refactor this code to not nest more than 3 if|for|do|while|switch statements.

<Med> Line 100: cpp:S3630 Replace "reinterpret_cast" with a safer operation.
<Med> Line 163: cpp:S5945 Use "std::string" instead of a C-style char array.
<Med> Line 179: cpp:S5827 Replace the redundant type with "auto".
<Med> Line 246: cpp:S5817 This function should be declared "const".
<Med> Line  25: cpp:S3230 Do not use the constructor's initializer list for data member "m_initialized". Use the in-class initializer instead.
<Med> Line  26: cpp:S3230 Do not use the constructor's initializer list for data member "m_dpTree". Use the in-class initializer instead.
<Med> Line 267: cpp:S5945 Use "std::string" instead of a C-style char array.
[...]

PR #4897, opened 2026-03-25
Data Product Catalog updates

timcanham:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5036 #3198 #3197 also |
|**_Has Unit Tests (y/n)_**| y |
|**_Documentation Included (y/n)_**| y |
|**_Generative AI was used in this contribution (y/n)_**| y |

---
## Change Description
This PR adds the following features:
1) Data product deletion - delete data products when they are confirmed downlinked
2) Data product reprioritization - re-prioritize a data product after initial priority assignment
3) Data product retransmission - marks the data product for retransmission if there were issues downlinking
4) Data product operations file - provides for batch processing of data product operations via an uplinked file

## Rationale

These provide additional features for data product maintenance

## Testing/Review Recommendations

Unit tests were added to test the features

## Future Work

Future possible work to segment the files, but that may be supplanted by CFDP.

## AI Usage (see [policy](https://github.com/nasa/fprime/blob/devel/AI_POLICY.md))

This was an excercise for me (@timcanham) to learn Claude, so the bulk of the updates were via Claude. I checked the resultant code and unit tests manually.

PR #4964, opened 2026-04-07
Fix static analysis blocker: bounds-guard m_directories in DpCatalog

sympact06:
| | |                                                                                                                                                                                                                  
  |:---|:---|                                                                                                                                                                                                              
  |**_Related Issue(s)_**| #4521 |                                                                                                                                                                                 
  |**_Has Unit Tests (y/n)_**| n) |                                                                                                                                                                  
                                                                                                                                                                                                                           
  ---                                                                                                                                                                                                                    
  ## Change Description                                                                                                                                                                                                    
                                                                                                                                                                                                                         
  Clamp the loop bound in `fillBinaryTree()` to `DP_MAX_DIRECTORIES` so that                                                                                                                                               
  `m_directories[dir]` is provably in-bounds for the static analyzer.                                                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                           
  ## Rationale                                                                                                                                                                                                             
                                                                                                                                                                                                                           
  Resolves the SonarQube **Blocker** finding:                                                                                                                                                                              
  > Line 399: Access of the field 'm_directories' at index 2, while it holds only 2 'class Fw::FileNameString' elements
                                                                                                                                                                                                                           
  No behavioral change the clamp is purely defensive
                                                                                                                                                                                                                           
  ## Future Work
                                                                                                                                                                                                                           
  Remaining non-blocker static analysis findings from #4521                                                                                                                                 
                                                                                                                                                                                                                           
  Claude Code (Anthropic, Opus 4.6) was used to:                                                                                                                                                                           
  - Analyze the static analysis finding and identify the root cause

First PR to fprime! Hope to contribute more and learn along the way


timcanham:
Let's merge https://github.com/nasa/fprime/pull/4897 first.

LeStarch:
Blocked per @timcanham

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -381,7 +381,10 @@ Fw::CmdResponse DpCatalog::fillBinaryTree() {
     FwSizeType totalFiles = 0;
 
     // get file listings from file system
-    for (FwSizeType dir = 0; dir < this->m_numDirectories; dir++) {
+    const FwSizeType numDirs = (this->m_numDirectories <= DP_MAX_DIRECTORIES)
```
thomas-bc: It may be more readable to use `std::min(this->m_numDirectories, DP_MAX_DIRECTORIES)` ?

But also this is provable / asserted in `configure()` :

https://github.com/nasa/fprime/blob/5843907fd9ab3760264f9ab2b63740c5e7a7b76b/Svc/DpCatalog/DpCatalog.cpp#L57

My vote would be to dispose of this static analysis warning as a false positive

thomas-bc: This doesn't feel like the right fix to me. If `m_numDirectories` is more than DP_MAX_DIRECTORY, something majorly wrong is going on. Probably coding error, or bitflips or whatever. Silently clamping the range doesn't sound right. I vote we either ASSERT, or dispose of static analysis as false positive.

LeStarch (approved):

PR #4967, opened 2026-04-07
Move Includes for ComCcsds into SubTopology

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

The ComCcsds subtopology relies on includes that were not part of the subtopology definition.


Ref/Top/RefTopologyDefs.hpp
thomas-bc, line 21: could/should each `#include PingEntries.hpp` also be moved within the SubtopologyDefs.hpp ?

thomas-bc (approved):

Ref/Top/RefTopologyDefs.hpp
LeStarch: Those are already part of a subtopology.

Issue #3134, opened 2025-01-14
ROSES - Rules Based Testing


LeStarch:
### Description

We should create a how-to guide on rules based testing. This should include the new rule integration with `fprime-util`.

### Deliverables (completed before closing issue):
- [x] Select deliverable type
  - [x] Reference
  - [ ] User Guide
  - [x] How-To
  - [ ] Tutorial
  - [x] Other (tool, fill-in, etc): `fprime-util new --rule`
- [ ] Create documentation
- [ ] Create example in examples repository
- [ ] Pass expert review
- [ ] Pass user review

PR #4977, opened 2026-04-08
Add common infrastructure for Rule-Based Testing

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Preface to https://github.com/nasa/fprime/issues/3134

The goal here is to provide a “standard" pattern for doing Rule-Based Testing. This does not mean restricting to only that pattern, but I will next document how to work with this specific pattern, and provide tooling to work with it.

- adds a TestUtils MACRO for defining Rule-Based Test rules. This enables the framework's `ASSERT_EVENTS_*` etc. macros to be usable within the rule body by... simply making the body of the rule delegate to a member function of the tester. this is all declared / handled by the macro, and users need simply to implement each body.
- refactors Svc.Ccsds.ApidManager to use said common infrastructure

## Future Work

- Write How-To Guide
- Identify potential tooling support. My plan for now:
    - `fprime-util new --rule-based-testing` --> bootstraps the RBT structure (TestState folder and class definition, Rules/*.cpp file)
    - I don't see much value in having a `fprime-util new --rule` for each rule, as we need to insert the MACROs inside the ComponentTester.hpp which would be very brittle and not much added value. Instruct users to copy-paste is fine imo.
    - Potentially `fprime-util new --rule-group` that creates the `RuleGroup.cpp` file and a TODO rule ?

Help think through the architecture and refactor the ApidManager test code to the new architecture


LeStarch (dismissed):

TestUtils/RuleBasedTesting.hpp
```diff
@@ -0,0 +1,74 @@
...
+//! \param TESTER_TYPE  The ComponentTester class (state passed to STest::Rule)
+//! \param GROUP_NAME   Rule group: appears in method names and the rule label
+//! \param RULE_NAME    Rule variant: appears in method names and the rule label
+// -----------------------------------------------------------------------
+#define FW_RBT_DEFINE_RULE(TESTER_TYPE, GROUP_NAME, RULE_NAME)                                \
+    bool GROUP_NAME##__##RULE_NAME##__precondition() const;                                   \
+    void GROUP_NAME##__##RULE_NAME##__action();                                               \
+    struct GROUP_NAME##__##RULE_NAME : public STest::Rule<TESTER_TYPE> {                      \
+        GROUP_NAME##__##RULE_NAME() : STest::Rule<TESTER_TYPE>(#GROUP_NAME "." #RULE_NAME) {} \
+        bool precondition(const TESTER_TYPE& tester) override {                               \
```
bocchino: I find it confusing to call the type `TESTER_TYPE` and the parameter `tester` here. According to STest conventions, the names should be something like `TestState` and `testState`. Also, if we generalize the naming, then this macro can be generally useful for testing with rules, not necessarily in the context of an F Prime Tester class.

bocchino (changes requested):
Looks great! I like the simplification into one macro. I just had one suggested revision to the naming in the macro.

TestUtils/RuleBasedTesting.hpp
bocchino: The general usage requirement for this macro, as I understand it, is that the rule preconditions and actions are implemented in the same class as the STest test state. This is a more general requirement than saying "the STest test state must be an F Prime tester class." The F Prime tester class meets this requirement.

TestUtils/RuleBasedTesting.hpp
```diff
@@ -47,28 +47,26 @@
 //! \def FW_RBT_DEFINE_RULE
 //!
 //! Defines everything needed for one rule inside a ComponentTester class:
```
LeStarch: This comment still discussed ComponentTester.

TestUtils/RuleBasedTesting.hpp
```diff
@@ -47,28 +47,26 @@
...
-//! 1) GroupName__RuleName__precondition() const — declaration
-//! 2) GroupName__RuleName__action()             — declaration
-//! 3) struct GroupName__RuleName : STest::Rule<TESTER_TYPE> — definition
+//! 1) GroupName__RuleName__precondition() const             — declaration
+//! 2) GroupName__RuleName__action()                         — declaration
+//! 3) struct GroupName__RuleName : STest::Rule<TEST_STATE>  — definition
 //!
 //! Place inside the tester class body; implement both method bodies in a
```
LeStarch: Same here.

LeStarch (changes requested):

LeStarch (approved):

bocchino (approved):
Looks good!

Issue #4896, opened 2026-03-25
Serialized size of CmdArgBuffer is not correct


bocchino:
Where it says `I32`, that should be `FwSizeStoreType`:

https://github.com/nasa/fprime/blob/048fe04062b91be7af138df7e09cdd15812dde96/Fw/Cmd/CmdArgBuffer.hpp#L25

This is probably benign, because size store type is not usually bigger than I32. However, we should double-check all these sizes. If FwSizeStoreType is set to U64, this could cause a problem.

bocchino:
For classes derived from StringBase, there is a constexpr function in the base class that computes the static serialized size for a given max string length. That way you don't have to write out the "max length plus size" computation every time. It would be good to do the same thing for LinearBufferBase.

Moferanoluwa:
This looks like it has already been resolved on `devel` (currently `35d4c0e881`).

`Fw/Cmd/CmdArgBuffer.hpp` no longer hand-codes the serialized size. It is now an alias:

```cpp
using CmdArgBuffer = LinearBufferTemplate<FW_CMD_ARG_BUFFER_MAX_SIZE>;
```

and the size comes from the constexpr helper on `LinearBufferBase` in `Fw/Types/Serializable.hpp`, which uses `FwSizeStoreType` rather than `I32`:

```cpp
static constexpr Serializable::SizeType STATIC_SERIALIZED_SIZE(
    Serializable::SizeType maxSize
) {
    return static_cast<Serializable::SizeType>(sizeof(FwSizeStoreType)) + maxSize;
}
```

That is the "constexpr function in the base class" approach @bocchino described above.

On the "we should double-check all these sizes" point: I checked every `.hpp` under `Fw/` at `upstream/devel` for a `SERIALIZED_SIZE` declaration paired with `I32` and found no remaining instances, so the `FwSizeStoreType = U64` concern should no longer apply here.

Flagging in case this can be closed.

---
<sub>Disclosure per [AI_POLICY.md](https://github.com/nasa/fprime/blob/devel/AI_POLICY.md): AI-assisted (Claude Code). Used to search the tree and check the current state; I reviewed the results against `devel` before posting.</sub>

thomas-bc:
Fixed per the above

PR #4978, opened 2026-04-09
Fix CmdArgBuffer SERIALIZED_SIZE computation

Sarah5567:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4896 |

---
## Change Description

Fix `CmdArgBuffer::SERIALIZED_SIZE`, which incorrectly used sizeof(I32) instead of sizeof(FwSizeStoreType).
To prevent this kind of mistake, this PR adds a STATIC_SERIALIZED_SIZE constexpr helper to LinearBufferBase and updates all buffer-derived classes to use it.

## Rationale

If FwSizeStoreType is configured larger than I32 (e.g. U64), `CmdArgBuffer::SERIALIZED_SIZE` would be too small, potentially causing serialization issues.

N/A

## Future Work

N/A

GitHub Copilot CLI (Claude)

### DIFF
--- a/Drv/Ports/DataTypes/DataBuffer.hpp
+++ b/Drv/Ports/DataTypes/DataBuffer.hpp
@@ -11,7 +11,7 @@ class DataBuffer : public Fw::SerializeBufferBase {
     enum {
         DATA_BUFFER_SIZE = 256,
         SERIALIZED_TYPE_ID = 1010,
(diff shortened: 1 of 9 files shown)


LeStarch:

```bash
```

LeStarch (approved):

thomas-bc (approved):

PR #4987, opened 2026-04-10
Add support for direct port calls

### LINKED ISSUE #3986: Provide a standard solution for "compacting" F Prime ports on constrained systems
| | |
|:---|:---|
|**_F´ Version_**| latest |
|**_Affected Component_**| fpp autocode |
---
## Feature Description

On a microcontroller project that I'm working on, the overhead of F Prime components and ports is too high. I have constrained code memory, constrained data memory, and stringent performance requirements.

Port definitions and port calls in F Prime tend to be large:

1. It take many instructions to execute each port call. These instructions both take up valuable code space and slow down microcontroller execution every time the port call is invoked. They cannot be optimized out even under Link Time Optimization because the port function pointer acts as an "optimization barrier." I've previously shared detailed examples of this with the F Prime team, but I'm happy to dig them up again if it would be useful.
2. Port definitions take up R/W data memory rather than R/O data memory, and they tend to be large: Each output port on our platform is 24 bytes, and each input port on our platform is 40 bytes. If we have 5 instances of a component with 49 input ports and 4 output ports (yes, we have a real-world use case for that), that's already over 10 KiB of mutable port data that will never need to change after initialization. For reference, the minimum R/W data memory per component on our platform is 16 bytes, so port definitions dominate the overhead of using F Prime.

We can contort our design to avoid these costs, but then the resulting codebase will be more difficult for us to develop and maintain, and less reusable by future missions.

I would like F Prime to find a way to "compact" port calls and port definitions, both in terms of number of instructions and mutability of data. The current approach of dynamically initializing port connections at runtime, even though the topologies are known at compile time, appears to be the underlying cause of these problems.

## Rationale

Better support for using F Prime for microcontroller software.

bocchino:
Closes #3986.


default/config/FpConfig.h
```diff
@@ -21,6 +21,9 @@ extern "C" {
 // Configuration switches
 // ----------------------------------------------------------------------
 
+// Enable strict assertions
+#define FW_STRICT_ASSERTIONS 1
+
```
Kronos3, line 28: Needs an ifndef guard

default/config/FpConfig.h
Kronos3: Can you fill out the comment here? Also we should follow the `\ \n...` pattern that is used by the other configuration flags in this file

default/config/FpConfig.h
bocchino: Done!

default/config/FpConfig.h
bocchino: Done! Should we set this to 1 by default?

LeStarch: Yes, it should default to the nominal case.

LeStarch (dismissed):

default/config/FpConfig.h
Kronos3: Actually I'd leave this off for now since it'll break UTs (I think? Unless there is already a mechanism that'll prevent UT breakage?) We may want to hook it up via CMake in a future PR.

bocchino: We could write `ifdef BUILD_UT 0 else 1`.

Kronos3: That should work but can you put the `ifdef BUILD_UT` inside the `ifndef FW_DIRECT_PORT_CALLS` so that externally setting direct port calls will override this? The integration tests for FppTestProject still build UTs while still using direct ports.

bocchino: Done!

Kronos3 (approved):
Looks good!

bocchino:
@LeStarch @thomas-bc I replaced `printf` with `Fw::Logger::log` as discussed here: https://github.com/nasa/fprime-tools/pull/305#issuecomment-4224806322.

thomas-bc:
All is ready on my end. Need review and merge of these:

- https://github.com/nasa/fprime-tools/pull/305
- https://github.com/fprime-community/fprime-workshop-led-blinker/pull/139

thomas-bc (approved):

PR #4999, opened 2026-04-11
CI: stop using runner.debug in reusable project builder

dcpagotto:
this fixes a workflow validation issue in the reusable project builder.

the current workflow sets VERBOSE_FLAG from an expression using runner.debug inside the env block. GitHub is rejecting that in the reusable workflow, which causes the ext-build workflows that call it to fail before they even start the actual build.

the fix here is to stop computing that flag in workflow expressions and instead set it inside each bash step from RUNNER_DEBUG. That keeps the same intent, but avoids the invalid workflow expression.

I checked the updated yaml locally to make sure it still parses cleanly, and this should unblock the external build workflows that were failing at workflow load time.

if there is a better repo-wide pattern for handling optional verbose flags in reusable workflows, I am happy to adjust, but this should be the safest direct fix for the current failure.

Thanks! Dhiego Pagotto...

### DIFF
--- a/.github/workflows/reusable-project-builder.yml
+++ b/.github/workflows/reusable-project-builder.yml
@@ -48,7 +48,6 @@ jobs:
     env:
       FPRIME_LOCATION: ${{ inputs.fprime_location }}
       TARGET_PLATFORM: ${{ inputs.target_platform }}
-      VERBOSE_FLAG: ${{ runner.debug == '1' && '--verbose' || '' }}
     steps:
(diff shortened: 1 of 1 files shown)


thomas-bc (commented):
LGTM. Confirmed it works by enabling verbose on a run. 

CI failure is expected, I'll go and fix the examples repo that's failing.

thomas-bc (approved):

PR #5033, opened 2026-04-17
Add FppTestProject Direct Port Call Integration Tests to CI

Kronos3:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Adds a build job with direct port call execution for our FPP integration tests

## Rationale

The standard FppTestProject uses BUILD_UT which disables direct port calls in the standard `FpConfig.h`. This option overrides that default for the topology tests since we don't need indirect port calls for those and allows us to test direct port calls in a unit-testy way.

Claude did it. I ran it manually as well, seems fine.

### DIFF
--- a/.github/workflows/fpp-tests.yml
+++ b/.github/workflows/fpp-tests.yml
@@ -21,7 +21,7 @@ jobs:
     name: Run FppTest
     runs-on: ubuntu-22.04
     steps:
-      - name: "Checkout F´ Repository"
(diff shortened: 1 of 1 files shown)


LeStarch (approved):
LGTM.  Minor question:  why JSON generation?

thomas-bc (approved):

Kronos3:
@LeStarch None other than the indirect ports has it.

PR #5037, opened 2026-04-18
PR 5036 - Event vs. ASSERT if DP file is corrupted

bitWarrior:
| | |
|:---|:---|
|**_5036_**|  |
|**_Has Unit Tests (Y)_**|  |

---
## Change Description

If DP file is corrupted, a new event named FileCorruptedDataError is generated vs. ASSERT

## Rationale

Add robust FSW response to off-nominal condition rather than crashing.

A new unit test called test_MalformedFile was added.  All other baseline unit test pass without modification. 

To test this implementation, I also created a DpState.dat file that contained all 0xFF:
00000000: ffff ffff ffff ffff ffff ffff ffff ffff  ................
00000010: ffff ffff ffff ffff ffff ffff ffff ffff  ................
00000020: ffff ffff ffff ffff                      ........

When DataProducts.dpCat.BUILD_CATALOG was called, the new FileCorruptedDataError event was generated and the FSW continued to operate nominally. 

## Future Work

N/A

AI was used to generate Unit Test for new functionality.

### DIFF
--- a/Svc/DpCatalog/DpCatalog.cpp
+++ b/Svc/DpCatalog/DpCatalog.cpp
@@ -204,18 +204,16 @@ Fw::CmdResponse DpCatalog::loadStateFile() {
         // the source buffer was specifically sized to hold the data
 
         // Deserialize the file directory index. If an error occurs processing the file,
-        // generate event and return EXECUTION_ERROR. 
+        // generate event and return EXECUTION_ERROR.
         Fw::SerializeStatus status = entryBuffer.deserializeTo(this->m_stateFileData[entry].entry.dir);
         if (status != Fw::FW_SERIALIZE_OK) {
-            this->log_WARNING_HI_FileCorruptedDataError
-                (this->m_stateFile, static_cast<I32>(status));
+            this->log_WARNING_HI_FileCorruptedDataError(this->m_stateFile, static_cast<I32>(status));
             stateFile.close();
             return Fw::CmdResponse::EXECUTION_ERROR;
         }
         status = entryBuffer.deserializeTo(this->m_stateFileData[entry].entry.record);
         if (status != Fw::FW_SERIALIZE_OK) {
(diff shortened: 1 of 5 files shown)


LeStarch (changes requested):
Looks good.

```bash
```

Svc/DpCatalog/test/ut/DpCatalogTester.cpp
```diff
@@ -756,4 +756,42 @@ void DpCatalogTester::test_ProcessFileInvalidDir() {
...
+    Fw::FileNameString dirs[DP_MAX_DIRECTORIES];
+    this->component.configure(dirs, 0, stateFile, 0, mockAllocator);
+
+    // 4. Dispatch the BUILD_CATALOG command
+    this->sendCmd_BUILD_CATALOG(0, 0);
+    this->component.doDispatch();
+
+    // 5. Command should generate event instead of ASSERT
+    ASSERT_CMD_RESPONSE_SIZE(1);
+    // ASSERT_CMD_RESPONSE(0, DpCatalogComponentBase::OPCODE_BUILD_CATALOG, 0, Fw::CmdResponse::EXECUTION_ERROR);
```
thomas-bc, line 785: Why is this commented out?

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -203,11 +203,20 @@ Fw::CmdResponse DpCatalog::loadStateFile() {
         // deserialization after this point should always work, since
         // the source buffer was specifically sized to hold the data
 
-        // Deserialize the file directory index
+        // Deserialize the file directory index. If an error occurs processing the file,
```
thomas-bc, line 206: comment above should probably be taken out as it contradicts the new logic: `deserialization after this point should always work`

thomas-bc (commented):
Few nits but otherwise LGTM

PR #5039, opened 2026-04-18
PR 5038 - Force Framers to setSize on Fw::Buffer before outputting it

bitWarrior:
| | |
|:---|:---|
|**_5038_**|  |
|**_Has Unit Tests (Y)_**|  |

---
## Change Description

Added a single `setSize()` call in each framer immediately before `dataOut_out`, trimming the buffer to the exact computed frame size. `frameSize` is already in scope in both handlers at that point and requires no additional computation.

**`Svc/FprimeFramer/FprimeFramer.cpp`**
```cpp
// Trim to actual frame size in case allocator returned a larger buffer
frameBuffer.setSize(static_cast<Fw::Buffer::SizeType>(frameSize));
```

**`Svc/Ccsds/SpacePacketFramer/SpacePacketFramer.cpp`**
```cpp
// Trim to actual frame size in case allocator returned a larger buffer
frameBuffer.setSize(static_cast<Fw::Buffer::SizeType>(frameSize));
```

## Rationale

Trim any extra data.

Created a test case `testOversizedAllocatorBufferIsTrimmed` to both `FprimeFramerTester` and `SpacePacketFramerTester`. Each test case overrides `from_bufferAllocate_handler` to return a buffer twice the requested size (simulating a `BufferManager` bin fallback), then asserts that the buffer received by `dataOut` reports `getSize()` equal to the exact expected frame size. Extra capacity is filled with sentinel byte `0xAB` so any stale-tail leak produces obviously wrong bytes rather than silent zeroes.

## Future Work

N/A

AI was used to write the unit test `testOversizedAllocatorBufferIsTrimmed for FprimeFramerTester and SpacePacketFramerTester.

### DIFF
--- a/Svc/Ccsds/SpacePacketFramer/SpacePacketFramer.cpp
+++ b/Svc/Ccsds/SpacePacketFramer/SpacePacketFramer.cpp
@@ -71,6 +71,8 @@ void SpacePacketFramer ::dataIn_handler(FwIndexType portNum, Fw::Buffer& data, c
     FW_ASSERT(status == Fw::FW_SERIALIZE_OK, status);
     status = frameSerializer.serializeFrom(data.getData(), data.getSize(), Fw::Serialization::OMIT_LENGTH);
     FW_ASSERT(status == Fw::FW_SERIALIZE_OK, status);
+    // Trim to actual frame size in case allocator returned a larger buffer
+    frameBuffer.setSize(static_cast<Fw::Buffer::SizeType>(frameSize));
 
     this->dataOut_out(0, frameBuffer, context);
     this->dataReturnOut_out(0, data, context);  // return ownership of the original data buffer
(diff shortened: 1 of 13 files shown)


LeStarch:

```bash
```

Also, there are a bunch of dp catalog changes. Please move those to another PR

LeStarch (changes requested):
See previous comment

thomas-bc (changes requested):
I'm seeing a few issues here

1. The branch seems to contain unrelated changes (Data Products)
2. From what I can tell, the test isn't actually testing anything. It's setting `m_useOversizedAlloc` but that value is not used anywhere. I see you've used AI to generate them - we should refine here.


Also, I agree that this is a good sanity check here, but `Svc.BufferManager` [does call setSize()](https://github.com/nasa/fprime/blob/devel/Svc/BufferManager/BufferManagerComponentImpl.cpp#L116-L117) before returning the buffer. We should tighten (requirement?) whether this is part of the BufferManager contract or not. I think it should be.

bitWarrior:
Fixed formatting issues.

thomas-bc:
This PR still includes changes that seem to come from #5037 

See list of files changed: https://github.com/nasa/fprime/pull/5039/changes

image

PR #5041, opened 2026-04-18
PR 5040-Potential DB file corruption on longer then shorter re-save DB file

bitWarrior:
| | |
|:---|:---|
|**_5040_**|  |
|**_Has Unit Tests (Y)_**|  |

---
## Change Description

Changed the open() call PrmDbImpl::PRM_SAVE_FILE_cmdHandler to use OPEN_CREATE with the OVERWRITE flag, which maps to O_WRONLY | O_CREAT | O_TRUNC on POSIX. This will overwrite all stale data from any previous save and ensures only the new prm data is saved into the file.

## Rationale

Fixes issue where old stale data from a larger file size could corrupt the current save. 

Run new unit test. Create a larger then smaller prm file. 

## Future Work

N/A

AI was used to create the new unit tests.

### DIFF
--- a/Svc/Ccsds/SpacePacketFramer/SpacePacketFramer.cpp
+++ b/Svc/Ccsds/SpacePacketFramer/SpacePacketFramer.cpp
@@ -71,6 +71,8 @@ void SpacePacketFramer ::dataIn_handler(FwIndexType portNum, Fw::Buffer& data, c
     FW_ASSERT(status == Fw::FW_SERIALIZE_OK, status);
     status = frameSerializer.serializeFrom(data.getData(), data.getSize(), Fw::Serialization::OMIT_LENGTH);
     FW_ASSERT(status == Fw::FW_SERIALIZE_OK, status);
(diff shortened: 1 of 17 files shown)


thomas-bc:
There seems to be other commits/changes not related to the described PR that are pushed to your branch. Can we fix this?

bitWarrior:
The changes are related to the core issue (belt and suspenders approach)
1) The modification of Svc/PrmDb/PrmDbImpl.cpp clears any state data opening the file
2) Without the change with 1 above, stale data could generate an ASSERT (verified during testing). Changing the ASSERT to an EVR adds robustness to F Prime with a grateful exit if we enter this condition even with the open fix to remove any stale data.

thomas-bc:
https://github.com/nasa/fprime/pull/5041/changes

This PR modifies 17 files. Changes seem to include commits from https://github.com/nasa/fprime/pull/5039 and https://github.com/nasa/fprime/pull/5037

image

Issue #5009, opened 2026-04-13
DpCatalog accepts truncated rogue .fdp files and sends header-derived paths instead of the scanned file


thomas-bc:
Credit to @20210607

# Summary

Svc::DpCatalog in nasa/fprime trusts unvalidated data-product headers from any .fdp file it finds in a managed directory, even when the file is truncated and even when the on-disk filename does not match the canonical Dp_<id>_<time>.fdp naming scheme. During catalog build, DpCatalog scans directory entries by extension, reads only Fw::DpContainer::Header::SIZE bytes from the discovered file, deserializes attacker-controlled id/priority/time/state metadata, and stores only that metadata plus the discovered file size. Later, when it starts transmission, it does not reuse the scanned file path; it reconstructs a fresh canonical path from the untrusted header fields and passes that path to fileOut.

As a result, a rogue or partially written .fdp file can poison the catalog in two security-relevant ways:

A header-only truncated file smaller than Fw::DpContainer::MIN_PACKET_SIZE is still accepted as a legitimate product and transmitted as if it were complete.
A renamed rogue file can cause DpCatalog to send a different canonical path than the file it actually scanned. If that canonical file exists, DpCatalog will transmit the different file. If it does not exist, DpCatalog still drives the send path for a file that was never scanned or validated.

Sarah5567:
@thomas-bc, I’ve opened a PR that addresses this issue. Could you review it when you have time?

PR #5042, opened 2026-04-19
DpCatalog - validate fdp files

Sarah5567:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5009 |

---
## Change Description

This change improves `Svc::DpCatalog` validation of data product files discovered during catalog builds.
The catalog now rejects malformed or inconsistent `.fdp` files before adding them to the catalog. Specifically, it now:
- Accepts only files whose final suffix is the data product extension.
- Rejects files smaller than `Fw::DpContainer::MIN_PACKET_SIZE`.
- Reads enough bytes to include the DP header and required hashes.
- Verifies the header hash before trusting header metadata.
- Verifies that the header `dataSize` matches the on-disk file size.
- Verifies that the discovered file path matches the canonical DP filename derived from the header metadata.
The unit test helper was also updated to generate complete DP packets, including payload data and hashes, so tests exercise the same validation path as real data product files.

## Rationale

Previously, `DpCatalog` trusted metadata from any file with a matching `.fdp` extension in a managed directory after reading only the DP header. This allowed truncated or renamed files to be cataloged.
A truncated file could be accepted even though it was smaller than the minimum valid DP packet size. A renamed file could also cause the catalog to scan one path but later transmit a different canonical path reconstructed from untrusted header metadata.
This change ensures that catalog entries are created only from structurally valid DP files whose on-disk path is consistent with the metadata used later for transmission.

Codex

### DIFF
--- a/Svc/DpCatalog/DpCatalog.cpp
+++ b/Svc/DpCatalog/DpCatalog.cpp
@@ -13,6 +13,7 @@
 #include "Fw/Types/StringUtils.hpp"
 #include "Os/File.hpp"
 #include "Os/FileSystem.hpp"
+#include "Utils/Hash/Hash.hpp"
 
 namespace Svc {
 static_assert(DP_MAX_DIRECTORIES > 0, "Configuration DP_MAX_DIRECTORIES must be positive");
@@ -408,11 +409,14 @@ Fw::CmdResponse DpCatalog::fillBinaryTree() {
         for (FwSizeType file = 0; file < filesRead; file++) {
             // only consider files with the DP extension
 
-            FwSignedSizeType loc =
-                Fw::StringUtils::substring_find(this->m_fileList[file].toChar(), this->m_fileList[file].length(),
-                                                DP_EXT, Fw::StringUtils::string_length(DP_EXT, sizeof(DP_EXT)));
+            const FwSizeType fileNameLength = this->m_fileList[file].length();
+            const FwSizeType dpExtLength = Fw::StringUtils::string_length(DP_EXT, sizeof(DP_EXT));
+            const FwSignedSizeType loc =
+                Fw::StringUtils::substring_find_last(this->m_fileList[file].toChar(), fileNameLength, DP_EXT,
+                                                     dpExtLength);
 
-            if (-1 == loc) {
+            // Only accept files whose final suffix is the data product extension
+            if ((-1 == loc) || (static_cast<FwSizeType>(loc) + dpExtLength != fileNameLength)) {
                 continue;
             }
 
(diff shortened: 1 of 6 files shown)


Svc/DpCatalog/test/ut/DpCatalogTester.cpp
```diff
@@ -756,4 +751,104 @@ void DpCatalogTester::test_ProcessFileInvalidDir() {
...
+    this->component.doDispatch();
+    ASSERT_CMD_RESPONSE_SIZE(1);
+    ASSERT_CMD_RESPONSE(0, DpCatalog::OPCODE_BUILD_CATALOG, 10, Fw::CmdResponse::OK);
+    ASSERT_EVENTS_DpFileAdded_SIZE(0);
+
+    this->sendCmd_START_XMIT_CATALOG(0, 11, Fw::Wait::NO_WAIT, false);
+    this->component.doDispatch();
+    ASSERT_CMD_RESPONSE_SIZE(2);
+    ASSERT_from_fileOut_SIZE(0);
+
```
LeStarch, line 798: I would have expected a test here that confirms the event indicating a too-small file emitted here.

Svc/DpCatalog/test/ut/DpCatalogTester.cpp
```diff
@@ -756,4 +751,104 @@ void DpCatalogTester::test_ProcessFileInvalidDir() {
...
```
LeStarch, line 829: Same here.

Svc/DpCatalog/test/ut/DpCatalogTester.cpp
```diff
@@ -756,4 +751,104 @@ void DpCatalogTester::test_ProcessFileInvalidDir() {
...
```
LeStarch, line 856: And here.

LeStarch (changes requested):
We are missing the check for events indicating each of the error cases.  Otherwise, I see improved validation!

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -526,13 +534,38 @@ int DpCatalog::processFile(Fw::String fullFile, FwSizeType dir) {
...
+    if (dataSize != expectedDataSize) {
+        this->log_WARNING_HI_FileReadError(fullFile, Os::File::BAD_SIZE);
+        return 0;
+    }
+
+    Fw::FileNameString canonicalFileName;
+    canonicalFileName.format(DP_FILENAME_FORMAT, this->m_directories[dir].toChar(), container.getId(),
+                             container.getTimeTag().getSeconds(), container.getTimeTag().getUSeconds());
+    if (canonicalFileName != fullFile) {
+        this->log_WARNING_HI_FileHdrDesError(fullFile, Fw::FW_SERIALIZE_FORMAT_ERROR);
```
LeStarch: On more thing, this should probably be a completely new event indicating the file name is invalid.

LeStarch (changes requested):

Sarah5567:
@LeStarch thanks for your feedback. I've addressed everything.

LeStarch:
Looks good to me, although I did revert the .gitignore change because it seem erroneous.

LeStarch (dismissed):

thomas-bc (dismissed):

Svc/DpCatalog/DpCatalog.cpp
```diff
@@ -526,13 +534,38 @@
...
 
+    const FwSizeType dataSize = container.getDataSize();
+    const FwSizeType expectedDataSize = fileSize - Fw::DpContainer::MIN_PACKET_SIZE;
+    if (dataSize != expectedDataSize) {
+        this->log_WARNING_HI_FileReadError(fullFile, Os::File::BAD_SIZE);
+        return 0;
+    }
+
+    Fw::FileNameString canonicalFileName;
+    canonicalFileName.format(DP_FILENAME_FORMAT, this->m_directories[dir].toChar(), container.getId(),
```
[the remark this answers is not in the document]
LeStarch: https://github.com/nasa/fprime/issues/5090

thomas-bc (approved):

PR #5046, opened 2026-04-20
Fix FileWorker compatibility with 32-bit baremetal platforms

celskeggs:
| | |
|:---|:---|

---
## Change Description

Fix FileWorker compatibility with 32-bit baremetal platforms. In particular:

- Types need to match between FileWorker.cpp and FileWorker.hpp. (Use FwSizeType everywhere, it is not equivalent to U64.)
- FileTester depends on POSIX-only filesystem implementations and therefore must be restricted to POSIX-compatible platforms.

## Rationale

When running `fprime-util build` in the top level directory on any platform, we expect the build to complete successfully.

n/a

## Future Work

n/a

Not used


thomas-bc (commented):
Thanks! We may want to exclude the [FileWorkerTestError UT](https://github.com/celskeggs/fprime/blob/f9880cb68ebab25f85889ed3f2b0ed3a4683dd8d/Svc/FileWorker/CMakeLists.txt#L41) from non-POSIX builds as well, since it depends on FileTester

celskeggs:
Looks like this collides with https://github.com/nasa/fprime/pull/5021, so I'll need to pull this back and make some fixes

thomas-bc:
@celskeggs we are trying to expedite a v4.2.2 hotfix for this — I figured you wouldn't mind if I took over your branch. Thanks!

Svc/FileWorker/CMakeLists.txt
```diff
@@ -39,18 +36,31 @@ register_fprime_ut(
...
+          Svc_FileWorker_test_FileTester
+        DEPENDS
+          STest
+        UT_AUTO_HELPERS
+        WORKING_DIRECTORY
+          "${CMAKE_CURRENT_LIST_DIR}"
+      )
+    else()
+      # Only print message if we're in BUILD_TESTING *and* FileTester isn't available
+      message(INFO "Svc/FileWorker: FileTester test dependency is only supported on Posix, skipping FileWorkerTestError")
```
LeStarch: INFO -> STATUS

LeStarch (changes requested):

LeStarch (approved):

PR #5058, opened 2026-04-24
Compile-time Override of Default Assert Hooks

### LINKED ISSUE #5057: Compile-time override of defaultReportAssert/defaultPrintAssert
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

I'd like to be able to override the default assertion hook at compile time rather than only allowing runtime override:

https://github.com/nasa/fprime/blob/ebd230659a3c9bd1e07ff42efc6b60561737a731/Fw/Types/Assert.cpp#L123-L132

These functions can be factored out into an F Prime implementation similar to `Fw_StringFormat`.

## Rationale

Assert hooks are registered at runtime rather than compile causing a large amount of code bloat for tiny systems. When manually overriding this, my text size drops by around 10k due to not using `fputs`. For deeply embedded systems it is useful to provide custom compile time hooks for these lower level abstractions.

Kronos3:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/5057 |

---
## Change Description

Implements https://github.com/nasa/fprime/issues/5057 (assuming it passes CCB, close this otherwise...)

## Rationale

See issue. TLDR, for small platform

## Future Work

Need to update platforms to point to this default assert implementation.

Claude code read the issue and drafted solution. I made some naming changes but in general looks pretty good.

Closes #5057


thomas-bc:
Remember to merge the following PRs upon merging of this current PRs:

- https://github.com/fprime-community/fprime-vxworks/pull/46
- https://github.com/fprime-community/fprime-zephyr/pull/46

Fw/Types/fputs_assert_hook.cpp
```diff
@@ -0,0 +1,77 @@
+// ======================================================================
+// \title  fputs_assert_hook.cpp
```
thomas-bc, line 2: nit: I would call this file `assert_hook_fputs.cpp` so it can be listed closer alongside `assert_hook.hpp`

Fw/Types/Assert.cpp
```diff
@@ -125,7 +62,7 @@ FW_ASSERT_NORETURN void defaultSwAssert(FILE_NAME_ARG file,
         defaultReportAssert(file, lineNo, numArgs, arg1, arg2, arg3, arg4, arg5, arg6, assertMsg,
                             static_cast<FwSizeType>(sizeof(assertMsg)));
         defaultPrintAssert(assertMsg);
-        assert(0);
+        defaultDoAssert();
     } else {
         registeredHook->reportAssert(file, lineNo, numArgs, arg1, arg2, arg3, arg4, arg5, arg6);
```
thomas-bc, line 67: blocking on design decision / discussion:

we should figure out what to do with the old runtime hook pattern. We probably don't want to keep 2 different patterns to override this behavior (this new link-time pattern, plus the old runtime registerHook pattern). The thing that needs to be considered is how to update `AssertFatalAdapter` which currently uses the registerHook to register itself at runtime.

Fw/Types/assert_hook.hpp
```diff
@@ -0,0 +1,61 @@
...
+#define FILE_NAME_ARG const CHAR*
+#endif
+
+namespace Fw {
+
+//! \brief handles an assertion
+//!
+//! Handle an assertion failure or FATAL. This function is called after the defaultPrintAssert has
+//! been called. The default implementation calls `assert(0)`.
+[[noreturn]] void defaultDoAssert();
```
thomas-bc, line 27: If we're in the business of refactoring things, we should pick better names for these functions. "reportAssert", "doAssert" etc. is confusing at best, and most likely just not correctly describing what it actually does. I think I recall Rob said something about this.

thomas-bc (commented):
I like that this leverages the new F´ link-time override mechanism. My recommendation is to consolidate on this and discard the old `registerHook` pattern altogether, but this will need some refactor of AssertFatalAdapter. TBD on how - will discuss offline.

thomas-bc:
Need discussion with Tim to confirm the pattern for splitting between AssertFatalAdapter/EventManager/FatalHandler

Need to find out if there is a way to reduce in a single pattern - it's not obvious. If not, it could be acceptable to keep both patterns in.

Fw/Types/fputs_assert_hook.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+#if FW_ASSERT_LEVEL == FW_FILEID_ASSERT
+#define fileIdFs "Assert: 0x%08" PRIx32 ":%" PRI_FwSizeType ""
+#else
+#define fileIdFs "Assert: \"%s:%" PRI_FwSizeType "\""
+#endif
+
+void Fw::defaultDoAssert() {
+    assert(0);
+    abort();
```
lntutor, line 20: The unconditional abort bypasses the existing FW_ASSERTIONS_ALWAYS_ABORT configuration. Its default value is 0, and FpConfig.h explicitly documents that disabled mode as allowing execution to continue after an assertion/FATAL handler decision; before this PR, assert(0) compiled out under NDEBUG and defaultSwAssert only forced abort inside #if FW_ASSERTIONS_ALWAYS_ABORT. I reproduced the two bodies with -DNDEBUG: the previous body returned normally, while this body terminated with SIGABRT (signal 6). Please preserve the configured behavior—e.g. keep the termination decision in defaultSwAssert and make the noreturn contract conditional—or treat this as a separately approved semantic change with coverage for both configuration values.

lntutor (changes requested):
The compile-time implementation pattern is promising, but the default do-assert implementation introduces an unconditional termination semantic change; see the inline finding.

thomas-bc:
Note: #5617 introduced two selection mechanisms for Os::RawTime (link time vs compile time). This is for optimization, and sets a precedent where I'm not too worried about this dual mechanism here either.

PR #4833, opened 2026-03-11
Refactor Ccsds.ApidManager to use `Fw::ArrayMap`

vsoulgard:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4338 |
|**_Has Unit Tests (y/n)_**| n |
|**_Documentation Included (y/n)_**| n |
|**_Generative AI was used in this contribution (y/n)_**| n |

---
## Change Description

Use `Fw::ArrayMap` instead of a custom dictionary in `Ccsds::ApidManager` class.

## Rationale

Best practice, simple code.

PR #5071, opened 2026-04-30
Fix ApidManager edge cases and throttle event

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Four in one:

1. Fixes a bug line 26:

```diff
- if (.... && receivedSeqCount != SEQUENCE_COUNT_ERROR)
+ if (.... && expectedSequenceCount != SEQUENCE_COUNT_ERROR)
```
this is intending to guard against doing setNextSequenceCount if we already got notification (through `expectedSequenceCount = SEQUENCE_COUNT_ERROR`) that the ApidTable is full

2. reorganize the code for clarity in `getAndIncrementSeqCount` - odd control flow is an artifact of when we used to have an inline data structure instead of a Map (see https://github.com/nasa/fprime/pull/4833)

3. Remove an FW_ASSERT in favor of an event. Even though with the bugfix 1), this code should never be reachable. Still a good practice not to assert here.

4. I also added a throttle on the event, but as I'm writing this, I wonder if maybe we specifically don't want to throttle, because these events could be valuable information

### DIFF
--- a/Svc/Ccsds/ApidManager/ApidManager.cpp
+++ b/Svc/Ccsds/ApidManager/ApidManager.cpp
@@ -22,12 +22,18 @@ ApidManager ::ApidManager(const char* const compName) : ApidManagerComponentBase
 // ----------------------------------------------------------------------
 
 U16 ApidManager ::validateApidSeqCountIn_handler(FwIndexType portNum, const ComCfg::Apid& apid, U16 receivedSeqCount) {
-    U16 expectedSequenceCount = this->getAndIncrementSeqCount(apid);
-    if (receivedSeqCount != expectedSequenceCount && receivedSeqCount != SEQUENCE_COUNT_ERROR) {
+    const U16 expectedSequenceCount = this->getAndIncrementSeqCount(apid);
+    if (expectedSequenceCount == SEQUENCE_COUNT_ERROR) {
+        // APID could not be tracked (table full); nothing to validate or sync against
+        return receivedSeqCount;
+    }
+    if (receivedSeqCount != expectedSequenceCount) {
         // Likely a packet was dropped or out of order
         this->log_WARNING_LO_UnexpectedSequenceCount(receivedSeqCount, expectedSequenceCount);
-        // Synchronize onboard count with received number so that count can keep going
-        this->setNextSeqCount(apid, this->calculateNextSeqCount(receivedSeqCount));
+        // Sync onboard count with received so counting can keep going. Insert cannot fail:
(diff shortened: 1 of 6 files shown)


Svc/Ccsds/ApidManager/ApidManager.fpp
```diff
@@ -17,7 +17,9 @@ module Ccsds {
         @ Received an unregistered APID
         event ApidTableFull(invalidApidValue: U16) \
             severity warning high \
-            format "APID Table is full, cannot generate or check sequence counts for APID: {}"
+            format "APID Table is full, cannot generate or check sequence counts for APID: {}" \
+            throttle 2 \
+            every {seconds=5}
```
thomas-bc: Had this at first, but I'm not convinced anymore that throttling this event is actually a good thing. Looking for feedback

Svc/Ccsds/ApidManager/ApidManager.cpp
```diff
@@ -48,17 +48,22 @@ U16 ApidManager ::getAndIncrementSeqCount(ComCfg::Apid::T apid) {
...
-    this->log_WARNING_HI_ApidTableFull(apid);
-    return SEQUENCE_COUNT_ERROR;
+    return seqCount;  // Return the current sequence count
 }
 
 void ApidManager::setNextSeqCount(ComCfg::Apid::T apid, U16 seqCount) {
     Fw::Success insertStatus = m_apidSequences.insert(apid, seqCount);
-    FW_ASSERT(insertStatus == Fw::Success::SUCCESS, static_cast<FwAssertArgType>(apid));
+    if (insertStatus != Fw::Success::SUCCESS) {
+        // This code should not be reachable with the if (expectedSequenceCount != SEQUENCE_COUNT_ERROR)
```
LeStarch: If it cannot be reached, then it should be an assert because it would be a failed software assumption.

LeStarch (changes requested):

Svc/Ccsds/ApidManager/ApidManager.cpp
thomas-bc: Ok. Note to self, check UTs for why they passed before and still pass. I suspect there should be an additional check missing.
Also, see if there's a way to simplify the complexity. Maybe helper function is not needed.

thomas-bc:
Closing and will reopen another PR

thomas-bc:
Superseded by https://github.com/nasa/fprime/pull/5091

PR #5073, opened 2026-04-30
Fix #5065: Remove assertion on command errors and handle queue full gracefully

### LINKED ISSUE #5065: Potential System Instability via Command Validation in CmdDispatcher
| | |
|:---|:---|
|**_F´ Version_**| v4.2.1 |
|**_Affected Component_**|CmdDispatcher |
---
## Problem Description

During a review of the HealthComponentImpl.cpp implementation, a potential reliability risk was identified concerning the HLTH_CHNG_PING and HLTH_PING_ENABLE command handlers. While these handlers currently perform logical validation and return Fw::CmdResponse::VALIDATION_ERROR, there is a configuration risk if the component's command dictionary is set to queueFullBehavior: "assert".

The F Prime Frameworks is used for mission-critical systems, an assertion triggered by command processing can lead to an unintended system reset. If an operator or a ground-side script sends invalid parameters (e.g., warningValue > fatalValue) or the uplink transmission is corrupted, the system should generate an event or increment a command error telemetry channel than risk a hard halt via assertion. If this occurs during a critical time in the mission, this could result in the loss of the spacecraft or mission. 

## Proposed Changes

Modify the CMD_DISPATCH logic. If the dispatch call returns a "Full" status, emit a WARNING_HI_CommandDroppedQueueFull event and ensure it always returns Fw::CmdResponse::EXECUTION_ERROR rather than allowing the dispatcher to hang or assert.

## Context / Environment

Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-6.14.0-37-generic-x86_64-with-glibc2.39
Python version: 3.12.3
CMake version: 3.26.0
Pip version: 26.0.1
Pip packages:
    fprime-tools==4.2.0
    fprime-gds==4.2.0
    fprime-fpp==3.3.0a1
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
[WARNING] Failed to retrieve submodule version information.

## How to Reproduce

1) Rapid-fire command sequence to trigger queueFullBehavior: "assert"

for i in range(25):
    # Flood the Health component with valid ping updates
    Svc.Health.HLTH_CHNG_PING("ActiveLogger", 100, 200)

2) The HLTH_CHNG_PING accepts logically invalid threshold configuration (warn >= fatal) without validation. 

fprime-cli command-send --dictionary <dict> \
    CdhCore.health.HLTH_CHNG_PING \
    --arguments="cmdDisp" "5" "3" \
    --tts-addr 127.0.0.1

## Expected Behavior

F Prime should gracefully respond to a command queue overflow or invalid opcode condition and not assert. This is especially important when a burst of commands are received.

linerfan5114:
Closes #5065

Changes
- Removed FW_ASSERT in compCmdStat_handler that caused system reset on command error
- Added CommandError WARNING_HI event for operator diagnostic visibility
- Added compCmdSendQueueFull input port for destination component queue full notification
- Added CommandDroppedQueueFull WARNING_HI event when destination queue is full
- Modified seqCmdBuff_overflowHook to return EXECUTION_ERROR instead of asserting
- Added m_numCmdQueueFull counter for tracking dropped commands due to queue full
- Added runCommandErrorResponse unit test

Testing
- Unit tests pass
- Added new test case for command error response flow


LeStarch:
This PR does not seem to address the problem as stated, nor do the changes make sense.

linerfan5114:
@LeStarch Thanks for reviewing. The issue #5065 describes that when a command handler returns VALIDATION_ERROR and the dictionary has queueFullBehavior: "assert", the FW_ASSERT in compCmdStat_handler causes a system reset. The fix removes that assertion and replaces it with a WARNING_HI event, so the system logs the error and continues operating instead of resetting. Let me know if I should rebase or if a different approach is preferred.

LeStarch:
@linerfan5114....first that issue is awaiting approval from the CCB so we need to approve it for implementation before we agree to accept a PR.  It likely will be approved, but that process is where we might recommend implementation specifics.

The assert you removed is actually correct.  The assert in question is the queue-overflow assert that is in the autocode layer.

The ideal implementation involves improving the autocoder to return a status on queue overflow and then let the caller (cmdDisp) deal with the message.  However, since this change isn't available the CCB is still reviewing this change.

@bitWarrior did I sum this up correctly?

Issue #2458, opened 2023-12-22
Vulnerabilities in FileUplink overwriting


rrieber:
| | |
|:---|:---|
|**_F´ Version_**| |
|**FileUplink**|   |
---
## Problem Description

CADRE discovered a vulnerability with FileUplink associated with overwriting existing files. Assume version-1 of a file is saved with the name <FILE.bin> and has the size X-bytes. Version-2 of <FILE.bin> is uplinked with a size of Y-bytes where X>Y, e.g. the new file is smaller than the old file. In this scenario, FileUplink successfully validates size, checksum, etc, then saves to disk in such a way that it partially overwrites the old version. The result is <FILE.bin> that has a size of X-bytes. The first Y-bytes originate in FILE.bin_v2 and the remaining (X minus Y)-bytes originate from FILE.bin_v1. And what is concerning is that no errors or warnings are raised.

CADRE is taking this as Use-As-Is (UAI) and operationally restricting file overwrites on uplink.

For JPLers, reference [CADRE issue 946](https://cadregitlab.jpl.nasa.gov/cadre/cadre-fsw/-/issues/946).

According to @csmith608, other FPrime missions at JPL have addressed this problem in their local code base, so there are solutions out in the wild.

## How to Reproduce

See above 

## Expected Behavior

The old file deleted prior to the new file being saved.

Joshua-Anderson:
A bit of context on why the existing behavior is desired by some projects:

FileUplink intended to be used for receiving file downlinks over unreliable links (i.e. radio). Because of this, the expectation is that files will need to be retransmitted, or transmitted partially over multiple activities. 

Example: Send a 3 MB files in 3 x 1 MB chunks, then later retransmit a 50 KB chunk in the middle of the file.

The goal was for FileUplink to be safe for data downlink - in that it would never destroy data, only overwrite.

Joshua-Anderson:
The problem as I view it is that there's two different use cases for FileUplink:

- Receiving critical files sent TO a deployment
- Receiving critical data sent FROM a deployment.

And you want different behaviors for both cases.

When receiving data from a deployment, you never want to destroy data, since you may never have another chance to transmit that data.

When sending files to a deployment, file integrity is paramount, since corrupted commands/configuration can cause irrecoverable damage.

Joshua-Anderson:
[...]

PR #5077, opened 2026-05-01
PR-5034 - Fix ground file upload processing of different file size for the same filename

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**| PR 5034 |

---
## Change Description

Added OVERWRITE parameter when opening file in Svc/FileUplink/File::open()
[WAS] this->osFile.open(path, Os::File::OPEN_CREATE)
[IS] this->osFile.open(path, Os::File::OPEN_CREATE, Os::File::OverwriteType::OVERWRITE);

## Rationale

This ensures that opening a file with the same filename with a smaller size than the previous version will not contain any stale data from the old file. Without this change, F Prime was reading in extra data from the previous file. At mimimum this wastes hardware resources. At worse, something can be embedded in this space that may be undesirable. 

I created a new unit test called FileUplinkOverwriteTests to validate this new functionality. 

## Future Work

N/A

I used Claude to generate the unit test, but I kinda wish I didn't becuase in the time it took to generate something usefull, I could have done it myself and saved some tokens, electrons and frustration.

### DIFF
--- a/Svc/FileUplink/CMakeLists.txt
+++ b/Svc/FileUplink/CMakeLists.txt
@@ -27,6 +27,7 @@ set(UT_SOURCE_FILES
   "${CMAKE_CURRENT_LIST_DIR}/FileUplink.fpp"
   ${CMAKE_CURRENT_LIST_DIR}/test/ut/FileUplinkTester.cpp
   ${CMAKE_CURRENT_LIST_DIR}/test/ut/FileUplinkMain.cpp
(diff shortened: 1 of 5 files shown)


LeStarch:
Per issue #2458, the correct fix for this issue is:

> The CCB is recommending these steps:
>
>  1.  Send size in start packet and if file exists, truncate to that size if larger.
>  2. Add parameter for full-file CRC
>  3. Add full-file CRC to end packet
>  4. Flag for when not including CRC

Since this is in no part the above solution, I will close this PR.

PR #5078, opened 2026-05-01
PR 5036 - Event vs. ASSERT if DP file is corrupted

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**| PR 5036 |

---
## Change Description

If the DP file is corrupted, a new event named FileCorruptedDataError is now generated vs. triggering an ASSERT. 

## Rationale

This change adds robustness when encountering an off-nominal condition where the DP files is corrupted, rather than crashing the FSW and not providing the ground an opprotunity to diagnose and correct the condition. 

A new unit test called test_MalformedFile was added. All other baseline unit test pass without modification.

To test this implementation, I also created a DpState.dat file that contained all 0xFF:
00000000: ffff ffff ffff ffff ffff ffff ffff ffff ................
00000010: ffff ffff ffff ffff ffff ffff ffff ffff ................
00000020: ffff ffff ffff ffff ........

When DataProducts.dpCat.BUILD_CATALOG was called, the new FileCorruptedDataError event was generated and the FSW continued to operate nominally.

## Future Work

N/A

N


LeStarch:
Blocked on other attempt to fix this problem.  See PR #5042

LeStarch:
The C++ code seems unaffected, so I will continue with a review.

LeStarch (commented):
There seems to be some discrepancy in this load routine.

Some errors fail the command, other doe not (they return success so as to "use the records that were read).

This PR adds to the "error" cases, but maybe need to be the success-and-warn case.

Svc/DpCatalog/test/ut/DpCatalogTester.cpp
```diff
@@ -751,6 +751,44 @@ void DpCatalogTester::test_ProcessFileInvalidDir() {
...
+    Fw::FileNameString dirs[DP_MAX_DIRECTORIES];
+    this->component.configure(dirs, 0, stateFile, 0, mockAllocator);
+
+    // 4. Dispatch the BUILD_CATALOG command
+    this->sendCmd_BUILD_CATALOG(0, 0);
+    this->component.doDispatch();
+
+    // 5. Command should generate event instead of ASSERT
+    ASSERT_CMD_RESPONSE_SIZE(1);
+    // ASSERT_CMD_RESPONSE(0, DpCatalogComponentBase::OPCODE_BUILD_CATALOG, 0, Fw::CmdResponse::EXECUTION_ERROR);
```
LeStarch: Probably don't want to leave this commented out.

LeStarch (changes requested):

Svc/DpCatalog/test/ut/DpCatalogTester.cpp
LeStarch: Cannot fix, DpCatalog is broken

LeStarch (approved):
Finally fixed.

PR #5079, opened 2026-05-01
PR 5038 - Force Framers to setSize on Fw::Buffer before outputting it

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**| PR 5038 |

---
## Change Description

Iadded a single setSize() call in each framer immediately before dataOut_out, trimming the buffer to the exact computed frame size. frameSize is already in scope in both handlers at that point and requires no additional computation.

Svc/FprimeFramer/FprimeFramer.cpp
// Trim to actual frame size in case allocator returned a larger buffer
frameBuffer.setSize(static_cast<Fw::Buffer::SizeType>(frameSize));

Svc/Ccsds/SpacePacketFramer/SpacePacketFramer.cpp
// Trim to actual frame size in case allocator returned a larger buffer
frameBuffer.setSize(static_cast<Fw::Buffer::SizeType>(frameSize));

This trim any extra data above what is expected. When a memory pool allocator returns a buffer larger than requested — which is common in fixed-block-size allocators — without the setSize trim the framed output would advertise an incorrect (oversized) length to the downstream ComManager, potentially causing it to transmit garbage padding bytes as part of the frame. The change ensures the framed buffer's reported size always exactly matches the true frame size regardless of allocator behavior, making FprimeFramer and SpacePacketFramer more complete, rather than relying on all allocator implementations to return exactly-sized buffers.

## Future Work

None.

Claude was used to generate unit tests.

### DIFF
--- a/Svc/Ccsds/SpacePacketFramer/test/ut/SpacePacketFramerTester.cpp
+++ b/Svc/Ccsds/SpacePacketFramer/test/ut/SpacePacketFramerTester.cpp
@@ -81,6 +81,33 @@ void SpacePacketFramerTester::testNominalFraming() {
     ASSERT_EQ(this->fromPortHistory_dataReturnOut->at(0).data.getSize(), sizeof(payload));
 }
 
+void SpacePacketFramerTester ::testOversizedAllocatorBufferIsTrimmed() {
+    U8 payload[16];
+    for (U32 i = 0; i < sizeof(payload); ++i) {
+        payload[i] = static_cast<U8>(STest::Random::lowerUpper(0, 0xFF));
+    }
+    Fw::Buffer data(payload, sizeof(payload));
+    ComCfg::FrameContext context;
+    context.set_apid(static_cast<ComCfg::Apid::T>(0x01));
+    this->m_nextSeqCount = 0;
+
+    // Signal the allocator handler to return a larger-than-needed buffer,
+    // simulating a BufferManager bin that is bigger than the exact packet size.
+    this->m_useOversizedAlloc = true;
+    this->invoke_to_dataIn(0, data, context);
+    this->m_useOversizedAlloc = false;
+
+    ASSERT_from_dataOut_SIZE(1);
+    ASSERT_from_dataReturnOut_SIZE(1);
+
+    const FwSizeType expectedFrameSize = sizeof(payload) + SpacePacketHeader::SERIALIZED_SIZE;
+
+    Fw::Buffer outBuffer = this->fromPortHistory_dataOut->at(0).data;
+    // If setSize() is missing from SpacePacketFramer, getSize() returns the
+    // oversized allocation (2 * expectedFrameSize) and this assertion fails.
+    ASSERT_EQ(outBuffer.getSize(), expectedFrameSize);
+}
+
 // ----------------------------------------------------------------------
 // Output port handler overrides
 // ----------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


Svc/Ccsds/SpacePacketFramer/test/ut/SpacePacketFramerTester.cpp
```diff
@@ -81,6 +81,33 @@ void SpacePacketFramerTester::testNominalFraming() {
...
+        payload[i] = static_cast<U8>(STest::Random::lowerUpper(0, 0xFF));
+    }
+    Fw::Buffer data(payload, sizeof(payload));
+    ComCfg::FrameContext context;
+    context.set_apid(static_cast<ComCfg::Apid::T>(0x01));
+    this->m_nextSeqCount = 0;
+
+    // Signal the allocator handler to return a larger-than-needed buffer,
+    // simulating a BufferManager bin that is bigger than the exact packet size.
+    this->m_useOversizedAlloc = true;
```
LeStarch, line 96: This doesn't actually do anything.  On line 122, you need to use this variable to provide a buffer >size.

LeStarch (changes requested):
One of the tests is flawed.  Otherwise looks good.

LeStarch (dismissed):
I went ahead and added the missing test code.

Svc/Ccsds/SpacePacketFramer/test/ut/SpacePacketFramerTester.cpp
```diff
@@ -92,7 +119,8 @@ U16 SpacePacketFramerTester ::from_getApidSeqCount_handler(FwIndexType portNum,
 }
 
 Fw::Buffer SpacePacketFramerTester ::from_bufferAllocate_handler(FwIndexType portNum, FwSizeType size) {
-    return Fw::Buffer(this->m_internalDataBuffer, size);
+    FwSizeType allocation = (this->m_useOversizedAlloc) ? sizeof(this->m_internalDataBuffer) : size;
+    return Fw::Buffer(this->m_internalDataBuffer, allocation);
```
thomas-bc, line 123: Implementation here is significantly different from that of FprimeFramerTester. No setsize or pushing history entry etc.
Should they differ?

Svc/Ccsds/SpacePacketFramer/test/ut/SpacePacketFramerTester.hpp
```diff
@@ -80,6 +81,12 @@ class SpacePacketFramerTester final : public SpacePacketFramerGTestBase {
     U8 m_internalDataBuffer[SpacePacketHeader::SERIALIZED_SIZE + MAX_TEST_PACKET_DATA_SIZE];
 
     U16 m_nextSeqCount;  // Sequence count to be returned by getApidSeqCount output port
+
+    // Backing store for the oversized allocation path
+    U8 m_oversizedDataBuffer[512];
```
thomas-bc: Not used - should either remove or use as intended when oversized allocation is requested.

thomas-bc (changes requested):

Svc/Ccsds/SpacePacketFramer/test/ut/SpacePacketFramerTester.cpp
LeStarch: Set size is not necessary, not is the memset.  History is only needed if it is used....which it is not.

thomas-bc (approved):

LeStarch (approved):

PR #5085, opened 2026-05-02
PR 5084 - Fixed pkt_length integer overflow and replaced ASSERT w/ event

bitWarrior:
…in SpacePacketDeframer + unit tests

| | |
|:---|:---|
|**_Related Issue(s)_**| PR 5084  |

---
## Change Description

For Issue 1, I added a guard to packetDataLength. The guard widens the arithmetic to U32 before adding 1, ensuring the overflow value 65536 correctly exceeds the available buffer size and the packet is rejected with an InvalidLength event.

For Issue 2, I replaced the ASSERT on the return status of deserializeTo(header). Before, the ASSERT would crash the  flight software if header deserialisation failed on a malformed or truncated incoming buffer. The fix replaces the ASSERT with a graceful check that emits an InvalidPacket warning event, drops the buffer, and returns — keeping the FSW operational.

## Rationale

These changes will improve the robustness of F Prime and make it less brittle to off-nominal conditions.

Execute new unit tests for off-nominal conditions: 1) testPacketDataLengthMaxU16Overflow 2) testBufferExactlyHeaderSize, 3) testBufferSmallerThanHeaderSize, 4) testBufferSingleByte. 

## Future Work

N/A

Claude was used to generate new unit tests.

### DIFF
--- a/Svc/Ccsds/SpacePacketDeframer/SpacePacketDeframer.cpp
+++ b/Svc/Ccsds/SpacePacketDeframer/SpacePacketDeframer.cpp
@@ -52,12 +52,29 @@ void SpacePacketDeframer ::dataIn_handler(FwIndexType portNum, Fw::Buffer& data,
 
     SpacePacketHeader header;
     Fw::SerializeStatus status = data.getDeserializer().deserializeTo(header);
-    // Length is valid, so deserialization here should succeed
-    FW_ASSERT(status == Fw::FW_SERIALIZE_OK, status);
+    // Deserialisation can still fail if the buffer is malformed despite passing the size check
+    if (status != Fw::FW_SERIALIZE_OK) {
+        this->log_WARNING_HI_InvalidPacket();
+        if (this->isConnected_errorNotify_OutputPort(0)) {
+            this->errorNotify_out(0, Svc::Ccsds::FrameError::SP_INVALID_PACKET);
+        }
+        this->dataReturnOut_out(0, data, context);  // Drop the packet
+        return;
+    }
 
-    // Space Packet protocol defines the Data Length as number of bytes minus 1
-    // so we need to add 1 to the length to get the actual data size
-    U16 pkt_length = static_cast<U16>(header.get_packetDataLength() + 1);
+    // Widen to U32 before adding 1 to prevent U16 truncation to 0 when packetDataLength == 0xFFFF (max U16 value).
+    // This is a undefined behavior condition in C++.
+    const U32 pkt_length_wide = static_cast<U32>(header.get_packetDataLength()) + 1U;
+    if (pkt_length_wide > static_cast<U32>(data.getSize() - SpacePacketHeader::SERIALIZED_SIZE)) {
+        FwSizeType maxDataAvailable = data.getSize() - SpacePacketHeader::SERIALIZED_SIZE;
+        this->log_WARNING_HI_InvalidLength(static_cast<U16>(pkt_length_wide), maxDataAvailable);
+        if (this->isConnected_errorNotify_OutputPort(0)) {
+            this->errorNotify_out(0, Svc::Ccsds::FrameError::SP_INVALID_LENGTH);
+        }
+        this->dataReturnOut_out(0, data, context);
+        return;
+    }
+    const U16 pkt_length = static_cast<U16>(pkt_length_wide);
     if (pkt_length > data.getSize() - SpacePacketHeader::SERIALIZED_SIZE) {
         FwSizeType maxDataAvailable = data.getSize() - SpacePacketHeader::SERIALIZED_SIZE;
         this->log_WARNING_HI_InvalidLength(pkt_length, maxDataAvailable);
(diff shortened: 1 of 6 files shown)


Svc/Ccsds/SpacePacketDeframer/SpacePacketDeframer.cpp
```diff
@@ -52,12 +52,29 @@ void SpacePacketDeframer ::dataIn_handler(FwIndexType portNum, Fw::Buffer& data,
...
+        return;
+    }
 
-    // Space Packet protocol defines the Data Length as number of bytes minus 1
-    // so we need to add 1 to the length to get the actual data size
-    U16 pkt_length = static_cast<U16>(header.get_packetDataLength() + 1);
+    // Widen to U32 before adding 1 to prevent U16 truncation to 0 when packetDataLength == 0xFFFF (max U16 value).
+    // This is a undefined behavior condition in C++.
+    const U32 pkt_length_wide = static_cast<U32>(header.get_packetDataLength()) + 1U;
+    if (pkt_length_wide > static_cast<U32>(data.getSize() - SpacePacketHeader::SERIALIZED_SIZE)) {
```
LeStarch: There is no reason for this redundant check.   Just leave pkt_length as the widened U32. and the below checks should pass correctly, while leaving only one return branch.

LeStarch (changes requested):
Please remove redundant invalid length block.

LeStarch (dismissed):

LeStarch (dismissed):

LeStarch (dismissed):

Svc/Ccsds/SpacePacketDeframer/SpacePacketDeframer.fpp
```diff
@@ -17,7 +17,7 @@ module Ccsds {
             format "Malformed packet received refusing to deframe"
 
         @ Deframing received an invalid frame length
-        event InvalidLength(transmitted: U16, actual: FwSizeType) \
+        event InvalidLength(transmitted: FwSizeType, actual: FwSizeType) \
```
thomas-bc, line 20: this changes from U16 to FwSizeType, but the code changed from U16 to U32

Svc/Ccsds/SpacePacketDeframer/test/ut/SpacePacketDeframerTestMain.cpp
```diff
@@ -21,6 +21,38 @@ TEST(SpacePacketDeframer, testDeframingIncorrectLength) {
     tester.testDeframingIncorrectLength();
 }
 
+// ----------------------------------------------------------------------
+// U16 overflow in pkt_length calculation
+// packetDataLength=0xFFFF wraps to 0 without the U32-widening fix,
```
thomas-bc: these comments speak to code that doesn't exist anymore which will be confusing to maintain, we should remove/rephrase

Svc/Ccsds/SpacePacketDeframer/test/ut/SpacePacketDeframerTester.cpp
```diff
@@ -106,9 +106,110 @@ void SpacePacketDeframerTester ::testDeframingIncorrectLength() {
...
+
+// ----------------------------------------------------------------------
+// Finding 3 — graceful handling of undersized buffers (no FW_ASSERT)
+// ----------------------------------------------------------------------
+
+void SpacePacketDeframerTester ::testBufferExactlyHeaderSize() {
+    // A buffer of exactly SERIALIZED_SIZE bytes passes the `>` check in the
+    // original FW_ASSERT version only if the check was `<` — the updated
+    // check uses `<=` so this must be caught and dropped gracefully.
+    // This is the tightest boundary case for the size guard.
```
thomas-bc: I don't understand this comment. There was no FW_ASSERT that I can find, only an `if (size)` check which hasn't changed. This makes me worry about the test code rationale here.

The size is also checked with `<=` so all three tests with sizes `SERIALIZED_SIZE`, `SERIALIZED_SIZE - 1` and `1` are functionally equivalent (maybe cause of wrong rationale). Should there be a test with a 0-size buffer?

Svc/Ccsds/SpacePacketDeframer/test/ut/SpacePacketDeframerTester.cpp
```diff
@@ -106,9 +106,110 @@ void SpacePacketDeframerTester ::testDeframingIncorrectLength() {
...
+    //   pkt_length = static_cast<U16>(0xFFFF + 1) = 0
+    //   length guard: 0 > (bufSize - SERIALIZED_SIZE) = false → PASSES
+    //   → empty buffer (size 0) forwarded to dataOut — silent corruption.
+    //
+    // With the fix (widen to U32 before adding 1):
+    //   pkt_length_wide = 65536
+    //   length guard: 65536 > 1 = true → REJECTED with InvalidLength.
+    //
+    // This test asserts the correct post-fix behavior.  If run against the
+    // unfixed code it will fail, acting as a regression guard.
```
thomas-bc: nit: these comments speak to code that doesn't exist anymore which will be confusing moving forward. I can remove.

thomas-bc (changes requested):

Svc/Ccsds/SpacePacketDeframer/SpacePacketDeframer.fpp
LeStarch: Since the comparison is automatically promoted, I am not worried about this.  This Event will show the actual comparison that is done.

thomas-bc: My worry is that this is the kind of assumption that prevent us from using `FwSizeType` smaller than U32.

thomas-bc (approved):

Svc/Ccsds/SpacePacketDeframer/SpacePacketDeframer.fpp
LeStarch: FwSizeType >= U32 is asserted in FPrimeBasicTypes.hpp and is an invariant that is required by the system.

PR #5087, opened 2026-05-02
PR 5086 - Improved frame detector arithmetic against integer underflo…

bitWarrior:
…w and overflow

| | |
|:---|:---|
|**_Related Issue(s)_**| PR 5086 |

---
## Change Description

I implemented the following changes:
1) Replaced the subtraction-based guard with an explicit addition-based check:
lengthField > MAX - overhead.

2) Added a static_assert to verify the two overhead constants don't overflow
when added together (compile-time, zero runtime cost).

3) Added a FW_ASSERT before hash_field_size that documents and enforces the
safety invariant at the point of use, so it cannot be silently broken by a
future refactor (See I do not hate ASSERTS :)>)

4) Moved data_to_crc_length to after both the MORE_DATA_NEEDED check and the
minimum-size check, so the subtraction can only execute when
expected_frame_length >= TRAILER_SIZE is guaranteed.

5) Widened data_to_crc_length from U16 to FwSizeType to match the type of
expected_frame_length and eliminate the truncation risk.

6) Updated the comment on the minimum-size check to explain why the ordering
matters.

## Rationale

The old implementation is brittle to malformed packets, which could contaminate CCSDS and F Prime packets .

Execute new unit tests CcsdsTcFrameDetectorUnderflowTest and FprimeFrameDetectorTest

## Future Work

N/A

Claude was used to generate new unit tests

### DIFF
--- a/Svc/FrameAccumulator/test/ut/detectors/FprimeFrameDetectorTest.cpp
+++ b/Svc/FrameAccumulator/test/ut/detectors/FprimeFrameDetectorTest.cpp
@@ -1,0 +1,290 @@
+// ======================================================================
+// \title  FprimeFrameDetectorTest.cpp
+// \brief  Unit tests for overflow-hardening fixes in FprimeFrameDetector
+//
+// These tests exercise FprimeFrameDetector::detect() directly (unlike the
+// FrameAccumulator tests which use a MockDetector). They cover:
+//
+//   Fix 1 — Overflow guard: near-max lengthField values must be rejected
+//            or handled safely before expected_frame_size is computed.
+//
+//            Implementation note: lengthField is a TokenType (U32, 32-bit).
+//            On a 64-bit host FwSizeType is 64-bit, so TokenType::max +
+//            OVERHEAD cannot overflow FwSizeType — the guard is dead on
+//            this platform. The tests below therefore verify the correct
+//            observable behaviour (NO_FRAME_DETECTED) for extreme inputs
+//            regardless of which check produces it (guard vs. capacity).
+//            On a 32-bit embedded target where FwSizeType == U32, the
+//            guard becomes the critical path and prevents the overflow.
+//
+//   Fix 2 — hash_field_size invariant: frames that pass the guard must
+//            reach the CRC stage without triggering the FW_ASSERT, proving
+//            the safety contract holds end-to-end.
+// ======================================================================
+
+#include <cstring>
+#include <limits>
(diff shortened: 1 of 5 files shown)


LeStarch (approved):

Svc/FrameAccumulator/test/ut/detectors/FprimeFrameDetectorTest.cpp
thomas-bc, line 1: I am worried about these AI generated tests. 

First, they don't share the same infrastructure as the pre-existing tests but rather re-define a bunch of helpers. That'll make it harder to maintain in the future. 

Also, some of the tests are duplicates. See for example [FprimeFrameDetector.TestBufferTooSmall](https://github.com/bitwarrior/fprime/blob/7505dd0329d78d0047bd0732c352afb868252d14/Svc/FrameAccumulator/test/ut/detectors/FprimeFrameDetectorTestMain.cpp#L72-L92) == [NewTests.InsufficientDataReturnsMoreDataNeeded](https://github.com/bitwarrior/fprime/blob/7505dd0329d78d0047bd0732c352afb868252d14/Svc/FrameAccumulator/test/ut/detectors/FprimeFrameDetectorTest.cpp#L254-L268)

And most importantly, I suspect most of these tests don't actually exercise what they describe to be testing:
[buildHeader()](https://github.com/bitwarrior/fprime/blob/7505dd0329d78d0047bd0732c352afb868252d14/Svc/FrameAccumulator/test/ut/detectors/CcsdsTcFrameDetectorUnderflowTest.cpp#L77) sets a passed in flags_and_scid token (either 0 or 0xFFFF). But the [CcsdsTcDetector matches on that token and rejects anything not equal to it](https://github.com/bitwarrior/fprime/blob/7505dd0329d78d0047bd0732c352afb868252d14/Svc/FrameAccumulator/FrameDetector/CcsdsTcFrameDetector.cpp#L40-L43). In most of these tests, the only assertion is on the output `NO_FRAME_DETECTED` (e.g. [here](https://github.com/bitwarrior/fprime/blob/7505dd0329d78d0047bd0732c352afb868252d14/Svc/FrameAccumulator/test/ut/detectors/CcsdsTcFrameDetectorUnderflowTest.cpp#L181)), which is returned because of malformed header token. This makes it that no test in this file exercise anything beyond Line 40 in the file.

thomas-bc: Captured in https://github.com/nasa/fprime/issues/5089

PR #5091, opened 2026-05-04
Fix ApidManager bug validating seqCount when table is full

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes a bug in ApidManager where validating a sequence count for a new APID would behave improperly if the Apid table is full.
Adds a test to cover this case.
Reduces complexity and odd code flow that was an artifact of migrating from inline data structures to map + odd choices by past me.

### DIFF
--- a/Svc/Ccsds/ApidManager/ApidManager.cpp
+++ b/Svc/Ccsds/ApidManager/ApidManager.cpp
@@ -30,8 +30,9 @@ U16 ApidManager ::validateApidSeqCountIn_handler(FwIndexType portNum, const ComC
     if (receivedSeqCount != expectedSequenceCount) {
         // Likely a packet was dropped or out of order
         this->log_WARNING_LO_UnexpectedSequenceCount(receivedSeqCount, expectedSequenceCount);
-        // Sync onboard count with received so counting can keep going. Insert cannot fail:
-        // getAndIncrementSeqCount above already tracked this APID, so this only updates it.
+        // Sync onboard count to what was received so counting can keep going. Insert cannot fail:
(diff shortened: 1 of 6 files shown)


Svc/Ccsds/ApidManager/ApidManager.cpp
```diff
@@ -42,23 +49,16 @@ U16 ApidManager ::getApidSeqCountIn_handler(FwIndexType portNum, const ComCfg::A
 
 U16 ApidManager ::getAndIncrementSeqCount(ComCfg::Apid::T apid) {
     U16 seqCount = 0;
-    // Find sequence count if exists, otherwise use 0
+    // Find current sequence count if APID is tracked; otherwise default to 0 (first count for a new APID)
     (void)m_apidSequences.find(apid, seqCount);
```
LeStarch, line 53: Are we sure that `find` does not touch `seqCount` in the case of a failed look-up?

thomas-bc: Yes: [docs](https://github.com/nasa/fprime/blob/1826e097e879d638ca70fa1a090d3ad03c7d84e2/Fw/DataStructures/docs/MapBase.md#L156-L161) say update value only if found, and confirmed in code: 
https://github.com/nasa/fprime/blob/1826e097e879d638ca70fa1a090d3ad03c7d84e2/Fw/DataStructures/ArraySetOrMapImpl.hpp#L178-L185

LeStarch (approved):

PR #5101, opened 2026-05-05
Fix CommandDispatcher sequence number wraparound error

### LINKED ISSUE #5082: CommandDispatcher's sequence number wraparound error
| | |
|:---|:---|
|**_F´ Version_**| 4.2.1 |
|**_Affected Component_**|  CommandDispatcher  |
---
## Problem Description

In the seqCmdBuff_handler function, the U32 counter m_seq is incremented unconditionally on every dispatched command. When m_seq wraps around to 0,  if any command sequence entry from the previous commanding cycle remains in m_sequenceTracker with a value of 0, then the command is rejected requiring the CMD_CLEAR_TRACKING command to clear the condition. 

The recommended change is to add a sentinel range around the max value 0xFFFFFFFF (4,294,967,295). 

CommandDispatcherImpl.hpp:
    // Guard for wrap-around
    static constexpr U32 CMD_SEQ_GUARD =
        std::numeric_limits<U32>::max() - CMD_DISPATCHER_SEQUENCER_TABLE_SIZE;

CommandDispatcherImpl.cpp:
[WAS] 
this->m_seq++;
[IS] 
this->m_seq++;
if (this->m_seq >= CMD_SEQ_GUARD) {
    this->m_seq = 0;
}

## Context / Environment

Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-6.14.0-37-generic-x86_64-with-glibc2.39
Python version: 3.12.3
CMake version: 3.26.0
Pip version: 26.1
Pip packages:
    fprime-tools==4.2.0
    fprime-gds==4.2.0
    fprime-fpp==3.3.0a3

## How to Reproduce

4,294,967,295 is allot of commands and has probably not been encountered in the wild. I stumbled on this testing another PR. I created a unit test that dispatches 4,294,967,295 commands to verify command sequence wrapping is handled.

txpham209:
Resolves #5082

| | |
|:---|:---|
|**_Related Issue(s)_**|  #5082 |

---
## Change Description

  This PR addresses the sequence number wraparound error in CommandDispatcherImpl as described in #5082.
  
## Rationale

  Fixes a bug where the command dispatcher can become deadlocked if m_seq wraps around to 0 and attempts to insert into m_sequenceTracker when an entry from the previous cycle remains. By resetting m_seq before it reaches the maximum U32 value minus the table size, we guarantee that duplicate sequence key collisions are avoided.
  
All existing unit tests in Svc/CmdDispatcher pass locally (fprime-util build --ut and fprime-util check). 

Use AI (gemini) to help familiarize myself with the codebase.

### DIFF
--- a/Svc/CmdDispatcher/CommandDispatcherImpl.cpp
+++ b/Svc/CmdDispatcher/CommandDispatcherImpl.cpp
@@ -121,6 +121,9 @@ void CommandDispatcherImpl::seqCmdBuff_handler(FwIndexType portNum, Fw::ComBuffe
 
     // increment sequence number
     this->m_seq++;
+    if (this->m_seq >= CMD_SEQ_GUARD) {
+        this->m_seq = 0;
+    }
 }
 
 void CommandDispatcherImpl ::run_handler(FwIndexType portNum, U32 context) {
--- a/Svc/CmdDispatcher/CommandDispatcherImpl.hpp
+++ b/Svc/CmdDispatcher/CommandDispatcherImpl.hpp
@@ -166,6 +166,10 @@ class CommandDispatcherImpl final : public CommandDispatcherComponentBase {
     //! \brief map from command sequence number to pending command state
     Fw::ArrayMap<U32, SequenceTrackerEntry, CMD_DISPATCHER_SEQUENCER_TABLE_SIZE> m_sequenceTracker;
 
+    //!< Guard for wrap-around
+    static constexpr U32 CMD_SEQ_GUARD =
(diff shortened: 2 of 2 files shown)


zimri-leisher:
U32 max is 4 billion... are we only vulnerable to this if we run 4 billion sequences? There are about a hundred other things that would fail if we got close to having ran that many seqs, so IMO this fix isn't worth the incremental added complexity.

EDIT--copying this comment over to the original issue because that's where the justification lies

LeStarch:
We decided that this isn't the correct fix.  Instead we decided that the correct fix would be to replacing `m_seq++` with a function that searches for an `m_seq` that is not in the table.

I am going to close this PR, but you are welcome to take a look at the suggested solution.

Issue #3598, opened 2025-05-13
ComQueue hits an ASSERT if used without one of the queues


celskeggs:
| | |
|:---|:---|
|**_F´ Version_**| devel |
|**_Affected Component_**| Svc.ComQueue |
---
## Problem Description

I'm working with a deployment that does not use file downlink. Therefore, I did not configure `configurationTable.entries[2]` when calling `comQueue.configure`. However, this resulted in `depth` being 0, which ultimately caused the following assertion:

```
Assert: "/workspaces/[...]/lib/fprime/Utils/Types/Queue.cpp:61" 0
```

This assertion was tripped when `ComQueue` tried to execute this line in `ComQueue::processQueue`:

```
if (queue.getQueueSize() == 0) {
```

## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-4.18.0-553.47.1.el8_10.x86_64-x86_64-with-glibc2.34
Python version: 3.9.21
CMake version: 3.24.2
Pip version: 21.3.1
Pip packages:
    fprime-tools==3.6.1
    fprime-gds==3.6.1
    fprime-fpp-*==3.0.0a8
Project submodules:
    https://github.com/nasa/fprime.git @ v3.6.0-92-g6be3d3a9b
    https://github.com/fprime-community/fprime-baremetal @ 162b7dd
```

## How to Reproduce

See above.

## Expected Behavior

ComQueue should function correctly even if one of its features is not required, and should not require allocating memory that will not be used, solely for the purpose of not crashing.

thomas-bc:
@celskeggs did you also modify the ComQueue ports constants here ?

https://github.com/nasa/fprime/blob/ae2940b6c0cff2fe7138f581f56b90cb896f60ce/default/config/AcConstants.fpp#L36-L40

Note that I'm not sure how FPP would behave with a 0-size port array...

thomas-bc:
Ah, from FPP User Guide

> each named port instance is actually an array of one or more port instances

So yes ComQueue likely isn't usable in its current form without one of the queue

celskeggs:
Yes, I tried that. Setting ComQueueBufferPorts to 0 caused an FPP error, as you predicted.

LeStarch:
I don't know that we'll easily overcome the zero-port-array issue.  Be we certainly could allow configurations to set queue depths to 0, or otherwise specifically disable that particular queue.

PR #5105, opened 2026-05-05
Enhance queue overflow condition check

### LINKED ISSUE #5096: Pre-emptive ComQueue popFront asserts on empty buffer queue when depth=0 and DROP_OLDEST
| | |
|:---|:---|
|**_F´ Version_**| |
---

## Problem Description

Pre-emptive popFront asserts on empty queue when depth=0 and DROP_OLDEST for buffer queues ([Svc/ComQueue/ComQueue.cpp:325-340](https://github.com/jpl-devin/fprime/blob/1826e097e879d638ca70fa1a090d3ad03c7d84e2/Svc/ComQueue/ComQueue.cpp#L325-L340))

The new pre-emptive overflow logic in ComQueue::enqueue() checks queue.getQueueSize() >= this->m_prioritizedList[i].depth. When a buffer queue is configured with depth=0 (which disables the queue) and overflowMode=QUEUE_DROP_OLDEST, the condition 0 >= 0 is always true, causing popFront() to be called on an empty queue. Queue::popFront() then calls m_internal.peek(message, m_message_size, 0) which fails because the circular buffer has no data, returning a non-OK status. The FW_ASSERT(dequeueStatus == Fw::FW_SERIALIZE_OK, ...) at line 335 then fires, crashing the system.

[Found by Devin](https://github.com/JPL-Devin/fprime/pull/38#pullrequestreview-4224281657)

LeStarch:
Add additional check for non-zero depth in queue overflow condition so as to prevent an assertion.

Fixes #5096.

| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix #5096

### DIFF
--- a/Svc/ComQueue/ComQueue.cpp
+++ b/Svc/ComQueue/ComQueue.cpp
@@ -325,7 +325,8 @@ bool ComQueue::enqueue(const FwIndexType queueNum, QueueType queueType, const U8
         for (FwIndexType i = 0; i < TOTAL_PORT_COUNT; i++) {
             if (this->m_prioritizedList[i].index == queueNum &&
                 this->m_prioritizedList[i].overflowMode == Types::QUEUE_DROP_OLDEST &&
-                queue.getQueueSize() >= this->m_prioritizedList[i].depth) {
+                queue.getQueueSize() >= this->m_prioritizedList[i].depth &&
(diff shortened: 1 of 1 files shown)


LeStarch:
Should fix: #3598 instead.

PR #5108, opened 2026-05-06
Add docs on Selecting Component, Port, and Command Kinds

### LINKED ISSUE #4884: ROSES - Document how to decide between Active/Passive/Queued component
### Description

This may be the number 1 question that we get asked during our tutorials. It __always__ comes up.

This needs to be documented, tradeoffs, etc.

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  #4884 |

---
## Change Description

Closes  #4884

### DIFF
--- a/docs/user-manual/framework/component-and-port-selection.md
+++ b/docs/user-manual/framework/component-and-port-selection.md
@@ -8,7 +8,7 @@ This guide assumes you have a basic understanding of the different component and
 > This document does not discuss output ports as the kind of port is determined on the input side of the connection.
 
 > [!IMPORTANT]
-> Since command and port kinds (`sync`, `async`, `guarded`) are very similar in their selection, we will focus on port kind selection and the same principles can be applied to command kind selection.
+> Commands are implemented in F Prime via ports. Thus they have the same kinds (`async`, `sync`, and `guarded`) and for the purposes of this document, they follow the same decision process as ports.
 
 **Table of Contents**
 - [Types of Work in F Prime Systems](#types-of-work-in-f-prime-systems)
@@ -40,7 +40,7 @@ Understanding the type of work your component will perform is the first step to 
 
 ## Component Selection for Cyclic Work
 
-When performing cyclic work, it is crucial to know whether all work in the cycle will complete before the cycle repeats, because this 'slip' indicates a failure to meet the cycle’s hard deadline. For example, a 10 Hz control update must happen every 100 ms. If one iteration takes longer than 100 ms, system control has been compromised. For this reason, cyclic work is almost always performed via synchronous invocations and therefore uses a sync port.
+When performing cyclic work, it is crucial to know whether all work in the cycle will complete before the cycle repeats, because this 'slip' indicates a failure to meet the cycle’s hard deadline. For example, a 10 Hz control update must happen every 100 ms. If one iteration takes longer than 100 ms, the cycle has slipped and system control has been compromised because it did not keep up with the expected control rate. For this reason, cyclic work is almost always performed via synchronous invocations and therefore uses a sync port. Asynchronous invocations are not used because this work would be run outside the rate group cycling context and thus slips and failure to reach deadlines would be hidden in another thread.
 
 > [!CAUTION]
 > Remember, `guarded` ports are also synchronous with an internal mutex to protect data.  These are not as common in cyclic work and a full discussion of `guarded` ports is outside the scope of this document.
@@ -80,13 +80,21 @@ sequenceDiagram
 
 When a component performing cyclic work also needs to accept some Event-Driven work, we require a queue to handle the asynchronous invocations, but adding a queue processing thread may disrupt the critical synchronous invocations of the core cyclic work of the component.  For this exact reason, we use a `queued` component. A `queued` component allows asynchronous events to be accepted while keeping a synchronous core.  In this model, the component dispatches the queue as part of the primary synchronous invocation (i.e. `Svc.Sched` handler) thereby moving the asynchronous work into the cycle.
 
+We do not typically used `sync` commands in this context because a synchronous command would block the command dispatch while executing, and also requires mutex protection to access shared data. These mutex accesses can disrupt the critical cyclic work by occurring at anytime whereas the queue is processed at a specific point in the cycle.
+
+Adding a thread to handle the asynchronous work would either mean the cyclic work is moved to this thread (again hiding slips and failures to meet deadlines) or the component thread and rate group thread would need concurrency protection that causes disruption.
+
+> [!IMPORTANT]
+> In most situations, you want a singular thread to execution to do all the work of the component. If a rate group is driving the component, use a queue to pull asynchronous work into that thread. If you have a Event-Driven components (see below) then all interaction should be `async` as to be performed on the component's thread.
+
+
 ```mermaid
 sequenceDiagram
     participant R as Rate Group
     participant C1 as Queued Component
     participant C2 as Event Source
 
-    C2-)C1: Asynchronous Invocation (e.g. command)
+    C2-)C1: Asynchronous Invocation (command or port)
     C2-)C1: Asynchronous Invocation ...
 
     loop Every Cycle (e.g. every 100ms)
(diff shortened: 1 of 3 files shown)


docs/user-manual/framework/component-and-port-selection.md
```diff
@@ -0,0 +1,236 @@
...
+
+This document will describe how to select the kind of component, port, and command to use when developing within the F Prime framework. We will focus on the component kinds (passive, queued, and active) and the critical port kinds (sync, async). We will begin by discussing the types of work performed by F Prime systems, then use that model to guide port and component kind selection.
+
+This guide assumes you have a basic understanding of the different component and port kinds. If you are unfamiliar with these concepts, please see [Core Constructs: Ports, Components, and Topologies](../overview/03-port-comp-top.md) for an introduction to these concepts.
+
+> [!NOTE]
+> This document does not discuss output ports as the kind of port is determined on the input side of the connection.
+
+> [!IMPORTANT]
+> Since command and port kinds (`sync`, `async`, `guarded`) are very similar in their selection, we will focus on port kind selection and the same principles can be applied to command kind selection.
```
thomas-bc: It could be super helpful instead of this quick note to add small section on Command Kind selection that basically says "everything we looked at for ports apply, except that you're in the specific scenario where the Event Source is the component that dispatches commands and the receiver is your component"?

docs/user-manual/framework/component-and-port-selection.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Types of Work in F Prime Systems
+
+Work in an F Prime system breaks down into three categories roughly driven by the timing requirements of the work.  These are:
+
+1. Cyclic Work: Cyclic work is how F Prime addresses hard deadlines. This work is performed on a repeating schedule driven by a [Rate Group](../design-patterns/rate-group.md). e.g. send a control update every 10ms.
+2. Event-Driven Work: Event-driven work is how F Prime addresses timely work lacking hard deadlines. e.g. dispatch commands reasonably quickly.
+3. Background Work: Background work is how F Prime addresses work without timing requirements. e.g. log telemetry to disk.
+
+Two other relevant terms are synchronous and asynchronous invocations (i.e. how a port executes).  Synchronous invocations happen immediately and block the caller until completion just like a typical function call. Asynchronous invocations are queued up until some point in the future when the receiver processes them. Synchronous ports are invoked synchronously and asynchronous ports are invoked asynchronously and backed by a queue.
```
thomas-bc:
```suggestion
Two other relevant terms are synchronous and asynchronous invocations (i.e. how a port executes).  Synchronous invocations happen immediately and block the caller until completion just like a typical function call. Asynchronous invocations are queued up until some point in the future when the receiver processes them. Synchronous ports are invoked synchronously and asynchronous ports are invoked asynchronously and backed by a queue in the receiver component.
```

Thoughts?

docs/user-manual/framework/component-and-port-selection.md
```diff
@@ -0,0 +1,236 @@
...
+
+> [!IMPORTANT]
+> Cyclic work is almost always performed via synchronous invocations while Event-Driven and Background work is typically performed via asynchronous invocations.
+
+Understanding the type of work your component will perform is the first step to selecting the appropriate component and port kinds. We will discuss component selection for each type of work.
+
+## Component Selection for Cyclic Work
+
+When performing cyclic work, it is crucial to know whether all work in the cycle will complete before the cycle repeats, because this 'slip' indicates a failure to meet the cycle’s hard deadline. For example, a 10 Hz control update must happen every 100 ms. If one iteration takes longer than 100 ms, system control has been compromised. For this reason, cyclic work is almost always performed via synchronous invocations and therefore uses a sync port.
```
thomas-bc: > If one iteration takes longer than 100 ms, system control has been compromised. For this reason, cyclic work is almost always performed via synchronous invocations

It's not obvious to me why A implies B here. My educated guess is that this is the simplest way to know that there has been a cycle slip? Might be worth calling out.

docs/user-manual/framework/component-and-port-selection.md
```diff
@@ -0,0 +1,236 @@
...
+    end
+```
+**Figure 1**: a rate group driven passive component that may call another component as part of its cyclic execution.
+
+> [!CAUTION]
+> This simple `passive` pattern breaks down when the cyclic component needs to accept some Event-Driven work (e.g. it processes some commands). This use case is described in the next section.
+
+### Queued Components for Cyclic Work
+
+When a component performing cyclic work also needs to accept some Event-Driven work, we require a queue to handle the asynchronous invocations, but adding a queue processing thread may disrupt the critical synchronous invocations of the core cyclic work of the component.  For this exact reason, we use a `queued` component. A `queued` component allows asynchronous events to be accepted while keeping a synchronous core.  In this model, the component dispatches the queue as part of the primary synchronous invocation (i.e. `Svc.Sched` handler) thereby moving the asynchronous work into the cycle.
```
thomas-bc, line 81: Queued components are hard and this may be worth elaborating a bit.

Why is it not ok to have a passive component do cyclic work synchronously and handle commands synchronously as well?
Why is adding a thread disruptive?

I'm guessing it's because otherwise you may have threads race for some state variables... but what if you don't, or if it's simple enough that you're ok adding a Mutex around it?

docs/user-manual/framework/component-and-port-selection.md
```diff
@@ -0,0 +1,236 @@
...
+    end
+```
+**Figure 2**: a rate group driven queued component that dispatches asynchronous events as part of its cyclic execution.
+
+> [!IMPORTANT]
+> In this model, it is **imperative** that you dispatch the queue in some synchronous implementation (i.e. the `Svc.Sched` handler); otherwise, the queue will fill but events will never process.
+
+## Event-Driven Work
+
+Since Event-Driven work is typically high-priority but without strict hard deadlines, this work is typically done via asynchronous invocations and uses the `async` port kind.  Since the component lacks another context to run in, we use an `active` component to dispatch the asynchronous work.
```
thomas-bc, line 115: Is sync Event-Driven work not recommended? I'm not seeing any reference to that in the doc. Could be worth calling out with a sentence or two if there are scenarios where this is acceptable.

docs/user-manual/framework/component-and-port-selection.md
```diff
@@ -0,0 +1,236 @@
...
+This model is constructed by having any number of `async` ports and commands attached to an `active` component. The thread scheduler handles the rest.
+
+```mermaid
+sequenceDiagram
+    participant E as Event Source
+    participant C1 as Active Component
+
+
+    E-)C1: Asynchronous Invocation (e.g. command)
+    E-)C1: Asynchronous Invocation ...
```
thomas-bc, line 131: There's a few diagram where there's two "Asynchronous Invocation" arrows. Is this to indicate there may be multiple that just queue up? 
Could be worth saying `(command or port)` instead of `(e.g. command)` ?

docs/user-manual/framework/component-and-port-selection.md
```diff
@@ -0,0 +1,236 @@
...
+        C1->>+H2: Synchronous Invocation
+        H2->>H2: Perform some work
+        H2-->>C1:
+        deactivate H2
+        deactivate C1
+    end
+```
+**Figure 6**: an active component that dispatches events as part of its thread lifecycle using a series of passive helper components for a more nuanced decomposition.
+
+### Passive Converter Pattern
```
thomas-bc: Do we usually call this an Adapter rather than a Converter?

docs/user-manual/framework/component-and-port-selection.md
```diff
@@ -0,0 +1,236 @@
...
+    participant D as Destination Component 
+
+    S->>+C: Synchronous Invocation
+    C->>C: Perform conversion
+    C->>D: Synchronous/Asynchronous Invocation
+    deactivate C
+```
+**Figure 7**: a passive converter component that performs a conversion inline with a port call.
+
+## Conclusion
```
thomas-bc, line 247: Thoughts on {adding / asking AI to add} a table where it tries and find an example for some of those patterns in our core framework ? Could be worth asking, and deciding based on the results if it's helpful or not.

thomas-bc (commented):
This is awesommeeeeeeee 🚀  I have a lot of questions because this is a real eye opener on the "unknown unknowns", things you didn't know you wanted to know. 

Top tier document!

Even for an experienced developer I think it really helps to have them laid out like that. We should definitely present that at our workshops.

docs/user-manual/framework/component-and-port-selection.md
LeStarch: I don't follow this suggestion.

docs/user-manual/framework/component-and-port-selection.md
LeStarch: Yes, good call!

docs/user-manual/overview/03-port-comp-top.md
```diff
@@ -226,3 +232,5 @@ topology. Alternate implementations can therefore easily be swapped, for example
 
 F´ is decomposed into a federation of discrete components organized into a system graph. The interconnections in this
 graph are handled through port calls such that the otherwise independent components can interact with each other.
+
+For an understanding of how to select port and component kinds, see [Selecting Component, Port, and Command Kinds](../framework/component-and-port-selection.md). For an understanding of how to select command kinds, see [Command Types](../framework/command-types.md).
```
thomas-bc: > see [Command Types](../framework/command-types.md)

here's the dead link - probably hallucination

thomas-bc (approved):

Issue #5118, opened 2026-05-08
Missing bounds check before memcpy in TlmRecv_handler resulting in a potential buffer overflow


bitWarrior:
| | |
|:---|:---|
|**_F´ Version_**| 4.2.1 |
|**_Affected Component_**|  TlmPacketizer |
---
## Problem Description

There is no bounds checking in TlmPacketizer ::TlmRecv_handler.

Line: 189
 U8* ptr = &this->m_fillBuffers[pkt].buffer.getBuffAddr()[entry.packetOffset[pkt]];
 (void)memcpy(ptr, val.getBuffAddr(), static_cast<size_t>(val.getSize())); 

There is no copy bounds checking before the memcpy. If a misbehaving or misconfigured sender provides a value larger than entry.channelSize, this data overflows into adjacent channel data in the same buffer.

## Context / Environment

Execute `fprime-util version-check` and share the output.

Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-6.14.0-37-generic-x86_64-with-glibc2.39
Python version: 3.12.3
CMake version: 3.24.2
Pip version: 24.0
Pip packages:
    fprime-tools==3.6.1
    fprime-gds==3.6.1
    fprime-fpp-*==2.3.1

## Expected Behavior

There should be bounds checking before the memcpy to ensure a buffer overflow condition does not exist.

Saunders-Trinity:
looking to tackle this one soon. i think i understand how to fix this

thomas-bc:
Fixed by #5131

PR #5123, opened 2026-05-08
Update TlmPacketizer.cpp

Saunders-Trinity:
|:---|:---|
|**_Related Issue(s)_**|  |
Issue #5118 
no
no
yes(testing)
---
## Change Description
Added a bounds check and copy safe logic to help get buffer overflow issue solved,

## Rationale
buffer overflow can be prevented via having bounds checking, limiting the length of bytes, and input validation. I used the bounds check mainly as thats what i know how to do.

Check to see if a buffer overflow happens on your local machine or make a custom test case.

## Future Work
None that I know of.

Used to test for more buffer overflow given different inputs and making sure if given a out of bounds input it returned the correct output.


Svc/TlmPacketizer/TlmPacketizer.cpp
```diff
@@ -187,7 +187,14 @@ void TlmPacketizer ::TlmRecv_handler(const FwIndexType portNum,
             this->m_fillBuffers[pkt].updated = true;
             this->m_fillBuffers[pkt].latestTime = timeTag;
             U8* ptr = &this->m_fillBuffers[pkt].buffer.getBuffAddr()[entry.packetOffset[pkt]];
-            (void)memcpy(ptr, val.getBuffAddr(), static_cast<size_t>(val.getSize()));
+            const size_t srcSize = static_cast<size_t>(val.getSize());
```
LeStarch, line 190: We should use `FwSizeType` for these constants as that is the native type.

Svc/TlmPacketizer/TlmPacketizer.cpp
```diff
@@ -187,7 +187,14 @@ void TlmPacketizer ::TlmRecv_handler(const FwIndexType portNum,
...
+            const size_t dstSize = static_cast<size_t>(entry.channelSize);
+            const size_t copySize = (srcSize <= dstSize) ? srcSize : dstSize;
+            (void)memcpy(ptr, val.getBuffAddr(), copySize);
```
LeStarch, line 193: Two notes:
1. We should check for overflow of copySize >= size_t.max() to prevent overflow.
2. If copySize > dstSize then we should emit a warning lo and drop the data.  This version fixes the bad overflow but produces a corrupt data packet (incomplete).

LeStarch (changes requested):
A few hopefully minor changes.  Thanks for the contribution!

Svc/TlmPacketizer/TlmPacketizer.cpp
Saunders-Trinity: oh i didnt catch that data packet, makes sense!

LeStarch:
I am not seeing either fix, just a merge from devel...perhaps a commit wasn't pushed?

thomas-bc:
@Saunders-Trinity this specific issue has been fixed by https://github.com/nasa/fprime/pull/5131 with an assert. Thanks for your contribution! This would have been the right fix in different circumstances, but because we don't expect to ever fall in that scenario, and bounding the copy would just result in an invalid buffer, we prefer to assert here. Closing this PR

PR #5125, opened 2026-05-09
Banish Ref to a project directory

### LINKED ISSUE #5124: Move Ref into a Project Directory
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

`Ref` is the only deployment which is still permitted to use its directory as a both a deployment directory and a project directory whose root is outside the project itself.

We already decided to:
1. Remove the `RPI` project for this reason
2. Put `FppTest` into `FppTestProject` for this reason

**Now:** the winter winds of malcontent blow toward Ref.  It is time for it to move.

We will put Ref in a folder called `TestDeploymentsProject` so that it is clear that it is a deployment inside a project for the purposes of testing. 

## Rationale

Maintaining a structure where the project and the include root (formally build root) are inconsistent causes considerable extra development in both the CMake system and the `fprime-util` implementation.

Additionally, this structure is only recommended within fprime to support a now strongly discouraged pattern of making deployments directly in a fork of time.

Ref, the bell tolls for thee.

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  #5124 |

---
## Change Description

Banish Ref from the domain of core F Prime and thus resolve #5124.

## Rationale

Let the crimes of Ref be known far and wide:
- Ref is a deployment, yet is wears the mask of a project.
- Ref claims the root of F Prime core as its own root
- Ref laughs in the face of conformity with F Prime norms

The CMake system buckles under the weight of Ref's flagrant disregard for F Prime standard.  `fprime-util` works hard simply to support Ref's selfish uniqueness.  It is time to end Ref's disregard for good programing practice.

Make CI pass.

## Future Work

Strip the CMake system of years of buttressing to support Ref's crimes against F Prime.

None....yet.


LeStarch:
Note: this change will break much of CI, and those failures will be addressed as the CI system runs through.

kevin-f-ortega:
Banish it

TestDeploymentsProject/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+#
+# This is the project that contains the set of test deployments. This includes:
+# - The 'Ref' deployment, which is a historical reference for testing and demonstration
+####
+
+# Allow CMake 3.0 - 4.0 projects to set the VERSION variable in the `project` call
+cmake_policy(SET CMP0048 NEW)
+# CMake basic setup: version consistent with requirements.txt, and define a project 
+cmake_minimum_required(VERSION 3.26)
+project(Ref VERSION 1.0.0 LANGUAGES C CXX)
```
thomas-bc, line 12: Is the project name kept intentionally? Or should we swap to TestDeploymentsProject here?

thomas-bc (commented):
LGTM - one quick question

TestDeploymentsProject/CMakeLists.txt
LeStarch: It was kept intentionally for now, since changing it would change the name of the binary we run and look for everywhere.  Thus, I expect to change that in the future should we ever develop more test deployments.

thomas-bc (approved):

PR #5121, opened 2026-05-08
Fixing Overwrite issue

Saunders-Trinity:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
Issue #5092 
Issue #5099 
Issue #5119 

|**_Has Unit Tests (y/n)_**|  |
|**_Documentation Included (y/n)_**|  |
|**_Generative AI was used in this contribution (y/n)_**|  |
no

---
## Change Description
addresses mutliple issues.

## Rationale
preventing over writing of components without the proper process.

## Testing/Review Recommendations
check for UT failures again.

## Future Work
dp catalog seems to have alot of issues as a file, not sure how that will be addressed.

PR #5127, opened 2026-05-10
PR-5119 - Fixed issue where invalid SET_LEVEL is processed

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**|  PR 5119 |

---
## Change Description

Added command VALIDATION_ERROR response and return in parameter validation logic in SET_LEVEL_cmdHandler if l"evel > MAX_CONFIGURABLE_TLMPACKETIZER_GROUP"

## Rationale

An invalid SET_LEVEL should be rejected with error response to user.

Created new unit test to verify invalid SET_LEVEL is rejected with VALIDATION_ERROR.

## Future Work

N/A

Claude was used to generate new unit test

### DIFF
--- a/Svc/TlmPacketizer/test/ut/TlmPacketizerTester.cpp
+++ b/Svc/TlmPacketizer/test/ut/TlmPacketizerTester.cpp
@@ -1718,6 +1718,48 @@ void TlmPacketizerTester ::sectionConfigParameterTest() {
     this->paramSave_SECTION_CONFIGS(0, 0);
 }
 
+void TlmPacketizerTester::pushAllChannels() {
+    // channel 10 — packet1, packet2, packet4
+    buff.resetSer();
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, buff.serializeFrom(static_cast<U32>(0xAB)));
+    this->invoke_to_TlmRecv(0, 10, ts, buff);
+
+    // channel 100 — packet1
+    buff.resetSer();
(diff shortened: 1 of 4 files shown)


Svc/TlmPacketizer/test/ut/TlmPacketizerTester.cpp
```diff
@@ -1718,6 +1718,140 @@ void TlmPacketizerTester ::sectionConfigParameterTest() {
...
+    this->stockConfiguration();
+    // packetList2 contains four packets at levels 1, 2, 2, 3
+    // (packet1 id=4, packet2 id=8, packet3 id=12, packet4 id=16)
+    this->component.setPacketList(packetList2, IGNORE_OMIT_LIST, 3);
+
+    Fw::Time ts;
+    Fw::TlmBuffer buff;
+
+    // Helper lambda to push one value to every channel used by packetList2
+    auto pushAllChannels = [&]() {
```
LeStarch: Don't use lambdas even in test code.  Please take the time to move this into a helper function.  It is easier for our maintainers who avoid lambdas in our flight code.

LeStarch (changes requested):
Refactoring change for the UT requested.

LeStarch:
Note: this one also collides with active work in another PR: #5121

Since this one is dedicated, and has a UT, I am want to accept this over #5121, but we need to remove the lambda first.

LeStarch (approved):

PR #5131, opened 2026-05-10
Fix memcpy without bounds checking (#5118)

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**| PR 5118 |
|**_Generative AI was used in this contribution (y/n)_**|  N |

---
## Change Description

Added bound check before memcpy

## Rationale

Bounds checking should be performed before a memcpy to avoid a potential buffer overflow.

Ironically, the existing unit test for TlmPacketizer has a bug where it was overwriting the buffer. Because there was no bound checking, the unit tests erroneously passed. When I implemented this change, the unit test failed (correctly) because it was overflowing the buffer. 

## Future Work

N/A

No


LeStarch (approved):

LeStarch:
This looks like the change I expected.  However, note, you are working on the same change as is seen in #5123.  We really should check open PRs before submitting competing work.

thomas-bc (approved):

Issue #5122, opened 2026-05-08
Add ability to limit path in assert to the last N characters of the file path


bevinduckett:
| | |
|:---|:---|
|**_F´ Version_**| devel |
|**_Affected Component_**| AssertAdaptor |
---
## Feature Description

Ability to specify a max character length, N, for the length of the file path string included in assert log messages. File  paths longer than N characters should be trimmed so that the last N characters (and not the first N) are reported in the assert message. 

The last N characters should be reported because they contain the most unique information (the first part will likely always be the path to the sandbox)

One option would be to use the same max size specified by AssertFatalAdapterEventFileSize (pro: single variable to set both, con: the name is misleading) . Recommendations on that would be appreciated

## Rationale
There's an assert in a sandbox I'm working on that is missing the required context information - all I can tell from the log message is that the assert is happening in an auto-generated file related to CmdDispatcher. 
```
0xffff800001128a30 (rateGroup3Comp): Assert: "<PATH_TO_MY_DEVELOPMENT_SANDBOX>/build-fprime-automatic-<PLATFORM>/F-Prime/Svc/CmdDispatcher/Command0xffff800001128a30 (rateGroup3Comp): FATAL 16797697 handled.
```
I could switch to using file hashes to get that information, but then there's an extra step in investigating the assert and I'd lose the current ability to capture the full context of the assert from the log file

bevinduckett:
fyi - @vincewoo

PR #5141, opened 2026-05-12
Truncate assert path

bevinduckett:
| | |
|:---|:---|
|**_Related Issue(s)_**|  #5122 |

---
## Change Description

1. Added the file path leading-truncation for the log() message from reportAssert() and added unit test coverage. 
2. Updated the truncation for the file path from the event argument to be leading-truncation (not trailing truncation)
3. Added custom config to the unit test so that file path truncation would be tested (custom configuration)
4. Updated the default value for AssertFatalAdapterEventFileSize  to be significantly less than the default value for FwAssertTextSize

## Rationale

See feature rationale in #5122

## Future Work

We could consider adding a function to StringUtils to do leading-truncation 

AI (Claude code, Sonnet 4.5) was used to assist with the unit test updates


bevinduckett:
Please use squash & merge when merging this PR (since intermediate commits don't all work)

thomas-bc:
I really like and want this feature, but since this implementation adds a copy in the critical assert-reporting code path, we'll need to get buy off from CCB

Svc/AssertFatalAdapter/AssertFatalAdapterComponentImpl.cpp
```diff
@@ -89,15 +90,30 @@
                                                    FwAssertArgType arg4,
                                                    FwAssertArgType arg5,
                                                    FwAssertArgType arg6) {
+
+    constexpr FwSizeType outputSize = FW_MIN(static_cast<FwSizeType>(AssertFatalAdapterEventFileSize),
+                                                    static_cast<FwSizeType>(FW_LOG_STRING_MAX_SIZE));
+    char output[outputSize];
```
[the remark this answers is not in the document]
bevinduckett: Removed output variable, so this finding is obsolete

Svc/AssertFatalAdapter/AssertFatalAdapterComponentImpl.cpp
```diff
@@ -89,15 +90,30 @@
...
+    FwSizeType len = Fw::StringUtils::string_length(file, FileNameStringSize);
+
+    // Calculate start index. If string is shorter than N, keep whole string.
+    FW_ASSERT(file != nullptr);
+    const char* start = (len > outputSize) ? file + (len - outputSize) : file;
```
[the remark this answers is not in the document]
bevinduckett: start has been updated to use CHAR (same as the msg variable), so this finding is ODD

Svc/AssertFatalAdapter/AssertFatalAdapterComponentImpl.cpp
```diff
@@ -89,15 +90,30 @@
                                                    FwAssertArgType arg4,
                                                    FwAssertArgType arg5,
                                                    FwAssertArgType arg6) {
+
+    constexpr FwSizeType outputSize = FW_MIN(static_cast<FwSizeType>(AssertFatalAdapterEventFileSize),
+                                                    static_cast<FwSizeType>(FW_LOG_STRING_MAX_SIZE));
+    char output[outputSize];
+    FwSizeType len = Fw::StringUtils::string_length(file, FileNameStringSize);
```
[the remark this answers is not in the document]
bevinduckett: parameter file will be checked by the string_length() function - `file` is checked in reportAssert() before being used

Svc/AssertFatalAdapter/AssertFatalAdapterComponentImpl.cpp
```diff
@@ -89,15 +90,30 @@
...
+
+    // Calculate start index. If string is shorter than N, keep whole string.
+    FW_ASSERT(file != nullptr);
+    const char* start = (len > outputSize) ? file + (len - outputSize) : file;
+
+    // Copy safely into output buffer (also ensures null-termination)
+    Fw::StringUtils::string_copy(output, start, outputSize + 1);
```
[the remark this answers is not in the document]
bevinduckett: OOD - string copy operation removed

bevinduckett:
@thomas-bc  - could you kick off the workflow checks again? I pushed an update to fix an UT failure (that update also removed the string copy since its primary purpose was ensuring the null character existed which the previous implementation treated as assumed)

Svc/AssertFatalAdapter/CMakeLists.txt
```diff
@@ -6,21 +6,21 @@
 #
 # Note: using PROJECT_NAME as EXECUTABLE_NAME
 ####
-
```
LeStarch, line 9: I appreciate you reworking this into the new structure!

Svc/AssertFatalAdapter/AssertFatalAdapterComponentImpl.cpp
```diff
@@ -89,15 +90,25 @@ void AssertFatalAdapterComponentImpl::reportAssert(FILE_NAME_ARG file,
                                                    FwAssertArgType arg4,
                                                    FwAssertArgType arg5,
                                                    FwAssertArgType arg6) {
+    constexpr FwSizeType outputSize = FW_MIN(static_cast<FwSizeType>(AssertFatalAdapterEventFileSize),
+                                             static_cast<FwSizeType>(FW_LOG_STRING_MAX_SIZE));
+    FwSizeType len = Fw::StringUtils::string_length(file, FileNameStringSize);
```
LeStarch: Unfortunately this code won't work.  This code assumes file is a string, but it is only guaranteed to be a string within the lines 105-105 (e.g. not the ID case).

Alll this string processing should be moved into the else clause there.

Svc/AssertFatalAdapter/AssertFatalAdapterComponentImpl.cpp
```diff
@@ -89,15 +90,25 @@ void AssertFatalAdapterComponentImpl::reportAssert(FILE_NAME_ARG file,
                                                    FwAssertArgType arg4,
                                                    FwAssertArgType arg5,
                                                    FwAssertArgType arg6) {
+    constexpr FwSizeType outputSize = FW_MIN(static_cast<FwSizeType>(AssertFatalAdapterEventFileSize),
```
LeStarch: Recommendation is to switch to `std::min` now, `FW_MIN` remains for backwards compatibility.

Svc/AssertFatalAdapter/AssertFatalAdapterComponentImpl.cpp
```diff
@@ -89,15 +90,25 @@ void AssertFatalAdapterComponentImpl::reportAssert(FILE_NAME_ARG file,
...
+                                             static_cast<FwSizeType>(FW_LOG_STRING_MAX_SIZE));
+    FwSizeType len = Fw::StringUtils::string_length(file, FileNameStringSize);
+
+    // Calculate start index. If string is shorter than N, keep whole string.
+    FW_ASSERT(file != nullptr);
```
LeStarch: Should check != nullptr before string length (although it likely checks itself)

default/config/AcConstants.fpp
```diff
@@ -54,11 +54,13 @@ constant FileNameStringSize = 240
...
+@ The size of a file name in an AssertFatalAdapter event (leading-truncation)
 @ Note: File names in assertion failures are also truncated by
-@ the constants FW_ASSERT_TEXT_SIZE and FW_LOG_STRING_MAX_SIZE, set
-@ in FpConfig.h.
-constant AssertFatalAdapterEventFileSize = FileNameStringSize
+@ the constants FwAssertTextSize (in this file) and FW_LOG_STRING_MAX_SIZE (set
+@ in FW_LOG_STRING_MAX_SIZE)
+@ Set much smaller than FwAssertTextSize so there's space for time stamp/assert 
+@ arguments in log message
+constant AssertFatalAdapterEventFileSize = 150
```
LeStarch: We should restor this to the default of FwFileStringSize

LeStarch (changes requested):
I think the changes will work, however; they need to be in the branch where FILE_ID_TYPE is a string.

thomas-bc:
@lestarch-autobot let the showdown begin!

lestarch-autobot (commented):
Leading truncation preserves the filename portion which is more useful for debugging. No security concerns.

Svc/AssertFatalAdapter/AssertFatalAdapterComponentImpl.cpp
```diff
@@ -89,15 +90,25 @@ void AssertFatalAdapterComponentImpl::reportAssert(FILE_NAME_ARG file,
                                                    FwAssertArgType arg4,
                                                    FwAssertArgType arg5,
                                                    FwAssertArgType arg6) {
+    constexpr FwSizeType outputSize = FW_MIN(static_cast<FwSizeType>(AssertFatalAdapterEventFileSize),
+                                             static_cast<FwSizeType>(FW_LOG_STRING_MAX_SIZE));
+    FwSizeType len = Fw::StringUtils::string_length(file, FileNameStringSize);
+
+    // Calculate start index. If string is shorter than N, keep whole string.
```
lestarch-autobot: **must fix** The null-pointer assertion is placed *after* the first dereference of .



Move the assert before the  call so the precondition is checked before any use:

default/config/AcConstants.fpp
lestarch-autobot: **could fix** The updated comment reads:
> the constants FwAssertTextSize (in this file) and FW_LOG_STRING_MAX_SIZE (set in FW_LOG_STRING_MAX_SIZE)

The parenthetical "(set in FW_LOG_STRING_MAX_SIZE)" is self-referential. It should indicate where the constant is defined, e.g. "(set in FpConfig.h)" or just omit the parenthetical entirely.

lestarch-autobot (commented):
Leading truncation is the right design: filenames (suffixes) carry more diagnostic value than directory prefixes. Reducing AssertFatalAdapterEventFileSize from 240 to 150 leaves adequate room for timestamps and assert arguments in the event string. CMakeLists modernization to keyword-based registration is consistent with current F Prime conventions.

Svc/AssertFatalAdapter/test/ut/AssertFatalAdapterTester.cpp
```diff
@@ -138,6 +148,94 @@ void AssertFatalAdapterTester::testAsserts() {
...
+    this->clearTextLogs();
+    this->component.reportAssert(exactPath, 200, 0, 0, 0, 0, 0, 0, 0);
+    ASSERT_EVENTS_AF_ASSERT_0_SIZE(1);
+    ASSERT_EVENTS_AF_ASSERT_0(0, exactPath, 200);
+    // Verify text log contains the exact path
+    const std::string& textLog2 = this->getLastTextLog();
+    ASSERT_NE(textLog2.find(exactPath), std::string::npos) << "Text log should contain exact-length path";
+    this->clearHistory();
+
+    // Test 3: Long path should be truncated (keeps trailing portion)
```
lestarch-autobot, line 181: **suggestion** These debug printf statements will pollute test runner output on every CI run. Consider removing them or gating behind a verbose flag before merge.



Good test coverage otherwise: tests short path (unchanged), exact-limit path, over-limit path with suffix verification, and multi-arg assert variant. The fillWithBytePattern helper creates identifiable patterns that make failures easy to diagnose.

lestarch-autobot (changes requested):

### Per-agent results

| F Prime C/C++ Design | 1 | 0 | 0 | 0 | 1 | No-Go |
| Documentation Currency | 0 | 0 | 1 | 0 | 0 | Go |
| **Totals** | 1 | 1 | 1 | 0 | 1 | **No-Go** |

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>

**F Prime C/C++ Design**
- Null-pointer assertion placed after first dereference of  in . The  must precede the  call. — [inline comment](https://github.com/nasa/fprime/pull/5141)

</details>

### Merge readiness
**Merge readiness: No-Go** — One must-fix: null check ordering bug in reportAssert (assert fires after the pointer is already dereferenced). Additionally: minor circular comment in AcConstants.fpp and debug printf in test code.

---

Good feature: leading truncation of assert file paths preserves the diagnostically-valuable filename suffix. The reduced buffer (150 vs 240) leaves room for timestamps/args. Well-tested with 4 distinct coverage scenarios. The CMakeLists modernization is a bonus cleanup. Just needs the ordering fix.

bocchino (commented):
This PR looks good! Having leading truncation for file paths will really help.

Instead of implementing leading truncation in the AssertFatalAdapter, I think it would be better to do as suggested in the PR comment and add it to the interfaces for StringUtils and the String classes. The char array manipulation is low-level and ideally should be encapsulated in the utility layer. Also, that way the new feature will be reusable.

LeStarch (dismissed):

LeStarch (dismissed):
Ready!

LeStarch (dismissed):

thomas-bc (dismissed):

thomas-bc (approved):

PR #5146, opened 2026-05-12
fix(FpySequencer): convert reachable asserts to recoverable errors

datqm:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**| y (existing 234 FpySequencerTester tests retained, all pass) |
|**_Generative AI was used in this contribution (y/n)_**| y (code generation, testing)|

---
## Change Description

Convert 12 reachable `FW_ASSERT`s in `Svc/FpySequencer` to recoverable errors so a malformed `.fpy` file or a transient OS file-API failure no longer aborts the FSW process.

**Sites converted:**

`FpySequencerStack.cpp` (8 sites):
| Method | Failure mode |
|---|---|
| `Stack::pop<T>()` | underflow on pop |
| `Stack::push<T>()` | overflow on push |
| `Stack::pop(U8*, size)` | underflow on bulk pop |
| `Stack::push(const U8*, size)` | overflow on bulk push |
| `Stack::pushZeroes()` | overflow on zero fill |
| `Stack::copy()` | dest bounds, src bounds, overlap (3 asserts) |
| `Stack::move()` | dest bounds, src bounds (2 asserts) |

`FpySequencerValidationState.cpp` (4 sites):
| Call site | Asserted condition |
|---|---|
| `validate()` | `Os::File::size()` returned `OP_OK` |
| `validate()` | `Os::File::position()` returned `OP_OK` |
| `readBytes()` | `expectedReadLen == actualReadLen` |
| `readBytes()` | `setBuffLen()` returned `FW_SERIALIZE_OK` |

**Approach for the stack sites:**

The Stack methods are called from 25+ sites in `FpySequencerDirectives.cpp`, so changing their signatures to return an error code would balloon the diff. Instead, add a single `bool hasBoundsError` flag on the `Stack` class. Each bounds- or overlap-violating op sets the flag and skips the underlying `memcpy`/`memset`. Value-returning `pop<T>()` returns a zero-initialized `T` so callers continue without UB. `dispatchStatement()` checks `stack.hasError()` at the top of the loop, emits `StackBoundsError(statementIdx)`, clears the flag, and returns `result_dispatchStatement_failure` so the sequence aborts cleanly before the next statement runs.

**Approach for the validation sites:**

`validate()` and `readBytes()` already have an `Fw::Success::FAILURE` return path. The four asserts are replaced with `log_WARNING_HI_FileApiError(filePath, stage)` + early `return Fw::Success::FAILURE`, where `stage` is a new `Fpy::FileApiStage` enum (`SIZE`, `POSITION`, `READ_LENGTH_MISMATCH`, `SET_BUFFER_LENGTH`) so the operator can tell which OS call failed.

**New telemetry (FPP):**

```fpp
enum FileApiStage : U8 {
    SIZE = 0
    POSITION = 1
    READ_LENGTH_MISMATCH = 2
    SET_BUFFER_LENGTH = 3
}

event StackBoundsError(statementIdx: U32) \
    severity warning high \
    format "Stack bounds violation while executing statement {}; sequence aborted"

event FileApiError(filePath: string size 60, stage: Svc.Fpy.FileApiStage) \
    severity warning high \
    format "Sequence {} validation failed at OS file API stage {}"
```

## Rationale

Each of the 12 `FW_ASSERT`s aborts the FSW process when it fires. They are reachable from:

1. **A malformed `.fpy` file.** The bytecode interpreter does not pre-validate that the offsets a `CALL`/`RETURN` frame, `STORE_REL_CONST_OFFSET`, `MEMCMP`, or stack-arithmetic directive will ask the stack to touch are within bounds. A hand-crafted sequence can request a `pop` larger than the current stack size or a `push` past `MAX_STACK_SIZE`. Any operator with command authority (or anyone who can get a sequence file uplinked) can take the spacecraft down with a sub-100-byte file.

2. **A transient OS file error.** `Os::File::size()` / `position()` can fail under concurrent `FileManager` operations on the same path, or on a real filesystem reporting `EIO`. The assumption that these calls always return `OP_OK` is incorrect on a live spacecraft.

This is the same bug class as PR #5136 (`CmdSequencer` `loadFile` reachable assert) which was accepted on `devel`. The fix follows the same minimum-invasive pattern: keep the public API stable, route the error to the existing `Fw::Success::FAILURE` / dispatcher-fail path, surface a telemetry event so the operator can tell what happened.

**Local verification done:**
- `fprime-util check` on `Svc/FpySequencer` → 234/234 unit tests pass (same as pre-patch).
- `fprime-util format --check --dirs Svc/FpySequencer` → exit 0.

**Suggested review focus:**
1. `FpySequencerRunState.cpp:21-30` — the new `stack.hasError()` check at the top of `dispatchStatement()`. Confirm the placement is correct: it runs *before* the next statement is fetched, so an erroneous op from the previous dispatch is always caught before any further reads.
2. `FpySequencerStack.cpp::pop<T>()` — returning `T{}` on underflow is the only place a "fake" value flows back into the interpreter. The dispatcher catches the error on the next call, so the only thing the caller does with that zero is potentially push it onto the stack again (which can re-flag the error, harmless) or feed it into a directive that runs to completion within the same dispatch. No directive does anything I/O-visible without going back through `dispatchStatement()` first.
3. The new `Fpy::FileApiStage` enum and the two new events are intentionally narrow — they tell the operator which OS call failed without leaking implementation details.

**Suggested additional testing the maintainer may want to run:**
- A negative test that constructs a `.fpy` file with a `MEMCMP` directive whose `size` field exceeds the current stack contents, runs it through `FpySequencer`, and asserts that the sequence ends in failure with `StackBoundsError` rather than crashing the test harness. I did not add this here because the existing `FpySequencerTester` harness asserts on `FW_ASSERT` via a death-test-style mechanism that would need adjustment; happy to add it in a follow-up if you want.

## Future Work

- Add the negative-path unit tests described above (one per Stack bounds class plus one for each file API stage).
- Consider whether the same defensive pattern should apply to the offset arithmetic *inside* directives (e.g., `LOAD_REL` `frameStart + offset` overflow) so the dispatcher catches the error before the Stack method is even called — currently those paths can still emit `FrameStartOutOfBounds` / `StackAccessOutOfBounds` `DirectiveErrorCode`s, which is fine, but worth confirming there is no remaining `FW_ASSERT` in that chain.

### DIFF
--- a/Svc/FpySequencer/FpySequencer.hpp
+++ b/Svc/FpySequencer/FpySequencer.hpp
@@ -87,6 +87,21 @@ class FpySequencer : public FpySequencerComponentBase {
         // the byte offset from the start of the stack where the current function's local variables begin.
         // analogous to a 'frame pointer'.
         Fpy::StackSizeType currentFrameStart = 0;
+        // Set true by any stack operation that would violate stack bounds
+        // (overflow on push, underflow on pop, or out-of-range offset on
+        // copy/move). The operation is skipped when set, and the executing
+        // sequence is aborted by the dispatcher on the next dispatchStatement
+        // call. This replaces an FW_ASSERT that an attacker-controlled
+        // .fpy bytecode could trigger to crash the FSW process.
+        bool hasBoundsError = false;
+
+        // True iff a stack operation since the last clearError() would have
+        // violated the stack bounds.
+        bool hasError() const { return this->hasBoundsError; }
+
+        // Clear the error flag (called by the dispatcher after surfacing
+        // the failure to the operator).
+        void clearError() { this->hasBoundsError = false; }
 
         // pops a value off of the top of the stack
         // converts it from big endian
--- a/Svc/FpySequencer/FpySequencerRunState.cpp
+++ b/Svc/FpySequencer/FpySequencerRunState.cpp
@@ -18,6 +18,18 @@ Signal FpySequencer::dispatchStatement() {
     // as that indicates eof
     FW_ASSERT(this->m_runtime.nextStatementIndex <= this->m_sequenceObj.get_header().get_statementCount());
 
+    // If the previous statement caused a stack bounds violation (e.g. a
+    // malformed .fpy bytecode pushed past MAX_STACK_SIZE), the Stack class
+    // skipped the bad op and set hasBoundsError. Abort the sequence here
+    // before the next statement runs, rather than letting downstream code
+    // execute on a corrupted stack. Previously the offending op was an
+    // FW_ASSERT that aborted the FSW process.
+    if (this->m_runtime.stack.hasError()) {
+        this->log_WARNING_HI_StackBoundsError(this->currentStatementIdx());
+        this->m_runtime.stack.clearError();
+        return Signal::result_dispatchStatement_failure;
+    }
+
     if (this->m_runtime.nextStatementIndex == this->m_sequenceObj.get_header().get_statementCount()) {
         return Signal::result_dispatchStatement_noMoreStatements;
     }
--- a/Svc/FpySequencer/FpySequencerStack.cpp
+++ b/Svc/FpySequencer/FpySequencerStack.cpp
@@ -5,8 +5,14 @@ namespace Svc {
 template <typename T>
 T FpySequencer::Stack::pop() {
     static_assert(sizeof(T) == 8 || sizeof(T) == 4 || sizeof(T) == 2 || sizeof(T) == 1, "size must be 1, 2, 4, 8");
-    FW_ASSERT(this->size >= sizeof(T), static_cast<FwAssertArgType>(this->size),
-              static_cast<FwAssertArgType>(sizeof(T)));
+    // Underflow guard: a malformed .fpy bytecode can request more bytes
+    // than the stack currently holds. Flag the error and return a zero
+    // value rather than asserting; the dispatcher aborts the sequence
+    // before the next statement runs.
+    if (this->size < sizeof(T)) {
+        this->hasBoundsError = true;
+        return T{};
+    }
     // first make a byte array which can definitely store our val
     U8 valBytes[8] = {0};
     // now move top of stack into byte array and shrink stack
@@ -57,8 +63,13 @@ F64 FpySequencer::Stack::pop<F64>() {
 template <typename T>
 void FpySequencer::Stack::push(T val) {
     static_assert(sizeof(T) == 8 || sizeof(T) == 4 || sizeof(T) == 2 || sizeof(T) == 1, "size must be 1, 2, 4, 8");
-    FW_ASSERT(this->size + sizeof(val) <= Fpy::MAX_STACK_SIZE, static_cast<FwAssertArgType>(this->size),
-              static_cast<FwAssertArgType>(sizeof(T)));
+    // Overflow guard: skip the write if it would exceed MAX_STACK_SIZE
+    // and let the dispatcher abort the sequence cleanly on the next
+    // statement. Previously an FW_ASSERT here aborted the FSW process.
+    if (this->size + sizeof(val) > Fpy::MAX_STACK_SIZE) {
+        this->hasBoundsError = true;
+        return;
+    }
     // first make a byte array which can definitely store our val
     U8 valBytes[8] = {0};
     // convert val to unsigned to avoid undefined behavior for bitshifts of signed types
@@ -114,7 +125,13 @@ void FpySequencer::Stack::push<F64>(F64 val) {
 // pops a byte array from the top of the stack into the destination array
 // does not convert endianness
 void FpySequencer::Stack::pop(U8* dest, Fpy::StackSizeType destSize) {
-    FW_ASSERT(this->size >= destSize, static_cast<FwAssertArgType>(this->size), static_cast<FwAssertArgType>(destSize));
+    if (this->size < destSize) {
+        this->hasBoundsError = true;
+        if (destSize > 0 && dest != nullptr) {
+            memset(dest, 0, destSize);
+        }
+        return;
+    }
     memcpy(dest, this->top() - destSize, destSize);
     this->size -= destSize;
 }
@@ -123,16 +140,20 @@ void FpySequencer::Stack::pop(U8* dest, Fpy::StackSizeType destSize) {
 // leaves the source array unmodified
 // does not convert endianness
 void FpySequencer::Stack::push(const U8* src, Fpy::StackSizeType srcSize) {
-    FW_ASSERT(this->size + srcSize <= Fpy::MAX_STACK_SIZE, static_cast<FwAssertArgType>(this->size),
-              static_cast<FwAssertArgType>(srcSize));
+    if (this->size + srcSize > Fpy::MAX_STACK_SIZE) {
+        this->hasBoundsError = true;
+        return;
+    }
     memcpy(this->top(), src, srcSize);
     this->size += srcSize;
 }
 
 // pushes zero bytes to the stack
 void FpySequencer::Stack::pushZeroes(Fpy::StackSizeType byteCount) {
-    FW_ASSERT(this->size + byteCount <= Fpy::MAX_STACK_SIZE, static_cast<FwAssertArgType>(this->size),
-              static_cast<FwAssertArgType>(byteCount));
+    if (this->size + byteCount > Fpy::MAX_STACK_SIZE) {
+        this->hasBoundsError = true;
+        return;
+    }
     memset(this->top(), 0, byteCount);
     this->size += byteCount;
 }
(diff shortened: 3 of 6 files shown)


thomas-bc:
FYI, see https://github.com/nasa/fprime/pull/5145#issuecomment-4456195332

It appears there is no human behind this user's interactions. I will close the other PR. Feel free to handle this one as you see fit.

Svc/FpySequencer/FpySequencerRunState.cpp
```diff
@@ -18,6 +18,18 @@ Signal FpySequencer::dispatchStatement() {
     // as that indicates eof
     FW_ASSERT(this->m_runtime.nextStatementIndex <= this->m_sequenceObj.get_header().get_statementCount());
 
+    // If the previous statement caused a stack bounds violation (e.g. a
+    // malformed .fpy bytecode pushed past MAX_STACK_SIZE), the Stack class
+    // skipped the bad op and set hasBoundsError. Abort the sequence here
+    // before the next statement runs, rather than letting downstream code
```
lestarch-autobot, line 24: **could fix** After a stack bounds violation is flagged, remaining stack operations within the *same statement* continue executing (each checks its own bounds independently but does not short-circuit on the prior error). This means a pop after a failed push returns zero, and downstream logic within that statement may use the garbage value before the dispatcher catches the flag at the next statement boundary.

The window is bounded (one statement), and no memory corruption occurs because each op independently validates bounds. However, for defense-in-depth, consider adding an early-return check at the top of each stack operation:

This would halt execution within the current statement immediately rather than running on stale/zero data.

lestarch-autobot (commented):
All C/C++ patterns are sound. Error-flag approach is clean: bounds checks precede every memcpy/memmove, inline member initializer for hasBoundsError, const-correct accessor. No CPP rule violations.

lestarch-autobot (commented):
New events are self-documented via FPP annotations. No stale documentation surfaces identified.

lestarch-autobot (commented):
Design is sound: deferred error-flag checked at statement dispatch boundary is a well-known pattern (similar to IEEE 754 sticky exception flags). The FileApiStage enum provides good diagnostic disambiguation. Converting ground-reachable asserts to recoverable errors aligns with F Prime safety philosophy.

Svc/FpySequencer/FpySequencerStack.cpp
```diff
@@ -5,8 +5,14 @@ namespace Svc {
...
     static_assert(sizeof(T) == 8 || sizeof(T) == 4 || sizeof(T) == 2 || sizeof(T) == 1, "size must be 1, 2, 4, 8");
-    FW_ASSERT(this->size >= sizeof(T), static_cast<FwAssertArgType>(this->size),
-              static_cast<FwAssertArgType>(sizeof(T)));
+    // Underflow guard: a malformed .fpy bytecode can request more bytes
+    // than the stack currently holds. Flag the error and return a zero
+    // value rather than asserting; the dispatcher aborts the sequence
+    // before the next statement runs.
+    if (this->size < sizeof(T)) {
+        this->hasBoundsError = true;
+        return T{};
```
lestarch-autobot, line 14: **must fix** New security-critical error paths lack test coverage.

This PR introduces recoverable error handling for 8+ previously-asserting code paths (stack overflow, underflow, out-of-range copy/move, file size/position failures, read-length mismatch, setBuffLen failure). None of these new error paths appear to have corresponding unit tests in this PR.

For security-hardening changes, test coverage is essential to verify:
1. The error flag is set correctly on each violation type
2. The dispatcher aborts the sequence on the next statement
3. The events are emitted with correct parameters
4. No memory corruption occurs (pop returns zero, push is skipped)

Expected test cases:
- Push beyond MAX_STACK_SIZE → hasBoundsError set, push skipped
- Pop from empty stack → hasBoundsError set, returns zero
- Copy/move with overlapping or out-of-bounds regions → hasBoundsError set
- dispatchStatement after bounds error → StackBoundsError event, failure signal
- File size/position API failure → FileApiError event, validation failure

lestarch-autobot (changes requested):

### Per-agent results

| Security Vulnerabilities | 0 | 0 | 1 | 0 | 0 | Go |
| Test Quality | 1 | 0 | 0 | 0 | 1 | No-Go |
| **Totals** | 1 | 0 | 1 | 0 | 1 | **No-Go** |

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>

**Test Quality**
- New security-critical error paths (stack bounds violations, file API failures) introduced by this PR have no corresponding unit test coverage. — [inline comment](https://github.com/nasa/fprime/pull/5146#pullrequestreview-4324101384)

</details>

### Merge readiness
**Merge readiness: No-Go** — Test Quality agent has 1 outstanding must-fix: the 8+ new recoverable error paths need unit tests to confirm correct flag behavior, event emission, and absence of memory corruption.

---

Excellent security hardening work — converting ground-reachable FW_ASSERTs to recoverable errors is exactly the right pattern for flight software handling untrusted input (.fpy bytecode). The implementation is clean and the deferred-error-flag design is sound. The only blocker is test coverage for the new paths.

Svc/FpySequencer/FpySequencerStack.cpp
```diff
@@ -114,7 +125,13 @@ void FpySequencer::Stack::push<F64>(F64 val) {
...
 void FpySequencer::Stack::pop(U8* dest, Fpy::StackSizeType destSize) {
-    FW_ASSERT(this->size >= destSize, static_cast<FwAssertArgType>(this->size), static_cast<FwAssertArgType>(destSize));
+    if (this->size < destSize) {
+        this->hasBoundsError = true;
+        if (destSize > 0 && dest != nullptr) {
+            memset(dest, 0, destSize);
+        }
+        return;
+    }
     memcpy(dest, this->top() - destSize, destSize);
```
LeStarch, line 135: There isn't a null pointer check here either....although that was preexisting.

Svc/FpySequencer/FpySequencerStack.cpp
```diff
@@ -142,34 +163,44 @@ U8* FpySequencer::Stack::top() {
...
+    }
+    if (srcOffset + copySize > Fpy::MAX_STACK_SIZE) {
+        this->hasBoundsError = true;
+        return;
+    }
     // Check for overlap: regions overlap if one starts before the other ends
     // No overlap if: destEnd <= srcStart OR srcEnd <= destStart
-    FW_ASSERT(copySize == 0 || (destOffset + copySize <= srcOffset) || (srcOffset + copySize <= destOffset),
-              static_cast<FwAssertArgType>(destOffset), static_cast<FwAssertArgType>(srcOffset));
+    if (!(copySize == 0 || (destOffset + copySize <= srcOffset) || (srcOffset + copySize <= destOffset))) {
```
LeStarch, line 181: This seems quite complicated.  It is easy to review now because the old code is here...but not so much after this is merged.

LeStarch (commented):
Some nits.

Svc/FpySequencer/FpySequencerRunState.cpp
LeStarch: I don't think this is real because a push or pop is exactly one statement.

zimri-leisher:
I'm on vacation without my laptop so don't really feel comfortable fully reviewing this. But the initial PR doesn't pass the smell test. The directive handlers check whether stack accesses are within bounds before calling push/pop. They fail gracefully by ending the sequence with an error code if the accesses are out of bounds. How are the push/pop asserts reachable?? If they are reachable, then the fix is to have the directive handler check before pushing or popping, leaving the asserts purely to catch coding errors. It is certainly not to let the push or pop fail.

Svc/FpySequencer/FpySequencerStack.cpp
zimri-leisher: Definitely should be a null pointer check here, although it would probably just be an assert, so the result would be the same (FSW crash). But at least we would have the assert line number.

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -226,13 +236,20 @@ Fw::Success FpySequencer::readBytes(Os::File& file,
...
+    // as a validation failure instead of aborting the FSW process.
+    if (expectedReadLen != actualReadLen) {
+        this->log_WARNING_HI_FileApiError(this->m_sequenceFilePath, Fpy::FileApiStage::READ_LENGTH_MISMATCH);
+        return Fw::Success::FAILURE;
+    }
 
     Fw::SerializeStatus serializeStatus =
         this->m_sequenceBuffer.setBuffLen(static_cast<Fw::Serializable::SizeType>(expectedReadLen));
-    FW_ASSERT(serializeStatus == Fw::FW_SERIALIZE_OK, serializeStatus);
+    if (serializeStatus != Fw::FW_SERIALIZE_OK) {
```
zimri-leisher, line 249: Is there any way that this could fail that is not a coding error?

zimri-leisher:
Had a chance to actually review this. Still agree with my earlier comment. The stack push/pops will only fail if there is a coding error. 

However, the events for the FileReadApi and also a missing null check are good changes.

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -87,6 +87,21 @@ class FpySequencer : public FpySequencerComponentBase {
         // the byte offset from the start of the stack where the current function's local variables begin.
         // analogous to a 'frame pointer'.
         Fpy::StackSizeType currentFrameStart = 0;
+        // Set true by any stack operation that would violate stack bounds
+        // (overflow on push, underflow on pop, or out-of-range offset on
+        // copy/move). The operation is skipped when set, and the executing
+        // sequence is aborted by the dispatcher on the next dispatchStatement
+        // call. This replaces an FW_ASSERT that an attacker-controlled
+        // .fpy bytecode could trigger to crash the FSW process.
+        bool hasBoundsError = false;
```
zimri-leisher, line 96: I recommend all changes related to these bounds errors be removed.

zimri-leisher (changes requested):

LeStarch:
Since the majority of this PR needs to be reverted, I am of the opinion we should close.  @zimri-leisher do you concur?

Issue #5120, opened 2026-05-08
Backport Linux UART driver telemetry to v3.6


nathancheek:
| | |
|:---|:---|
|**_F´ Version_**| `3.6.4` |
|**_Affected Component_**| `Drv/LinuxUartDriver` |
---
## Feature Description

The LinuxUartDriver provides `BytesSent` and `BytesRecv` telemetry channels, but doesn't write to them. This was fixed in F Prime v4, see #3918 and #3948, but it would be nice to backport this to v3.6.

I have cherry-picked the fix on top of v3.6.4 and done some light testing. That work is available [here](https://github.com/nathancheek/fprime/tree/fix/linux-uart-driver-telemetry-backport), and I can open a PR soon. It seems like `FwSizeType` does not work for telemetry channel types in v3, but if I revert the channels to `U32` type, the test deployment compiles and functions as expected.

## Rationale

Users of the LinuxUartDriver expect the `BytesSent` and `BytesRecv` telemetry channels to have relevant data written to them. Bytes sent/received are useful metrics for debugging and understanding the state of the system.

SterlingPeet:
GPDM needs this work debugging instrument manager components at GT

PR #5150, opened 2026-05-13
Backport LinuxUartDriver rategroup-driven telemetry

nathancheek:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5120 |

---
## Change Description

Backported #3980 to v3.6.x, making it so LinuxUartDriver won't spam the telemetry system every time it receives or sends a chunk of bytes. Instead, telemetry updates happen via rate group call.

Also updated the `send_handler()` header to match the existing fpp definition, changing `NATIVE_INT_TYPE` to `FwIndexType`, `serBuffer` to `sendBuffer`, and adding a missing `override` keyword.

## Rationale

Sending an update to the telemetry system every time bytes are sent/received could cause undue load on the telemetry system.

I have confirmed that a test deployment sends incrementing UART RX/TX byte counts to F Prime GDS every time the rate group fires. It might be worth testing on more deployments / more architectures.

## Future Work

N/A

N/A

### DIFF
--- a/Drv/LinuxUartDriver/LinuxUartDriver.cpp
+++ b/Drv/LinuxUartDriver/LinuxUartDriver.cpp
@@ -297,13 +297,18 @@ LinuxUartDriver ::~LinuxUartDriver() {
 // Handler implementations for user-defined typed input ports
 // ----------------------------------------------------------------------
 
-Drv::SendStatus LinuxUartDriver ::send_handler(const NATIVE_INT_TYPE portNum, Fw::Buffer& serBuffer) {
+void LinuxUartDriver ::run_handler(FwIndexType portNum, U32 context) {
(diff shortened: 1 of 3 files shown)


thomas-bc (approved):

thomas-bc:
LGTM and in line with what we have now in devel (with the exception of U32 vs FwSizeType for the types of those telemetry channels, but I understand not wanting to change that for our purposes here)

SterlingPeet:
I think I am good with the changes, but I would like to better understand why the macos CI is failing.  Specifically I want to know that its not due to this change, though I can't see why it would be.

Since this is backporting a bug fix from v4, I intend to ignore the AI code review suggestions.  We aren't trying to improve the code beyond the targeted maintenance bug fix.

SterlingPeet (approved):
Seems the macos failure was intermittent.

PR #5156, opened 2026-05-14
ci: declare contents:read on pip-check workflow

arpitjain099:
Adds a workflow-level `permissions: contents: read` block. The job here only checks out the repository and runs its tests / validation; no GitHub API call beyond the initial checkout is needed.

CVE-2025-30066 (the March 2025 `tj-actions/changed-files` supply-chain compromise) is the canonical motivation: a tampered third-party action exfiltrated `GITHUB_TOKEN` from workflow logs and the leaked token retained whatever scope was issued at the workflow level. Per-workflow caps bound that runtime authority irrespective of repo or org default, give drift protection if the default ever widens, and register with OpenSSF Scorecard's Token-Permissions check (which only credits explicit per-workflow declarations).

### DIFF
--- a/.github/workflows/pip-check.yml
+++ b/.github/workflows/pip-check.yml
@@ -15,6 +15,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: ${{ !contains(github.ref, 'devel') && !contains(github.ref, 'release/')}}
 
+permissions:
+  contents: read
+
 jobs:
   pip-install:
     runs-on: ${{ matrix.runner }}


thomas-bc:
@arpitjain099 thanks for pushing best practice. I believe that most of our workflow would actually benefit from having only `contents: read` — I don't think any need more than that, except for the few that already list different permissions (e.g. https://github.com/nasa/fprime/blob/e38e23d434404d5941efc964454e4d268c556721/.github/workflows/codeql-jpl-standard.yml#L27-L29). Would you want to do the update to all of those that don't specify anything yet? Or should I push to your branch ?

LeStarch (approved):

PR #5158, opened 2026-05-14
ci: declare contents:read on fpp-to-json workflow

arpitjain099:
Adds a workflow-level `permissions: contents: read` block. The job here only checks out the repository and runs its tests / validation; no GitHub API call beyond the initial checkout is needed.

CVE-2025-30066 (the March 2025 `tj-actions/changed-files` supply-chain compromise) is the canonical motivation: a tampered third-party action exfiltrated `GITHUB_TOKEN` from workflow logs and the leaked token retained whatever scope was issued at the workflow level. Per-workflow caps bound that runtime authority irrespective of repo or org default, give drift protection if the default ever widens, and register with OpenSSF Scorecard's Token-Permissions check (which only credits explicit per-workflow declarations).


thomas-bc:
Thanks for pushing best practice. Closing this PR, please consolidate all updates in one PR here: https://github.com/nasa/fprime/pull/5156

PR #5170, opened 2026-05-19
TlmChan hash improvements

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**|  4.2.1 |

---
## Change Description

Add a secret seed to TlmChan hash processing and replaced simple modulo hash with more complex logic.

## Rationale

Better downlink processing w/ processing cap's on run_handler to avoid threat overruns during off-nominal processing.

Execute new unit test.

## Future Work

None.

Claude was used to create new unit tests. Also checked my hash math.


LeStarch:
@lestarch-autobot please?

Svc/TlmChan/TlmChan.cpp
```diff
@@ -45,12 +60,81 @@ TlmChan::TlmChan(const char* name) : TlmChanComponentBase(name), m_activeBuffer(
...
+    std::random_device rd;
+
+    // check for hardware platform specific entropy (random source)
+    if (rd.entropy() > 0.0) {
+        // True hardware entropy available
+        rdVal = static_cast<U32>(rd());
+    } else {
+        // Platform has no entropy source
+        // Fall back to a hash of the type name as a static distinguisher
+        rdVal = static_cast<U32>(std::hash<const char*>{}(__FILE__) ^ __LINE__);
```
lestarch-autobot: **could fix** Entropy fallback is compile-time deterministic.

When , this fallback resolves to a value known at compile time ( and  are fixed). On bare-metal targets that also lack a high-resolution clock and have deterministic stack layout, the combined seed () may be predictable across boots, partially re-enabling the collision-attack vector this PR mitigates.

Consider using a platform-provided hardware register (e.g., free-running timer counter, ADC noise sample) as an additional entropy source when .

Svc/TlmChan/TlmChan.cpp
```diff
@@ -13,6 +13,8 @@
 #include <Fw/FPrimeBasicTypes.hpp>
 #include <Fw/Types/Assert.hpp>
 #include <Svc/TlmChan/TlmChan.hpp>
+#include <chrono>
+#include <random>
```
lestarch-autobot: **must fix** CPP-25: `<chrono>` and `<random>` introduce C++ STL usage in flight code.

F Prime bans STL headers in production flight components to ensure portability across bare-metal toolchains that may not provide full standard-library support. `std::random_device`, `std::chrono::steady_clock`, and `std::hash` are all STL facilities. Consider using F Prime OSAL primitives (e.g., `Os::RawTime` for timestamps) and a platform-abstracted entropy source instead.

cc @LeStarch @thomas-bc — please confirm whether STL usage is acceptable in this component given its security-hardening purpose.

Svc/TlmChan/TlmChan.cpp
```diff
@@ -45,12 +60,81 @@ TlmChan::TlmChan(const char* name) : TlmChanComponentBase(name), m_activeBuffer(
...
+    if (rd.entropy() > 0.0) {
+        // True hardware entropy available
+        rdVal = static_cast<U32>(rd());
+    } else {
+        // Platform has no entropy source
+        // Fall back to a hash of the type name as a static distinguisher
+        rdVal = static_cast<U32>(std::hash<const char*>{}(__FILE__) ^ __LINE__);
+    }
+
+    const uintptr_t raw = reinterpret_cast<uintptr_t>(&seed);
```
lestarch-autobot: **could fix** CPP-10: `reinterpret_cast` used without inline justification.

The cast from `&seed` to `uintptr_t` is used to harvest stack-address entropy for hash-seed generation. While the security purpose is clear from surrounding comments, F Prime style calls for a brief inline rationale on any `reinterpret_cast` in production code.

Svc/TlmChan/TlmChan.fpp
```diff
@@ -1,7 +1,20 @@
...
-
   @ A component for storing telemetry
   active component TlmChan {
+#-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --
+#Events
+#-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --
+
+      @Epoch Processing cap reached; one or more telemetry entries were deferred this cycle
+    event TlmChanEpochProcessingCapReached(
+        numDeferred: U32 @< Entries skipped (dropped) this invocation
```
lestarch-autobot, line 10: **must fix** The SDD at `Svc/TlmChan/docs/sdd.md` needs updating for: (1) new `TlmChanEpochProcessingCapReached` event, (2) new `timeCaller`/`eventOut`/`eventOutText` ports, (3) changed hash algorithm (seeded Murmur3/Wang replacing linear modulo), and (4) `TLMCHAN_MAX_ENTRIES_PER_RUN` processing cap in `Run_handler`.

The SDD currently documents the old hash behavior and does not mention the processing-cap or the new event. These are externally-observable behavioral changes that downstream integrators need to understand.

Svc/TlmChan/TlmChan.cpp
```diff
@@ -45,12 +60,81 @@ TlmChan::TlmChan(const char* name) : TlmChanComponentBase(name), m_activeBuffer(
...
+    // non-deterministic random source for seed
+    U32 seed = 0;
+
+    const I64 timeNs = static_cast<I64>(
+        std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::steady_clock::now().time_since_epoch())
+            .count());
+    const U32 foldedNs = static_cast<U32>(timeNs) ^ (static_cast<U32>(static_cast<U64>(timeNs) >> 32));
+
+    U32 rdVal = 0;
+    std::random_device rd;
```
lestarch-autobot: **suggestion** The entropy-gathering logic uses platform-specific C++ STL facilities (`std::chrono`, `std::random_device`) that bypass F Prime's OSAL abstraction layer.

F Prime's design pattern is to route all platform interactions through the OSAL (`Os/`) so that platform-specific behavior is encapsulated and testable via stubs. Consider encapsulating entropy collection behind an `Os::Entropy` or similar OSAL interface so bare-metal ports can provide their own implementation without modifying this component.

Svc/TlmChan/test/ut/TlmChanTester.cpp
```diff
@@ -135,6 +140,88 @@ void TlmChanTester::runOffNominal() {
...
+    // This test only has meaning when the per-run cap is set below the total
+    // bucket count.  If TLMCHAN_MAX_ENTRIES_PER_RUN equals TLMCHAN_HASH_BUCKETS
+    // the guard can never fire.
+    if (TLMCHAN_MAX_ENTRIES_PER_RUN >= TLMCHAN_HASH_BUCKETS) {
+        printf(
+            "SKIP: TLMCHAN_MAX_ENTRIES_PER_RUN (%u) >= TLMCHAN_HASH_BUCKETS (%u); "
+            "guard cannot fire with current config\n",
+            static_cast<unsigned>(TLMCHAN_MAX_ENTRIES_PER_RUN), static_cast<unsigned>(TLMCHAN_HASH_BUCKETS));
+        return;
+    }
```
lestarch-autobot, line 153: **suggestion** The `TlmChanEpochProcessingCapReached` event is a new FPP element added by this PR. The `runProcGuard()` test exercises the happy path (cap fires correctly), but there is no explicit test for the case where `TLMCHAN_MAX_ENTRIES_PER_RUN == TLMCHAN_HASH_BUCKETS` (cap cannot fire). The early-return guard at the top handles this, but adding a brief `ASSERT_EVENTS_SIZE(0)` after `doRun` in that configuration would confirm the event is correctly suppressed.

lestarch-autobot (changes requested):

### Per-agent results

| Security Vulnerabilities | 0 | 0 | 1 | 0 | 0 | Go |
| F Prime C/C++ Design | 1 | 0 | 1 | 0 | 1 | No-Go |
| Documentation Currency | 1 | 0 | 0 | 0 | 1 | No-Go |
| Design | 0 | 1 | 0 | 0 | 0 | Go |
| **Totals** | 2 | 2 | 2 | 0 | 2 | **No-Go** |

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>
<summary>Outstanding must-fix items (2)</summary>

**F Prime C/C++ Design**
- CPP-25: `<chrono>` and `<random>` STL headers in flight code — [link](https://github.com/nasa/fprime/pull/5170#pullrequestreview-4323903585)

**Documentation Currency**
- SDD at `Svc/TlmChan/docs/sdd.md` needs update for new event, ports, hash algorithm, and processing cap — [link](https://github.com/nasa/fprime/pull/5170#pullrequestreview-4323904546)

</details>

### Merge readiness
**Merge readiness: No-Go** — F Prime C/C++ Design and Documentation Currency agents each have 1 outstanding must-fix item.

---

The hash-hardening approach is solid security work; the main blockers are STL portability and doc currency.

default/config/TlmChanImplCfg.hpp
```diff
@@ -46,8 +46,11 @@ enum {
     TLMCHAN_HASH_MOD_VALUE = 99,      // !< The modulo value of the hashing function.
                                       // Should be set to a little below the ID gaps to spread the entries around
 
-    TLMCHAN_HASH_BUCKETS = 500  // !< Buckets assignable to a hash slot.
-                                // Buckets must be >= number of telemetry channels in system
+    TLMCHAN_HASH_BUCKETS = 500,  // !< Buckets assignable to a hash slot.
+                                 // Buckets must be >= number of telemetry channels in system
+
+    // Maximum number of updated telemetry entries Run_handler will serialize per invocation.
+    TLMCHAN_MAX_ENTRIES_PER_RUN = (TLMCHAN_HASH_BUCKETS / 2),
```
LeStarch: It seems that we are mixing it two changes here.  Hash function safety and runtime bounding.

Svc/TlmChan/TlmChan.cpp
```diff
@@ -192,50 +259,87 @@ void TlmChan::Run_handler(FwIndexType portNum, U32 context) {
...
     }
 
-    // lock mutex long enough to modify active telemetry buffer
-    // so the data can be read without worrying about updates
+    // Lock mutex long enough to swap the active buffer so the inactive buffer
+    // can be read without worrying about concurrent updates.
     this->lock();
     this->m_activeBuffer = 1 - this->m_activeBuffer;
-    // set activeBuffer to not updated
+    // Clear the new active buffer's updated flags so it is clean for incoming
```
LeStarch, line 288: This is a clear functional change, that I do not believe was approved by the CCB.

Svc/TlmChan/TlmChan.cpp
```diff
@@ -45,12 +60,81 @@ TlmChan::TlmChan(const char* name) : TlmChanComponentBase(name), m_activeBuffer(
...
     this->m_tlmEntries[1].free = 0;
+
+    // ------- Set random telemetry hash seed -------
+
+    // use timestamp, hardware random source and stack address for
+    // non-deterministic random source for seed
+    U32 seed = 0;
+
+    const I64 timeNs = static_cast<I64>(
+        std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::steady_clock::now().time_since_epoch())
```
LeStarch: You've introduced a dependency on `std::chrono` here that was intentionally removed from standard components.....due to the fact that some compilers don't have a working implementation.

Could we switch to F Prime getTime and use microseconds for the seed?

Svc/TlmChan/TlmChan.cpp
LeStarch: Random device seems like something that should not be pulled in from standard.  Perhaps we need an OSAL random at some point?

Svc/TlmChan/TlmChan.cpp
LeStarch: I know you are trying to use the stack location as a source of "random" but this is not well explaineed with a comment.

Svc/TlmChan/TlmChan.cpp
```diff
@@ -45,12 +60,81 @@ TlmChan::TlmChan(const char* name) : TlmChanComponentBase(name), m_activeBuffer(
...
+        // Fall back to a hash of the type name as a static distinguisher
+        rdVal = static_cast<U32>(std::hash<const char*>{}(__FILE__) ^ __LINE__);
+    }
+
+    const uintptr_t raw = reinterpret_cast<uintptr_t>(&seed);
+    const U32 foldedStack = static_cast<U32>(raw ^ (raw >> 32));
+
+    seed = foldedNs ^ rdVal ^ foldedStack;
+
+    // check for a zero seed
```
LeStarch: Why does a zero seed matter?

LeStarch (changes requested):
My biggest concerns are:
1. A functional change without CCB approval (limiting)
2. Use of stl for time and random.  This really should be in the OSAL since these services are not always implemented.

We can absolve item 1, if the configuration delivered is set to HASH_BUCKETS so the behavior is identical.  However, we **must** update the SDD to catch the new value.

Two is a blocker in my opinion because it will break this component on some platforms (potentially).

Svc/TlmChan/TlmChan.cpp
LeStarch: Or just use the time directly, no?

Svc/TlmChan/TlmChan.cpp
LeStarch: Not here.  We can use `getTime()` to work around chrono, but probably need a random added to the OSAL.

Svc/TlmChan/TlmChan.cpp
LeStarch: Hey nice work @lestarch-autobot! Good catch!

Svc/TlmChan/test/ut/TlmChanTester.cpp
LeStarch: Ohhhh, good catch!

Svc/TlmChan/TlmChan.cpp
bitWarrior: Using something hardware specific make this change hardware dependent which I do not want to do for a core F Prime component.

Svc/TlmChan/TlmChan.cpp
bitWarrior: getTime() should work.

Svc/TlmChan/TlmChan.fpp
bitWarrior: Yes, the SDD will need to be updated.

Svc/TlmChan/TlmChan.cpp
bitWarrior: A seed of 0 reverts to the original linear-modulo behavior I want to avoid. I can add a comment or two to make it more clear.

Svc/TlmChan/TlmChan.cpp
bitWarrior: Fair point. I could add a comment to something like "Cast the address of a local variable to enforce hardware entropy; the stack address low-order bits that vary per run even when other hardware resources are deterministic.

Svc/TlmChan/TlmChan.cpp
bitWarrior: Yes. It is something I am planning for in the encryption up/down implementation. I did not want to make this PR more complex than it is.

Svc/TlmChan/TlmChan.cpp
bitWarrior: Yes we can switch. The concern is for deployments such as VxWorks, the microseconds may not vary as I would like between runs. I was not aware we are moving away from std::chrono so I will change it.

Svc/TlmChan/TlmChan.cpp
bitWarrior: Huh? Well since this has not been through CCB you are technically correct. Please review the advisory.

default/config/TlmChanImplCfg.hpp
bitWarrior: Actually they are the same.  Adding a processing cap is a layered defense by preventing the TlmChan run_handler from taking over CPU processing during a burst of telemetry updates (hardware anomaly, runaway component, software fault, cyber-attack, etc.) and starving higher-priority tasks.

lestarch-autobot (changes requested):
### Per-agent results

| Security Vulnerabilities | 0 | 0 | 1 | 0 | 1 | Go |
| F Prime C/C++ Design | 1 | 0 | 1 | 0 | 2 | No-Go |
| Documentation Currency | 1 | 0 | 0 | 0 | 1 | No-Go |
| Design | 0 | 1 | 0 | 0 | 1 | Go |
| **Totals** | 2 | 2 | 2 | 0 | 6 | **No-Go** |

### CI safety: Go

Both CI-safety reviewers (security, supply-chain) report Go. No runner-safety or vulnerability concerns.

### Merge readiness: No-Go

**2 outstanding must-fix findings** block merge:

1. **CPP-25 — STL in flight code** (`fprime-code-review`): `#include <chrono>` and `#include <random>` introduce C++ STL usage in flight code. F Prime bans STL features for determinism, binary-size, and portability reasons. Replace with `Os::` / `Fw::` APIs or compile-time constants.

2. **SDD stale** (`stale-documentation-review`): `Svc/TlmChan/docs/sdd.md` §3.5 describes the old `(id % MOD) % SLOTS` hash. The PR replaces this with Murmur3/Wang hashing with a random seed and adds a processing cap with `TlmChanEpochProcessingCapReached` event. The SDD, port table, and requirements section need updating.

**Additional open items** (non-blocking but recommended):
- `could fix`: CPP-10 — `reinterpret_cast<uintptr_t>(&seed)` needs inline justification comment
- `could fix`: Entropy fallback is compile-time deterministic when `std::random_device` is unavailable
- `suggestion`: Entropy-gathering logic ties the component to platform-specific STL facilities
- `suggestion`: `TlmChanEpochProcessingCapReached` event needs dedicated test coverage

---

Re-review run 2 — HEAD unchanged from run 1. All prior findings remain open.

Svc/TlmChan/TlmChan.cpp
```diff
@@ -45,12 +60,81 @@ TlmChan::TlmChan(const char* name) : TlmChanComponentBase(name), m_activeBuffer(
...
 }
 
 TlmChan::~TlmChan() {}
+constexpr size_t FwChanIdSize = sizeof(FwChanIdType);
 
 FwChanIdType TlmChan::doHash(FwChanIdType id) {
-    return (id % TLMCHAN_HASH_MOD_VALUE) % TLMCHAN_NUM_TLM_HASH_SLOTS;
+    // Validate input before use
+    FW_ASSERT(TLMCHAN_NUM_TLM_HASH_SLOTS > 0);
+    FwChanIdType result;
```
lestarch-autobot, line 113: **suggestion** CPP-19: Variable `result` declared without initializer.

Every variable must have an explicit initializer at declaration. Although `result` is assigned in every branch of the `if`/`else if`/`else` chain below, the F Prime style rule requires initialization at the point of declaration to prevent latent UB if the control flow is later refactored.

```suggestion
    FwChanIdType result = 0;
```

Svc/TlmChan/TlmChan.cpp
```diff
@@ -45,12 +60,81 @@ TlmChan::TlmChan(const char* name) : TlmChanComponentBase(name), m_activeBuffer(
...
+
+    if (FwChanIdSize >= 4) {
+        U32 h = static_cast<U32>(id) ^ static_cast<U32>(this->m_hashSeed);
+
+        // Murmur3 32-bit
+        h ^= (h >> 16);
+        h *= 0x85EBCA6B;
+        h ^= (h >> 13);
+        h *= 0xC2B2AE35;
+        h ^= (h >> 16);
```
lestarch-autobot: **could fix** CPP-30: Murmur3 finalizer uses unnamed numeric literals.

The constants `0x85EBCA6B` and `0xC2B2AE35` are the Murmur3 32-bit finalizer multipliers, and `0x2993U`/`0xE877U` (lines 125-127) are the Wang 16-bit constants. Per CPP-30, numeric literals should be named constants with a derivation comment so reviewers can verify provenance without external lookup.

Consider:
```cpp
static constexpr U32 MURMUR3_C1 = 0x85EBCA6B;  // Murmur3 32-bit finalizer constant 1
static constexpr U32 MURMUR3_C2 = 0xC2B2AE35;  // Murmur3 32-bit finalizer constant 2
static constexpr U16 WANG16_C1 = 0x2993U;       // Wang 16-bit hash constant 1
static constexpr U16 WANG16_C2 = 0xE877U;       // Wang 16-bit hash constant 2
```

Svc/TlmChan/TlmChan.cpp
```diff
@@ -192,50 +259,87 @@ void TlmChan::Run_handler(FwIndexType portNum, U32 context) {
...
+            // be cleared by the next buffer swap (see lock section above), so
+            // the sample is intentionally dropped for this cycle.  This bounds
+            // Run_handler's worst-case execution time and prevents it from
+            // starving higher-priority tasks during a telemetry burst caused by
+            // a hardware anomaly, runaway component, software fault, or
+            // cyber-attack.
+            // ------------------------------------------------------------------
+            if (entriesProcessed >= TLMCHAN_MAX_ENTRIES_PER_RUN) {
+                entriesDeferred++;
+                continue;
```
lestarch-autobot, line 329: **must fix** Behavioral regression: processing cap silently drops telemetry for existing deployments.

The new `TLMCHAN_MAX_ENTRIES_PER_RUN` cap (default = `TLMCHAN_HASH_BUCKETS / 2`) means Run_handler will now skip (drop) entries beyond the cap each cycle. Existing deployments that previously relied on all updated telemetry channels being downlinked every cycle will silently lose data when the cap fires.

The WARNING_HI event is a good mitigation signal, but the default value of `BUCKETS / 2` means any deployment with more than 250 active telemetry channels will immediately hit this cap on upgrade — a backwards-incompatible behavioral change.

**Recommended action**: Either (a) set the default to `TLMCHAN_HASH_BUCKETS` (cap disabled by default, opt-in hardening), or (b) document this as a breaking change and require deployers to consciously tune the constant, or (c) add a compile-time opt-out `#ifndef TLMCHAN_DISABLE_PROC_CAP`.

cc @LeStarch @thomas-bc — needs human adjudication on backward-compatibility policy for this behavioral change.

lestarch-autobot (changes requested):

Svc/TlmChan/TlmChan.cpp
lestarch-autobot: Fixed in 6f4bd9d1fd5ed80f6efb9f81623b421e3aeab488. Constants are now named (`MURMUR3_C1`, `MURMUR3_C2`, `WANG16_C1`, `WANG16_C2`) in the header.

lestarch-autobot (changes requested):
## Automated review summary  (run 4)

### Per-agent results

| Security Vulnerabilities | 0 | 0 | 1 | 0 | 0 | Go |
| F Prime C/C++ Design | 1 | 1 | 1 | 0 | 1 | No-Go |
| Documentation Currency | 1 | 0 | 0 | 0 | 1 | No-Go |
| Design | 1 | 1 | 0 | 0 | 1 | No-Go |
| **Totals** | **3** | **3** | **2** | **0** | **3** | |

### CI safety: Go

Both security-review and supply-chain-review report Go. No CI runner risk.

### Merge readiness: No-Go

3 outstanding must-fix findings across 3 agents:
- **fprime-code-review**: `<ctime>` and `<random>` STL headers in flight code (CPP-25)
- **stale-documentation-review**: SDD not updated for new event, ports, hash algorithm, processing cap
- **design-review**: Processing cap default (`TLMCHAN_HASH_BUCKETS / 2`) silently drops telemetry for existing deployments with >250 active channels

### Since last run

| Metric | Count |
|---|---:|
| Resolved | 1 |
| Still open | 8 |
| Newly added | 0 |
| Incorrect fix | 0 |
| Improperly resolved | 0 |
| Disagreements | 0 |

**Resolved**: Magic number constants (Murmur3/Wang multipliers) now named in header (`MURMUR3_C1`, `MURMUR3_C2`, `WANG16_C1`, `WANG16_C2`).

### Prompt-injection pre-check: clean

bitWarrior:
All issues resolved. Please run autobot and merge. This is an important improvement to F Prime

LeStarch (dismissed):

LeStarch (approved):

PR #5181, opened 2026-05-20
Support non-default recordSize and prmId field lengths in PrmDb (#4716)

nathancheek:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4716 |

---
## Change Description

Updated PrmDbImpl to use configurable types `FwSizeStoreType` and `FwPrmIdType` when handling `recordSize` and `prmId` fields.

Updated unit tests to support these configurable types.  The `PrmFileReadError` test was specifically updated to successfully test the `PrmReadError::RECORD_SIZE_VALUE` scenario for both `U16` and `U32` `FwPrmIdType` cases.

## Rationale

Some projects have unique sizing requirements that differ from the standard F Prime configuration. `FwSizeStoreType` and `FwPrmIdType` should be used where relevant.

These changes have been tested on macOS. Unit tests were confirmed to pass with `FwPrmIdType` configured as both `U16` and `U32`. It might be worth testing on more deployments / more architectures.

## Future Work

N/A

N/A

### DIFF
--- a/Svc/PrmDb/PrmDbImpl.cpp
+++ b/Svc/PrmDb/PrmDbImpl.cpp
@@ -51,7 +51,7 @@ namespace Svc {
     }
 
     void PrmDbImpl::clearDb() {
-        for (I32 entry = 0; entry < PRMDB_NUM_DB_ENTRIES; entry++) {
+        for (FwPrmIdType entry = 0; entry < PRMDB_NUM_DB_ENTRIES; entry++) {
             this->m_db[entry].used = false;
             this->m_db[entry].id = 0;
         }
@@ -64,7 +64,7 @@ namespace Svc {
         // search for entry
         Fw::ParamValid stat = Fw::ParamValid::INVALID;
 
-        for (I32 entry = 0; entry < PRMDB_NUM_DB_ENTRIES; entry++) {
+        for (FwPrmIdType entry = 0; entry < PRMDB_NUM_DB_ENTRIES; entry++) {
             if (this->m_db[entry].used) {
                 if (this->m_db[entry].id == id) {
                     val = this->m_db[entry].val;
@@ -91,7 +91,7 @@ namespace Svc {
         bool existingEntry = false;
         bool noSlots = true;
 
-        for (NATIVE_INT_TYPE entry = 0; entry < PRMDB_NUM_DB_ENTRIES; entry++) {
+        for (FwPrmIdType entry = 0; entry < PRMDB_NUM_DB_ENTRIES; entry++) {
             if ((this->m_db[entry].used) && (id == this->m_db[entry].id)) {
                 this->m_db[entry].val = val;
                 existingEntry = true;
@@ -101,7 +101,7 @@ namespace Svc {
 
         // if there is no existing entry, add one
         if (!existingEntry) {
-            for (I32 entry = 0; entry < PRMDB_NUM_DB_ENTRIES; entry++) {
+            for (FwPrmIdType entry = 0; entry < PRMDB_NUM_DB_ENTRIES; entry++) {
                 if (!(this->m_db[entry].used)) {
                     this->m_db[entry].val = val;
                     this->m_db[entry].id = id;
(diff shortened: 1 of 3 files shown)


LeStarch:
@lestarch-autobot

Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -321,7 +321,7 @@ namespace Svc {
 
             // sanity check value. It can't be larger than the maximum parameter buffer size + id
             // or smaller than the record id
-            if ((recordSize > FW_PARAM_BUFFER_MAX_SIZE + sizeof(U32)) or (recordSize < sizeof(U32))) {
+            if ((recordSize > FW_PARAM_BUFFER_MAX_SIZE + sizeof(FwPrmIdType)) or (recordSize < sizeof(FwPrmIdType))) {
```
lestarch-autobot, line 324: (Anchored above the offending line; the diff does not include line 364.)

**future work** (CPP-28) At line 364, `static_cast<U32>(readSize)` is compared against `recordSize - sizeof(parameterId)` where `recordSize` is now `FwSizeStoreType`. The `U32` cast is a bare fixed-size type left over from the pre-refactor code; it should use the configurable type for consistency with this PR's direction.

Preexisting — the PR did not introduce this, but worth aligning in a follow-up.

Svc/PrmDb/PrmDbImpl.cpp
```diff
@@ -164,7 +164,7 @@ namespace Svc {
                     return;
                 }
                 // serialize record size = id field + data
-                U32 recordSize = static_cast<U32>(sizeof(FwPrmIdType) + this->m_db[entry].val.getBuffLength());
+                FwSizeStoreType recordSize = static_cast<FwSizeStoreType>(sizeof(FwPrmIdType) + this->m_db[entry].val.getBuffLength());
```
lestarch-autobot, line 167: **could fix** `Svc/PrmDb/docs/sdd.md` §3.2 describes the parameter file format with a hardcoded record-size width:

> Record Size | **4** | Id type size + number of bytes in parameter value

With this PR the record-size field is now `sizeof(FwSizeStoreType)` (default `U16` = 2 bytes, not 4). The SDD table should be updated to reflect the configurable width — e.g. `sizeof(FwSizeStoreType)` — so the documented file format stays accurate.

lestarch-autobot (approved):
### Run 1

### Per-agent results

| F Prime C/C++ Design | 0 | 0 | 0 | 1 | 0 | Go |
| Documentation Currency | 0 | 0 | 1 | 0 | 1 | Go |
| **Totals** | 0 | 0 | 1 | 1 | 1 | **Go** |

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

### Merge readiness
**Merge readiness: Go** — all agents completed with zero outstanding must-fix items.

---

Clean run on a focused type-correctness fix — ready for the mission.

LeStarch (approved):
@SterlingPeet looking for your review too!

SterlingPeet (changes requested):
I think it looks good.  The auto-bot brings up some good points, and I think it would be ideal to fix them up because they are small easy fixes.

Svc/PrmDb/PrmDbImpl.cpp
SterlingPeet: Lets fix this one because we have outstanding GDS work that should reference this documentation.

Svc/PrmDb/PrmDbImpl.cpp
SterlingPeet: This seems like it should be quick, just remove the static cast, probably.  I haven't referenced to code directly though.

nathancheek: I believe you still want the static cast, since `readSize` is of type `FwSignedSizeType`, not of type `FwSizeStoreType`. But the static cast should be updated from `static_cast<U32>(readSize)` to `static_cast<FwSizeStoreType>(readSize)` to match what it's being compared to.

I have made this change and re-pushed.

Svc/PrmDb/PrmDbImpl.cpp
nathancheek: I updated this to match the wording of the next row in the table. I also updated the previous row to indicate that while 0xA5 is the default delimiter value, it is technically set by the `PRMDB_ENTRY_DELIMITER` enumerator in `config/PrmDbImplTesterCfg.hpp`.

Svc/PrmDb/PrmDbImpl.cpp
SterlingPeet: I concur, this is the correct solution.

SterlingPeet (approved):
Excellent work, this looks good to go.

PR #5196, opened 2026-05-22
AosFramer: make spacecraftId and vcId configurable

shan-zh-26:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Added the changes in AosFramer to support configurable spacecraftId and vcId.

## Rationale

Different projects need different spacecraftId and vcId.

Unit test.

### DIFF
--- a/Svc/Ccsds/AosFramer/AosFramer.cpp
+++ b/Svc/Ccsds/AosFramer/AosFramer.cpp
@@ -16,8 +16,7 @@ namespace Ccsds {
 // Component construction and destruction
 // ----------------------------------------------------------------------
 
-AosFramer ::AosFramer(const char* const compName)
-    : AosFramerComponentBase(compName), m_spacecraftId(ComCfg::SpacecraftId) {
+AosFramer ::AosFramer(const char* const compName) : AosFramerComponentBase(compName) {
     // Default to FECF on, Max Sized if you don't override w/ another configure call
     configure(ComCfg::AosMaxFrameFixedSize, true);
 }
@@ -36,6 +35,12 @@ void AosFramer::configure(const U32 fixedFrameSize,
     FW_ASSERT(fixedFrameSize > AOSHeader::SERIALIZED_SIZE + M_PDUHeader::SERIALIZED_SIZE +
                                    (frameErrorControlField ? AOSTrailer::SERIALIZED_SIZE : 0),
               static_cast<FwAssertArgType>(fixedFrameSize));
+
+    // Spacecraft ID is 10 bits (per CCSDS 732.0-B-5 Section 4.1.2.2)
+    FW_ASSERT((spacecraftId & 0xFC00) == 0, static_cast<FwAssertArgType>(spacecraftId));
+
+    // Virtual Channel ID is 6 bits (per CCSDS 732.0-B-5 Section 4.1.2.3)
+    FW_ASSERT((vcId & 0xC0) == 0, static_cast<FwAssertArgType>(vcId));
 
     // AOS Framer must be provided with a protocol to use for Idle Packets
     // Currently, only SPP idle packing is supported
--- a/Svc/Ccsds/AosFramer/AosFramer.cpp
+++ b/Svc/Ccsds/AosFramer/AosFramer.cpp
@@ -16,14 +16,19 @@ namespace Ccsds {
 // Component construction and destruction
 // ----------------------------------------------------------------------
 
-AosFramer ::AosFramer(const char* const compName) : AosFramerComponentBase(compName) {
+AosFramer ::AosFramer(const char* const compName)
+    : AosFramerComponentBase(compName), m_spacecraftId(ComCfg::SpacecraftId) {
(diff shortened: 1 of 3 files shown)


Svc/Ccsds/AosFramer/test/ut/AosFramerTester.cpp
Willmac16, line 1: Was the formatting wrong before, or why are we getting such large git diffs here

Svc/Ccsds/AosFramer/AosFramer.cpp
```diff
@@ -16,14 +16,19 @@ namespace Ccsds {
 // Component construction and destruction
 // ----------------------------------------------------------------------
 
-AosFramer ::AosFramer(const char* const compName) : AosFramerComponentBase(compName) {
+AosFramer ::AosFramer(const char* const compName)
+    : AosFramerComponentBase(compName), m_spacecraftId(ComCfg::SpacecraftId) {
```
Willmac16: I don't think you need to init the SCID here, pretty sure the configure call covers that for you

Svc/Ccsds/AosFramer/AosFramer.hpp
Willmac16: Consider setting the struct default to 0 since we're now overriding it via configure for every instance

Willmac16 (changes requested):

Svc/Ccsds/AosFramer/test/ut/AosFramerTester.cpp
shan-zh-26: The formatting was not wrong, but I am adding a new testcase to the existing unit test. It is using a for loop to iterate through these testcases which has different spacecraft ID and vcId, and previous code just got tabs in front of them now. 
**Update**: I reverted the change since it does not increase Line/Branch Coverage, it only provided  a different set of configs other than default values.

LeStarch (dismissed):

Svc/Ccsds/AosFramer/AosFramer.cpp
```diff
@@ -151,7 +163,7 @@ void AosFramer ::setup_header(const ComCfg::FrameContext& context) {
 
     // GVCID (Global Virtual Channel ID) (Standard 4.1.2.2 and 4.1.2.3)
     U16 globalVcId = static_cast<U16>(context.get_vcId() << AOSHeaderSubfields::virtualChannelIdOffset);
-    globalVcId |= static_cast<U16>((ComCfg::SpacecraftId & 0x00FF) << AOSHeaderSubfields::spacecraftIdLsbOffset);
+    globalVcId |= static_cast<U16>((m_spacecraftId & 0x00FF) << AOSHeaderSubfields::spacecraftIdLsbOffset);
     globalVcId |= static_cast<U16>((Tfvn::AOS & 0x3) << AOSHeaderSubfields::frameVersionOffset);
```
thomas-bc, line 155: likely need 
```
    // GVCID (Global Virtual Channel ID) (Standard 4.1.2.2 and 4.1.2.3)
    U16 globalVcId = static_cast<U16>((context.get_vcId() << AOSHeaderSubfields::virtualChannelIdOffset) |
                                      ((m_spacecraftId & 0x00FF) << AOSHeaderSubfields::spacecraftIdLsbOffset) |
                                      ((Tfvn::AOS & 0x3) << AOSHeaderSubfields::frameVersionOffset));

```

Willmac16 (approved):

Svc/Ccsds/AosFramer/AosFramer.cpp
thomas-bc: for RHEL8 CI failure

shan-zh-26: Hi Thomas, looking at the log. The RHEL8 CI build was failing with -Werror=conversion errors at lines 166 and 178  might be relate to When bitwise AND operations mix U16 variables with int literals (0x00FF, 0x0300),  C++ integer promotion rules cause the U16 to be promoted to int, the casting this back to U16 triggers conversion warnings. I added the change and lets see if it fixes the problem.

LeStarch: I don't think this is going to fix this.   The issue is this operator ` |=` is always promoted to `int` and thus the assignment: `U16 |=` is always restricting an `int` to a `U16`.

To fix, we must have the assignment `=` broken away from the `|`.  Hence thomas' change above.

LeStarch: https://eel.is/c%252B%252Bdraft/conv.prom

LeStarch:
@shan-zh-26 you should use @thomas-bc's code to avoid the implicit promotion inside the `|=`

Svc/Ccsds/AosFramer/AosFramer.cpp
shan-zh-26: thanks, i added the suggested changes by thomas!

LeStarch (approved):

PR #5204, opened 2026-05-27
Added Os::CountingSemaphore osal module (#5173)

bevinduckett:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/5173  |
|**_Has Unit Tests (y/n)_**|  Yes |
|**_Documentation Included (y/n)_**|  N/A |
|**_Generative AI was used in this contribution (y/n)_**|  yes |

---
## Change Description

Added Os::CountingSemaphore to the OS abstraction layer 

## Rationale

See rationale in https://github.com/nasa/fprime/issues/5173 

## Testing/Review Recommendations

I've verified all UTs pass & the reference deployment builds & runs. 

## Future Work

Once this is added to fprime, various OSs will need need to add CountingSemaphore s 

## AI Usage (see [policy](https://github.com/nasa/fprime/blob/devel/AI_POLICY.md))

AI was used to assist with documentation, code review, and unit test implementation. The AI tool used was Claude Sonnet 4.5 - edits were reviewed by a human.

Issue #3603, opened 2025-05-14
ComQueue may assert when receiving comStatusIn message


celskeggs:
| | |
|:---|:---|
|**_F´ Version_**| devel |
|**_Affected Component_**| ComQueue |
---
## Problem Description

ComQueue has some ports marked as "drop" and some ports that assert when full:

```
      @ Port for receiving the status signal
      async input port comStatusIn: Fw.SuccessCondition

      @ Port array for receiving Fw::ComBuffers
      async input port comPacketQueueIn: [ComQueueComPorts] Fw.Com drop
```

If the queue is filled up by messages that could be dropped, and then a single message comes in that cannot be dropped, ComQueue triggers an assertion:

```
Assert: "/workspaces/[...]/build-fprime-automatic-[...]/F-Prime/Svc/ComQueue/ComQueueComponentAc.cpp:947" 8
```

## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-4.18.0-553.47.1.el8_10.x86_64-x86_64-with-glibc2.34
Python version: 3.9.21
CMake version: 3.24.2
Pip version: 21.3.1
Pip packages:
    fprime-tools==3.6.1
    fprime-gds==3.6.1
    fprime-fpp-*==3.0.0a9
Project submodules:
    https://github.com/nasa/fprime.git @ v3.6.0-94-g41cadfd21
    https://github.com/fprime-community/fprime-baremetal @ 162b7dd
```

## How to Reproduce

1. Transmit lots of data
2. Lose a connection to a ByteStream
3. Let the ByteStream re-establish a connection and send a Ready message
4. Observe the Ready message overflow the queue

## Expected Behavior

If any async input ports are drop, then no async input ports should assert.

thomas-bc:
@celskeggs the `comStatusIn` can not be dropped because that's what releases a buffer from its internal queues. If it's dropped, the flow is halted. Are you asking that `comPacketQueueIn` should not be dropped? Or maybe that `comStatusIn` could "kick a message out of the queue" ?

celskeggs:
You have multiple options for how to fix this bug, and I'm flexible on which one you choose.

The point is that it is generally unreasonable to define a component to have both `drop` and `assert` async input ports, given that F' intermingles the messages.

I would consider it reasonable to kick a `drop` message out of the queue when receiving an `assert` message. I would also consider it reasonable to define separate queues for the different classes of messages.

Issue #3512, opened 2025-04-17
Queue Implementation Type at Component Construction


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Allow queued and passive components to choose queue type at construction time.

This would allow users to select queue implementations per-component.

LeStarch:
@kevin-f-ortega you are the champion of this ticket during the CCB!

bevinduckett:
Consider https://github.com/nasa/fprime/issues/5359 when dispositioning this ticket

Kronos3:
Related to https://github.com/nasa/fpp/issues/1023

PR #5205, opened 2026-05-27
Message Queue with per priority memory allocation

bevinduckett:
| | |
|:---|:---|
|**_Related Issue(s)_**|  #3603 |
|**_Has Unit Tests (y/n)_**|  Yes |
|**_Documentation Included (y/n)_**|  Yes |
|**_Generative AI was used in this contribution (y/n)_**| Yes   |

---
## Change Description

This PR adds:
- Os::Generic::AtomicQueue: Lock-free MPMC circular buffer using atomic operations for O(1) enqueue/dequeue. ISR-safe (platform-dependent) with optional counting semaphore support for blocking operations. Serves as building block for PriorityMemQueue.
- Os::Generic::PriorityMemQueue: ISR-safe, SMP-safe priority queue supporting up to 16 priority levels with per-priority AtomicQueue instances, independent memory pools, and static configuration system. Includes lock-free priority tracking via atomic bitmasks and rule-based unit tests
- priority_buffer_analyzer.py: CMake-integrated Python script that parses topology dictionaries to auto-generate message size constants per component/priority, simplifying buffer sizing with automatic overhead accounting.

The code in this PR depends on https://github.com/nasa/fprime/pull/5204 and cannot be merged until that PR is complete 

## Rationale

F' allows a component to mark some ports as drop and some ports as assert on full. This allows "drop" ports to fill up the message queue and cause "assert" ports to fail to send a message leading to an assert. E.g. If the queue is filled up by messages that could be dropped, and then a single message comes in that cannot be dropped, ComQueue triggers an assertion:
```
Assert: "/workspaces/[...]/build-fprime-automatic-[...]/F-Prime/Svc/ComQueue/ComQueueComponentAc.cpp:947" 8
```
Adding a message queue with per priority memory allocation resolves that 
```
   @ Port for receiving the status signal
      async input port comStatusIn: Fw.SuccessCondition priority 2

      @ Port array for receiving Fw::ComBuffers
      async input port comPacketQueueIn: [ComQueueComPorts] Fw.Com priority 1 drop
```

Separating messages by priority is also useful for ports with the same drop behavior when the different messages have different priorities (e.g. it would be bad for EventManager to drop a WARNING HI event because DIAGNOSTIC events consumed memory pool). 

There's unit tests and PriorityMemQueue has been tested in a private deployment. I'd like to add it to a deployment in fprime-community but I'm not sure which 

## Future Work

F' components should be updated to use different priorities when sending messages with different urgencies or with different queue full behavior. 

Claude code - Sonnet 4.5. Used to assist with code generation under human supervision (careful review for the flight code, less careful review for the unit test code) and with documentation


Os/Generic/docs/sdd.md
```diff
@@ -39,3 +66,251 @@ When an index is pulled from this structure, the root is removed as it is the hi
 `heapify` starts at the newly ill-ordered root. It iteratively swaps this node with the highest-priority child until this node is the largest of the three (parent, left child, and right child) or until this node is swapped into a leaf position without children. The max-heap invariant is now restored.
 
 
+## Os::PriorityMemQueue
+
+Os::PriorityMemQueue is an ISR-safe and SMP-safe, priority-based memory queue implementation for F´ using VxWorks message queues (msgQ). This implementation leverages the VxWorks msgQ API to provide robust multi-core synchronization without custom spinlock management, eliminating interrupt starvation issues while maintaining ISR safety.
```
thomas-bc: Lots of notes about VxWorks which doesn't seem right - likely the docs were generated before AtomicQueue was introduced? We need to update

Os/Generic/Types/AtomicQueue.cpp
```diff
@@ -0,0 +1,354 @@
...
+                         Fw::MemAllocator& allocator,
+                         FwEnumStoreType allocatorId) {
+    FW_ASSERT(numBuffers > 0, static_cast<FwAssertArgType>(numBuffers));
+    FW_ASSERT(bufferSize > 0, static_cast<FwAssertArgType>(bufferSize));
+
+    this->m_capacity = numBuffers;
+    this->m_bufferSize = bufferSize;
+    this->m_allocator = &allocator;
+    this->m_allocatorId = allocatorId;
+    this->m_id = allocatorId;  // Store for diagnostic logging
```
thomas-bc: Is this using the same `allocatorId` on purpose ?

Os/Generic/PriorityMemQueue.hpp
```diff
@@ -0,0 +1,294 @@
...
+#include "Os/Queue.hpp"
+
+namespace Os {
+namespace Generic {
+
+// Constants
+namespace Queue {
+// JPL heritage implementation supported up to 32 priorities per message queue, limiting to 16 to reduce memory
+// footprint
+constexpr static FwSizeType MAX_PRIORITIES = 16;
```
thomas-bc: This should be clearly documented as it will limit usable priorities if we want to use standard components with PriorityMemQueue

Also need to remove the comment about JPL heritage

Os/Generic/test/ut/PriorityMemQueueTests.cpp
```diff
@@ -0,0 +1,1899 @@
...
+}  // namespace Test
+}  // namespace Ref
+
+// ======================================================================
+// Configuration Tests
+// ======================================================================
+
+// Test configuration with single queue and single priority
+TEST(PriorityMemQueueConfig, SingleQueueSinglePriority) {
+    printf("\n=== TEST: SingleQueueSinglePriority ===\n");
```
thomas-bc: Googletest can handle the logging, we shouldn't need to do any of that

Os/Generic/test/ut/PriorityMemQueueInputValidationTests.cpp
```diff
@@ -0,0 +1,343 @@
...
+}
+
+// ========================================================================
+// Test: isPriorityEnabled() indirectly via enable/disable behavior
+// ========================================================================
+TEST_F(PriorityMemQueueInputValidation, IsPriorityEnabledIndirect) {
+    Os::Generic::PriorityMemQueue queue;
+
+    // Configure multiple priorities
+    Os::Generic::PriorityMemQueue::QueuePriorityConfig priorityConfigs[] = {{0, 128, 10}, {1, 128, 10}, {2, 128, 10}};
```
thomas-bc: There's a lot of duplicated setup code - it'd be nice if AI could use a GTest fixture or helper functions to handle this and reduce the verbosity a bit

Os/Generic/PriorityMemQueue.hpp
```diff
@@ -0,0 +1,294 @@
...
+    Os::CountingSemaphore* m_notEmptySem;  // Counting semaphore signaling messages available
+    FwEnumStoreType m_id;                  // Queue identifier
+    FwEnumStoreType m_allocatorId;         // Allocator ID for memory operations
+    std::atomic<U32> m_priorityMask;       // Bit mask of enabled priorities
+    std::atomic<U32> m_nonEmptyMask;       // Bit mask of priorities with messages (optimization)
+    std::atomic<U32>* m_highWaterMarks;    // Pointer to array of per-priority high water marks
+
+    //! \brief Constructor to initialize members
+    PriorityMemQueueHandle()
+        : m_atomicQueues(nullptr), m_maxPriority(0), m_notEmptySem(nullptr), m_highWaterMarks(nullptr) {}
```
thomas-bc: Not all member variables are initialized properly. Could use the  `= 0`/`=nullptr` on the definition rather than constructor init list

thomas-bc (changes requested):

Os/Generic/docs/sdd.md
```diff
@@ -39,3 +66,251 @@ When an index is pulled from this structure, the root is removed as it is the hi
...
+
+The key components are:
+
+1. **VxWorks msgQ** - One msgQ per priority level for ISR-safe + SMP-safe message storage
+2. **Atomic Priority Tracking** - Lock-free bitmasks to track which priorities have messages
+3. **Counting Semaphore** - Notification mechanism for blocking receive operations
+4. **PriorityMemQueue** - Main queue implementation that implements Os::QueueInterface
```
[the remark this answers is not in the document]
thomas-bc: Resolving - captured in https://github.com/nasa/fprime/pull/5205#discussion_r3376188934

thomas-bc (commented):
Also I agree with your note in your PR, we really ought to be testing this in an actual deployment - we should tag up

cmake/autocoder/scripts/priority_buffer_analyzer.py
```diff
@@ -0,0 +1,693 @@
...
+            return ' + '.join(size_expressions)
+
+        except (AttributeError, KeyError) as e:
+            logger.warning(f"Warning: Could not calculate internal port size: {e}\n{port_instance.__dict__!r}")
+            raise
+    
+    def _generate_header_preamble(self, lines: List[str]) -> None:
+        """Generate header file preamble with includes"""
+        lines.append("#ifndef PRIORITY_BUFFER_SIZES_AC_HPP")
+        lines.append("#define PRIORITY_BUFFER_SIZES_AC_HPP")
```
thomas-bc, line 649: Lots of places in this Python script that seems to be generating C++ code. Why? 

And how are users expected to use this script?

Os/Generic/Types/AtomicQueue.cpp
bevinduckett: Yes - the allocator ID is usually the component instance ID (at least for queues created during component init). So storing that ID to report in asserts provides a link to which component's queue asserted (particularly useful for asserts that occur during send operations where the thread/task ID doesn't identify receiving queue)

Os/Generic/Types/AtomicQueue.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+    FwSizeType enq = this->m_enqueuePos.load(std::memory_order_relaxed);
+    FwSizeType deq = this->m_dequeuePos.load(std::memory_order_relaxed);
+    FwSignedSizeType diff = static_cast<FwSignedSizeType>(enq) - static_cast<FwSignedSizeType>(deq);
+    FW_ASSERT(diff >= 0, static_cast<FwAssertArgType>(enq), static_cast<FwAssertArgType>(deq));
+    FW_ASSERT(static_cast<FwSizeType>(diff) <= this->m_capacity, static_cast<FwAssertArgType>(diff),
+              static_cast<FwAssertArgType>(this->m_capacity));
+
+    // Approximate size (may be slightly stale in concurrent access)
+    return static_cast<FwSizeType>(diff);
```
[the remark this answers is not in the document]
bevinduckett: Resolved - assert was removed and comment added

Os/Generic/PriorityMemQueue.cpp
```diff
@@ -0,0 +1,777 @@
...
+        // Explicitly destroy AtomicQueues before deallocation
+        for (FwSizeType i = 0; i < arraySize; ++i) {
+            this->m_atomicQueues[i].~AtomicQueue();
+        }
+        allocator.deallocate(allocatorId, this->m_atomicQueues);
+        this->m_atomicQueues = nullptr;
+    }
+    this->m_maxPriority = 0;
+    this->m_allocatorId = 0;
+}
```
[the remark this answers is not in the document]
bevinduckett: Resolved - removed explicit destructor calls

Os/Generic/PriorityMemQueue.cpp
```diff
@@ -0,0 +1,777 @@
...
+    // Only call this in test environments after all queues are destroyed
+    if (s_configsUsed != nullptr) {
+        // Get allocator (same as used in config())
+        Fw::MemAllocator& allocator = Fw::MemAllocatorRegistry::getInstance().getAnAllocator(
+            Fw::MemoryAllocation::MemoryAllocatorType::OS_GENERIC_PRIORITY_QUEUE);
+        FwEnumStoreType allocatorId = 0;
+        
+        // Deallocate the tracking array
+        allocator.deallocate(allocatorId, s_configsUsed);
+        s_configsUsed = nullptr;
```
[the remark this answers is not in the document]
bevinduckett: Resolved - added allocatorId to the static class data and deallocation now uses the same ID as allocation

Os/Generic/PriorityMemQueue.hpp
```diff
@@ -0,0 +1,296 @@
...
+    PriorityMemQueue();
+
+    //! \brief default queue destructor
+    virtual ~PriorityMemQueue();
+
+    //! \brief copy constructor is forbidden
+    PriorityMemQueue(const QueueInterface& other) = delete;
+
+    //! \brief copy constructor is forbidden
+    PriorityMemQueue(const QueueInterface* other) = delete;
```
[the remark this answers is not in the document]
bevinduckett: Resolved by cb51d28bf

Os/Generic/PriorityMemQueue.hpp
bevinduckett: Fixed - all member variables are now initialized

Os/Generic/docs/sdd.md
bevinduckett: Fixed - updated the SDD to strip VxWorks information and add info for AtomicQueue

Os/Generic/PriorityMemQueue.hpp
bevinduckett: Removed comment about JPL, made this comment in the header more visible, increased limit to 32 and improved memory allocation efficiency when a component uses  sparse priorities (e.g. priority 0 and 31)

bevinduckett: Not sure if this comment means that it should be noted somewhere other than header

Os/Generic/test/ut/PriorityMemQueueTests.cpp
bevinduckett: Stripped all UT printfs & added assertion messages

Os/Generic/test/ut/PriorityMemQueueInputValidationTests.cpp
bevinduckett: Resolved - UT was refactored to be more efficient in 3b7525385 & 73d54561e

Os/Generic/PriorityMemQueue.cpp
```diff
@@ -0,0 +1,777 @@
...
+        U32 enabledPriorities = this->m_handle.m_priorityMask.load(std::memory_order_acquire);
+        U32 nonEmptyHint = this->m_handle.m_nonEmptyMask.load(std::memory_order_acquire);
+        
+        // Combine masks: only scan priorities that are both enabled and likely non-empty
+        U32 scanMask = enabledPriorities & nonEmptyHint;
+        
+        // Scan priorities and attempt dequeue
+        if (scanAndDequeue(this->m_handle, destination, capacity, actualSize, priority, scanMask)) {
+            return QueueInterface::Status::OP_OK;
+        }
```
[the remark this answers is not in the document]
bevinduckett: Resolved - nonEmptyMask has been removed, the performance trade off (relatively minor given most messages block rather than poll) was worth the simplification

Os/Generic/docs/sdd.md
bevinduckett: Resolved - updated SDD to replace VxWorks notes with AtomicQueue description

cmake/autocoder/scripts/priority_buffer_analyzer.py
bevinduckett: This script generates a header file to make sizing per priority message queues simpler, see the " Using `priority_buffer_analyzer.py` for Message Size Calculation" section of Os/Generic/docs/sdd.md

bevinduckett (commented):
@thomas-bc  - please kick off the CI jobs when you get a chance

thomas-bc:
To fix the REHL8 build warning, you need to explicitly cast to the appropriate type _after_ the addition

```diff
-         FwQueuePriorityType expectedPriority = 2 - static_cast<FwQueuePriorityType>(i);
+         FwQueuePriorityType expectedPriority = static_cast<FwQueuePriorityType>(2 - i);
```

thomas-bc:
For the print format strings (failing macOS UTs), you can use the `PRI_*` variables, e.g. `PRI_FwSizeType` 

https://github.com/nasa/fprime/blob/ddb970c6d6f4f9f86b402d85475977c46f141e74/Fw/Test/UnitTestAssert.cpp#L47

bevinduckett:
@thomas-bc  - I've pushed an update to fix the CI failures

Os/Generic/PriorityMemQueue.hpp
```diff
@@ -0,0 +1,314 @@
...
+//! \brief AtomicQueue-based priority queue implementation
+//!
+//! \warning This Priority Queue is ISR safe and SMP safe
+//!
+//! An OS-agnostic implementation of a priority queue using AtomicQueue for ISR-safe + SMP-safe operation.
+//! Each priority has its own AtomicQueue. Uses counting semaphores for blocking operations.
+//!
+//! \warning allocates memory using MemAllocator
+
+class PriorityMemQueue : public Os::QueueInterface {
```
lestarch-autobot, line 113: [Design] **must fix** **Human design adjudication required.** This PR introduces a new framework-level `Os::Queue` implementation (`PriorityMemQueue`) with a global static configuration system, a new lock-free data structure (`AtomicQueue`), and a CMake-integrated Python code generator. This is a significant architectural addition to the OSAL layer.

Key design questions requiring human input:
1. Should this coexist alongside `Os::PriorityQueue` or replace it?
2. Is the global static `configure()` pattern appropriate vs. extending the `create()` interface or using FPP-modeled configuration?
3. The dependency on unmerged PR #5204 (`CountingSemaphore`, `MemAllocatorRegistry`) — is that design accepted?
4. Should the `priority_buffer_analyzer.py` live in `cmake/autocoder/scripts/` or be part of `fprime-tools`?

Os/Generic/PriorityMemQueue.cpp
```diff
@@ -0,0 +1,756 @@
...
+            for (FwSizeType q = p + 1; q < currentConfig->numPriorities; ++q) {
+                PriorityMemQueue::QueuePriorityConfig* qConfig = &priorityConfigs[q];
+                FW_ASSERT(pConfig->priority != qConfig->priority, static_cast<FwAssertArgType>(i),
+                          currentConfig->instanceId, pConfig->priority);
+            }
+        }
+    }
+}
+
+void PriorityMemQueue::configure(QueueConfig* queueConfigs,
```
lestarch-autobot, line 291: [Design] **suggestion** The static global configuration pattern (`s_configs`, `s_configsUsed`, `s_configured`) introduces hidden ordering requirements and global mutable state. F Prime typically favors component-level or build-time configuration over runtime global registries.

Concerns: (1) `configure()` must be called before any `create()` — implicit ordering constraint; (2) global statics couple all queue instances; (3) `resetConfig()` hack needed for testing suggests the pattern is test-unfriendly.

Consider passing configuration directly to `create()` via an extended signature or a configuration object, keeping queue instances self-contained.

```suggestion
// Alternative: Accept config directly at create() time via a builder or config parameter
// rather than requiring a global configure() call before any queue creation.
void PriorityMemQueue::configure(QueueConfig* queueConfigs,
```

Os/Generic/PriorityMemQueue.hpp
```diff
@@ -0,0 +1,314 @@
...
+
+// Constants
+namespace Queue {
+// Maximum number of priority levels supported per queue (0-31).
+// Limited to 32 to:
+// - Fit priority bitmask in a 32-bit atomic for efficient lock-free operations
+// - Meet typical F' component needs (most use 2-4 priorities)
+// WARNING: Standard F' components may use priorities outside this range.
+// Ensure FPP priority assignments are within [0, MAX_PRIORITIES-1].
+constexpr static FwSizeType MAX_PRIORITIES = 32;
```
lestarch-autobot, line 31: [Design] **could fix** PR description states "supporting up to 16 priority levels" but the code defines `MAX_PRIORITIES = 32`. The description should be updated to match the implementation, or vice versa, to avoid confusion for users sizing their systems.

Os/Generic/docs/sdd.md
```diff
@@ -4,7 +4,8 @@ This package provides generic implementations of various OSAL modules. These are
 
 Available implementations:
 
-1. [Os::PriorityQueue](#ospriorityqueue)
+1. [Os::PriorityQueue](#ospriorityqueue) - Heap-based priority queue implementation using Os::Mutex for thread safety
+2. [Os::PriorityMemQueue](#osprioritymemqueue) - ISR-safe, heap-based priority queue with per-priority memory pools and configuration
```
lestarch-autobot, line 8: [Documentation] **must fix** `Os/docs/sdd.md` §4 "Generic Services" table lists only `PriorityQueue` but must also list `PriorityMemQueue`.

(Anchored here because `Os/docs/sdd.md` is not in this PR's diff.) The parent OSAL SDD at `Os/docs/sdd.md` line 63 has a table of generic services that currently reads:

| **PriorityQueue** | Heap-based priority queue implementation |

This PR introduces `PriorityMemQueue` as a second generic service. Add an entry so porters discover both queue options.

Os/Generic/CMakeLists.txt
```diff
@@ -1,6 +1,46 @@
...
+  SOURCES
+    "${CMAKE_CURRENT_LIST_DIR}/PriorityMemQueue.cpp"
+  HEADERS
+    "${CMAKE_CURRENT_LIST_DIR}/PriorityMemQueue.hpp"
+  DEPENDS
+    Fw_Types
+    Os_Generic_Types
+    Os_CountingSemaphore
+)
+register_fprime_implementation(
```
lestarch-autobot, line 16: [Documentation] **could fix** `docs/how-to/implement-osal.md` line 239 mentions only `Os_Generic_PriorityQueue`; should note `Os_Generic_PriorityMemQueue` as an ISR-safe alternative.

(Anchored here because `docs/how-to/implement-osal.md` is not in this PR's diff.) The note currently states F´ provides a single platform-independent queue. With this PR, porters now have two options — the existing mutex-based queue and the new ISR-safe atomic queue. A brief update would help platform developers choose.

cmake/autocoder/scripts/priority_buffer_analyzer.py
```diff
@@ -0,0 +1,815 @@
...
+    ComponentInstance as fpm_ComponentInstance,
+)
+from fprime_python_model.semantics import port_instance as fpm_port_instance
+from fprime_python_model.semantics import command as fpm_command
+from fprime_python_model.semantics.interface_instance import (
+    InterfaceComponentInstance as fpm_InterfaceComponentInstance,
+)
+from fprime_python_model.semantics import types_values as fpm_types_values
+from fprime_python_model import fpp_ast as fpm_fpp_ast
+from fprime_python_model.semantics.symbol import Symbol as fpm_Symbol
```
lestarch-autobot, line 40: [Supply Chain] **could fix** `fprime_python_model` is not available on PyPI and is not listed in any dependency manifest (`requirements.txt`, `pyproject.toml`, etc.) in this repository.

This build-time script imports extensively from this package but there is no pinned version or hash, and the package's provenance cannot be verified from the repository alone. The PR description references a dependency on PR #5204 — if that PR introduces this package, the dependency should be tracked in a manifest with a version pin once published.

Until the package is available and pinned, this script will fail with `ModuleNotFoundError` at build time.

Os/Generic/Types/AtomicQueue.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+    // Allocate contiguous buffer memory for all slots
+    FwSizeType totalBufferSize = numBuffers * bufferSize;
+    void* bufferMem = allocator.checkedAllocate(allocatorId, totalBufferSize, 64);
+    FW_ASSERT(bufferMem != nullptr, static_cast<FwAssertArgType>(numBuffers), static_cast<FwAssertArgType>(bufferSize));
+    this->m_bufferMemory = static_cast<U8*>(bufferMem);
+
+    // Initialize all slots with placement new and assign buffer pointers
+    for (FwSizeType i = 0; i < numBuffers; ++i) {
+        FW_ASSERT(i < numBuffers, static_cast<FwAssertArgType>(i), static_cast<FwAssertArgType>(numBuffers));
```
lestarch-autobot: [Security] **suggestion** Integer overflow in allocation size computation.

`numBuffers * bufferSize` and `numBuffers * sizeof(Slot)` can silently overflow on platforms where `FwSizeType` is narrower than the product, leading to under-allocation and subsequent out-of-bounds writes during slot initialization.

```suggestion
    // Allocate slot array (with overflow check)
    FW_ASSERT(numBuffers <= std::numeric_limits<FwSizeType>::max() / sizeof(Slot),
              static_cast<FwAssertArgType>(numBuffers), static_cast<FwAssertArgType>(sizeof(Slot)));
    FwSizeType slotsSize = numBuffers * sizeof(Slot);
    void* slotMem = allocator.checkedAllocate(allocatorId, slotsSize, alignof(Slot));
    FW_ASSERT(slotMem != nullptr, static_cast<FwAssertArgType>(numBuffers), static_cast<FwAssertArgType>(bufferSize));
    this->m_slots = static_cast<Slot*>(slotMem);

    // Allocate contiguous buffer memory for all slots (with overflow check)
    FW_ASSERT(numBuffers <= std::numeric_limits<FwSizeType>::max() / bufferSize,
              static_cast<FwAssertArgType>(numBuffers), static_cast<FwAssertArgType>(bufferSize));
    FwSizeType totalBufferSize = numBuffers * bufferSize;
    void* bufferMem = allocator.checkedAllocate(allocatorId, totalBufferSize, 64);
    FW_ASSERT(bufferMem != nullptr, static_cast<FwAssertArgType>(numBuffers), static_cast<FwAssertArgType>(bufferSize));
    this->m_bufferMemory = static_cast<U8*>(bufferMem);
```

Os/Generic/Types/AtomicQueue.hpp
```diff
@@ -0,0 +1,206 @@
...
+//! used (instead of forcing U64) to support platforms with 32-bit native word size where
+//! 64-bit atomics may not be lock-free or require expensive emulation.
+class AtomicQueue {
+    friend class ::AtomicQueueWrapAroundTest;  // Test-only accessor for counter manipulation
+  public:
+    //! \brief AtomicQueue constructor
+    AtomicQueue();
+
+    //! \brief AtomicQueue destructor
+    ~AtomicQueue();
```
lestarch-autobot, line 58: [C++ Design] **must fix** `AtomicQueue` declares a destructor managing resources (`m_slots`, `m_bufferMemory`, `m_notFullSem`) but does not delete or define copy/move operations (CPP-17).

An accidental copy would shallow-copy the owning pointers, causing double-free on destruction. Since `AtomicQueue` is not intended to be copied or moved, explicitly delete all four operations.

```suggestion
    ~AtomicQueue();

    // Rule of Five: prevent accidental copy/move of resource-managing class
    AtomicQueue(const AtomicQueue&) = delete;
    AtomicQueue& operator=(const AtomicQueue&) = delete;
    AtomicQueue(AtomicQueue&&) = delete;
    AtomicQueue& operator=(AtomicQueue&&) = delete;
```

Os/Generic/PriorityMemQueue.hpp
```diff
@@ -0,0 +1,314 @@
...
+//! \brief critical data stored for priority queue
+//!
+//! The priority queue uses AtomicQueue for ISR-safe + SMP-safe operation.
+//! Each priority has its own AtomicQueue. Priority tracking uses atomic
+//! bitmasks and a counting semaphore for receive notification.
+struct PriorityMemQueueHandle : public QueueHandle {
+    // Sparse priority support: map priority→index into m_atomicQueues
+    // Reduces memory waste when using non-consecutive priorities (e.g., {0, 15, 31})
+    I8 m_priorityMap[Queue::MAX_PRIORITIES];         // Priority→index mapping (-1 = unused)
+    Types::AtomicQueue* m_atomicQueues = nullptr;    // Array sized to actual configured priorities
```
lestarch-autobot: [C++ Design] **must fix** `PriorityMemQueueHandle` holds owning pointers (`m_atomicQueues`, `m_notEmptySem`, `m_highWaterMarks`) but defines neither a destructor nor deleted copy/move operations (CPP-17).

If the handle is ever copied (even implicitly through aggregate initialization), the resulting double-ownership leads to use-after-free. Delete the copy/move ops to make the resource-owning intent explicit.

```suggestion
    Types::AtomicQueue* m_atomicQueues = nullptr;    // Array sized to actual configured priorities

    // Rule of Five: prevent accidental copy/move of resource-owning handle
    PriorityMemQueueHandle(const PriorityMemQueueHandle&) = delete;
    PriorityMemQueueHandle& operator=(const PriorityMemQueueHandle&) = delete;
    PriorityMemQueueHandle(PriorityMemQueueHandle&&) = delete;
    PriorityMemQueueHandle& operator=(PriorityMemQueueHandle&&) = delete;
```

Os/Generic/PriorityMemQueue.cpp
```diff
@@ -0,0 +1,756 @@
...
+    Types::AtomicQueue* atomicQueue =
+        resolvePriorityQueue(this->m_handle, priority, this->m_handle.m_id, s_requirePrioritySizing);
+
+    // Check for sizing problem
+    if (size > atomicQueue->getBufferSize()) {
+        return QueueInterface::Status::SIZE_MISMATCH;
+    }
+
+    // Send message using AtomicQueue
+    bool success;
```
lestarch-autobot, line 652: [C++ Design] **suggestion** Local variable `success` declared without an initializer (CPP-19).

Although both branches of the `if/else` assign it, CPP-19 requires every variable to carry an explicit initializer. Brace-init to `false` is zero-cost and prevents future refactoring mistakes.

```suggestion
    bool success = false;
```

Os/Generic/PriorityMemQueue.cpp
```diff
@@ -0,0 +1,756 @@
...
+    if (queueConfigs != nullptr) {
+        validateQueueConfigs(queueConfigs, numQueueConfigs);
+    }
+    // Get the memory allocator configured for priority queues
+    Fw::MemAllocator& allocator = Fw::MemAllocatorRegistry::getInstance().getAnAllocator(
+        Fw::MemoryAllocation::MemoryAllocatorType::OS_GENERIC_PRIORITY_QUEUE);
+    // Allocate memory for tracking used configurations
+    if (numQueueConfigs > 0) {
+        FwSizeType expSize = numQueueConfigs * sizeof(std::atomic<bool>);
+        s_configsUsed = reinterpret_cast<std::atomic<bool>*>(
```
lestarch-autobot: [C++ Design] **could fix** `reinterpret_cast` from `void*` where `static_cast` is well-defined and preferred (CPP-10).

`static_cast<T*>(void_ptr)` is the standard idiom for converting allocator returns; `reinterpret_cast` is a stronger escape hatch and requires inline justification. This file already uses `static_cast` from `void*` in `AtomicQueue.cpp:68`; prefer consistency.

```suggestion
        s_configsUsed = static_cast<std::atomic<bool>*>(
```

lestarch-autobot (changes requested):

### Per-agent results

| Security Vulnerabilities | 0 | 1 | 0 | 0 | 1 | Go |
| F Prime C/C++ Design | 2 | 1 | 1 | 0 | 4 | No-Go |
| Documentation Currency | 1 | 0 | 1 | 0 | 2 | No-Go |
| Design | 1 | 1 | 1 | 0 | 3 | No-Go |
| **Totals** | 4 | 3 | 4 | 0 | 11 | **No-Go** |

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | 1 could-fix — fprime_python_model not in any manifest and not on PyPI |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>
<summary>Outstanding must-fix items (4)</summary>

**F Prime C/C++ Design**
- `AtomicQueue` declares a destructor but does not delete or define copy/move operations (CPP-17) — [link](https://github.com/nasa/fprime/pull/5205#discussion_r3437301240)
- `PriorityMemQueueHandle` holds owning pointers but defines neither destructor nor deleted copy/move operations (CPP-17) — [link](https://github.com/nasa/fprime/pull/5205#discussion_r3437301258)

**Documentation Currency**
- `Os/docs/sdd.md` §4 "Generic Services" table must also list `PriorityMemQueue` — [link](https://github.com/nasa/fprime/pull/5205#discussion_r3437286823)

**Design**
- **Human design adjudication required.** New framework-level `Os::Queue` implementation with global static config, lock-free data structure, and CMake-integrated code generator — [link](https://github.com/nasa/fprime/pull/5205#discussion_r3437283775)

</details>

### Merge readiness
**Merge readiness: No-Go** — F Prime C/C++ Design has 2, Documentation Currency has 1, and Design has 1 outstanding must-fix items.

---

Priority queues for priority missions — a few items to address before this is flight-ready.

thomas-bc:
I am getting AddressSanitizers errors in the UTs on macOS

```
[ RUN      ] PriorityMemQueueInputValidation.MaxPriorityEdgeAllConfigs
=================================================================
==94543==ERROR: AddressSanitizer: stack-use-after-scope on address 0x00016f96fd40 at pc 0x000100668c24 bp 0x00016f96f420 sp 0x00016f96f418
READ of size 4 at 0x00016f96fd40 thread T0
```
```
Address 0x00016f96fd40 is located in stack of thread T0 at offset 1056 in frame
    #0 0x0001005ea1c8 in PriorityMemQueueInputValidation_MaxPriorityEdgeAllConfigs_Test::TestBody() PriorityMemQueueInputValidationTests.cpp:187
```
```
HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
      (longjmp and C++ exceptions *are* supported)
SUMMARY: AddressSanitizer: stack-use-after-scope PriorityMemQueue.cpp:536 in Os::Generic::PriorityMemQueue::teardownInternal()
Shadow bytes around the buggy address:
```

thomas-bc:
~~Coverage failure is due to a bug in gcov, believe it or not~~ Or maybe the same failure as the macOS sanitizer... just confusing since ubuntu UTs do pass

```
Got negative hit value in gcov line 'branch  1 taken -103 (fallthrough)' caused by a
bug in gcov tool, see
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=68080. Use option
--gcov-ignore-parse-errors with a value of negative_hits.warn,
or negative_hits.warn_once_per_file.
```

Os/Generic/Types/AtomicQueue.cpp
thomas-bc: I guess I'm confused because `m_id` is not used. I hear that this applies to `PriorityMemQueue`, but this isn't used in AtomicQueue and I wonder if it is a remnant of a past implementation // wonder if we should simply scrub it. I've confirmed locally that removing it -> everything seems to work fine.

Os/Generic/PriorityMemQueue.cpp
```diff
@@ -0,0 +1,756 @@
...
+    }
+    this->m_handle.m_notEmptySem = new (semMem) Os::CountingSemaphore(static_cast<U32>(0));
+
+    // Create the priority queues based on configuration
+    if (queueConfig != nullptr) {
+        // Use the found configuration to create multiple priority queues
+        return createConfiguredQueues(queueConfig, allocator, id);
+    } else {
+        // No configuration found, create a single default priority queue
+        return createDefaultQueue(depth, messageSize, allocator, id);
```
thomas-bc, line 434: Not a blocker for now, but architecturally there is a bit of a concern: FPP-specified queue size is being ignored when a C++ `queueConfig` is specified. This is the best we can do without FPP support, but we should capture a follow-up ticket for CCB to think through this. Can tie in with https://github.com/nasa/fprime/issues/3512

Os/Generic/Types/AtomicQueue.cpp
thomas-bc: but I also get that they don't really mean the same thing, so up to you.

Os/Generic/PriorityMemQueue.cpp
```diff
@@ -0,0 +1,756 @@
...
+
+        for (I32 p = this->m_handle.m_maxPriority; p >= 0; --p) {
+            FwQueuePriorityType testPriority = static_cast<FwQueuePriorityType>(p);
+            if ((enabledPriorities & priorityBitMask(testPriority)) == 0) {
+                continue;
+            }
+            // Look up priority in sparse map
+            I8 index = this->m_handle.getPriorityIndex(testPriority);
+            if (index < 0) {
+                continue;  // Priority not configured
```
thomas-bc: Should priority not configured after initialization be allowed? Should it assert instead?

thomas-bc (commented):
Ok - this looks great! Just a last few questions. I'll trigger Devin with Opus 4.8 and confirm we're all set here

thomas-bc:
I will also try to figure out why Coverage is reporting UT failures even though the UT checks are passing...

cmake/autocoder/scripts/priority_buffer_analyzer.py
thomas-bc: ok for now - this is a best-effort helper

Os/Generic/PriorityMemQueue.hpp
thomas-bc: CCB approved

cmake/autocoder/scripts/priority_buffer_analyzer.py
lestarch-autobot: [Supply Chain] **Improperly resolved.** This finding still applies on f02af8b.

`cmake/autocoder/scripts/priority_buffer_analyzer.py` is unchanged at the current head: it still imports `fprime_python_model`, which is not on PyPI and not pinned in any manifest, so provenance remains unverified and the build-time import is unresolved. (Noted the maintainer reply; leaving the call to a code owner.)

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

Os/Generic/PriorityMemQueue.cpp
lestarch-autobot: [C++ Design] **Improperly resolved.** This finding still applies on f02af8b.

`Os/Generic/PriorityMemQueue.cpp` is unchanged at the current head: the local is still declared `bool success;` (line 627) without an initializer (CPP-19). Resolving the thread did not add the brace-init.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

Os/Generic/docs/sdd.md
lestarch-autobot: [Documentation] **Improperly resolved.** This finding still applies on f02af8b.

`Os/docs/sdd.md` is unchanged at the current head: the §4 "Generic Services" table at line 63 still lists only `PriorityQueue` and omits the new `PriorityMemQueue`. Resolving the thread did not update the SDD.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

Os/Generic/CMakeLists.txt
lestarch-autobot: [Documentation] **Improperly resolved.** This finding still applies on f02af8b.

`docs/how-to/implement-osal.md` is unchanged at the current head: line 239 still references only `Os_Generic_PriorityQueue` and does not mention the new `Os_Generic_PriorityMemQueue`. Resolving the thread did not update the guide.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

Os/Generic/PriorityMemQueue.hpp
lestarch-autobot: [Design] **Improperly resolved.** This finding still applies on f02af8b.

Re-checked `Os/Generic/PriorityMemQueue.hpp`/`.cpp` at the current head: the new framework-level `Os::Queue` implementation, the global static `configure()` pattern, and the `cmake/autocoder/scripts` code-generator placement are all unchanged, so the design questions raised here are still open in the code. The thread was resolved but no code change addresses them.

Os/Generic/PriorityMemQueue.hpp
lestarch-autobot: [Design] **Improperly resolved.** This finding still applies on f02af8b.

`Os/Generic/PriorityMemQueue.hpp` is unchanged at the current head: the code still defines `MAX_PRIORITIES = 32` while the PR description still says "up to 16 priority levels". Resolving the thread did not reconcile the mismatch.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

lestarch-autobot (changes requested):

### Per-agent results

| Security Vulnerabilities | 0 | 1 | 0 | 0 | 1 | Go |
| F Prime C/C++ Design | 2 | 1 | 1 | 0 | 4 | No-Go |
| Documentation Currency | 1 | 0 | 1 | 0 | 2 | No-Go |
| Design | 1 | 1 | 1 | 0 | 3 | No-Go |
| **Totals** | 4 | 3 | 4 | 0 | 11 | **No-Go** |

<details>

| Security Vulnerabilities | 0 | 1 | 0 | 0 | 0 | 0 |
| Supply Chain / Runner Safety | 0 | 1 | 0 | 0 | 1 | 0 |
| F Prime C/C++ Design | 0 | 4 | 0 | 0 | 1 | 0 |
| Documentation Currency | 0 | 2 | 0 | 0 | 2 | 0 |
| Design | 0 | 3 | 0 | 0 | 2 | 0 |
| Test Quality | 0 | 0 | 0 | 0 | 0 | 0 |

</details>

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | 1 could-fix — fprime_python_model not in any manifest and not on PyPI |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>
<summary>Outstanding must-fix items (4)</summary>

**F Prime C/C++ Design**
- `AtomicQueue` declares a destructor but does not delete or define copy/move operations (CPP-17) — [link](https://github.com/nasa/fprime/pull/5205#discussion_r3437301240)
- `PriorityMemQueueHandle` holds owning pointers but defines neither destructor nor deleted copy/move operations (CPP-17) — [link](https://github.com/nasa/fprime/pull/5205#discussion_r3437301258)

**Documentation Currency**
- `Os/docs/sdd.md` §4 "Generic Services" table must also list `PriorityMemQueue` — [link](https://github.com/nasa/fprime/pull/5205#discussion_r3437286823)

**Design**
- **Human design adjudication required.** New framework-level `Os::Queue` implementation with global static config, lock-free data structure, and CMake-integrated code generator — [link](https://github.com/nasa/fprime/pull/5205#discussion_r3437283775)

</details>

### Merge readiness
**Merge readiness: No-Go** — F Prime C/C++ Design has 2, Documentation Currency has 1, and Design has 1 outstanding must-fix items. Six previously-resolved threads still carry live findings on this head; improper-resolution replies were posted for maintainer adjudication (the token could not un-resolve, so the resolution state remains the maintainer's call).

---

Course held since last orbit: the same four must-fix beacons are still lit — clear them and we're cleared for final approach.

Os/Generic/PriorityMemQueue.cpp
thomas-bc: I will note that this has been causing lots of issues in UTs - thankfully AI solved them fairly easily, but it shows that the pattern is a bit of a footgun

Since this is configure-code, could we make copies instead of relying on globals to outlive the queues?

> the borrow-not-copy contract of configure() is a footgun — worth either documenting loudly or having configure() copy into owned storage, but that's a design call for the PR author since it changes flight-side memory behavior.

thomas-bc: cc @bevinduckett

Os/Generic/PriorityMemQueue.hpp
lestarch-autobot: [Design] **Adjudicated.** Design question resolved by @thomas-bc (CCB approved) on f02af8b.

The `design-needs-human-adjudication` finding exists to route the design direction — a new framework-level `Os::Queue` implementation, the global static `configure()` pattern, the dependency on #5204, and the `cmake/autocoder/scripts` code-generator placement — to a code owner *before* deeper review is worthwhile. A core maintainer has now CCB-approved it, so this design gate is cleared and I'm closing out the finding. (Supersedes my prior improperly-resolved reply on this thread, which predated accounting for the CCB approval.)

Os/Generic/test/ut/PriorityMemQueueTests.cpp
```diff
@@ -0,0 +1,2208 @@
...
+
+    // Run for 100ms
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    stopThreads.store(true);
+
+    senderThread.join();
+    togglerThread.join();
+
+    // Verify no crash and some operations succeeded
+    EXPECT_GT(sendAttempts.load(), 0);
```
lestarch-autobot, line 1974: [Test Quality] **suggestion** Concurrent enable/disable test asserts only harness state, not queue behavior.

`sendSuccesses` records whether the queue actually accepted sends while priority 1 was toggled, but it is never asserted; the lone `EXPECT_GT(sendAttempts...)` only confirms the sender loop ran. The test would still pass if every `send()` failed. Assert the observable outcome too (the TSAN/no-crash value is unaffected).

```suggestion
    EXPECT_GT(sendAttempts.load(), 0);
    EXPECT_GT(sendSuccesses.load(), 0) << "queue accepted no sends during concurrent enable/disable";
```

lestarch-autobot (changes requested):

### Per-agent results

| Security Vulnerabilities | 0 | 1 | 0 | 0 | 1 | Go |
| F Prime C/C++ Design | 2 | 1 | 1 | 0 | 4 | No-Go |
| Documentation Currency | 1 | 0 | 1 | 0 | 2 | No-Go |
| Design | 1 | 1 | 1 | 0 | 1 | Go |
| **Totals** | 4 | 4 | 4 | 0 | 10 | **No-Go** |

<details>

| Security Vulnerabilities | 0 | 1 | 0 | 0 | 0 | 0 |
| Supply Chain / Runner Safety | 0 | 1 | 0 | 0 | 0 | 0 |
| F Prime C/C++ Design | 0 | 4 | 0 | 0 | 0 | 0 |
| Documentation Currency | 0 | 2 | 0 | 0 | 0 | 0 |
| Design | 2 | 1 | 0 | 0 | 0 | 0 |
| Test Quality | 0 | 0 | 1 | 0 | 0 | 0 |

</details>

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | 1 could-fix — fprime_python_model not in any manifest and not on PyPI |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>
<summary>Outstanding must-fix items (3)</summary>

**F Prime C/C++ Design**
- `AtomicQueue` declares a destructor managing resources but does not delete or define copy/move operations (CPP-17) — [link](https://github.com/nasa/fprime/pull/5205#discussion_r3437301240)
- `PriorityMemQueueHandle` holds owning pointers but defines neither a destructor nor deleted copy/move operations (CPP-17) — [link](https://github.com/nasa/fprime/pull/5205#discussion_r3437301258)

**Documentation Currency**
- `Os/docs/sdd.md` §4 "Generic Services" table must also list `PriorityMemQueue` — [link](https://github.com/nasa/fprime/pull/5205#discussion_r3437286823)

</details>

### Merge readiness
**Merge readiness: No-Go** — F Prime C/C++ Design has 2 and Documentation Currency has 1 outstanding must-fix item. Design cleared its prior must-fix this run; CI safety remains Go (security and supply-chain both clean).

---

Two decks still flagged on the board — seal the copy/move hatches and log `PriorityMemQueue` in the SDD, and this one's cleared for final approach.

Os/Generic/Types/test/ut/AtomicQueue/AtomicQueueTest.cpp
```diff
@@ -0,0 +1,741 @@
...
+    void deallocate(const FwEnumStoreType identifier, void* ptr) override {
+        (void)identifier;
+        (void)ptr;
+    }
+};
+
+// Partial failure allocator - fails after N successful allocations
+class PartialFailAllocator : public Fw::MemAllocator {
+  public:
+    PartialFailAllocator(U32 failAfter) : m_failAfter(failAfter), m_allocCount(0) {}
```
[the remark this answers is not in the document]
bevinduckett: Added

Os/Generic/Types/AtomicQueue.cpp
bevinduckett: got it - I was conflating how it was used in PriorityMemQueue. Agreed, m_id is redundant since the allocator ID is already being stored, I've removed it

Os/Generic/PriorityMemQueue.cpp
bevinduckett: I don't disagree with the autobot comment - nor that it's a foot-gunny pattern.  But the goal of PriorityMemQueue is to not change the Queue interface, so it can't be added to create like AI suggests. 
Copying the config locally is feasible - it's not memory efficient but it's a relatively small array of small structs so that seems reasonable.

bevinduckett: Updated to copy config

Os/Generic/PriorityMemQueue.cpp
bevinduckett: Created https://github.com/nasa/fprime/issues/5359 & added a comment on #3512 linking the two

Os/Generic/PriorityMemQueue.cpp
bevinduckett: yes - will this to an assert

thomas-bc (approved):

thomas-bc:
Zephyr CI has been acting up in the past few days - failures here can be ignored.

PR #5210, opened 2026-05-28
AosDeframer Integer Overflow Issue

bitWarrior:
| | |
|:---|:---|

---
## Change Description

This PR improves the robustness of packet size arithmetic in AosDeframer and adds a defensive assertion to protect the  memcpy of the packet header from corrupting memory. This PR also expands test coverage for malformed EPP `lol=4` inputs. It also resolves a pre-existing `FIXME` comment that had been tracking this area as needing attention.

## Rationale

**Integer promotion fix in `sizeEppPacket`**

The expression `headerLength + packetDataLength` was computed in `U32` arithmetic regardless of the declared type of `totalPacketSize`. In C++, `U8 + U32` is always evaluated as `U32`; the result is only widened when stored. On platforms where `FwSizeType` is 64 bits, a `packetDataLength` value near `U32_MAX` would silently wrap to a small integer before being stored, causing a severely undersized allocation later in `appendToSpanningPacket`.

Fixed by casting both operands to `FwSizeType` before the addition so the arithmetic is performed at the correct width on all targets:

```cpp
// Before
totalPacketSize = headerLength + packetDataLength;

// After
totalPacketSize = static_cast<FwSizeType>(headerLength)
               + static_cast<FwSizeType>(packetDataLength);
```

The existing wraparound guard (`packetDataLength > max<FwSizeType>() - headerLength`) is retained. It is the primary protection on 32-bit flight targets where `FwSizeType` is also `U32`; without it the cast-safe addition would still wrap on those platforms.

**Defensive assertion before `memcpy` in `appendToSpanningPacket`**

Added an `FW_ASSERT` immediately before the `memcpy` that copies the accumulated header bytes into the dynamically allocated packet buffer:

```cpp
FW_ASSERT(vc.spanningPacket.bytesReceived <= vc.spanningPacket.buffer.getSize(), ...);
::memcpy(vc.spanningPacket.buffer.getData(),
         vc.spanningPacket.headerBuf,
         vc.spanningPacket.bytesReceived);
```

This converts any future regression in the sizing path into an immediate, diagnosable assert failure rather than silent memory corruption — important given the safety-critical deployment context.

**Resolved `FIXME` comment**

Removed the pre-existing `FIXME` that flagged `packetSize` as untrusted and noted the need for a mission-defined limit. Replaced with a `TODO` pointing to the specific additions required (a `ComCfg::AosMaxPacketSize` constant and an `OversizedPacket` FPP event) to re-enable a hard size cap as a further layer of defense.

**`#include <limits>` added**

Required by the `std::numeric_limits<FwSizeType>` call in the overflow guard.

**`sizeSppPacket` — comment added**

The analogous addition in `sizeSppPacket` is safe today because the SPP data-length field is 16 bits (max sum well within `U32`). A comment documents this assumption so a future refactor does not silently break it.

Execute the new unit test that verifies proper handling of an overflow condition. 

## Future Work

None.

AI was used to generate the new unit tests to verify the proper handling of an overflow condition. It also helped generate this PR.

### DIFF
--- a/Svc/Ccsds/AosDeframer/AosDeframer.cpp
+++ b/Svc/Ccsds/AosDeframer/AosDeframer.cpp
@@ -9,9 +9,9 @@
 // - Space Packet Protocol (SPP) extraction (CCSDS 133.0-B-2)
 // - Encapsulation Packet Protocol (EPP) extraction (CCSDS 133.1-B-3)
 // ======================================================================
+#include "Svc/Ccsds/AosDeframer/AosDeframer.hpp"
 #include <cstring>
 #include <limits>
-#include "Svc/Ccsds/AosDeframer/AosDeframer.hpp"
 #include "Svc/Ccsds/Types/EppLengthOfLengthEnumAc.hpp"
 #include "Svc/Ccsds/Types/EppProtocolIdEnumAc.hpp"
 #include "Svc/Ccsds/Types/SpacePacketHeaderSerializableAc.hpp"
--- a/Svc/Ccsds/AosDeframer/AosDeframer.cpp
+++ b/Svc/Ccsds/AosDeframer/AosDeframer.cpp
@@ -10,11 +10,8 @@
 // - Encapsulation Packet Protocol (EPP) extraction (CCSDS 133.1-B-3)
 // ======================================================================
 
-include "Svc/Ccsds/AosDeframer/sDemer#include <cstri
-1
-
-
-ng>#include <limits>
+#include <cstring>
+#include <limits>
 #include "Svc/Ccsds/Types/EppLengthOfLengthEnumAc.hpp"
 #include "Svc/Ccsds/Types/EppProtocolIdEnumAc.hpp"
 #include "Svc/Ccsds/Types/SpacePacketHeaderSerializableAc.hpp"
--- a/Svc/Ccsds/AosDeframer/AosDeframer.cpp
+++ b/Svc/Ccsds/AosDeframer/AosDeframer.cpp
@@ -319,25 +319,10 @@ FwSizeType AosDeframer::appendToSpanningPacket(AosDeframerVc& vc, U8* data, FwSi
             return 0;
         }
 
-        // Reject packets that exceed the mission-defined maximum before touching the allocator.
-        // packetSize is derived from an untrusted length field read straight off the wire
-        // (especially EPP lol=4, which can claim up to ~4 GB). Capping here turns an implausible
-        // size into an explicit, telemetered rejection rather than relying on the allocator to
-        // refuse the request -- that fallback only emits a generic SpanningPacketAllocFailed and
-        // depends on the allocator pool being smaller than the claimed size, which is not a
-        // guarantee on all targets.
-        if (packetSize > AosDeframer_MaxPacketSize) {
-            this->log_WARNING_HI_OversizedPacket(vc.virtualChannelId, vc.spanningPacket.context.get_pvn(), packetSize,
-                                                 AosDeframer_MaxPacketSize);
-            // The claimed size is untrusted, so we cannot reliably seek past the packet;
-            // abandon any accumulated header state and stop processing this frame.
-            this->abandonSpanningPacket(vc);
-            return 0;
-        }
-
-        // Try to allocate a buffer for the whole packet
+        // Try to allocate a buffer for the whole packet. If this size is invalid (too large) or if the buffer
+        // manager is out of memory, this is handled below.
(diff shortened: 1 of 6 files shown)


Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -491,6 +502,9 @@ FwSizeType AosDeframer::sizeSppPacket(U8* payloadStart, FwSizeType payloadSize)
     }
 
     // Per CCSDS 133.0-B-2 Section 4.1.3.5.2, packet data length = (actual length - 1)
+    // packetDataLength is a 16-bit field (max 65535); SERIALIZED_SIZE is a small constant.
```
LeStarch, line 498: Add a static assertion to guarantee this.

Svc/Ccsds/AosDeframer/AosDeframer.cpp
```diff
@@ -313,14 +314,18 @@ FwSizeType AosDeframer::appendToSpanningPacket(AosDeframerVc& vc, U8* data, FwSi
...
         // packet)
-        // FIXME: packetSize is untrusted (read straight from the wire) and could have unintended side effects,
-        // especially with EPP. We should skip if packetSize exceeds a mission-defined limit instead of letting the
-        // allocator handle any requests, which has a side effect of abandoning spanning packets on failure
         const FwSizeType packetSize = sizePacket(vc, vc.spanningPacket.headerBuf, vc.spanningPacket.bytesReceived);
         if (packetSize == 0) {
             return 0;
         }
 
+        // TODO(security): Add a mission-defined upper bound check here once
```
LeStarch: Why add a todo?  Why not just fix it?

LeStarch (changes requested):

LeStarch:
@Willmac16

Svc/Ccsds/AosDeframer/AosDeframer.cpp
bitWarrior: Added static assert

Svc/Ccsds/AosDeframer/AosDeframer.cpp
bitWarrior: Fixed TODO

Svc/Ccsds/AosDeframer/AosDeframer.fpp
```diff
@@ -1,15 +1,23 @@
...
         constant NumVcs = 1
 
+        @ Mission-defined upper bound on a single reassembled packet, in bytes.
+        @ Packets whose computed size exceeds this are rejected before allocation and
+        @ reported via the OversizedPacket event. This is a sanity ceiling against
+        @ corrupt/untrusted length fields (notably EPP lol=4, which can claim ~4 GB),
+        @ not a tight limit. Missions MUST tune this to exceed their largest legitimate
+        @ packet while staying well below sizes that indicate corruption or attack.
+        @ Default: 256 MiB.
+        constant MaxPacketSize = 0x10000000
```
LeStarch: If missions must tune this, then it belongs in a mission specific config.

Svc/Ccsds/AosDeframer/AosDeframer.fpp
```diff
@@ -1,15 +1,23 @@
 module Svc {
-module Ccsds {
+    module Ccsds {
```
LeStarch: This is a negative change to the formatting.

Svc/Ccsds/AosDeframer/AosDeframer.fpp
```diff
@@ -22,7 +30,7 @@ module Ccsds {
         include "AosDeframerTelem.fppi"
 
         ###############################################################################
-        # Standard AC Ports: Required for Channels, Events, Commands, and Parameters  #
+#Standard AC Ports : Required for Channels, Events, Commands, and Parameters #
```
LeStarch: Another bad formatting change.

Svc/Ccsds/AosDeframer/AosDeframer.fpp
```diff
@@ -41,7 +49,6 @@ module Ccsds {
 
         @Port to set the value of a parameter
         param set port prmSetOut
-
     }
-}
+    }
```
LeStarch: And here.

LeStarch (dismissed):
Made the fixes myself.

LeStarch (dismissed):

default/config/AosDeframerCfg.fpp
thomas-bc, line 1: empty file? should remove

LeStarch (approved):

Issue #5220, opened 2026-06-01
fprime-prm-write GDS tool does not create CRC32 header in PrmDb .dat file


Specky26846:
| | |
|:---|:---|
|**_F´ Version_**| 4.2.0 |
|**_Affected Component_**| fprime-prm-write (tool) |
---
## Problem Description

The `fprime-prm-write` tool for creating a .dat file input for PrmDb does not add the necessary CRC32 checksum header for each component instance. In params.py (in fprime-gds), the `parsed_json_to_dat()` function needs to be updated to append checksums in the .dat binary file, as required by the PrmDb component.

## Context / Environment

```
Operating System: Darwin
CPU Architecture: arm64
Platform: macOS-26.5-arm64-arm-64bit-Mach-O
Python version: 3.14.5
CMake version: 3.26.0
Pip version: 26.1.1
Pip packages:
    fprime-tools==4.2.0
    fprime-gds==4.2.0
    fprime-fpp==3.3.0a3
```

## How to Reproduce

Generally followed the instructions from the [fprime-prm-write README](https://github.com/nasa/fprime-gds/tree/devel/src/fprime_gds/common/tools).

1.  Created a JSON file for some of the parameters of the Ref project from the `RecvBuffApp`, `SendBuffApp`, and `TypeDemo` components:
```
{
      "Ref.recvBuffComp": {
          "parameter1": 20,
          "parameter2": -5
      },
  
      "Ref.sendBuffComp": {
          "parameter3": 25,
          "parameter4": 99.99
      },
      
      "Ref.typeDemo": {
          "CHOICE_PRM": "RED",
          "CHOICES_PRM": ["ONE", "BLUE"],
          "CHOICE_PAIR_PRM": {
              "firstChoice": "RED",
              "secondChoice": "BLUE"
        }
    }
}
```
2. I ran `fprime-prm-write dat <json file> --dictionary <path to compiled FPrime dict>` to create the PrmDb input .dat file.
3. I ran the F Prime GDS and used the FileHandling.prmDb.PRM_LOAD_FILE command to import my new .dat file into the GDS.

## Expected Behavior

I saw the following two warnings in the Events tab of the GDS:

2026-06-01T18:47:38.078Z | 0x500300c | FileHandling.prmDb.PrmFileBadCrc | WARNING_HI | Parameter file failed CRC. Read: 0xa5000000 Computed: 0x7692ff28
-- | -- | -- | -- | --
[...]

PR #5229, opened 2026-06-03
Add PrmDb integration test workflow

Specky26846:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5220 |

---
## Change Description

Adds a new workflow for testing parameters in the GDS. Currently this workflow sets up F Prime/Python, builds the Ref project, creates a test parameter JSON file, uses the `fprime-prm-write` tool to create a dat file, inputs it into the GDS, and checks that the parameter load was successful.

## Rationale

Creates a common integration test workflow for GDS features, such as parameters (included now), and later Data Products.

More tests could be written to test complex parameter types. This should also be tested against a change to the Ref project. Additional ideas in section below.

## Future Work

The next step would be to add a further check to ensure that the new PrmDb.dat file parameters match the input. A similar test for the `fprime-prm-write seq` subcommand may be useful as well. Further work may include Data Products integration testing.

AI  was used to create an outline for this .yml file, and I filled in the implementation manually as a learning exercise.


.github/workflows/ext-gds-integration.yml
thomas-bc, line 1: I would suggest renaming this file `ext-gds-integration.yml` so that we can use it as a common InT workflow for GDS things, such as parameters, but also Data Products down the line, etc.

.github/workflows/prmdb-integration-test.yml
```diff
@@ -0,0 +1,156 @@
+name: PrmDb Integration Test
```
thomas-bc:
```suggestion
name: GDS Integration
```

.github/workflows/prmdb-integration-test.yml
```diff
@@ -0,0 +1,156 @@
...
+              echo "GDS failed to start"
+              exit 1
+            fi
+
+      # Step 8: Load parameters into PrmDb
+      - name: "Load Parameters via GDS"
+        working-directory: ./Ref
+        run: |
+          # Sending GDS fprime-cli commands
+          fprime-cli commands send FileHandling.prmDb.PRM_LOAD_FILE "test_params.dat" MERGE
```
thomas-bc: this doesn't look like the right command, should be
`fprime-cli command-send`

.github/workflows/prmdb-integration-test.yml
```diff
@@ -0,0 +1,156 @@
...
+          exit 1
+          fi
+          if grep -q "PrmFileBadCrc" gds.log; then
+            echo "ERROR: Found PrmFileBadCrc"
+            cat gds.log
+            exit 1
+          fi
+
+      # Step 10: Cleanup
+      - name: "Stop GDS"
```
thomas-bc: This is handled by GitHub Actions - shouldn't be needed

.github/workflows/prmdb-integration-test.yml
```diff
@@ -0,0 +1,156 @@
...
+        working-directory: ./Ref
+        run: |
+          fprime-util generate -f
+          fprime-util build
+
+      # Step 5: Create test parameter file
+      - name: "Create Test Parameter JSON"
+        working-directory: ./Ref
+        run: |
+          # TODO: Create a simple JSON file with test parameters
```
thomas-bc: there's a few TODOs in this file that looks like they can be removed

thomas-bc (changes requested):

.github/workflows/ext-gds-integration.yml
```diff
@@ -0,0 +1,141 @@
...
+            }
+          }
+          EOF
+
+      # Step 6: Generate .dat file
+      - name: "Generate Parameter .dat File"
+        working-directory: ./TestDeploymentsProject/Ref
+        run: |
+          fprime-prm-write dat test_params.json \
+          --dictionary build-artifacts/Linux/Ref/dict/RefTopologyDictionary.json \
```
thomas-bc, line 72: path is outdated - needs TestDeploymentsProject

I'd encourage you to test this workflow in a fork by having the following to the top of your YAML file:

```
on:
  push:
    branches: [ devel, release/**, patch-1 ]
  pull_request:
    branches: [ devel, release/** ]
```

This will allow you to run the workflow on your fork without having to wait for us to approve CI to run.

thomas-bc (changes requested):

thomas-bc:
Closing for now to not overuse our CI runners. Please reopen when the CI is passing on your fork (see above recommendation)

Issue #4941, opened 2026-04-02
Implement a Data Product Compression Processing component


kubiak-jpl:
| | |
|:---|:---|
|**_F´ Version_**| 4.2 |
|**_Affected Component_**| DpWriter |
---
## Feature Description

A data product processing component that is called by DpWriter and compresses the data portion of a Data Product. This work would be split into two components that I'm calling `DpCompressProc` and `ZlibCompress`.

DpCompressProc would be responsible for
- Splitting the received product into smaller chunks for compression
- Sending these chunks to a compression component which would compress this data
- Re-formatting the data section with smaller compressed sections
- Dealing with chunks of a data that cannot be compressed
- Perform this processing in place in the data product

ZlibCompress
- Compress received chunks using zlib (DEFLATE algorithm, algorithm behind gzip)
  - Zlib was chosen because it is widely available. It would also be good to implement another backend with better compression ratios such as lzma.

## Rationale

1. Serve as a reference implementation of a data product processing component
2. Provide a plug-and-play system for compressing data products
3. Work through any issues with decompressing the products inside the fprime data product decoding tools

kubiak-jpl:
I've implemented a proof of concept in my branch here
https://github.com/kubiak-jpl/fprime-kubiak/tree/dev/dpcompressproc

This branch adds `DpCompressProc` and `DpZLibCompressor` components which compress the data section of a data product. I updated the Ref deployment in that branch so that the DpDemo can request a compressed data product.

Here's a screenshot with some of the EVRs. ZLib deflate was able to compress the product data from 1212 bytes to 263 bytes.
<img width="2549" height="1234" alt="Image" src="https://github.com/user-attachments/assets/818c45e7-d994-4d19-b170-c8909eeb8f14" />

I also updated the data product decoding scripts to automatically decode these products here
https://github.com/kubiak-jpl/fprime-gds-kubiak/tree/dev/compressed_data_products

The raw data product without decompressing it
```
$ fprime-dp decode -b Dp_00002576_1775763717_00163306.fdp -d ../build-artifacts/Linux/Ref/dict/RefTopologyDictionary.json -o compressed_product -z
$ cat compressed_product
...
  "Records": [
    {         
      "Record": {
        "record_id": 67125248,
        "record_name": "DataProducts.dpCompressProc.CompressionRecord",
        "record_type_name": "U8Type",
        "record_type": "<class 'fprime_gds.common.models.serialize.numerical_types.U8Type'>",
        "is_array": true,
[...]

PR #5235, opened 2026-06-04
Add Lossless Data Product Compression Components

kubiak-jpl:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4941 |

---
## Change Description

This PR makes the following changes
- Adds the DpCompressProc component
  - This component plugs into the DpWriter Proc interface and is responsible for creating compressed data products. It uses a port call to a second compression component for the actual compression
- Adds the DpZLibCompressor component
  - Uses libz to compress chunks of data
  - DpZLibCompressor requires libz to compile. This is a very common library, even in embedded contexts, but I currently don't have any CMake stuff to guard against it. @LeStarch Any thoughts here?
- Adds compressed data product pipeline to TestDeployment and DpDemo subtopology
- Adds command arguments to DpDemo to trigger compression of demo data products via command
- Updates SerializeFrom to use `memmove` instead of `memcpy`
  - DpCompressProc re-serializes data into an existing buffer and can encounter situations where memory overlap occurs between input buffer and the destination serialized buffer. In those situations memcpy would execute with overlapping buffers and trigger undefined behavior
  - This change allows me to use `serializeFrom` method calls which make the algorithm to re-write the data product cleaner. However, I can also revert this change and modify my code to use `memmove` directly

## Rationale

Implements lossless compressible data products seamlessly into the F Prime Data Product subsystem

The algorithm to modify the data product in place is non-trivial. I believe I have sufficiently thought through the state machine and have unit tests to cover all the cases. But it could use thorough review.

I had this integrated into the Ref deployment before that was moved.

## Future Work

1. Implement a compressor using LZMA for better compression ratios

N/A


LeStarch (dismissed):

kubiak-jpl:
@LeStarch Looks like some of the failures are due to missing zlib.h. Do you want me to find a solution to compiling on platforms without zlib?

kubiak-jpl:
I'm a little confused by this error in the RHEL8 build
```
/__w/fprime/fprime/Svc/DpCompressProc/DpCompressProc.cpp:358:31: error: conversion from ‘int’ to ‘FwSizeStoreType’ {aka ‘short unsigned int’} may change value [-Werror=conversion]
             uncompressed_size += chunk_size;
```
Because both `uncompress_size` and `chunk_size` are `FwSizeStoreType`

thomas-bc:
@kubiak-jpl this is a type of warning we're used to seeing on the RHEL8 compilers. `FwSizeStoreType` is a U16 but the addition in `+=` auto-promotes it to U32, so there's an implicit cast down back into U16. I think `uncompressed_size = static_cast<FwSizeStoreType>(uncompressed_size + chunk_size)` should fix it

kubiak-jpl:
> @kubiak-jpl this is a type of warning we're used to seeing on the RHEL8 compilers. `FwSizeStoreType` is a U16 but the addition in `+=` auto-promotes it to U32, so there's an implicit cast down back into U16. I think `uncompressed_size = static_cast<FwSizeStoreType>(uncompressed_size + chunk_size)` should fix it

I didn't realize that could happen. I made your requested change

thomas-bc:
For the missing zlib.h - hopefully installing zlib/zlib-devel around here (in all 3 jobs) should fix it
https://github.com/nasa/fprime/blob/c310b326c89d35ecc3c52202e4ba343974ad0b3e/.github/workflows/build-test-rhel8.yml#L55

Svc/Subtopologies/DataProducts/DataProducts.fpp
```diff
@@ -57,6 +57,34 @@ module DataProducts{
         DataProducts::dpBufferManager.cleanup();
         """
     }
+
+    instance dpCompressProc: Svc.DpCompressProc base id DataProductsConfig.BASE_ID + 0x04000
+
+    instance dpZLibCompressor: Svc.DpZLibCompressor base id DataProductsConfig.BASE_ID + 0x05000
```
thomas-bc: CI is exposing that the zlib dependency introduced by the `DipZLibCompressor` component is not a standard library that's available on most distributions (ubuntu, RHEL8, macos runners all don't have it).

Adding this component to the DataProducts subtopology adds a new system dependency on zlib for systems that want to use this subtopology. I don't think this is desirable as a default option.

Likely the best way forward here is to make the inclusion of this component optional (build option, other subtopology, or something else). That would be awesome, but it's possibly more work than you want to be doing right now.

Given that the change is only adding port connections, a good-and-easy solution would be to add it in the topology of one of our references, but leave it out of the subtopology by default.  https://github.com/nasa/fprime-examples/tree/pr-5235 would be a good place.

There may be other options too.

@kubiak-jpl thoughts? TL;DR multiple options, let me know what you prefer

thomas-bc (changes requested):

Svc/Subtopologies/DataProducts/DataProducts.fpp
kubiak-jpl: Another option would be to conditionally compile the component as a no-op if zlib isn't available. That would prevent breaking these other deployments, but it might add more confusion about whether compression is actually enabled or not.

Maybe there's a middle ground where the DpCompressProc component stays in the subtopology, but we omit the compressor. Then users could add compressors their system supports. Do subtopologies support exporting ports yet?

thomas-bc: Yeah conditionally compiling a feature with a no-op component doesn't really respect the intended component-based design.

And yes subtopologies can export ports, for the top-level topology to easily connect to: https://nasa.github.io/fpp/fpp-users-guide.html#Defining-Topologies_Topology-Ports

kubiak-jpl: How about this, I can update the DataProduct subtopology to export the DpWriter Proc port and I'll write a subtopology with DpCompressProc and DpZLibCompressor as an example of how to enable compression. Does that work? Would it make sense for the compression subtopology to exist in the fprime repo?

thomas-bc: Yes that's a great idea! Amazing.

If it were me I would keep the DpProc and DpZLib components in the same (optional) subtopology. That way to enable compression, all one needs to do is 

```diff
import Subtopologies.DataProducts
+ import Subtopologies.DpCompression

[....]

+ DataProducts.procBufferSendOut[0] -> DpCompression.procRequest

```

Or is there a use-case for having a dpCompressProc component in the DataProducts subtopology even if it's not connected to anything?

thomas-bc: And this is actually a really nifty pattern for all DpProcessing systems...

kubiak-jpl: Sounds good. I'll also add some cmake logic to only expose the DpZLibCompressor component if zlib is found. Something with `find_library`

kubiak-jpl: I've made the change. I add the DpCompress subtopology to the Ref deployment, but commented it out. Let me know if you want to leave it like that, or just remove it from Ref.

Svc/DpCompressProc/DpCompressProc.cpp
```diff
@@ -0,0 +1,358 @@
...
+    container.deserializeHeader();
+
+    FwSizeType prm_chunk_size = paramGet_CHUNK_SIZE(param_valid);
+    FW_ASSERT((param_valid == Fw::ParamValid::DEFAULT) || (param_valid == Fw::ParamValid::VALID), param_valid);
+
+    if (prm_chunk_size > container.getDataSize()) {
+        prm_chunk_size = container.getDataSize();
+    }
+    const FwSizeType max_chunk_size = prm_chunk_size;
+    FW_ASSERT(max_chunk_size != 0);
```
lestarch-autobot: [Security] **must fix** Ground-reachable assert: `CHUNK_SIZE` parameter set to 0 triggers DoS.

The `CHUNK_SIZE` param is ground-settable. If ground sets it to 0, `max_chunk_size` becomes 0 and this assert fires, crashing the spacecraft. Replace with a validation that returns early.

```suggestion
    if (prm_chunk_size == 0 || prm_chunk_size > container.getDataSize()) {
        prm_chunk_size = container.getDataSize();
    }
    const FwSizeType max_chunk_size = prm_chunk_size;
    if (max_chunk_size == 0) {
        return;
    }
```

Svc/DpZLibCompressor/DpZLibCompressor.cpp
```diff
@@ -0,0 +1,164 @@
...
+voidpf DpZLibCompressor::zlib_alloc_fn(voidpf opaque, uInt items, uInt size) {
+    FW_ASSERT(opaque != nullptr);
+    ZLibCtx* ctx = reinterpret_cast<ZLibCtx*>(opaque);
+
+    FW_ASSERT(ctx->bump_allocator <= ctx->zlib_alloc_buffer.getSize(),
+              static_cast<FwAssertArgType>(ctx->bump_allocator),
+              static_cast<FwAssertArgType>(ctx->zlib_alloc_buffer.getSize()));
+    const FwSizeType free_space = ctx->zlib_alloc_buffer.getSize() - ctx->bump_allocator;
+
+    const FwSizeType alloc_size = items * size;
```
lestarch-autobot: [Security] **suggestion** Potential 32-bit integer overflow in `items * size` before widening to `FwSizeType`.

Both `items` and `size` are `uInt` (32-bit). Their product overflows in 32-bit arithmetic before assignment to 64-bit `FwSizeType`. Cast to `FwSizeType` before multiplying.

```suggestion
    const FwSizeType alloc_size = static_cast<FwSizeType>(items) * static_cast<FwSizeType>(size);
```

Svc/DpCompressProc/DpCompressProc.fpp
```diff
@@ -0,0 +1,63 @@
+module Svc {
+    @ Generic Data Product compression helper component
+    passive component DpCompressProc {
```
lestarch-autobot, line 3: [Documentation] **must fix** `docs/user-manual/framework/data-products.md` §2.2 states the data product processor component "is not yet developed." This PR introduces `Svc::DpCompressProc`, which is exactly that processor.

The user-manual actively misdirects users into thinking no processor exists. Update §2.2 item 4 to reference `Svc::DpCompressProc` (and its SDD) instead of the "not yet developed" placeholder.

Svc/DpCompressProc/docs/sdd.md
```diff
@@ -0,0 +1,175 @@
...
+     procBufferSendOut    compressChunk
+ +--------+    +--------------+    +----------------+
+ |DpWriter| -> |DpCompressProc| -> |DpZLibCompressor|
+ +--------+    +--------------+    +----------------+
+```
+
+## Port Descriptions
+| Name | Description |
+|---|---|
+|compressChunk| Pass chunks of the original file data to a compressor. See detailed documentation|
```
lestarch-autobot, line 48: [Documentation] **must fix** The port table omits the primary input port `procRequest: Fw.BufferSend`. This is the entry point through which `DpWriter` sends buffers for compression processing.

```suggestion
## Port Descriptions
| Name | Description |
|---|---|
|procRequest| Synchronous input port receiving `Fw::Buffer` data product containers from `DpWriter` for compression processing|
|compressChunk| Pass chunks of the original file data to a compressor. See detailed documentation|
```

Svc/DpCompressProc/docs/sdd.md
```diff
@@ -0,0 +1,175 @@
...
+
+The buffer, an `Fw::Buffer`, is a chunk of the original data product container. The compressor will overwrite this buffer if it is able to compress the product. The compressor must also update the size of the buffer if compression took place.
+
+min\_compression represents the minimum amount of compression necessary for the compression to be useful. In order to return a compression data product, with the necessary compression record headers prepended, some chunks may need to be compressed with sufficient free space to accommodate the addition of these buffers. The value is a maximum number of bytes that can be returned in the `Fw::Buffer`
+
+In some cases it is necessary to place the compressed chunk at an offset within the passed `Fw::Buffer`. The write offset informs the compressor where to place the compressed data within the `Fw::Buffer` to accommodate any necessary headers.
+// TODO: Now that I'm thinking about it more, is this necessary? Would it be more straightforward to always place the buffer at the beginning of the `Fw::Buffer` and memmove it as necessary?
```
lestarch-autobot: [Documentation] **must fix** The SDD contains an unresolved design question (`// TODO: Now that I'm thinking about it more, is this necessary?`). Open TODOs in shipped SDDs signal incomplete design. Resolve the question and remove the TODO before merging.

Svc/DpCompressProc/docs/sdd.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+## Port Descriptions
+| Name | Description |
+|---|---|
+|compressChunk| Pass chunks of the original file data to a compressor. See detailed documentation|
+
+## Component States
+
+None. DpCompressProc is stateless between invocations
```
lestarch-autobot, line 52: [Documentation] **must fix** The SDD has no Events section. The FPP model declares two ground-facing events (`CompressionComplete` diagnostic, `DidNotCompress` activity low) that operators will see in telemetry. Add an Events table documenting both.

Svc/DpZLibCompressor/docs/sdd.md
```diff
@@ -0,0 +1,42 @@
...
+|---|---|
+|compressChunk| Receive chunks of data to compress. See DpCompressProc for more information|
+|bufferCompressionGet| Get an `Fw::Buffer`. This compression buffer is used as a temporary buffer to store the results of the compression|
+|bufferCompressionReturn| Return the compression buffer |
+|bufferZLibGet| Get an `Fw::Buffer`. This zlib buffer is used as a memory pool to implement malloc/free for zlib through an internal bump allocator|
+|bufferZLibReturn| Return the zlib buffer |
+
+## Component States
+
+None. DpZLibCompressor is stateless between invocations
```
lestarch-autobot, line 20: [Documentation] **must fix** The SDD has no Events section. The FPP model declares 7 events (`ZLibCompressionBadBuffer`, `ZLibAllocBadBuffer`, `ZLibInitError`, `ZLibDeflateError`, `BufferTooBigForZLib`, `ZLibNoCompression`, `ZLibCompression`, `ZLibMemoryUsage`) — several are `warning low` severity that operators need to understand for diagnosing compression failures. Add an Events table.

Svc/DpCompressProc/docs/sdd.md
```diff
@@ -0,0 +1,175 @@
...
+|   Record 2  |
+|   Metadata  |
+| Chunks 2 & 3|
++-------------+
+|  Compressed |
+|   Record 3  |
+|   Metadata  |
+| Chunk 3 (C) | 
++-------------+
+```
```
lestarch-autobot, line 160: [Documentation] **could fix** The compressed-product diagram labels the last record as "Chunk 3 (C)" but per the input diagram (lines 120-132) Chunk 3 is uncompressed — the second compressible chunk is Chunk 4.

```suggestion
|  Compressed |
|   Record 3  |
|   Metadata  |
| Chunk 4 (C) | 
+-------------+
```

Svc/Subtopologies/DpCompression/DpCompression.fpp
```diff
@@ -0,0 +1,61 @@
...
+            DpCompression::Allocation::memAllocator,
+            ConfigObjects::DpCompression_dpZLibCompressorBufferManager::bins
+        );
+        """
+        phase Fpp.ToCpp.Phases.tearDownComponents """
+        DpCompression::dpZLibCompressorBufferManager.cleanup();
+        """
+    }
+
+    topology Subtopology {
```
lestarch-autobot, line 34: [Documentation] **suggestion** `docs/user-manual/design-patterns/subtopologies.md` §"F Prime core subtopologies" lists CdhCore, ComCcsds, ComFprime, DataProducts, and FileHandling but does not mention the new `Svc.DpCompression` subtopology.

```suggestion
    topology Subtopology {
```

Add a bullet to the list in `subtopologies.md`:
```md
- [Svc.DpCompression](../../../Svc/Subtopologies/DpCompression): Data product compression components (requires zlib)
```

Fw/Types/Serializable.cpp
```diff
@@ -252,7 +252,7 @@ SerializeStatus LinearBufferBase::serializeFrom(const U8* buff,
     }
     FW_ASSERT(this->getBuffAddr());
     // copy buffer to our buffer
-    (void)memcpy(&this->getBuffAddr()[this->m_serLoc], buff, static_cast<size_t>(length));
+    (void)memmove(&this->getBuffAddr()[this->m_serLoc], buff, static_cast<size_t>(length));
```
lestarch-autobot, line 255: [Documentation] **could fix** The `memcpy` → `memmove` change means `LinearBufferBase::serializeFrom(const U8*, ...)` now supports overlapping source and destination buffers. The doc comment on the declaration in `Serializable.hpp` (line ~264) does not note this relaxed precondition. Consider adding a note that overlapping regions are supported.

Svc/DpCompressProc/DpCompressProc.cpp
lestarch-autobot: [C++ Design] **must fix** CPP-4 (`FW_ASSERT` on untrusted input): `max_chunk_size` derives from ground-settable parameter `CHUNK_SIZE`.

If a ground operator sets `CHUNK_SIZE` to 0, this assert fires and crashes the component. Validate and return early (or emit a warning event) instead of asserting on a value controllable via ground command.

```suggestion
    if (max_chunk_size == 0) {
        return;
    }
```

Svc/DpCompressProc/DpCompressProc.cpp
```diff
@@ -0,0 +1,358 @@
...
+        FW_ASSERT(chunk_size_tmp <= std::numeric_limits<FwSizeStoreType>::max(),
+                  static_cast<FwAssertArgType>(chunk_size_tmp));
+        const FwSizeStoreType chunk_size = static_cast<FwSizeStoreType>(chunk_size_tmp);
+        // a const U8 * and I need to mutable U8*
+        U8* data_offset = data_buffer.getData() + (data_buffer.getSize() - data_deser.getDeserializeSizeLeft());
+
+        Fw::Buffer compression_buffer(data_offset, chunk_size);
+
+        // Jump the deserializer to mark the data as read
+        data_deser.deserializeSkip(chunk_size);
```
lestarch-autobot: [C++ Design] **must fix** CPP-32 (ignored return value): `deserializeSkip` returns `Fw::SerializeStatus` which is discarded here.

This pattern repeats for `serializeSkip` (line 241) and multiple `serializeFrom` calls (lines 245, 253, 266, 270, 338). Serialization failures indicate buffer overflows and must be checked or explicitly `(void)`-cast with justification.

```suggestion
        Fw::SerializeStatus stat = data_deser.deserializeSkip(chunk_size);
        FW_ASSERT(stat == Fw::FW_SERIALIZE_OK, stat);
```

Svc/DpZLibCompressor/DpZLibCompressor.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    // Note: Skipping call to deflateEnd. Per the documentation this
+    // call frees the dynamically allocated data structures, but this
+    // is not necessary with the bump allocator.
+    // The call to deflateEnd does not flush any additional output data
+}
+
+voidpf DpZLibCompressor::zlib_alloc_fn(voidpf opaque, uInt items, uInt size) {
+    FW_ASSERT(opaque != nullptr);
+    ZLibCtx* ctx = reinterpret_cast<ZLibCtx*>(opaque);
```
lestarch-autobot, line 141: [C++ Design] **could fix** CPP-10 (`reinterpret_cast` without justification): add an inline comment explaining why this cast is required.

The zlib allocator callback API passes an opaque `void*` that must be cast back to the concrete context type — this is a valid use, but per CPP-10 the justification must be documented inline.

```suggestion
    ZLibCtx* ctx = reinterpret_cast<ZLibCtx*>(opaque);  // zlib allocator API passes opaque void*
```

TestDeploymentsProject/Ref/DpDemo/DpDemo.hpp
```diff
@@ -68,6 +71,7 @@ class DpDemo final : public DpDemoComponentBase {
     DpDemo_ColorEnum selectedColor;
     U32 numRecords;
     U32 dpPriority;
+    Fw::DpCfg::ProcType dpProc;
```
lestarch-autobot, line 74: [C++ Design] **suggestion** CPP-19 (uninitialized member): `dpProc` is not explicitly initialized in the constructor.

While the FPP-generated enum type has a default constructor, the existing pattern in `DpDemo::DpDemo()` explicitly initializes all other members. Add an in-class initializer for consistency and clarity.

```suggestion
    Fw::DpCfg::ProcType dpProc{};
```

TestDeploymentsProject/Ref/DpDemo/DpDemo.fpp
```diff
@@ -69,7 +69,7 @@ module Ref {
         async command SelectColor(color: ColorEnum) opcode 0
 
         @ Command for generating a DP
-        sync command Dp(reqType: DpReqType, $priority: U32)
+        sync command Dp(reqType: DpReqType, $priority: U32, proc: Fw.DpCfg.ProcType)
```
lestarch-autobot, line 72: [Architecture] **future work** `sync command Dp` on `active component DpDemo` bypasses thread isolation.

Preexisting — this PR only adds the `proc` parameter. An active component's commands should typically be `async` so they dispatch through the component's queue and thread (see *Component Selection for Event-Driven Work* in the selection guide). The `sync` kind forces the command dispatcher to block while `Dp_cmdHandler` executes, coupling execution contexts. Consider converting to `async`.

Fw/Types/Serializable.cpp
lestarch-autobot, line 255: [Design] **must fix** **Human design adjudication required.** Framework-wide `memcpy`→`memmove` in `LinearBufferBase::serializeFrom` to support one component's in-place buffer rewriting.

This PR crosses the Fw↔Svc boundary: it modifies core serialization used by every component in every deployment to accommodate `DpCompressProc`'s overlapping-buffer use case. The author notes this could be reverted (using `memmove` directly in `DpCompressProc`). Additionally, this PR introduces two new `Svc/` components, a new subtopology, a new external dependency (zlib), and the framework concept of compressed data products — all warranting design-owner input before deeper review.

`memmove` is functionally safe wherever `memcpy` is, but carries a performance penalty on every serialization call framework-wide. The design question: should the framework accommodate overlapping buffers in `serializeFrom`, or should `DpCompressProc` handle this locally?

default/config/DpCfg.fpp
```diff
@@ -13,8 +13,10 @@ module Fw {
     @ A bit mask for selecting the type of processing to perform on
     @ a container before writing it to disk.
     enum ProcType: U8 {
+      @ No Processing
+      PROC_TYPE_NONE = 0x00
       @ Processing type 0
-      PROC_TYPE_ZERO = 0x01
+      PROC_TYPE_ZLIB_DEFLATE = 0x01
```
lestarch-autobot, line 19: [Design] **could fix** Rename of `PROC_TYPE_ZERO` → `PROC_TYPE_ZLIB_DEFLATE` and addition of `PROC_TYPE_NONE` changes the framework's public enum API without acknowledgment in the PR description.

No in-repo code references `PROC_TYPE_ZERO` by name (only the definition), so compilation breakage risk is minimal. However, per design-behavioral-regression criteria, changes to framework defaults should be (a) acknowledged in the PR description, (b) justified, and (c) assessed for backward compatibility — especially for downstream deployments or ground-system scripts that may reference these names. A one-line note in the PR description would close this gap.

Svc/DpCompressProc/DpCompressProc.fpp
```diff
@@ -0,0 +1,63 @@
...
+
+        output port compressChunk: CompressChunk
+
+        @ Chunk size to use when passing data to the compression backend
+        param CHUNK_SIZE: FwSizeStoreType default 32*1024
+        
+        @ Number of bytes to use in a compression chunk
+        param ENABLE: Fw.Enabled default Fw.Enabled.ENABLED
+
+        event CompressionComplete(dp_id: FwDpIdType,
```
lestarch-autobot, line 15: [Test Quality] **must fix** New event `CompressionComplete` has no `ASSERT_EVENTS_CompressionComplete` reference in `test/ut/`.

The test exercises `invoke_to_procRequest` and verifies data integrity, but never asserts this diagnostic event was emitted with the expected `dp_id`, `initial_size`, `final_size` payload. A wrong or missing emission would go undetected.

Svc/DpCompressProc/DpCompressProc.fpp
```diff
@@ -0,0 +1,63 @@
...
+                                  initial_size: FwSizeType
+                                  final_size: FwSizeType) \
+            severity diagnostic id 0 \
+            format "Compressed Data Product {} from {} to {} bytes"
+
+        event DidNotCompress(dp_id: FwDpIdType,
+                             data_size: FwSizeType) \
+            severity activity low \
+            format "Unable to reduce size of Data Product {}. Uncompressed size {}" \
+            throttle 10
```
lestarch-autobot, line 25: [Test Quality] **must fix** New event `DidNotCompress` has no `ASSERT_EVENTS_DidNotCompress` reference in `test/ut/`, and `throttle 10` is unverified.

The `Uncompressible` test exercises the code path emitting this event but never asserts emission or payload (`dp_id`, `data_size`). Additionally, no test fires the event >10 times to confirm throttle suppression.

Svc/DpCompressProc/DpCompressProc.fpp
```diff
@@ -0,0 +1,63 @@
...
+
+        sync input port procRequest: Fw.BufferSend
+
+        output port compressChunk: CompressChunk
+
+        @ Chunk size to use when passing data to the compression backend
+        param CHUNK_SIZE: FwSizeStoreType default 32*1024
+        
+        @ Number of bytes to use in a compression chunk
+        param ENABLE: Fw.Enabled default Fw.Enabled.ENABLED
```
lestarch-autobot, line 13: [Test Quality] **must fix** `param ENABLE` disabled path never exercised in tests.

`test_chunks_helper` always calls `paramSet_ENABLE(Fw::Enabled::ENABLED, ...)`. The `DISABLED` branch in `procRequest_handler` (early return, bypass compression) is never tested — a regression breaking the bypass would pass CI undetected.

Svc/DpZLibCompressor/DpZLibCompressor.fpp
```diff
@@ -0,0 +1,106 @@
...
+            throttle 5
+
+        @ Unable to allocate an alloc buffer for ZLib
+        event ZLibAllocBadBuffer($size: FwSizeType) \
+            severity warning low \
+            format "Unable to allocate a buffer of size {} for ZLib alloc" \
+            throttle 5
+
+        @ ZLib error during deflateInit
+        event ZLibInitError(err: I32,
```
lestarch-autobot, line 30: [Test Quality] **must fix** `ZLibInitError` event path exercised by `InsufficientZLibMem` test but event emission never asserted.

The test correctly expects `UNCOMPRESSED` return, but `compress()` has no `ASSERT_EVENTS_ZLibInitError` for this path. A silent regression removing the event would go undetected.

Svc/DpZLibCompressor/DpZLibCompressor.fpp
```diff
@@ -0,0 +1,106 @@
...
+                            msg: string) \
+            severity warning low \
+            format "ZLib Error during initialization {} {}" \
+            throttle 5
+
+        @ ZLib error during deflate
+        event ZLibDeflateError(err: I32,
```
lestarch-autobot, line 37: [Test Quality] **must fix** No test triggers the `ZLibDeflateError` code path.

This event fires when `deflate()` returns an error other than `Z_OK`/`Z_STREAM_END`. No existing test exercises this failure — consider providing a compression buffer smaller than the deflated output while `deflateInit` succeeds.

Svc/DpZLibCompressor/DpZLibCompressor.fpp
```diff
@@ -0,0 +1,106 @@
...
+                               msg: string) \
+            severity warning low \
+            format "ZLib Error during compression {} {}" \
+            throttle 5
+
+        @ Buffers are too large to process with zlib due to the
+        @ type differences between FwSizeType and uInt
```
lestarch-autobot, line 44: [Test Quality] **must fix** No test triggers the `BufferTooBigForZLib` code path.

This event fires when `in_buffer.getSize() > std::numeric_limits<uInt>::max()`. On 64-bit platforms where `FwSizeType` > 32 bits, a buffer exceeding 4 GiB would hit this guard. A targeted test (or a mock narrowing `uInt` range) is needed.

Svc/DpZLibCompressor/test/ut/DpZLibCompressorTester.cpp
```diff
@@ -0,0 +1,102 @@
...
+    const CompressionAlgorithm alg = invoke_to_compressChunk(0, uncomp_buffer, min_compression, write_offset);
+
+    ASSERT_EQ(alg, exp_alg);
+    if (alg != CompressionAlgorithm::UNCOMPRESSED) {
+        ASSERT_LT(uncomp_buffer.getSize(), min_compression + write_offset);
+
+        std::vector<U8> uncomp_data = zlib_inflate(uncomp_buffer, data.size(), write_offset);
+        ASSERT_EQ(uncomp_data.size(), data.size());
+
+        ASSERT_EQ(memcmp(uncomp_data.data(), data.data(), data.size()), 0);
```
lestarch-autobot, line 66: [Test Quality] **must fix** Diagnostic events `ZLibCompression`, `ZLibMemoryUsage`, `ZLibNoCompression` emitted during tests but never asserted.

Success-path tests trigger `ZLibCompression` + `ZLibMemoryUsage`; uncompressed-path tests trigger `ZLibNoCompression`. The `compress()` helper checks data integrity but never verifies these event payloads. Add e.g. `ASSERT_EVENTS_ZLibCompression(0, data.size(), compressed_size)` after the success branch.

Svc/DpZLibCompressor/DpZLibCompressor.fpp
```diff
@@ -0,0 +1,106 @@
...
+
+        ##############################################################################
+        #### Uncomment the following examples to start customizing your component ####
+        ##############################################################################
+
+        @ Unable to allocate a compression buffer for ZLib
+        event ZLibCompressionBadBuffer($size: FwSizeType) \
+            severity warning low \
+            format "Unable to allocate a buffer of size {} for compression" \
+            throttle 5
```
lestarch-autobot, line 21: [Test Quality] **must fix** `throttle 5` on warning events `ZLibCompressionBadBuffer` / `ZLibAllocBadBuffer` never verified.

Tests assert these events fire once but never invoke the path >5 times to confirm throttle suppression. The same applies to `ZLibInitError`, `ZLibDeflateError`, `BufferTooBigForZLib` once those paths gain tests.

Svc/DpCompressProc/test/ut/DpCompressProcTestMain.cpp
```diff
@@ -0,0 +1,146 @@
...
+// \author kubiak
+// \brief  cpp file for DpCompressProc component test main function
+// ======================================================================
+
+#include "AbstractState.hpp"
+#include "DpCompressProcTester.hpp"
+#include "STest/STest/Random/Random.hpp"
+#include "Svc/DpCompressProc/test/ut/Rules/Testers.hpp"
+
+TEST(Nominal, Compressible) {
```
lestarch-autobot, line 12: [Test Quality] **could fix** Six test cases follow an identical `set_chunk_state` → `CompressTest()` pattern differing only in chunk configurations.

Consider GTest parameterized tests (`TEST_P` / `INSTANTIATE_TEST_SUITE_P`) or a helper accepting the chunk vector directly. This reduces ~60 lines to a data table and makes adding new configurations trivial.

lestarch-autobot (changes requested):

### Per-agent results

| Security Vulnerabilities | 1 | 1 | 0 | 0 | 2 | No-Go |
| F Prime C/C++ Design | 2 | 1 | 1 | 0 | 4 | No-Go |
| Documentation Currency | 5 | 1 | 2 | 0 | 8 | No-Go |
| Architecture | 0 | 0 | 0 | 1 | 1 | Go |
| Test Quality | 8 | 0 | 1 | 0 | 9 | No-Go |
| **Totals** | 17 | 3 | 5 | 1 | 26 | **No-Go** |

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>
<summary>Outstanding must-fix items (17)</summary>

**Security Vulnerabilities**
- Ground-reachable assert: `CHUNK_SIZE` parameter set to 0 triggers DoS — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437282266)

**F Prime C/C++ Design**
- CPP-4: `FW_ASSERT` on untrusted ground-settable `CHUNK_SIZE` parameter — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437295391)
- CPP-32: ignored `deserializeSkip` return value — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437295402)

**Documentation Currency**
- `docs/user-manual/framework/data-products.md` §2.2 states processor "is not yet developed" — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437290343)
- Port table omits primary input port `procRequest: Fw.BufferSend` — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437290348)
- SDD contains unresolved TODO design question — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437290364)
- DpCompressProc SDD missing Events section — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437290367)
- DpZLibCompressor SDD missing Events section — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437290370)

**Design**
- **Human design adjudication required.** Framework-wide `memcpy`→`memmove` in `LinearBufferBase::serializeFrom` — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437296829)

**Test Quality**
- New event `CompressionComplete` untested — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437299725)
- New event `DidNotCompress` untested, `throttle 10` unverified — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437299731)
- `param ENABLE` disabled path never exercised — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437299735)
- `ZLibInitError` event path exercised but emission never asserted — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437299737)
- `ZLibDeflateError` code path untested — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437299743)
- `BufferTooBigForZLib` code path untested — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437299749)
- Diagnostic events emitted but never asserted in tests — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437299755)
- `throttle 5` on warning events never verified — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437299757)

</details>

### Merge readiness
**Merge readiness: No-Go** — 5 of 7 reviewers report outstanding must-fix items (17 total across Security, C++ Design, Documentation, Design, and Test Quality).

---

Compression is a worthy addition — clear the open findings and this payload is ready for orbit.

Fw/Types/Serializable.cpp
kubiak-jpl: @LeStarch @thomas-bc Thoughts here? I mentioned this change in the PR but could use an explicit ack or nack

thomas-bc: Discussed with @LeStarch - I think we're happy to move to `memmove`. Our specs don't call out that you're not supposed to have overlap, so actually respecting that by using `memmove` seems like a good _move_! 

Tagging @bocchino because he has expressed concerns. Rob, my understanding from a few minutes of googling is the overhead should be negligeable. Thoughts?

bocchino: I think it's OK to make the change if the overhead is negligible. I agree it's safer/more defensive.

I was wondering what code needs this change. Ordinarily deserialization operates on distinct buffers. However, I see this issue is addressed in the main PR comments (it makes the code cleaner) so I think it's OK.

Svc/DpCompressProc/DpCompressProc.cpp
kubiak-jpl: Done

Svc/DpZLibCompressor/DpZLibCompressor.cpp
kubiak-jpl: Done

Svc/DpCompressProc/DpCompressProc.fpp
kubiak-jpl: Updated with a pointer to DpCompressProc

Svc/DpCompressProc/docs/sdd.md
kubiak-jpl: Removed the TODO. Decided that the offset isn't strictly necessary but it does simplify the logic in DpCompressProc

Svc/DpCompressProc/docs/sdd.md
kubiak-jpl: Done

Svc/DpZLibCompressor/docs/sdd.md
kubiak-jpl: Done

Svc/DpCompressProc/docs/sdd.md
kubiak-jpl: Done

Svc/Subtopologies/DpCompression/DpCompression.fpp
kubiak-jpl: Updated

Svc/DpCompressProc/DpCompressProc.cpp
kubiak-jpl: Done

Svc/DpCompressProc/DpCompressProc.cpp
kubiak-jpl: Done

Svc/DpZLibCompressor/DpZLibCompressor.cpp
kubiak-jpl: I think this is clear enough from context

TestDeploymentsProject/Ref/DpDemo/DpDemo.hpp
kubiak-jpl: I don't want to make this change within the context of this PR

TestDeploymentsProject/Ref/DpDemo/DpDemo.fpp
kubiak-jpl: Not changing this behavior within this PR

default/config/DpCfg.fpp
kubiak-jpl: @LeStarch @thomas-bc Thoughts here? I originally updated the default DpCfg.fpp because the PROC types was in the Ref deployment. Not that it's in an optional sub-topology this isn't a required change.

I think I'd still like to keep it though

Svc/DpCompressProc/DpCompressProc.fpp
kubiak-jpl: Done

Svc/DpCompressProc/DpCompressProc.fpp
kubiak-jpl: Done

Svc/DpCompressProc/DpCompressProc.fpp
kubiak-jpl: Done

Svc/DpZLibCompressor/DpZLibCompressor.fpp
kubiak-jpl: Done

Svc/DpZLibCompressor/DpZLibCompressor.fpp
kubiak-jpl: It's unclear how to trigger a ZLibDeflateError event. It's a defensive event that I have not identified a path to reach. The proposed test case results in a Z_OK return code, which does not trigger this event.

Svc/DpZLibCompressor/DpZLibCompressor.fpp
kubiak-jpl: This is a defensive event. Not going to test it because it requires playing with F Prime types

Svc/DpZLibCompressor/test/ut/DpZLibCompressorTester.cpp
kubiak-jpl: Done

Svc/DpZLibCompressor/DpZLibCompressor.fpp
kubiak-jpl: Not testing event throttles

Svc/DpCompressProc/test/ut/DpCompressProcTestMain.cpp
kubiak-jpl: Not implementing

Fw/Types/Serializable.cpp
kubiak-jpl: I'm not going to take any action here

kubiak-jpl:
@thomas-bc @LeStarch I've addressed all the autobot comments. Ready for another round of review

thomas-bc:
https://github.com/nasa/fprime/pull/5319 - looks like the TestUtils/Optional was moved/modified in `devel` so it's now breaking the UTs here

TestDeploymentsProject/Ref/DpDemo/DpDemo.fpp
thomas-bc, line 72: Because there's a new argument, we need to update the InT scripts to account for it:

https://github.com/nasa/fprime/blob/3067220ec68fd8e5dc056f8190b2113e6b4d9340/TestDeploymentsProject/Ref/DpDemo/test/int/dp_demo_integration_test.py#L11

https://github.com/nasa/fprime/blob/3067220ec68fd8e5dc056f8190b2113e6b4d9340/TestDeploymentsProject/Ref/DpDemo/test/int/dp_demo_integration_test.py#L32

Svc/DpZLibCompressor/DpZLibCompressor.fpp
thomas-bc: Agreed

default/config/DpCfg.fpp
thomas-bc: I think this is fine

TestDeploymentsProject/Ref/Top/topology.fpp
```diff
@@ -20,6 +20,7 @@ module Ref {
     instance ComCcsds.Subtopology
     instance FileHandling.Subtopology
     instance DataProducts.Subtopology
+    #instance DpCompression.Subtopology
```
thomas-bc, line 23: Is the topology not included here on purpose? Ideally we'd need this topology building in a deployment in CI, with a GDS decompression test.

Since `zlib` isn't necessarily available, https://github.com/nasa/fprime-examples would be a good place to add it instead of `Ref/`

Happy to capture in a follow up ticket.

thomas-bc (approved):
LGTM - one comment on CI testing

TestDeploymentsProject/Ref/Top/topology.fpp
kubiak-jpl: I wanted to include an example but didn't want to bring in the zlib dependency. Happy to move it to fprime-examples if that the right home. You can assign the issue ticket to me

thomas-bc: https://github.com/nasa/fprime/issues/5358

thomas-bc (dismissed):
Opened follow-up ticket for integration tests: https://github.com/nasa/fprime/issues/5358

Svc/DpCompressProc/DpCompressProc.cpp
lestarch-autobot: [Security] Fixed in d0d0a29.

Svc/DpZLibCompressor/DpZLibCompressor.cpp
lestarch-autobot: [Security] Fixed in d0d0a29.

Svc/DpCompressProc/DpCompressProc.fpp
lestarch-autobot: [Documentation] Fixed in d0d0a29.

Svc/DpCompressProc/docs/sdd.md
lestarch-autobot: [Documentation] Fixed in d0d0a29.

Svc/DpCompressProc/docs/sdd.md
lestarch-autobot: [Documentation] Fixed in d0d0a29.

Svc/DpCompressProc/docs/sdd.md
lestarch-autobot: [Documentation] Fixed in d0d0a29.

Svc/DpZLibCompressor/docs/sdd.md
lestarch-autobot: [Documentation] Fixed in d0d0a29.

Svc/DpCompressProc/docs/sdd.md
lestarch-autobot: [Documentation] Fixed in d0d0a29.

Svc/Subtopologies/DpCompression/DpCompression.fpp
lestarch-autobot: [Documentation] Fixed in d0d0a29.

Fw/Types/Serializable.cpp
lestarch-autobot: [Documentation] **Improperly resolved.** This finding still applies on d0d0a29.

`Fw::Serializable::serializeFrom`/`serialize` doc comments in Serializable.hpp still do not state that overlapping source/destination buffers are now supported via `memmove`; the documentation gap still applies on the new head.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

Svc/DpCompressProc/DpCompressProc.cpp
lestarch-autobot: [C++ Design] Fixed in d0d0a29.

Svc/DpCompressProc/DpCompressProc.cpp
lestarch-autobot: [C++ Design] Fixed in d0d0a29.

Svc/DpZLibCompressor/DpZLibCompressor.cpp
lestarch-autobot: [C++ Design] **Improperly resolved.** This finding still applies on d0d0a29.

The `reinterpret_cast<ZLibCtx*>(opaque)` at DpZLibCompressor.cpp:141 is unchanged on the new head; the CPP-10 finding still applies. A brief `// zlib opaque round-trips our ZLibCtx*` comment or a `static_cast` via `void*` would satisfy it.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

TestDeploymentsProject/Ref/DpDemo/DpDemo.hpp
lestarch-autobot: [C++ Design] **Improperly resolved.** This finding still applies on d0d0a29.

`dpProc` still has no in-class initializer / constructor init on the new head (DpDemo.hpp:71, DpDemo.cpp ctor initializes every other member but not `dpProc`); CPP-19 still applies.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

TestDeploymentsProject/Ref/DpDemo/DpDemo.fpp
lestarch-autobot: [Architecture] **Improperly resolved.** This finding still applies on d0d0a29.

`DpDemo` is still `active` with a `sync command Dp` on the new head; the preexisting thread-isolation observation still applies (future-work, non-blocking).

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

Fw/Types/Serializable.cpp
lestarch-autobot: [Design] **Improperly resolved.** This finding still applies on d0d0a29.

The framework-wide `memcpy`->`memmove` change in Fw/Types/Serializable.cpp is still present on the new head. Maintainers discussed it in-thread, but this `design-needs-human-adjudication` finding requires an explicit design-owner sign-off recorded on the PR before it is closed.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

default/config/DpCfg.fpp
lestarch-autobot: [Design] **Improperly resolved.** This finding still applies on d0d0a29.

The public `DpCfg` enum surface still renames `PROC_TYPE_ZERO` -> `PROC_TYPE_ZLIB_DEFLATE` on the new head; the compatibility-affecting rename still applies. A note in the PR body / a compatibility comment would address it.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

Svc/DpCompressProc/DpCompressProc.fpp
lestarch-autobot: [Test Quality] Fixed in d0d0a29.

Svc/DpCompressProc/DpCompressProc.fpp
lestarch-autobot: [Test Quality] Fixed in d0d0a29.

Svc/DpCompressProc/DpCompressProc.fpp
lestarch-autobot: [Test Quality] Fixed in d0d0a29.

Svc/DpZLibCompressor/DpZLibCompressor.fpp
lestarch-autobot: [Test Quality] Fixed in d0d0a29.

Svc/DpZLibCompressor/DpZLibCompressor.fpp
lestarch-autobot: [Test Quality] **Improperly resolved.** This finding still applies on d0d0a29.

No test on the new head triggers the `ZLibDeflateError` code path (DpZLibCompressorTestMain.cpp still stops at `InsufficientZLibMem`); the missing failure-path coverage still applies.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

Svc/DpZLibCompressor/DpZLibCompressor.fpp
lestarch-autobot: [Test Quality] **Improperly resolved.** This finding still applies on d0d0a29.

No test on the new head triggers the `BufferTooBigForZLib` code path; the missing failure-path coverage still applies.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

Svc/DpZLibCompressor/test/ut/DpZLibCompressorTester.cpp
lestarch-autobot: [Test Quality] Fixed in d0d0a29.

Svc/DpZLibCompressor/DpZLibCompressor.fpp
lestarch-autobot: [Test Quality] **Improperly resolved.** This finding still applies on d0d0a29.

No test on the new head fires `ZLibCompressionBadBuffer`/`ZLibAllocBadBuffer` more than 5 times to confirm throttle suppression; the throttle-verification gap still applies.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

Svc/DpCompressProc/test/ut/DpCompressProcTestMain.cpp
lestarch-autobot: [Test Quality] **Improperly resolved.** This finding still applies on d0d0a29.

The six `set_chunk_state` -> `CompressTest()` cases in DpCompressProcTestMain.cpp remain structurally identical on the new head; the maintainability observation still applies (non-blocking).

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate.

Svc/DpCompressProc/DpCompressProc.cpp
```diff
@@ -0,0 +1,363 @@
...
+                              CompressionMetadata::SERIALIZED_SIZE, compressed_size);
+                    std::memmove(data_buffer.getData() + compression_header_size, data_buffer.getData(),
+                                 uncompressed_size);
+
+                    // Serialize the header bytes to the front of the data
+                    serializeCompressionHeader(data_reser, uncompressed_size,
+                                               CompressionMetadata(CompressionAlgorithm::UNCOMPRESSED));
+
+                    // Move the serializer past the uncompressed chunk manually
+                    data_reser.serializeSkip(uncompressed_size);
```
lestarch-autobot: [C++ Design] **must fix** CPP-32 (ignored return value): `serializeSkip` return discarded.

`data_reser.serializeSkip(uncompressed_size)` discards its `SerializeStatus`. The downstream `FW_ASSERT(ser_loc <= deser_loc, ...)` does NOT catch a failed skip (a no-op skip leaves `ser_loc` smaller, so the invariant still holds), leaving the compressed layout silently corrupt. Assert the status like the `deserializeSkip` fix in the same file.

```suggestion
                    Fw::SerializeStatus ser_stat = data_reser.serializeSkip(uncompressed_size);
                    FW_ASSERT(ser_stat == Fw::FW_SERIALIZE_OK, ser_stat);
```

Svc/DpZLibCompressor/test/ut/DpZLibCompressorTester.cpp
```diff
@@ -0,0 +1,130 @@
...
+                                                     const I8 compression_level) {
+    paramSet_ZLibBufferSize(zlib_alloc_size, Fw::ParamValid::VALID);
+    paramSet_CompressionLevel(compression_level, Fw::ParamValid::VALID);
+
+    this->component.loadParameters();
+
+    std::vector<U8> buffer_data = data;
+    Fw::Buffer uncomp_buffer(buffer_data.data(), buffer_data.size());
+    const CompressionAlgorithm alg = invoke_to_compressChunk(0, uncomp_buffer, min_compression, write_offset);
+    ASSERT_EVENTS_ZLibInitError_SIZE(1);
```
lestarch-autobot, line 126: [Test Quality] **could fix** `ZLibInitError` asserted by count only, payload unchecked.

`ASSERT_EVENTS_ZLibInitError_SIZE(1)` confirms the event fired but not its payload. The event carries the zlib return code, so the test can pass with the wrong error reported. Assert the payload via the generated `ASSERT_EVENTS_ZLibInitError` accessor.

lestarch-autobot (changes requested):

### Per-agent results

| Security Vulnerabilities | 1 | 1 | 0 | 0 | 0 | Go |
| F Prime C/C++ Design | 3 | 1 | 1 | 0 | 3 | No-Go |
| Documentation Currency | 5 | 1 | 2 | 0 | 1 | Go |
| Architecture | 0 | 0 | 0 | 1 | 1 | Go |
| Test Quality | 8 | 0 | 2 | 0 | 5 | No-Go |
| **CI safety** | — | — | — | — | — | **Go** — Security and Supply Chain both clean |
| **Totals** | 18 | 3 | 6 | 1 | 12 | **No-Go** |

<details>

| Security Vulnerabilities | 2 | 0 | 0 | 0 | 0 | 0 |
| F Prime C/C++ Design | 2 | 2 | 1 | 0 | 2 | 0 |
| Documentation Currency | 7 | 1 | 0 | 0 | 1 | 0 |
| Design | 0 | 2 | 0 | 0 | 2 | 0 |
| Architecture | 0 | 1 | 0 | 0 | 1 | 0 |
| Test Quality | 5 | 4 | 1 | 0 | 4 | 0 |
| **Totals** | 16 | 10 | 2 | 0 | 10 | 0 |

</details>

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>
<summary>Outstanding must-fix items (5)</summary>

**F Prime C/C++ Design**
- CPP-32: ignored `serializeSkip` return value in the in-place uncompressed rewrite — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3532231315)

**Design**
- **Human design adjudication required.** Framework-wide `memcpy`→`memmove` in `Fw::Serializable` serialization — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437296829)

**Test Quality**
- `ZLibDeflateError` code path untested — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437299743)
- `BufferTooBigForZLib` code path untested — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437299749)
- `throttle 5` on warning events never verified — [link](https://github.com/nasa/fprime/pull/5235#discussion_r3437299757)

</details>

### Merge readiness
**Merge readiness: No-Go** — 3 of 7 reviewers report outstanding must-fix items (5 total across C++ Design, Design, and Test Quality). CI safety is **Go**: both Security and Supply Chain are clean. Most run-1 findings were addressed (16 resolved); the remaining threads were closed without a code change, so they were re-flagged for maintainer adjudication.

---

The compressor has cleared most checks — resolve these last few and it's cleared for docking.

Svc/DpCompressProc/DpCompressProc.cpp
thomas-bc: @kubiak-jpl last `must-fix` — thoughts? I'm unclear why it's flagging specifically this one and not the other serialization calls that also discard statuses

kubiak-jpl: It's easy enough for me to do a scrub of these calls and add the asserts. I'll take care of it today

kubiak-jpl: Done. I added the serialization asserts in a few places throughout the DpCompressProc component

thomas-bc (approved):

Svc/Subtopologies/DpCompression/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/DpCompressionConfig/")
```
LeStarch, line 1: There doesn't appear to be a zlib check here in this subtopology.

LeStarch (approved):

Issue #4436, opened 2025-11-18
EXCLUDE_FROM_ALL By Default


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

When a project builds the `all` target, they end up building three categories of things:

1. All modules in F Prime not listed as `EXCLUDE_FROM_ALL`
2. All modules in all fprime libraries not listed as `EXCLUDE_FROM_ALL`
3. All modules in their project

It is likely that only item 3 here is intended.  We should likely mark all modules `EXCLUDE_FROM_ALL` in the first two categories above (fprime, fprime-libraries).

LeStarch:
1. Just build project
2.

LeStarch:
I don't understand the CCB resolution above, so I recommend it goes back to CCB

thomas-bc:
@LeStarch i believe I remember that the decision was to build only project. So essentially approving your recommendation 3) only

PR #5245, opened 2026-06-05
EXCLUDE_FROM_ALL by default

jrussino:
| | |
|:---|:---|
|**_Related Issue(s)_**|  https://github.com/nasa/fprime/issues/4436 |
|**_Generative AI was used in this contribution (y/n)_**|  n |

---
## Change Description

Adds the EXCLUDE_FROM_ALL for fprime core modules and libraries in FPrime cmake.

## Rationale

From https://github.com/nasa/fprime/issues/4436 :

> When a project builds the `all` target, they end up building three categories of things:
> 
>     1. All modules in F Prime not listed as `EXCLUDE_FROM_ALL`
> 
>     2. All modules in all fprime libraries not listed as `EXCLUDE_FROM_ALL`
> 
>     3. All modules in their project
> 
> 
> It is likely that only item 3 here is intended. We should likely mark all modules `EXCLUDE_FROM_ALL` in the first two categories above (fprime, fprime-libraries).

None

## Future Work

It may be advisable to remove any redundant instances of the EXCLUDE_FROM_ALL flag that are currently used for individual modules / libraries.

AI not used for this work.


cmake/FPrime.cmake
```diff
@@ -23,7 +23,7 @@ endif()
 # for adding libraries from the _fprime_packages directory
 if (IS_DIRECTORY "${FPRIME_PROJECT_ROOT}/_fprime_packages")
     if (EXISTS "${FPRIME_PROJECT_ROOT}/_fprime_packages/packages.cmake")
-        include("${FPRIME_PROJECT_ROOT}/_fprime_packages/packages.cmake")
+        include("${FPRIME_PROJECT_ROOT}/_fprime_packages/packages.cmake" EXCLUDE_FROM_ALL)
```
LeStarch, line 26: Include statements should not be passing `EXCLUDE_FROM_ALL`

cmake/FPrime.cmake
```diff
@@ -246,7 +246,7 @@ function(fprime_setup_included_code)
             set(__FPRIME_NO_UT_GEN__ ON)
         endif()
         fprime_cmake_status("[LIBRARY] Adding modules from ${LIBRARY_NAME}")
-        include("${LIBRARY_MANIFEST}")
+        include("${LIBRARY_MANIFEST}" EXCLUDE_FROM_ALL)
```
LeStarch, line 249: Same here

LeStarch (changes requested):
We at minimum need to remove the erroneous changes, fix the merge conflict with API.cmake, and get CI to pass.

LeStarch:
Closing this PR, we are reviewing whether EXCLUDE_FROM_ALL is the correct approach.  CCB will re-review the issue.  We can reopen this if we decide this is truly the right approach.

Issue #5144, opened 2026-05-12
The TlmChan's Run_handler is inefficient


bitWarrior:
| | |
|:---|:---|
|**_F´ Version_**| 4.2.1 |
|**_Affected Component_**| TlmChan |
---
## Feature Description

The TlmChan's Run_handler performs two full scans of the entire channel table every cycle, regardless of how many channels were actually updated. This is inefficient. Adding a second linked list that tracks which channel is actually updated can be easily implemented. 

## Rationale

The TlmChan's Run_handler executes at 10Hz. Depending on the size of the FSW's CTDB and the ratio of high rate of updating telemetry channels, this change can significantly reduce the per-cycle execution time of TlmChan, reducing the risk of task timing overflows, additional CPU cycles for application processing, etc.

thomas-bc:
CCB: could use an `Fw::Set`

thomas-bc:
Superseded by https://github.com/nasa/fprime/issues/5253 which adds more implementation notes.

PR #5246, opened 2026-06-05
Optimize tlmchan run handler

youio:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5144 |
|**_Documentation Included (y/n)_**| n  |

---
## Change Description

This PR adds an ArraySet to the TlmSet structure for storing updated channels. When a channel is updated in TlmRecv_handler, a pointer to its corresponding entry is stored in the active set. Later, Run_handler loops through the set and sends out all updated channels. 

## Rationale

This change improves the runtime complexity of Run_handler from O(n) where n is the total number of channel ids to O(m) where m is the number of updated channels since the last run. This ties the runtime complexity of the function to its frequency as m will decrease as call frequency increases.

This update does not change functionality so the old unit tests should cover its behavior.

## Future Work

While this change significantly improves the efficiency of Run_handler, it does introduce inefficiencies elsewhere. For instance it decreases memory efficiency by adding two new buffers of size TLMCHAN_HASH_BUCKETS. It also introduces some runtime complexity to TlmRecv_handler equal to O(m) as ArraySet loops through all elements on each insert. Using linked list logic or a hash map would remove these inefficiencies, but I'm assuming these should be avoided in favor of determinism. If that is not the case, this solution could be improved in the future using a linked list or hash map implementation.

AI was used for code base understanding and debugging.


thomas-bc:
Closing because we have closed the original issue and opened https://github.com/nasa/fprime/issues/5253 which we think is a better solution to the problem

Issue #5269, opened 2026-06-10
Fpy pip dependency


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.2 |
|**_Affected Component_**| requirements.txt |
---
## Feature Description
* Add fprime-fpy package as a dependency of FPrime

## Rationale

Feature request from FPrime industry users

PR #5270, opened 2026-06-10
Add Fpy dependency back to requirements.txt

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5269 |

---
## Change Description

* Add fprime-fpy==0.4.0 as a pip requirement of the latest devel FPrime
* This transitively adds only one new dependency: lark>=1.2.2 (not a large package) so this should not increase install times noticeably

## Rationale

Industry users of FPrime requesting Fpy version be pinned to FPrime version. I was told that @LeStarch was okay with this change

None

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,6 +20,7 @@ Flask-RESTful==0.3.10
 fprime-fpl-layout==1.0.4
 fprime-fpl-write-pic==1.0.4
 fprime-fpp==3.3.0a5
+fprime-fpy==0.4.0
 fprime-gds==4.2.0
 fprime-tools==4.3.0a3
 fprime-visual==1.0.2


zimri-leisher:
CC @Brian-Campuzano

LeStarch (approved):

LeStarch:
As long as the dependencies continue to pass the installation checks, I am happy.  @zimri-leisher we will need a decent integration test somewhere to ensure that the pinned version here matches the actual C++ code. 

Would you propose something?

zimri-leisher:
@LeStarch in order to integration test this, we would need a deployment which has an FpySequencer.
I currently maintain the zimri-leisher/fprime-fpy-testbed repo, which has such a deployment and is up to date with latest devel. It actually has three sequencers connected by a SeqDispatcher so I can test sequence calling.
I'm open to ideas on how you want to proceed with this. 
1. We could move fprime-fpy-testbed into fprime-community, and have Github Actions clone it in commits to F Prime and run the Fpy integration tests against it (actually, probably just a small subset of the integration tests, as the full ones take several minutes).
2. We could add an FpySequencer to the Ref deployment, or even swap the CmdSequencer. I know this will require some further discussion, but I'm definitely open to finding out what you would need for this to happen.
3. We could add a second FpyRef deployment to nasa/fprime, which would just be a copy of fprime-fpy-testbed.


I think my ideal option is #2, with the compromise that we would only add an FpySeq, not swap out the existing one. For those who use Ref to start a project, they would be able to simply comment out whichever sequencer they don't want.

Issue #5287, opened 2026-06-12
Backport FileDownlink timeout removal to v3.6


nathancheek:
| | |
|:---|:---|
|**_F´ Version_**| `v3.6.4` |
|**_Affected Component_**| `Svc.FileDownlink` |
---
## Problem Description

A buffer ownership bug was fixed in `v4.2.0`, see https://github.com/nasa/fprime/pull/4555. It would be nice to backport this fix to `v3.6.x`.

This bug was found by implementing and placing a `ThrottledBufferRepeater` downstream of a `FileDownlink` component, and setting the throttle timing high enough that it triggered a `FileDownlink` timeout. This brought to light a bug in `bufferReturn_handler()`, where `FileDownlink` doesn't properly handle receiving buffers while in `COOLDOWN` mode.

The fix in #4555 should keep us from hitting the `COOLDOWN` bug in normal conditions, since `FileDownlink` won't enter `COOLDOWN` mode until the buffer has been returned. However, I would also like to update `bufferReturn_handler()` so it knows how to handle returned buffers while in `COOLDOWN` mode. Here's how `bufferReturn_handler()` currently handles returned buffers based on the component's mode:
* `IDLE`: Does nothing
* `CANCEL`: Sends appropriate next packet
* `WAIT`: Sends appropriate next packet
* `DOWNLINK`: asserts, but this is okay since the `DOWNLINK` mode is a temporary mode only used while inside of a `bufferReturn_handler()` call, so we should never be in `DOWNLINK` mode at the start of a `bufferReturn_handler()` call
* `COOLDOWN`: asserts, not ideal! But would require the downstream component to return a packet at an unexpected time.

I think returned buffers should be handled the same way in both `IDLE` mode and `COOLDOWN` mode. Seeing as how `bufferReturn_handler()` already has a non-asserting offramp for buffers returned during `IDLE` mode, I propose we do the same for `COOLDOWN` mode. I believe this `COOLDOWN` fix would also be useful for F Prime `v4`... if there is any concurrence I can open an issue and PR to pull that change into `dev`.

I will open a PR shortly with the #4555 fix cherry-picked on top of v3.6.4, along with a fix so `bufferReturn_handler()` handles `COOLDOWN` mode.

## How to Reproduce

1. Place a component downstream of `FileDownlink` that takes longer than `FileDownlink`'s `m_timeout`, but shorter than `m_timeout + m_cooldown`, to return.
[...]

PR #5288, opened 2026-06-12
Backport FileDownlink timeout removal to v3.6

nathancheek:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5287 |

---
## Change Description

This backports the #4555 bugfix to v3.6.x. It also updates `FileDownlink` to handle returned buffers during `COOLDOWN` the same way they are handled during `IDLE` mode, rather than asserting.

## Rationale

Downstream throttling should not cause `FileDownlink` to assert.

I've tested these `FileDownlink` changes on Linux and the component seems to be operating as expected. Unfortunately, this PR changes the `FileDownlink::configure()` function header, removing the `timeout` argument, which breaks CI builds. Question: Should we change the external test deployments so they build? Should we override `FileDownlink::configure()` so it has a backwards-compatible function header with a dummy `U32` argument where `timeout` used to be?

## Future Work

I propose bringing forward the handling of returned buffers in `COOLDOWN` mode to F Prime `v4`, making `bufferReturn_handler()` handle returns in `COOLDOWN` mode the same way it handles returns in `IDLE` mode.

N/A

### DIFF
--- a/Ref/Top/RefTopology.cpp
+++ b/Ref/Top/RefTopology.cpp
@@ -48,7 +48,6 @@ NATIVE_INT_TYPE rateGroup3Context[Svc::ActiveRateGroup::CONNECTION_COUNT_MAX] = 
 // A number of constants are needed for construction of the topology. These are specified here.
 enum TopologyConstants {
     CMD_SEQ_BUFFER_SIZE = 5 * 1024,
-    FILE_DOWNLINK_TIMEOUT = 1000,
     FILE_DOWNLINK_COOLDOWN = 1000,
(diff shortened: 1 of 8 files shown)


nathancheek:
I just pushed changes to overload the `configure()` function, allowing it to remain backwards compatible with existing deployments.

SterlingPeet:
Backwards compatibility is the correct choice.  I'll kick the CI.

SterlingPeet:
@LeStarch I was out at MSFC delivering a spacecraft last week, so I missed the CCB.  Does this need to wait or is there an async approval process?  Should be low stakes because this is a backport.

Otherwise I am good to approve it.

PR #5295, opened 2026-06-15
Update FPP to v3.0.0a9; revise FppTest

bocchino:
This PR does the following:
* Update FPP to v3.0.0a9. This release revises the validity check for generated enums as described in this PR: https://github.com/nasa/fpp/pull/1011.
* Revise FppTest to use the new code generation and avoid undefined behavior in C++.
* Enable some tests that were disabled by mistake in FppTest.


bocchino:
Looks like RHEL8 is complaining about unnecessary comparisons.

bocchino:
Here is an example of the error that occurs on RHEL 8:
```
/__w/fprime/fprime/build-fprime-automatic-native/F-Prime/default/config/FileDispatchPortEnumAc.cpp:129:32: error: comparison is always true due to limited range of data type [-Werror=type-limits]
       return ((serialTypeValue >= SEQUENCE_FILE_PORT) && (serialTypeValue <= MAX_FILE_DISPATCH_PORTS));
                ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```
In this case `serialTypeValue` is unsigned, and `SEQUENCE_FILE_PORT` is 0, so the compiler is complaining about a redundant comparison.

If we could disable these warnings, that would be helpful. It would keep the code generation simple and uniform across all types.

bocchino:
> If we could disable these warnings, that would be helpful. It would keep the code generation simple and uniform across all types.

On second thought, let me put in the code gen to avoid the unnecessary comparisons. It should be straightforward.

thomas-bc (dismissed):

bocchino:
I made one small refinement to the code gen. I'm waiting for the 3.3.0a9 release to build, and then I'll update this PR and it will be ready to merge!

bocchino:
@LeStarch @thomas-bc It looks like something has gone wrong with the spell check in CI: https://github.com/nasa/fprime/actions/runs/27634586750/job/81718254466?pr=5295.

LeStarch (approved):

PR #5301, opened 2026-06-17
Fix FW_NUM_ARRAY_ELEMENTS for FPP array types

### LINKED ISSUE #5155: FW_NUM_ARRAY_ELEMENTS produces an unexpected result when called on FPP arrays
| | |
|:---|:---|
|**_F´ Version_**| v4.2.1ish |
|**_Affected Component_**| Fw/Types |
---
## Problem Description

This code produces an unexpected result:

```
# In Example.fpp
array Example = [32] U32

// In Example.cpp
Example array = {
    1234,
    5678,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
};

constexpr U32 arrayLength = FW_NUM_ARRAY_ELEMENTS(array);
```

The length is 33 instead of 32. This is of course because FW_NUM_ARRAY_ELEMENTS uses sizeof, which isn't appropriate for FPP arrays. (FPP arrays are not plain C arrays and have other memory allocated.)

## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-4.18.0-553.117.1.el8_10.x86_64-x86_64-with-glibc2.28
Python version: 3.12.13
CMake version: 3.26.0
Pip version: 26.0.1
Pip packages:
    fprime-tools==4.2.0
    fprime-gds==4.0.2.dev63+g8731b5b83
    fprime-fpp==3.3.0a1
Project submodules:
    https://github.com/nasa/fprime.git @ v4.2.1-26-g7e26e25554
    https://github.com/fprime-community/fprime-baremetal @ f27f574
    https://github.com/fprime-community/fprime-vorago @ fe67818
```

## How to Reproduce

See above.

## Expected Behavior

FW_NUM_ARRAY_ELEMENTS should either (a) cause a compile-time error if used on a FPP array, or (b) should produce the expected/correct result if used on a FPP array.

arpitjain099:
Fixes #5155.

This switches FW_NUM_ARRAY_ELEMENTS to a C++ helper that still handles raw C arrays and also recognizes FPP-generated array types by reading their compile-time SIZE. In C builds the macro still uses sizeof exactly as before.

I added a regression test in FppTestProject/FppTest/array/ArrayToStringTest.cpp that checks FW_NUM_ARRAY_ELEMENTS on several FPP array types.

Verification:
- python3 -m cmake -S . -B build-5155-ut -G "Unix Makefiles" -DBUILD_TESTING=ON -DCMAKE_BUILD_TYPE=Debug
  - fails early in this environment because fprime-version-check is not installed
- python3 -c "<focused compile script>"
  - pass: compiled and ran C++ and C checks against Fw/Types/BasicTypes.h
  - confirmed raw arrays still report element count and FPP-style arrays report SIZE

### DIFF
--- a/Fw/Types/BasicTypes.h
+++ b/Fw/Types/BasicTypes.h
@@ -36,6 +36,7 @@ extern "C" {
 #endif
 #include <config/FPrimeNumericalConfig.h>
 #include <inttypes.h>  // Standard integer types and printf macros
+#include <stddef.h>    // size_t
 
 // Compiler checks
 #if defined(__GNUC__) || defined(__llvm__) || defined(PLATFORM_OVERRIDE_GCC_CLANG_CHECK)
@@ -87,7 +88,30 @@ typedef double F64;   //!< 64-bit floating point (double). Required for compiler
 /*----------------------------------------------------------------------------*/
 /* Useful macro definitions                                                   */
 /*----------------------------------------------------------------------------*/
+#ifdef __cplusplus
+}
+
+namespace Fw {
+namespace BasicTypes {
+
+template <typename T, size_t N>
+constexpr size_t fwNumArrayElements(const T (&)[N]) {
+    return N;
+}
+
+template <typename ArrayType>
+constexpr auto fwNumArrayElements(const ArrayType&)
+    -> decltype(ArrayType::SIZE, sizeof(typename ArrayType::ElementType), size_t{}) {
+    return static_cast<size_t>(ArrayType::SIZE);
+}
+
+}  // namespace BasicTypes
+}  // namespace Fw
+
+#define FW_NUM_ARRAY_ELEMENTS(a) (::Fw::BasicTypes::fwNumArrayElements(a))  //!< number of elements in an array
+#else
 #define FW_NUM_ARRAY_ELEMENTS(a) (sizeof(a) / sizeof((a)[0]))  //!< number of elements in an array
+#endif
 // Deprecated: prefer std::max/std::min from <algorithm> in C++ code.
 // Kept for C compatibility only.
 #define FW_MAX(a, b) (((a) > (b)) ? (a) : (b))  //!< MAX macro (deprecated in C++, use std::max)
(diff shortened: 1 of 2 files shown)


Fw/Types/BasicTypes.h
```diff
@@ -87,7 +88,30 @@ typedef double F64;   //!< 64-bit floating point (double). Required for compiler
...
+template <typename T, size_t N>
+constexpr size_t fwNumArrayElements(const T (&)[N]) {
+    return N;
+}
+
+template <typename ArrayType>
+constexpr auto fwNumArrayElements(const ArrayType&)
+    -> decltype(ArrayType::SIZE, sizeof(typename ArrayType::ElementType), size_t{}) {
+    return static_cast<size_t>(ArrayType::SIZE);
+}
```
bocchino, line 106: I don't think we should include this code. It allows `FW_NUM_ARRAY_ELEMENTS` to be applied to any type that defines a member `::SIZE`. But that's neither necessary nor sufficient for being an array type. Also, if there's no `::SIZE` member, there will be a template error explosion.

Could we convert this code into a static assertion that says "not a primitive array"? That would limit the application of `FW_NUM_ARRAY_ELEMENTS` to C-style arrays, which was the original intent.

LeStarch: I agree with this approach specified by @bocchino

LeStarch:
Closing this PR as it was not the desired fix.  We (CCB) have selected the static-error approach.

arpitjain099:
> Closing this PR as it was not the desired fix.  We (CCB) have selected the static-error approach.

@LeStarch noted, thank you for the update.

Fw/Types/BasicTypes.h
```diff
@@ -87,7 +88,30 @@ typedef double F64;   //!< 64-bit floating point (double). Required for compiler
...
+#ifdef __cplusplus
+}
+
+namespace Fw {
+namespace BasicTypes {
+
+template <typename T, size_t N>
+constexpr size_t fwNumArrayElements(const T (&)[N]) {
+    return N;
+}
```
bocchino, line 100: This template does look like a correct solution to the issue raised in #5155. @arpitjain099 You could resubmit the PR with this change, together with a static assertion for the catch-all template.

PR #5303, opened 2026-06-17
Restrict FW_NUM_ARRAY_ELEMENTS to C-style arrays

### LINKED ISSUE #5155: FW_NUM_ARRAY_ELEMENTS produces an unexpected result when called on FPP arrays

arpitjain099:
Follow-up to #5301. That one was closed in favor of the static-error approach the CCB picked, and @bocchino suggested keeping the C-style array template and turning the catch-all into a static assertion. This does that.

In C++ `FW_NUM_ARRAY_ELEMENTS` now resolves to a small constexpr helper with two overloads:
- `const T (&)[N]` returns `N` for real C arrays
- a catch-all whose body just `static_assert`s

So a raw array still reports its element count, and anything that isn't a C array (FPP array types, pointers, etc.) is now a compile error with a readable message instead of the old sizeof result that was wrong for FPP arrays (#5155). The assert keys off `sizeof(T) == 0` so it only trips when the catch-all is actually selected, no template-error explosion.

The C build is unchanged and still uses the sizeof macro.

I went through every current use of the macro in the tree and they all pass raw C arrays, so nothing else needed touching. Added a TypesTest case for the C-array result.

Verified locally:
- compiled and ran the C-array cases (1D, 2D, single element), counts are right
- confirmed calling it on an FPP-like type (one with `::SIZE`) now fails to compile with the new message
- C path still compiles and returns the right count

Fixes #5155.


LeStarch (dismissed):

LeStarch:
@arpitjain099, thanks for submitting a follow-on PR!  This made it easy to review!

LeStarch:
The issue with CI is this:  `BasicTypes.hpp` includes `BasicTypes.h` in its own `extern C` block and as a result the templates are still in C linkage.

Since `BasicTypes.hpp` is intended for including in C++ code, I might recommend that you move these C++ definitions into `BasicTypes.hpp` and leave only the C function in a C-only block

e.g. 
```
// Only define FW_NUM_ARRAY_ELEMENTS for C types in this C header
#if not __c_plus_plus
#define FW_NUM_ARRAY_ELEMENTS(a) sizeof(a)/sizeo(a[0])
#endif
```

Everything else can be in BasicTypes.hpp where it is native C++

arpitjain099:
Thanks, that was the linkage issue exactly. Moved the templates and the C++ macro into BasicTypes.hpp (native C++, after the extern "C" include) and left BasicTypes.h with just the C sizeof macro under `#ifndef __cplusplus`, like you suggested.

Checked the include scenarios locally: C-array use through BasicTypes.hpp still gives the right count, a non-array (type with `::SIZE`) now fails to compile with the static_assert message, and the C path is unchanged. Every current macro user in the tree reaches it via BasicTypes.hpp/FPrimeBasicTypes.hpp, so nothing else needed touching.

LeStarch (dismissed):

LeStarch:
```use of 'this' pointer is only allowed within the evaluation of a call to a 'constexpr' member function```

Since we are now passing to a templated function, we get errors when dealing with member variables.

**Comment by arpitjain099** [comment]
Good catch, that was the by-reference helper forcing evaluation of `this`. Switched to the old-school unevaluated sizeof trick: a declared-only `fwArrayElementCountHelper(const T (&)[N])` that returns `char[N]`, called inside `sizeof`. Since sizeof never evaluates its operand, `FW_NUM_ARRAY_ELEMENTS(this->m_member)` works again and stays a constant expression, same as the original macro. A non-array still won't compile (no matching overload).

One tradeoff to flag: with this form the error for a non-array is a plain "no matching function for fwArrayElementCountHelper" instead of the nice static_assert text. Getting both the friendly message and the unevaluated/member-safe behavior in C++14 is doable with a type trait, but it runs into the static-data-member definition rules, so I went with the simpler sizeof version. Happy to switch to the trait if you'd rather have the message.

Added a TypesTest case that uses a member array through `this->` in a constant expression so this path stays covered. Built it locally against C++14 (the repo default) for the array, member, 2D, and rejection cases.

**Comment by LeStarch** [comment]
I think it may be a harder sell without a good clean error message.  However, I have not seen how complicated the type-trait approach might be.

**Comment by arpitjain099** [comment]
Agreed on the error message. Turns out the trait isn't bad, so I went with it. It's just:

```cpp
template <typename T>
struct ArrayElementCount {
    static_assert(sizeof(T) == 0, "FW_NUM_ARRAY_ELEMENTS may only be used on a primitive (C-style) array");
    static const std::size_t value = 0;
};
template <typename T, std::size_t N>
struct ArrayElementCount<T[N]> {
    static const std::size_t value = N;
};
#define FW_NUM_ARRAY_ELEMENTS(a) (static_cast<std::size_t>(::Fw::BasicTypes::ArrayElementCount<decltype(a)>::value))
```

`decltype(a)` is unevaluated, so this->member in a constant expression still works (kept the test for that), and reading `::value` through the cast is a constant-expression read, so there's no out-of-line definition needed under C++14. I checked that with a link test where the result is bound to a `const std::size_t&` param, no undefined reference.

Now a non-array (or a decayed pointer) fails with:

```
error: static assertion failed due to requirement 'sizeof(FppLike) == 0': FW_NUM_ARRAY_ELEMENTS may only be used on a primitive (C-style) array
```

Built locally against C++14 for the array, 2D, member, pointer, and non-array cases, and it's clang-format clean.

bocchino:
Looks good! But why assert `sizeof(T) == 0`? Why not simply assert `false` in the non-array case? Otherwise it seems the non-array template would accept a non-array type of size 0, e.g., an empty struct. Also the error
```
[static assertion failed due to requirement 'sizeof(FppLike) == 0'` ](error: static assertion failed due to requirement 'sizeof(FppLike) == 0': FW_NUM_ARRAY_ELEMENTS may only be used on a primitive (C-style) array)
```
could be misleading. It suggests that `FW_NUM_ARRAY_ELEMENTS` may be called on all and only types that have size zero -- which is not the intent.

arpitjain099:
Good points. Switched the condition to `std::is_array<T>::value`, so the failed requirement now reads as the real intent:

```
error: static assertion failed due to requirement 'std::is_array<FppLike>::value': FW_NUM_ARRAY_ELEMENTS may only be used on a primitive (C-style) array
```

On the empty-struct question: that one's actually safe either way, since C++ guarantees `sizeof` of any complete type is at least 1, so `sizeof(T) == 0` never holds and the assert always fired for non-arrays (I confirmed an empty struct is rejected). But you're right that the requirement text was misleading, and is_array is clearer regardless.

I kept it as a dependent condition rather than a plain `static_assert(false)` on purpose: a non-dependent false assert in an uninstantiated primary template is ill-formed on the older compilers in CI (GCC 8 on RHEL8, GCC 11 on ubuntu), so it would fire even for the array case. is_array gives us both the dependence and the clear message.

bocchino:
That error message looks good! I think we have a workable solution. But now do we need the templates at all? It seems it would be simpler to have the C++ version of the macro assert that the argument is an array and then compute the size the C way.

arpitjain099:
Good call. Dropped the two-template trait for a single helper that does exactly that:

```cpp
template <typename T>
constexpr std::size_t numArrayElements() {
    static_assert(std::is_array<T>::value, "FW_NUM_ARRAY_ELEMENTS may only be used on a primitive (C-style) array");
    return std::extent<T>::value;
}
#define FW_NUM_ARRAY_ELEMENTS(a) (::Fw::BasicTypes::numArrayElements<decltype(a)>())
```

It still keys off `decltype(a)` and takes no runtime argument, so it stays a constant expression and works on member arrays like `this->member` (the reason a by-reference helper didn't work). Same error message as before. Verified array, 2D, this->member, and the non-array case against C++14.

bocchino (approved):
This looks good to me! Thanks for working through this issue so carefully.

LeStarch (approved):

Issue #5307, opened 2026-06-18
Overflow in POP_EVENT


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.2 |
|**_Affected Component_**| FpySequencer  |
---
## Problem Description

Calculation of needed size in POP_EVENT directory can overflow for large message sizes.

PR #5308, opened 2026-06-18
Fix overflow in POP_EVENT directive in FpySequencer

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**|  #5307|
|**_Has Unit Tests (y/n)_**|  y|

---
## Change Description

* Fix an overflow in POP_EVENT directive allowing a read out of bounds
* Add an assertion so the read would have crashed before getting there
* Add a UT to test that this doesn't happen again


zimri-leisher:
@LeStarch  I think this PR is ready for your review, it failed a check but I think it just is a flaky UT? Seemed to time out.

LeStarch (approved):

LeStarch:
@zimri-leisher it was "cancelled" which is usually a glitch in the running system.  I re-ran it and it worked fine!

Issue #5314, opened 2026-06-18
FpySequencer no cmd response if running VALIDATE after first RUN with NO_BLOCK


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.2 |
|**_Affected Component_**|  FpySequencer |
---
## Problem Description

* Under certain combinations of actions, you won't get a cmd response from VALIDATE

## How to Reproduce

1. Run RUN with NO_BLOCK on a fresh boot of FpySequencer
2. Run VALIDATE, observe no cmd response

PR #5315, opened 2026-06-18
FpySequencer Fix no response from validate, refactor to prevent this class of issu…

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5314 |

---
## Change Description

* Fix issue #5314, if you run RUN with NO_BLOCK and then run VALIDATE, you don't get a response from VALIDATE
* Refactor the state machine and member variables slightly to prevent this class of issue from occurring
* The refactor brought up some other issues with the design of the SEQ_BASE_DIR param, which necessitated making it a compile time constant shared among all sequencers
  * The refactor made it so that the seq file path, block state and sequence arguments are all modified together, at once
  * This meant that the file path had to be set the moment the user calls the RUN cmd, e.g.
  * But because setting the file path can fail (due to truncation when the SEQ_BASE_DIR param would make the result too large), we need to allow it to gracefully fail
  * Right now, there are only two ways that a sequence can fail to run: in the validation state (header issues, crc, etc) or in the run state (early exit with error, overflow, etc)
  * I decided to refactor to move the file path concatenation into the validate() method, which can just return Fw::Success:FAILURE if there is overflow.
  * However, by moving the calculation of the actual sequence file path into the validate() method, we can't report the actual sequence file path in the seqStartOut port until we've already started validating
  * This is a problem, because it means there would be a brief window where the sequence is actually occupied (running VALIDATE) but hasn't reported that it is occupied (through seqStartOut)
  * But if we make the SEQ_BASE_DIR path a compile time constant, then both sides know what the path prefix is, so I don't need to include it in the seqStartOut port call
  * This brings up some more fundamental issues about sequence identity. Also, I think the existing CmdSequencer should use the SEQ_BASE_DIR compile time constant too, but that's for later.

**Important note:** this makes it somewhat more difficult to call sequences at absolute paths, because there is no option to turn off the suffix. However, I think this is okay, as it's a compile time constant, so as long as you know what FSW you're running, you know the path prefix, and you can always just back out of the prefix directory.

One small other change was adding single quotes around file names in event format strings, so that you can differentiate between the path and the event msg.

Also, updated the state machine SDD

Used for writing tests and some final edits. Mostly human written (a rarity for me these days)


Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -645,29 +620,31 @@ class FpySequencer : public FpySequencerComponentBase {
...
 
-    // assigned by the user via cmd
-    // length is FileNameStringSize
-    Fw::FileNameString m_sequenceFilePath;
+    // the execution arguments of the sequence:
+    // * the file name
+    // * the block state
+    // * the sequence runtime args
+    // See the struct member docstrings for more info
+    Svc::FpySequencer_SequenceExecutionArgs m_sequenceExecArgs;
```
zimri-leisher, line 633: to begin running the sequence, we need these three args (filename, block state and seq runtime args). Store them all in a single struct so updates to them can be "atomic"

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -645,29 +620,31 @@ class FpySequencer : public FpySequencerComponentBase {
...
+
+    // the full (including the SEQ_BASE_DIR prefix) path
+    // to the sequence file. Note: not updated until validate() is run
+    Fw::FileNameString m_fullSequenceFilePath;
```
zimri-leisher, line 637: the file name provided in the sequence exec args is not the "final" filename that we actually try to open. we need to prefix with seq base dir first.

Svc/FpySequencer/FpySequencerStateMachine.fppi
```diff
@@ -50,12 +54,8 @@ state machine SequencerStateMachine {
     ####################
     @ simply raises the "entered" signal
     action signalEntered
-    @ sets the current sequence file path member var
-    action setSequenceFilePath: SequenceExecutionArgs
-    @ sets the block state of the sequence to be run
-    action setSequenceBlockState: SequenceExecutionArgs
-    @ sets the arguments to pass to the sequence
-    action setSequenceArguments: SequenceExecutionArgs
+    @ sets the sequence execution arguments (seq path, block state, seq args)
```
zimri-leisher, line 57: this is a clear example of how i made the args "atomic"--you can only update all three at once

Svc/FpySequencer/FpySequencerValidationState.cpp
```diff
@@ -23,20 +23,48 @@ void FpySequencer::deallocateBuffer(Fw::MemAllocator& allocator) {
...
+// returns the base directory against which sequence file paths are resolved.
+// defaults to the SEQ_BASE_DIR config constant; overridable in tests/deployments
+const char* FpySequencer::getSequenceBaseDir() const {
+    return Fpy::SEQ_BASE_DIR;
+}
+
 // loads the sequence in memory, and does header/crc/integrity checks.
 // return SUCCESS if sequence is valid, FAILURE otherwise
 Fw::Success FpySequencer::validate() {
-    FW_ASSERT(this->m_sequenceFilePath.length() > 0);
```
zimri-leisher, line 29: honestly no idea how this assert got in here. it was a clear mistake. the sequencer should not crash if the path length is zero. this has been here for a while.

It is impossible to send an empty string arg in fprime-gds, which I think is the wrong behavior, but was in practice preventing this from occurring.

Svc/FpySequencer/docs/sdd.md
```diff
@@ -49,14 +49,15 @@ The following diagram represents the states of the `FpySequencer`.
 stateDiagram-v2
   direction LR
   state "IDLE
-entry / clearBreakpoint, clearSequenceFile
```
zimri-leisher, line 52: state machine sdd hadn't been updated in a while

zimri-leisher:
@LeStarch @thomas-bc whenever you get a sec, this PR is ready for review

zimri-leisher:
TODO make sure we can call abs paths

EDIT--updated description to make it clear that you will have to prefix "../../" etc out of the SEQ_BASE_DIR prefix if you want to call something outside of seq base dir. See desc for rationale

Svc/FpySequencer/FpySequencerStateMachine.cpp
```diff
@@ -409,9 +355,11 @@ void FpySequencer::Svc_FpySequencer_SequencerStateMachine_action_report_seqStart
...
 ) {
     if (this->isConnected_seqStartOut_OutputPort(0)) {
-        // report that the sequence started to internal callers
-        // NOTE: Sequence Arguments would be cleared if a VALIDATION command is sent, not a full RUN command.
-        this->seqStartOut_out(0, this->m_sequenceFilePath, this->m_sequenceArgs);
+        // report that the sequence started to internal callers.
+        // we report the raw operator-supplied path rather than m_fullSequenceFilePath
+        // because this action runs in the VALIDATING entry, before validate() resolves
+        // the base dir prefix into m_fullSequenceFilePath
+        this->seqStartOut_out(0, this->m_sequenceExecArgs.get_filePath(), this->m_sequenceExecArgs.get_runArgsBuf());
```
LeStarch, line 362: I think this breaks encapsulation.

The expected contract of this port is that it includes the sequence file path that was run.  The remote receiver should not need to know the internals of how FpySequencer constructs the file path in order to have that information.  Nor should the remote component need to know how to handle operator commands to this particular component.

LeStarch (commented):
I am not sure I agree with this change.   It pushes the need to understand internal details of this component (how full paths are constructed, what command arguments look like) through a port interface and into any user of that port.

This means that any time this component is adjusted, remote components need to be adjusted in parallel to ensure that they remain in-sync with this component.

It would be much better, in my opinion, to send out the full-file path on the seq start port as that is really the expected contract of that port.  There are several other ways that could be explored to accomplish this:

1. Compute the full file path where it was, store the path and the status, check status in validate and emit an error.  
2. Consider only "starting a sequence" if it passes validate
3. Allow for errors that can occur before validate runs (like we did before).

zimri-leisher:
Fair point. I think I was imagining a future in which all sequence ecosystem components would use the SEQ_BASE_DIR prefix, so everyone would be "speaking the same language" when it comes to sequence paths.


> 1. Compute the full file path where it was, store the path and the status, check status in validate and emit an error.
> 2. Consider only "starting a sequence" if it passes validate
> 3. Allow for errors that can occur before validate runs (like we did before).

I'll consider these options again. Another option which I think might be the best, is a new "seqRunReject" port back to the SeqDispatcher.

lntutor (commented):
Reviewed exact head 99db18382c209ab0183e29ff38b4014c019e4d9f. The focused Svc/FpySequencer check builds and passes all 266 tests, including ValidateAfterNoBlockRun, and the unified SequenceExecutionArgs update fixes the stale NO_BLOCK state that suppressed the later VALIDATE response.

I also traced cancellation ordering in the generated component: internal state-machine signals run at priority 9 while CANCEL is priority 8, so the entered/validate/result chain resolves m_fullSequenceFilePath before cancellation can report it. I did not find an additional correctness blocker beyond the existing seqStartOut path-contract discussion.

LeStarch:
This PR has been unmaintained for over a month, and the number of conflicts above show that it has begun to code rot.  I will close it for now.  It can be opened up once it is back in the foreground.

PR #5352, opened 2026-07-02
fix: only resolve path variables for absolute DEPENDS entries

### LINKED ISSUE #3803: DEPENDS Will Resolve Local Files
| | |
|:---|:---|
|**_F´ Version_**| |
---
## Problem Description

The `resolve_path_variables` function will detect local files and resolve them.  If DEPENDS specifies a dependency on name that collides with a local file, this breaks.

mvanhorn:
## Summary
`fprime__process_module_setup` ran `resolve_path_variables` over the entire `DEPENDS` list, which treated plain module dependency names as file paths and could mangle them when a local file of the same name existed.

## Changes
For the `DEPENDS` control set, resolve path variables only for entries that are absolute paths; leave non-absolute entries (module names) untouched. All other control sets keep the existing behavior.

## Testing
Added `TestDependsCollision` test deployment and a `test_feature.py` case exercising a module name that collides with a local file.

Fixes #3803


thomas-bc:
Need confirmation from @LeStarch that this is the intended behavior. I was taking an educated guess in the ticket comments, but you may have other insights?

LeStarch:
I don't think this is the fix we want...as it will not attempt to resolve local paths. I don't know there is a solution outside of changing the DEPEND behavior like we did for SOURCES/AUTOCODER_INPUTS.

I'll close this for now.

LeStarch:
I reopened this....as it caught my attention that this isn't just an AI PR slung at us.  So let's think a bit more.

We have 2 options:

1. We can split DEPENDS into two categories:

DEPENDS: dependencies on targets
LINKS: dependencies on files (.a and .so)

Where depends is used whole-sale, and LINKS resolves paths.

This is a clean, but breaking solution.

2. We could fix this by checking if the DEPENDS is a target, resolving if it is NOT a target.  

This isn't great because a target can be defined later on, and still collide with a same-named file....but it reduces the scope.

@mvanhorn thoughts?

mvanhorn:
Option 1 feels like the right destination. The root problem in #3803 is that DEPENDS overloads two meanings (target names and file paths), and every fix inside that ambiguity is a heuristic. Splitting the intent into DEPENDS (targets, used wholesale) and LINKS (files, path-resolved) removes the guessing entirely, and it's consistent with what you already did for SOURCES/AUTOCODER_INPUTS.

The breakage could be staged so it lands gently:

1. Introduce LINKS with resolution semantics.
2. Keep DEPENDS accepting file-ish entries for one release, but emit a deprecation warning when an entry contains a path separator or resolves to an existing file, pointing at LINKS.
3. After the deprecation window, DEPENDS becomes targets-only.

Option 2 works as a transitional heuristic, but the later-defined-target collision you mention is the same class of ambiguity that produced #3803 - it just moves where the surprise happens, and it's order-dependent, which makes it hard to debug when it bites.

If you're good with direction 1, I'm happy to rework this PR into the LINKS split plus deprecation warning (or split it into two if you'd rather see LINKS land separately). If you'd prefer the smaller scope of option 2 for now, I can implement that instead with a docs note about the collision caveat.

LeStarch:
I am going to close this PR in that case given that it does not address the desired path forward.

PR #5354, opened 2026-07-02
Fixed error when running BufferLogger UT as root

bevinduckett:
| | |
|:---|:---|
|**_Related Issue(s)_**|  none |
|**_Has Unit Tests (y/n)_**|  yes |
|**_Documentation Included (y/n)_**|  n/a|
|**_Generative AI was used in this contribution (y/n)_**|  Yes |

---
## Change Description

Updated the BufferLogger unit test to skip LogFileValidation when run as root (not uncommon when running inside a container) 

## Rationale

Done because in UNIX root skips file read/write permission checks. 

Verify BufferLogger UT  still compiles for all targeted OSs & passes when  executed as root (I verified in my dev container)  

## Future Work

N/A 

Used Google to lookup how to skip a test case & the macros for detecting unix systems


LeStarch:
@bevinduckett seems like there are multiple merge conflicts here.

Also, instead of defining a helper flag (UNISTD_INCLUDEDED) I'd recommend we just repeat the same #if check in both places.  That way you don't have to find the first one, read that it defined a new flag, and then find again to pickup the second.

LeStarch:
@bevinduckett let me know if you'd like help resolving the above!

bevinduckett:
Hi @LeStarch , it should be fixed. Can you kick off the jobs?

LeStarch (approved):

thomas-bc (approved):

PR #5368, opened 2026-07-08
fix(Ccsds): reject Space Packet type/APID mismatches in deframer

### LINKED ISSUE #5290: SpacePacketDeframer does not reject invalid primary-header control fields
SpacePacketDeframer currently validates buffer size, header deserialization, and packet data length, but it does not reject packets whose CCSDS primary-header control fields are invalid.

I do not mean the APID enum-value question in #5281. That issue already covers APID validity.

I mean the other control fields in the 6-byte primary header:
- packet version number
- packet type
- secondary header flag
- sequence flags

Code path:
- `Svc/Ccsds/SpacePacketDeframer/SpacePacketDeframer.cpp`
- `dataIn_handler(...)`

Current behavior in that handler:
1. reject undersized buffers
2. deserialize `SpacePacketHeader`
3. validate `packetDataLength`
4. extract APID and sequence count
5. forward the payload through `dataOut_out`

The implementation comment lists the full primary-header structure:
- packet version number
- packet type
- secondary header flag
- APID
- sequence flags
- sequence count
- packet data length

But the code only enforces the length side of that contract. It never checks whether those non-APID control fields contain values that are valid for the protocol and for the component's intended use.

That leaves a class of malformed packets that are structurally long enough to pass through the deframer and reach downstream logic as if they were valid payload packets.

Why I think this is a bug:
- The SDD says the component "validates the Space Packet header and extracts the payload":
  - `Svc/Ccsds/SpacePacketDeframer/docs/sdd.md`
- The existing unit tests cover nominal packets, undersized buffers, and invalid length handling, but I do not see coverage for invalid primary-header control fields:
  - `Svc/Ccsds/SpacePacketDeframer/test/ut/SpacePacketDeframerTester.cpp`

A concrete example would be a packet with:
- valid serialized header bytes
- valid `packetDataLength`
- invalid packet version number, or invalid sequence flags for the supported Space Packet mode

Today that appears to flow through the component, because `dataIn_handler` does not reject it before building `contextCopy` and calling `dataOut_out`.

Expected behavior:
- reject packets whose primary-header control fields violate the CCSDS packet contract enforced by this component
- emit the existing invalid-packet error path (`InvalidPacket` and `errorNotify`) instead of forwarding the payload

I have not attached a patch yet. I wanted to check first whether maintainers want `SpacePacketDeframer` to validate only length, or the full set of primary-header control fields that this component depends on.

Abhishek21g:
## Summary
Validates CCSDS Space Packet **Packet Type** against known APIDs (`FW_PACKET_COMMAND` / `FW_PACKET_TELEM`) before forwarding payloads. Extends #5290 beyond packet version number checks.

## Testing
`Svc_Ccsds_SpacePacketDeframer_ut_exe`: 11/11 pass (incl. new `testInvalidPacketTypeForApid`).

Fixes #5290

### DIFF
--- a/Svc/Ccsds/SpacePacketDeframer/SpacePacketDeframer.cpp
+++ b/Svc/Ccsds/SpacePacketDeframer/SpacePacketDeframer.cpp
@@ -28,6 +28,23 @@ bool isValidPacketVersionNumber(const SpacePacketHeader& header) {
     return pvn == static_cast<U16>(ComCfg::Pvn::SPACE_PACKET_PROTOCOL);
 }
 
+bool isValidPacketTypeForApid(const SpacePacketHeader& header) {
+    const U16 packetIdentification = header.get_packetIdentification();
+    const U16 packetType = (packetIdentification & SpacePacketSubfields::PktTypeMask) >> SpacePacketSubfields::PktTypeOffset;
+    const U16 apidValue = packetIdentification & SpacePacketSubfields::ApidMask;
+    if (!ComCfg::Apid::isValid(apidValue)) {
+        return true;
+    }
+    const ComCfg::Apid::T apid = static_cast<ComCfg::Apid::T>(apidValue);
+    if (apid == ComCfg::Apid::FW_PACKET_COMMAND) {
+        return packetType == 1U;
+    }
+    if (apid == ComCfg::Apid::FW_PACKET_TELEM) {
+        return packetType == 0U;
+    }
+    return true;
+}
+
 }  // namespace
 
 // ----------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


thomas-bc:
Please use the Pull Request template when submitting a PR.

As per [CCSDS blue book $ 4.1.3.3.2.3](https://ccsds.org/Pubs/133x0b2e2.pdf):

> The exact definition of [packet type] ‘telemetry Packets’ and ‘telecommand Packets’ needs to be established by the project that uses this protocol.

Because of that explicit vagueness in the protocol, I don't think this change is desired. Some validation has already been implemented for #5290 - I don't think restricting further is a better model. While I certainly hear that this should hold most of the time, who's to tell how the packetType field is going to be used by projects.

Thoughts?

Abhishek21g:
Thanks @thomas-bc — fair point on CCSDS §4.1.3.3.2.3 leaving packet type semantics project-defined. Withdrawing this; a global command/telemetry restriction is too opinionated for the framework core.

Abhishek21g:
Closing per above.

Issue #5243, opened 2026-06-05
Add Timer Source Selection to RawTime API.


bdshenker:
| | |
|:---|:---|
|**_F´ Version_**| 4.x|
|**_Affected Component_**|  Os|
---
## Feature Description
Currently, the Os::RawTime API provides no mechanism to select between different hardware timer sources.
We propose to add an optional timer source selection parameter to the RawTime API.

The RawTimeSource enumeration would be defined in the platform configuration (e.g. in `fprime-vxworks`). The default definition would contain only the single RAWTIME_DEFAULT enumerator:

  namespace Os {
      enum RawTimeSource {
          RAWTIME_DEFAULT = 0     // Platform's default timer (current behavior)
      };
  }

Add RawTimeSource parameter to RawTime constructor (defaulting to RAWTIME_DEFAULT). Initial implementation could define only RAWTIME_DEFAULT to maintain existing behavior. 
For backward compatibility, existing code without explicit source selection should continue to work.
Platform implementations can add source-specific values as needed and ensure that these values are properly processed.

This request is only scoped to the Os::RawTime API. Other Os APIs, e.g., File, etc, should be evaluated on a case-by-case basis before adding similar timer selection features.

## Rationale

Our VA41630 platform provides multiple timer peripherals:
1) Cascaded Timer Peripherals (current default) - 64-bit counter at APB1 frequency,
2) ARM SysTick - 24-bit counter at SysClk frequency (wraps ~200 msec)

Different use cases benefit from different timer characteristics—short high-resolution profiling needs SysTick, while long-duration timekeeping needs the 64-bit  counter.

LeStarch:
CCB:
1. The enumeration should be provided by the platform provider (author of the RawTime implementation)
2. Uses of raw time should be updated to take the specified clock
3. Each raw time shall provide a default clock selections.

bdshenker:
This update was merged as part of commit https://github.com/nasa/fprime/commit/2a5178389109ad887d57897e86298f39ab6e58ad. Therefore, this ticket can now be closed.

thomas-bc:
Thanks for circling back on the ticket. Closing.

PR #5370, opened 2026-07-08
Add optional per-member timing for PassiveRateGroup (#5293)

### LINKED ISSUE #5293: Request support for timing individual cycle group members
| | |
|:---|:---|
|**4.x**| |
|**Svc**|  |
---
## Feature Description

To be able to measure the execution time of individual components within a cycle group, we propose  to add functionality similar to the following for each iteration of CycleIn_handler() in PassiveRategroup.cpp:

```
              // Capture start time for this component
              componentStartTime.now();
              // Invoke the component
              this->RateGroupMemberOut_out(port, this->m_contexts[port]);
              // Capture end time for this component
              componentEndTime.now();
              // Calculate component execution time
              (void)componentEndTime.getDiffUsec(componentStartTime, componentTime);
              // Track max time per component
              if (componentTime > this->m_componentMaxTime[port]) {
                  this->m_componentMaxTime[port] = componentTime;
              }
```
Header file will need the following:

`U32 m_componentMaxTime[NUM_RATEGROUPMEMBEROUT_OUTPUT_PORTS];  // It should be sized to at least 50 members.`

This enhancement is contingent upon tickets 5243 and 5244 being approved and completed first. This will enable us to develop an implementation optimized for our platform.

We welcome consideration of other solutions.

## Rationale

We need the ability to measure the execution time of individual components within a cycle group in production deployment. Given the performance constraints of the VA41630 platform, the solution must minimize CPU overhead. After this feature is implemented, we will employ the timer implementation that best aligns with our platform's performance constraints and architectural requirements.

Abhishek21g:
## Summary

Implements the compile-time configuration approach discussed in #5293 (per @celskeggs): `PASSIVE_RATE_GROUP_ENABLE_COMPONENT_TIMING` defaults to **off** so timing code is eliminated at compile time for deployments that do not need it.

When enabled in project `PassiveRateGroupCfg.hpp`:
- Measures execution time for each connected `RateGroupMemberOut` port per cycle
- Tracks per-port high-water marks in `m_componentMaxTime[]`
- Emits `ComponentMaxTimeUpdated` activity event when a port sets a new max (infrequent snapshot, not continuous telemetry proliferation)

Also adds `Log` / `LogText` ports to `PassiveRateGroup` (required for events).

## Rationale

Addresses JPL flight-software need to measure individual rate-group member execution time on resource-constrained platforms (VA41630 class), without taxing users who do not need instrumentation.

Follows maintainer thread: avoids tinker-toy topology boilerplate; uses compile-time flag instead of always-on overhead.

## Verification

```
cd TestDeploymentsProject/build-fprime-automatic-native-ut
ninja Svc_PassiveRateGroup_ut_exe
./bin/Darwin/Svc_PassiveRateGroup_ut_exe
```

All existing PassiveRateGroup UT tests pass with timing disabled (default).

## Notes

- Depends on #5243/#5244 for optimized timer selection on target platforms (as noted in issue)
- Projects enable timing by setting `PASSIVE_RATE_GROUP_ENABLE_COMPONENT_TIMING` to `1` in their config override

Closes #5293

### DIFF
--- a/Svc/PassiveRateGroup/PassiveRateGroup.cpp
+++ b/Svc/PassiveRateGroup/PassiveRateGroup.cpp
@@ -15,10 +15,17 @@
 #include <Fw/Types/Assert.hpp>
 #include <Os/Console.hpp>
 #include <Svc/PassiveRateGroup/PassiveRateGroup.hpp>
+#include <config/PassiveRateGroupCfg.hpp>
 
 namespace Svc {
 PassiveRateGroup::PassiveRateGroup(const char* compName)
-    : PassiveRateGroupComponentBase(compName), m_cycles(0), m_maxTime(0), m_numContexts(0) {}
+    : PassiveRateGroupComponentBase(compName), m_cycles(0), m_maxTime(0), m_numContexts(0) {
+#if PASSIVE_RATE_GROUP_ENABLE_COMPONENT_TIMING
+    for (FwIndexType port = 0; port < NUM_RATEGROUPMEMBEROUT_OUTPUT_PORTS; port++) {
+        this->m_componentMaxTime[port] = 0;
+    }
+#endif
+}
 
 PassiveRateGroup::~PassiveRateGroup() {}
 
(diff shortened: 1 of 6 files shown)


timcanham:
I think emitting an event whenever a max is reached could result in event spamming if the timing is variable and increasing. We should find a telemetry channel based approach. I wouldn't recommend approving this until it is addressed.

Abhishek21g:
Thanks @timcanham — agreed events would spam. Withdrawing in favor of telemetry approach; happy to help.

Abhishek21g:
Closing per above — telemetry approach is the right path for #5293.

PR #5371, opened 2026-07-08
Telemetry port cycle times on Svc.PassiveRateGroup

Kronos3:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5293, #5370 |

---
## Change Description

Adds a telemetry channel that will be optionally emitted (enabled at compile-time) to measure the execution time on each output port of the PassiveRateGroup.

## Rationale

Useful for measuring performance metrics for embedded systems

## Future Work

We may want to telemeter max cycle time instead (or parameterize that). I've found the raw cycle times on every cycle to be very useful even though it incurs an overall runtime overhead.

No AI was used


LeStarch (commented):
I like this PR and how it is implemented....but why narrow the output from U32 to U16?  This means long-duration RGs might overflow this.  Probably rare, but not impossible.

Kronos3:
@LeStarch Fair. This is the channel I've been using on my SAMD21 which has very tight memory buffers but I'll probably end up manually changing the definition to `U16` whenever I want to enable this channel anyway. I'll update it to `U32` by default.

LeStarch (approved):

PR #5378, opened 2026-07-08
Add POP_SERIALIZABLE to DeserializeDirectives

Lex-ari:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5327 |

---
## Change Description

Adds the missing `POP_SERIALIZABLE` case to `deserializeDirective` in `Svc/FpySequencer/FpySequencerRunState.cpp`, plus a `deserialize_popSerializable` unit test.

## Rationale

PR #5327 added the `POP_SERIALIZABLE` directive's handler and `dispatchDirective` case but not its `deserializeDirective` case, so the directive fails to decode at run time.

Added unit tests to the FpySequencer to check deserialization of the popSerializable command.

## Future Work

Implementing POP_SERIALIZABLE on the Fpy-backend side.

Claude Code was used to identify the missing deserialize case and wrote implementation and unit tests.


zimri-leisher (changes requested):
This was an unfortunate miss on the original PR and on my review. All other directives have this unit test, but ideally we should use CPP semantics to prevent this from being possible. I think if we removed the default case in the switch statement, it would be a compiler error if there were a directive ID that weren't handled. @Lex-ari can you try that? Make sure to still include the warning event and return Fw::SUCCESS::FAILURE, but just do it after the switch case.

Once you do that, please confirm (by removing a switch case) that the code fails to compile, to make sure this works.

Lex-ari:
> This was an unfortunate miss on the original PR and on my review. All other directives have this unit test, but ideally we should use CPP semantics to prevent this from being possible. I think if we removed the default case in the switch statement, it would be a compiler error if there were a directive ID that weren't handled. @Lex-ari can you try that? Make sure to still include the warning event and return Fw::SUCCESS::FAILURE, but just do it after the switch case.
> 
> Once you do that, please confirm (by removing a switch case) that the code fails to compile, to make sure this works.

Can confirm that removing the default case produces a compile error. Will make this change.

zimri-leisher (approved):

zimri-leisher:
Thanks. Looks good to me

Lex-ari:
@LeStarch @thomas-bc Any work on our end to make this work with the integration tests?

Lex-ari:
merge does it.

zimri-leisher (approved):

thomas-bc (approved):

thomas-bc:
Zephyr is expected to misbehave right now - no worries

Lex-ari:
yay

Issue #5400, opened 2026-07-11
Dictionary constant for default time base


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.2 |
|**_Affected Component_**| Fw.TimeBase |
---
## Feature Description

A default value for Fw.TimeBase should be included in the dictionary.

## Rationale

Various ground data tools (fprime-gds, Fpy, fprime-seqgen) want to construct time objects and send them to the flight software. However, they often force the user to input which time base they should use, or simply make a guess. This would allow each deployment to specify which timebase ground systems should assume by default.

thomas-bc:
CCB: should match the behavior of current `fprime-seqgen --timebase <SPECIFY_TIMEBASE>` - defaults should be intentional, accepting this variable would force projects to have to use it - some may not want to.

PR #5401, opened 2026-07-11
Add FW_TIME_BASE_DEFAULT constant to dict

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5400 |

---
## Change Description

* Add FW_TIME_BASE_DEFAULT constant to FPP and dictionary. Defaults to TB_WORKSTATION_TIME

## Rationale

Allows ground data tools not to guess when choosing a default time base. Saves a decent bit of time for anyone who has to manually input times again and again. This was present in Fpy, users have to pass TB_WORKSTATION_TIME to the `time()` function again and again when constructing ISO 8601 dates.

None

### DIFF
--- a/default/config/FpConfig.fpp
+++ b/default/config/FpConfig.fpp
@@ -94,3 +94,6 @@ dictionary enum TimeBase : FwTimeBaseStoreType {
     TB_SC_TIME = 3,          @< Time as reported by the spacecraft clock.
     TB_DONT_CARE = 0xFFFF    @< Don't care value for sequences. If FwTimeBaseStoreType is changed, value should be changed (Required)
 } default TB_NONE;
+
+@ The default time base that various ground software systems should assume the flight software uses
+dictionary constant FW_TIME_BASE_DEFAULT = TimeBase.TB_WORKSTATION_TIME


default/config/FpConfig.fpp
```diff
@@ -94,3 +94,6 @@ dictionary enum TimeBase : FwTimeBaseStoreType {
     TB_SC_TIME = 3,          @< Time as reported by the spacecraft clock.
     TB_DONT_CARE = 0xFFFF    @< Don't care value for sequences. If FwTimeBaseStoreType is changed, value should be changed (Required)
 } default TB_NONE;
```
thomas-bc, line 96: Should we use the value as the actual `default` of the `TimeBase` enum?

thomas-bc (changes requested):

default/config/FpConfig.fpp
zimri-leisher: I thought about this for a while. I decided against it, because I think they're doing two unrelated things. One is configuring what the default ground software option for TIME_BASE should be. The other is configuring what the default value should be when you construct this enum, and I just don't think those are necessarily the same thing.

thomas-bc: ah I see - I did not catch that nuance. thanks

thomas-bc: would be worth clarifying in the docstring not to confuse the next reader :D

zimri-leisher: That's fair, I can clarify that!

thomas-bc:
CCB recommended another approach, closing.

Issue #4273, opened 2025-10-07
SeqDispatcher cancel sequences cmd


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.1.0 |
|**_Affected Component_**| SeqDispatcher |
---
## Feature Description

Add a command to cancel all sequences dispatched by the dispatcher, or just a specific sequence

## Rationale

This would be useful to guarantee no other seqs are running during critical periods

PR #5423, opened 2026-07-16
Add ABORT by name cmd to SeqDisp

Lex-ari:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4273, [lestarch-autobot/fprime#201](https://github.com/lestarch-autobot/fprime/issues/201l) |

---
## Change Description

- Sequencers now have a telemetry channel indicating their currently running sequence.
- Added an `ABORT` command to SeqDisp that cancels any running sequence matching a given file name. 
  - Uses a seqCancelOut port, as well as events (SequenceAborted / AbortSequenceNotFound).

## Rationale

Operators want telemetry indicating the currently running sequence.
Operators also want a way to cancel a sequence by name rather than finding the specific sequencer running a sequence.

`ABORT` intentionally does not clear it's own state table, but rather lets sequencers clear themselves via seqDoneIn. This may produce many cmdResponses as a result.

## Future Work

None. (But watch for an `ABORT_ALL` sequence PR coming!)

Claude Code was used to add ports and functions to the SeqDisp, and check for packet usage on updated telemetry channels. All code was reviewed.


Svc/CmdSequencer/CmdSequencerImpl.cpp
```diff
@@ -336,6 +338,7 @@ void CmdSequencerComponentImpl ::CS_START_cmdHandler(FwOpcodeType opcode, U32 cm
 
     this->m_blockState = Svc::BlockState::NO_BLOCK;
     this->m_runMode = RUNNING;
+    this->tlmWrite_CS_CurrentSequence(this->m_sequence->getStringFileName());
```
zimri-leisher, line 343: I would be careful about the order here. Why are you putting this before the first step? It's not clear to me where this statement should go, and it's worth justifying.

zimri-leisher: In fact, I think maybe we should put the performcmd_Step last here, after we report the sequence started. Just be consistent with the order in the other three sites. This component is a mess! How are there three places where a sequence can start??

Svc/SeqDispatcher/SeqDispatcherCommands.fppi
```diff
@@ -15,3 +15,9 @@ async command RUN_ARGS(
 
 @ Logs via Events the state of each connected command sequencer
 async command LOG_STATUS() opcode 2
+
+@ Aborts any running sequence matching the given file name
```
zimri-leisher: I would change the wording to make it super clear it can abort multiple. So like "Aborts all running sequences matching the given file name"
Also, I think the command should be CANCEL. It's called CANCEL on both of the sequencers.

zimri-leisher (changes requested):

Svc/CmdSequencer/CmdSequencerImpl.cpp
```diff
@@ -339,8 +339,8 @@ void CmdSequencerComponentImpl ::CS_START_cmdHandler(FwOpcodeType opcode, U32 cm
     this->m_blockState = Svc::BlockState::NO_BLOCK;
     this->m_runMode = RUNNING;
     this->tlmWrite_CS_CurrentSequence(this->m_sequence->getStringFileName());
-    this->performCmd_Step();
     this->log_ACTIVITY_HI_CS_CmdStarted(this->m_sequence->getLogFileName());
+    this->performCmd_Step();
```
zimri-leisher, line 345: @LeStarch @thomas-bc this change in particular I want your eyes on. I think this was a mistake in the original CmdSequencer, or at least an inconsistency.

TestDeploymentsProject/Ref/Top/RefPackets.fppi
```diff
@@ -287,4 +287,6 @@ telemetry packets RefPackets {
 
 } omit {
   CdhCore.cmdDisp.CommandErrors
+  # String channel; not suited to fixed-layout packets
```
LeStarch: Why?  String should have a fixed layout.

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -185,4 +185,31 @@ void SeqDispatcher::LOG_STATUS_cmdHandler(const FwOpcodeType opCode, /*!< The op
...
+    bool canceled = false;
+    for (FwIndexType idx = 0; idx < SeqDispatcherSequencerPorts; idx++) {
+        // only slots actively running the named sequence are candidates
+        const bool running = this->m_entryTable[idx].state != SeqDispatcher_CmdSequencerState::AVAILABLE;
+        if (running && this->m_entryTable[idx].sequenceRunning == fileName) {
+            if (this->isConnected_seqCancelOut_OutputPort(idx)) {
+                this->seqCancelOut_out(idx);
+            }
+            // Entry table is cleared via seqDoneIn_handler
+            this->log_ACTIVITY_HI_SequenceCanceled(static_cast<U16>(idx),
```
LeStarch: I think these changes need to be in the if-connected block.  You cannot cancel a sequence without the port connected.

Svc/SeqDispatcher/SeqDispatcher.cpp
```diff
@@ -185,4 +185,31 @@ void SeqDispatcher::LOG_STATUS_cmdHandler(const FwOpcodeType opCode, /*!< The op
...
+                                                   Fw::LogStringArg(this->m_entryTable[idx].sequenceRunning));
+            this->tlmWrite_canceledCount(++this->m_canceledCount);
+            canceled = true;
+        }
+    }
+
+    if (canceled) {
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::OK);
+    } else {
+        this->log_WARNING_LO_CancelSequenceNotFound(Fw::LogStringArg(fileName));
```
LeStarch, line 212: It also fails if the port is not connected.

LeStarch (changes requested):

Svc/CmdSequencer/CmdSequencerImpl.cpp
LeStarch: Why?  The perform step is just moved after an event rather than before.  Why is this relevant?

zimri-leisher: I think it could be confusing for the sequence to execute a command before the CS_CmdStarted event comes through. This isn't directly relevant to the PR, but I suggested the change to Alex because I noticed it while reviewing the other places where sequences can start, for consistency.

Svc/SeqDispatcher/SeqDispatcher.cpp
Lex-ari: Right I agree with this and moved it into the If loop.

I'm wondering, is there ever a case where the entryTable status would mark itself as RUNNING without a port attached? This is handled by the getNextAvailableSequenceIdx() function. If this is the case, should I instead fo a FW_ASSERT here for an impossible case?

Svc/SeqDispatcher/SeqDispatcher.cpp
Lex-ari: Is this handled above when the logic is placed inside the if loop?

Lex-ari:
Also curious if I should include cmdSeq.CS_CurrentSequence in RefPackets - and if that might break any of the CI tests

LeStarch (approved):

Lex-ari:
🎉

Issue #5424, opened 2026-07-17
Potential overflow in time arithmetic in FpySequencer


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.2 |
|**_Affected Component_**| FpySequencer  |
---
## Problem Description

From Claude:

"""
32-bit overflow in checkStatementTimeout — the timeout can miss, leaving you stuck with the safety net enabled. FpySequencerRunState.cpp:716-718:
cppU64 currentUSeconds = currentTime.getSeconds() * 1000000 + currentTime.getUSeconds();
getSeconds() returns U32, so the multiply happens in 32-bit unsigned arithmetic and wraps every 4294.97 seconds of time-base value — for an epoch-based TB_WORKSTATION_TIME it's wrapped from the start, permanently. Two failure modes: if the wrapped value crosses a 2³² boundary between dispatch and the check, currentUSeconds - dispatchUSeconds underflows in U64 to an enormous number → spurious timeout that kills a healthy statement (any sleep or command await spanning one of those ~71.6-minute boundaries). Conversely, true elapsed time aliases mod 2³² µs, so a statement that's been hung for 4300 seconds reads as ~5 seconds elapsed — a genuinely dead command await can sail past its configured timeout and stay stuck. Fix: static_cast<U64>(currentTime.getSeconds()) * 1000000ULL on both lines.
"""

This bug hasn't been manually confirmed, but looks real.

zimri-leisher:
@Abhishek21g you've been helpful on a few issues, if you're up for it this one would be good for you too, I think.

Abhishek21g:
Opened PR #5426 with the fix — `static_cast<U64>(getSeconds())` before the multiply, plus a unit test with seconds > 4294 to exercise the overflow path. Thanks for the tag @zimri-leisher!

PR #5426, opened 2026-07-17
fix(FpySequencer): prevent 32-bit overflow in checkStatementTimeout

Abhishek21g:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5424 |

---
## Change Description

Widen `getSeconds()` (U32) to `U64` before multiplying by `1000000` in `FpySequencer::checkStatementTimeout()`:

```cpp
U64 currentUSeconds = static_cast<U64>(currentTime.getSeconds()) * 1000000 + currentTime.getUSeconds();
U64 dispatchUSeconds = static_cast<U64>(this->m_runtime.currentStatementDispatchTime.getSeconds()) * 1000000 +
                       this->m_runtime.currentStatementDispatchTime.getUSeconds();
```

Unit tests (fail on unfixed code, pass with this change):
- `checkStatementTimeoutU32OverflowMissed` — dispatch@0s, now@4300s, timeout=10s
- `checkStatementTimeoutU32OverflowSpurious` — dispatch@4294s, now@4295s, timeout=10s

## Rationale

`getSeconds()` returns `U32`. Multiplying by the `int` literal `1000000` is performed in 32-bit unsigned arithmetic and wraps every `2^32` µs (~4294.97 s). For epoch-based `TB_WORKSTATION_TIME` that wrap is already past from t=0, so:

1. **Missed timeout:** elapsed aliases mod `2^32` µs (e.g. 4300 s → ~5.03 s) and a hung statement can sail past `STATEMENT_TIMEOUT_SECS`.
2. **Spurious timeout:** crossing the product wrap boundary (4294→4295) makes wrapped `current < dispatch`; `U64` subtraction underflows to a huge elapsed and kills a healthy 1 s wait.

The first UT in this PR (seconds≈5000 with a small delta) did **not** catch this — both times stay in the same wrap window so the difference is still correct. These replacements do.

To confirm the UTs fail without the fix (repro zimri's check):

```bash
# on this branch, temporarily drop the casts, then rebuild/run FpySequencer UTs
# expected: checkStatementTimeoutU32OverflowMissed FAIL (got noTimeout)
#           checkStatementTimeoutU32OverflowSpurious FAIL (got statementTimeout)
```

Standalone arithmetic mirror of the exact expressions (same results):

| Case | Buggy elapsed | Fixed elapsed | timeout=10s |
|------|---------------|---------------|-------------|
| 0→4300 s | 5,032,704 µs (~5 s) → **no timeout** | 4,300,000,000 µs → **timeout** | proves miss |
| 4294→4295 s | U64 underflow → **timeout** | 1,000,000 µs → **no timeout** | proves spurious |

## Future Work

None planned.

Issue analysis and initial draft from AI; overflow failure modes and unit-test cases validated manually against the U32 multiply wrap. PR description filled to match the project template.

IAMAI

### DIFF
--- a/Svc/FpySequencer/FpySequencerRunState.cpp
+++ b/Svc/FpySequencer/FpySequencerRunState.cpp
@@ -728,8 +728,8 @@ Signal FpySequencer::checkStatementTimeout() {
         return Signal::result_checkStatementTimeout_noTimeout;
     }
 
-    U64 currentUSeconds = currentTime.getSeconds() * 1000000 + currentTime.getUSeconds();
-    U64 dispatchUSeconds = this->m_runtime.currentStatementDispatchTime.getSeconds() * 1000000 +
+    U64 currentUSeconds = static_cast<U64>(currentTime.getSeconds()) * 1000000 + currentTime.getUSeconds();
+    U64 dispatchUSeconds = static_cast<U64>(this->m_runtime.currentStatementDispatchTime.getSeconds()) * 1000000 +
                            this->m_runtime.currentStatementDispatchTime.getUSeconds();
 
     U64 timeoutUSeconds = static_cast<U64>(timeout * 1000000.0f);
--- a/Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
+++ b/Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
@@ -2138,23 +2138,35 @@ TEST_F(FpySequencerTester, checkStatementTimeout) {
     ASSERT_EQ(result, Signal::result_checkStatementTimeout_statementTimeout);
 }
 
-TEST_F(FpySequencerTester, checkStatementTimeoutLargeSeconds) {
-    // Seconds value > 4294 triggers 32-bit overflow without the U64 cast (issue #5424)
-    Fw::Time testTime(TimeBase::TB_WORKSTATION_TIME, 0, 5000, 500000);
+TEST_F(FpySequencerTester, checkStatementTimeoutU32OverflowMissed) {
+    // Without widening getSeconds() to U64 before * 1000000, the product wraps every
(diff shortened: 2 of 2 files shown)


zimri-leisher:
I checked the unit test you added, and it doesn't fail even with the current FpySequencer. Please make sure that you can prove this is not a false positive with a unit test before implementing.

thomas-bc:
@zimri-leisher is this change desired or are you recommending we close the PR ?

@Abhishek21g please make sure you use the PR template to submit PRs

Abhishek21g:
@zimri-leisher you're right — the first UT was a false positive. Both times sat in the same `2^32` wrap window, so the *difference* was still correct even with the U32 multiply.

Replaced it with two cases that diverge without the cast (pushed in bea8f88):

1. **Missed timeout** — dispatch@0s, now@4300s, timeout=10s  
   Buggy elapsed ≈ 5.03s → `noTimeout`; fixed → `statementTimeout`
2. **Spurious timeout** — dispatch@4294s, now@4295s (crosses the product wrap), timeout=10s  
   Buggy U64 underflow → `statementTimeout`; fixed (1s elapsed) → `noTimeout`

Quick way to re-check on this branch: temporarily remove the `static_cast<U64>(...)` casts in `checkStatementTimeout`, rebuild FpySequencer UTs — both new tests should fail; restore casts and they pass.

@thomas-bc — change is still desired if the overflow is considered real (arithmetic above matches the issue description). Also updated the PR body to the project template. Happy to close if you'd rather park until after CCB / manual confirmation.

zimri-leisher:
@thomas-bc yes, I actually asked for it. It was a bug found by Claude, I didn't verify it myself but thought it would be a good one for someone else to bang out. I'm double checking the UT now.

zimri-leisher (approved):
Confirmed the UTs fail without the change, and succeed with it.

thomas-bc (approved):

LeStarch (approved):

PR #5138, opened 2026-05-11
CCSDS File Delivery Protocol - CfdpManager

Brian-Campuzano:
| | |
|:---|:---|
|**_Related Issue(s)_**| #2768 |
|**_Has Unit Tests (y/n)_**| y |
|**_Documentation Included (y/n)_**| y |
|**_Generative AI was used in this contribution (y/n)_**| y |

---
## Change Description

This PR adds the CfdpManager component, an F´ implementation of the CCSDS File Delivery Protocol (CFDP) standard ported from [NASA's Core Flight System (cFS) CFDP Application v3.0.0](https://github.com/nasa/CF/releases/tag/v3.0.0). CfdpManager provides both Class 1 (unacknowledged) and Class 2 (acknowledged) file transfer capabilities, designed to replace the standard FileUplink and FileDownlink components with guaranteed file delivery support.

**Attribution and License:**
- **Substantial portions ported from**: NASA Core Flight System (cFS) CFDP (CF) Application v3.0.0
- **Original License**: Apache License 2.0
- **Original Copyright**: Copyright (c) 2019 United States Government as represented by the Administrator of the National Aeronautics and Space Administration
- **NASA Docket**: GSC-18,447-1

**Ported components** (retain original NASA copyright):
- Core CFDP engine and transaction management logic (`Engine.cpp`, `Transaction.hpp`, `TransactionTx.cpp`, `TransactionRx.cpp`)
- Protocol state machines for transmit and receive operations
- Utility functions for file handling and resource management (`Utils.cpp`, `Channel.cpp`)
- Chunk and gap tracking for Class 2 transfers (`Chunk.cpp`, `Clist.cpp`)

**New F´-specific implementations**:
- CfdpManager component wrapper with F´ ports, commands, events, telemetry, and parameters
- Object-oriented PDU encoding/decoding classes based on F´ `Serializable` interface
- Timer implementation using F´ time primitives
- FileHandlingCfdp subtopology for integration

See [ATTRIBUTION.md](Svc/Ccsds/CfdpManager/ATTRIBUTION.md) for complete file-by-file attribution breakdown.

**Summary of changes:**
- Full PDU support: Metadata, FileData, EOF, FIN, ACK, NAK
- Multi-channel support with configurable throttling and timer parameters
- Port-based file transfer interface compatible with existing F´ components (e.g., DpCatalog)
- Comprehensive unit tests covering Class 1/2 transactions, PDU handling, and edge cases
[...]

Issue #4909, opened 2026-03-27
Update App<->Com interface (ComQueue&Router) to split PacketDescriptor from Payload


thomas-bc:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

F´ currently uses a single payload buffer when passing data from the app layer to the comms stack of components (in ComQueue for outgoing comms, and receiving from FprimeRouter for incoming comms).

From that payload buffer, the first `FwPacketDescriptorType` bytes are the packet descriptor. This is not easy to work with.

We should update those interfaces to accept an arbitrary payload buffer, and a Packet Descriptor argument.

Here's an example of how this would simplify for example FileDownlink:

```diff
// FileDownlink.cpp
-    // Serialize the packet descriptor FW_PACKET_FILE to the buffer
-    Fw::SerializeStatus status = buffer.getSerializer().serializeFrom(static_cast<FwPacketDescriptorType>(Fw::ComPacketType::FW_PACKET_FILE));
-    Fw::Buffer offsetBuffer(buffer.getData() + sizeof(FwPacketDescriptorType),
-                            buffer.getSize() - static_cast<Fw::Buffer::SizeType>(sizeof(FwPacketDescriptorType)));
-    status = filePacket.toBuffer(offsetBuffer);
-    this->bufferSendOut_out(0, buffer);
+    this->bufferSendOut_out(0, buffer, ComCfg::Apid::FW_PACKET_FILE);
```

Kronos3:
If you put the Apid before the buffer can you pipe this output port into a serial port in the com interface since the serial representation will be kept the same?

thomas-bc:
yes that's a very accurate observation.... but I wonder if it is desirable to rely on that? Sounds like a means to delay a required change. idk. happy to make it so.

thomas-bc:
note: the App<->Com interface was documented here: https://fprime.jpl.nasa.gov/devel/docs/how-to/develop/custom-uplink-downlink-data/ – I'm still hoping we can improve the model here, but at least it's workable

PR #5433, opened 2026-07-20
Fix/router context preserving file uplink

Brian-Campuzano:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Updated `FprimeRouter` to maintain buffer `FrameContext` for multi-path uplink designs. This also necessitated changing `FileUplink` to use `Svc.ComDataWithContext` instead of `Fw.BufferSend`, so the `ComCfg::FrameContext` associated with an uplinked buffer is preserved across the full round trip rather than being dropped.

- **`Svc/Interfaces/Router.fpp`**: `fileOut` and `fileBufferReturnIn` changed from `Fw.BufferSend` to `Svc.ComDataWithContext`.
- **`Svc/FprimeRouter`**: all return paths (command, file, unknown) now return the buffer with the context it was received with, instead of substituting a fresh empty context. `fileBufferReturnIn_handler` gains the context parameter and echoes it. The component holds no added state.
- **`Svc/FileUplink`**: `bufferSendIn`/`bufferSendOut` changed to `Svc.ComDataWithContext`; the received context is threaded through and echoed on every return path.

## Rationale

`FprimeRouter` previously discarded the `FrameContext` on every return path, substituting a default-constructed one. This makes it impossible for an upstream component to correlate a returned uplink buffer with the context it was routed under (e.g. virtual channel / APID). By carrying the context on the file-uplink ports and echoing it back unchanged, downstream/upstream components can rely on the context surviving the round trip. This is a prerequisite for multi-path uplink designs where the return must be routed back to the source path by its context.

- **FprimeRouter UT**: a sentinel `vcId` is stamped on input and asserted preserved on `fileOut`, `unknownDataOut`, and `dataReturnOut` (covering the command, unknown, unknown-unconnected, and file-return paths).
- **FileUplink UT**: `sendFilePacket` stamps a sentinel `vcId` and asserts it is preserved on `bufferSendOut`, so every file-packet test exercises context preservation.
- **Integration**: the `Ref` deployment (which wires `ComCcsds.fileUplinkOut -> FileHandling.fileUplinkBufferSendIn` and the matching return) builds clean with the converted ports -- no topology edits were required, since the type change propagates through the subtopologies' re-exported ports.
- Review focus: confirm no other consumers of `Router.fileOut` / `FileUplink.bufferSendIn` remain on `Fw.BufferSend`.

## Future Work

Will need to make a similar update to CfdpManager in #5138 depending on the order of the PRs closing.

AI was used for unit-test authoring and codebase investigation (verifying the buffer round-trip contract and topology impact).


LeStarch:
The issue with this fix is that it assumes downstream components outside the uplink chain should have to deal with context when they really shouldn't.  The proposed fix we had was to store a map of buffer -> context in the router, and pull the context out of the map to send back downstream.  That way the fix is localize to the uplink subsystem.

LeStarch (changes requested):
See my comment in the conversation.

thomas-bc:
Also clashes with https://github.com/nasa/fprime/issues/4909 - agreed application components should not have to deal with com `context` objects

Brian-Campuzano:
@LeStarch @thomas-bc  FprimeRouter now keeps a small fixed-size buffer→context table so it can restore the original `FrameContext` when a file buffer returns on `fileBufferReturnIn`. The `fileOut`/`fileBufferReturnIn` ports stay `Fw.BufferSend`, so downstream file consumers (FileUplink, CfdpManager) are unchanged and the context-preservation concern is kept entirely inside the router.
 
The table is sized to the uplink buffer pool (`ComCfg.RouterBufferContextTableSize`), which is the hard upper bound on how many buffers can be outstanding at once, so overflow should be unreachable by construction. For robustness the router still handles two edge cases: (a) the table is full when a buffer is handed off, and (b) a returned buffer isn't found in the table. This PR currently emits a `WARNING_HI` event and safely degrades (returns the buffer with an empty context, losing only that buffer's routing context) rather than asserting.
 
 Would you prefer a hard `FW_ASSERT` on these invariants, treating them as should-never-happen design guarantees, or is the event + graceful-degrade the preferred approach? I am of two minds here.

Brian-Campuzano:
One inconsistency that I noticed during this refactor was that the port `fileBufferReturnIn` takes back ownership for buffers sent on both `fileOut` and `unknownDataOut`, so the file prefix is misleading. Something like `bufferReturnIn` would match the actual contract.

Caveat: it's defined in the `Svc.Router` interface, so renaming is interface-breaking (all implementers, topologies, and out-of-tree users). Given the doc comment already notes the real scope, this may not be worth the churn. Flagging so it's a conscious call by team.

Brian-Campuzano:
cFS Reference build failure root cause

The `External Repo: cFS Reference` failure isn't in this PR's code, it's a config duplication issue. FprimeRouter now sizes its table from `ComCfg::RouterBufferContextTableSize`, which this PR adds to `default/config/ComCfg.fpp`. But the `fprime_cfs_reference` deployment doesn't use fprime's default config. It resolves `ComCfg` from the `fprime_app` copy, which doesn't have the new constant, so the build fails at:

```
FprimeRouter.hpp:79: error: 'RouterBufferContextTableSize' is not a member of 'ComCfg'
```

This cascades into every `m_bufferContextTable` reference. Any downstream deployment that ships its own `ComCfg` copy hits the same wall until the constant is mirrored. Note that I lack write access to both `fprime_app` and its parent `fprime_cfs_reference`, so I can't push the mirrored constant or the submodule bump myself.

Question on approach: rather than adding this to the shared `ComCfg`, which forces every deployment's `ComCfg` copy to stay in sync, should I move the constant into a new `FprimeRouterCfg.fpp` under `default/config`? That keeps FprimeRouter's config self-contained and avoids the cross-repo `ComCfg` duplication. Happy to do that here if preferred.

lntutor (changes requested):
The fixed table is still undersized for the default topology, so overflow is reachable during normal ownership flow.

`FileHandlingConfig.QueueSizes.fileUplink` is 10, but `FileUplink.bufferSendIn` is asynchronous. Once the active task dequeues one message, that queue slot is free while the handler still owns the dequeued buffer; ownership is not returned until `bufferSendOut` at the end of `bufferSendIn_handler`. Therefore the file path alone can have 1 buffer in service plus 10 queued buffers = 11 outstanding contexts. `RouterBufferContextTableSize` is 10, so the 11th insertion fails without overflowing the FileUplink queue.

This is also not the uplink buffer-pool size described in the code/docs. Both bundled communication subtopologies configure 20 buffers of 2048 bytes and 30 buffers of 3000 bytes, and `BufferManager` falls through to a larger free bin. The table is additionally shared with `unknownDataOut`, so its bound must cover outstanding ownership across both handoff paths, not only the FileUplink queue.

I built and ran the exact-head FprimeRouter tests locally (10/10 pass). `TestFileOutContextTableFull` demonstrates the resulting behavior directly: with 11 distinct outstanding file buffers, the 11th context is not recorded and returns with the default context. That test is currently validating a state the shipped topology can reach, rather than an unreachable defensive path.

Please size the table from a documented upper bound on all outstanding handed-off buffers (or make that bound a component/project-specific configuration) and add a regression that relates the capacity to the default topology's queue/pool ownership bound.

There is also still a separate build blocker from placing the constant in shared `ComCfg`: the cFS external integration fails because its legitimate `ComCfg` override has no `RouterBufferContextTableSize`. Moving this setting to component-specific configuration would avoid requiring every downstream `ComCfg` copy to acquire an unrelated router constant.

Brian-Campuzano:
Good catches @lntutor. I've pushed a fix.

On the async point: agreed, and it's worth stating why the async handler doesn't change the bound. `FileUplink` never manufactures buffers. Every buffer it holds on `bufferSendIn` was allocated upstream from the uplink `commsBufferManager` pool, so the number of contexts that can ever be outstanding is capped by that pool, not by the `FileUplink` queue depth. The async handler only means one buffer can be in service while its queue slot is already free (the +1 you noted), but it still came from the same finite pool. Async can't exhaust anything the pool doesn't already bound.

That said, you're right on the substance and I was wrong to call overflow unreachable by construction. The table was 10, but the uplink pool in both bundled subtopologies is 20 @ `2048B` + 30 @ `3000B` = 50 buffers, and the table is shared across both the `fileOut` and `unknownDataOut` handoff paths. So the real upper bound on outstanding contexts is the pool total, well above 10, and overflow is reachable during normal ownership flow. `TestFileOutContextTableFull` was exercising a state the shipped topology can actually reach, exactly as you describe.

Fix in the latest push:

* The table is now sized from the uplink pool total (50), the documented upper bound on outstanding handed-off buffers across both paths, rather than a queue depth.
* The constant moved out of shared `ComCfg` into a self-contained `FprimeRouterCfg.fpp`. This resolves the cFS Reference build blocker: downstream `ComCfg` copies no longer need to carry an unrelated router constant.
* The overflow regression now relates its capacity to the pool-ownership bound rather than a number the default topology reaches in normal flow, and the SDD documents the derivation.

LeStarch (approved):

PR #5445, opened 2026-07-21
Space packet seq flags

shan-zh-26:
| | |
|:---|:---|
|**_Related Issue(s)_**| n |

---
## Change Description

This PR adds support for configurable sequence flags in the CCSDS Space Packet Protocol implementation.

## Rationale

Previously, both components hardcoded the sequence flags to `0x3` (unsegmented). This change enables support for segmented packets by making the flags configurable through the `FrameContext`.

Unit tests updated in both `SpacePacketFramerTester` and `SpacePacketDeframerTester`.

- sequence flags extraction/encoding logic
- unit tests


shan-zh-26:
There was a **cFS Build Test** failure (set_sequenceFlags not found) is likely caused by fprime-community/fprime_app's own ComCfg.fpp override lacking **sequenceFlags** field.

default/config/ComCfg.fpp
```diff
@@ -48,6 +48,7 @@ module ComCfg {
         comQueueIndex: FwIndexType  @< Queue Index used by the ComQueue, other components shall not modify
         apid: Apid                  @< 11 bits APID in CCSDS
         hasSecHdr: bool             @< Secondary header flag for SpacePacketFramer
+        sequenceFlags: U8           @< 2 bit Sequence flags (0b00=continuation, 0b01=first, 0b10=last, 0b11=unsegmented)
```
lntutor, line 51: Adding this field changes the generated `ComCfg::FrameContext` API, so projects that override `ComCfg.fpp` no longer provide the `get_sequenceFlags`/`set_sequenceFlags` methods used by these components. This is already visible in the cFS build failure, and the current [`fprime-community/fprime_app` override](https://github.com/fprime-community/fprime_app/blob/devel/fprime_config/config/ComCfg.fpp) still omits the field. Please land a tandem `fprime_app` update and pin compatible revisions in the cFS test before merging this core change; otherwise downstream builds against `devel` break immediately.

Svc/Ccsds/SpacePacketFramer/SpacePacketFramer.cpp
```diff
@@ -51,11 +51,11 @@ void SpacePacketFramer ::dataIn_handler(FwIndexType portNum, Fw::Buffer& data, c
...
     U16 sequenceCount = this->getApidSeqCount_out(0, apid, 0);  // retrieve the sequence count for this APID
-    U16 packetSequenceControl = 0;
-    packetSequenceControl |=
-        0x3 << SpacePacketSubfields::SeqFlagsOffset;  // Sequence Flags 0b11 = unsegmented User Data
-    packetSequenceControl |=
-        sequenceCount & static_cast<U16>(SpacePacketSubfields::SeqCountMask);  // 14 bit sequence count
+    const U8 seqFlags = context.get_sequenceFlags();
+    // 2 bit sequence flags | 14 bit sequence count
+    U16 packetSequenceControl = static_cast<U16>(
+        ((static_cast<U16>(seqFlags) << SpacePacketSubfields::SeqFlagsOffset) & SpacePacketSubfields::SeqFlagsMask) |
```
lntutor, line 57: `sequenceFlags` is an unconstrained `U8`, but only values 0 through 3 are valid. This mask silently aliases every other value to a different valid meaning—for example, `4` becomes continuation (`00`) and `5` becomes first (`01`)—so bad context data can mislabel packet segmentation without any signal. Please model the field as a four-value enum (defaulting to unsegmented), or explicitly reject values above 3 before encoding and add invalid-input coverage.

lntutor (commented):
I found two issues that should be addressed before this lands: the generated context API currently breaks a maintained downstream configuration, and invalid sequence-flag values are silently encoded as different valid meanings.

default/config/ComCfg.fpp
shan-zh-26: Thanks, here's the update to fprime_app. Please have a look: https://github.com/fprime-community/fprime_app/pull/10

LeStarch (approved):

PR #5476, opened 2026-07-22
Add configuration to Async I2c interfaces

Kronos3:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

1. Adds a configuration item for the number of async I2c port numbers in the Async I2C FPP interfaces
2. Adds an AsyncSyncI2c interface

## Rationale

1. Async interfaces need multiple ports to know which client to reply to
2. AsyncSync is useful for small devices with no context like the samd21.

## Future Work

None


Drv/Interfaces/AsyncSyncI2c.fpp
```diff
@@ -0,0 +1,27 @@
+module Drv {
```
LeStarch: Do we really need 2 interfaces 1 guarded and 1 sync?

LeStarch (approved):

PR #5490, opened 2026-07-22
Move PrmDb filename in FileHandling subtopology into FileHandlingConfig

### LINKED ISSUE #4425: PrmDb filename in subtopology should be configurable
| | |
|:---|:---|
|**_F´ Version_**| https://github.com/nasa/fprime/releases/tag/v4.0.0a1 |
|**_Affected Component_**| Svc/Subtoopologies/FileHandling  |
---
## Problem Description

The filename for the PrmDb is in the subtopology. This makes it so user projects can't use the subtopologies if they want to use another file name or location.

https://github.com/nasa/fprime/blob/devel/Svc/Subtopologies/FileHandling/FileHandling.fpp#L37

## Context / Environment

```
--INSERT HERE--
```

## How to Reproduce

1. 
2.
3.

## Expected Behavior

This should be a constant at least in the config directory, or users put it in the topology cpp files.

rvaccone:
| | |
|:---|:---|
|**_Related Issue(s)_**| Fixes #4425 |

---
## Change Description

Moves the PrmDb file name out of the FileHandling subtopology's `configComponents` phase, where it was hardcoded as `"PrmDb.dat"`, into a new `FileHandlingConfig.Paths.prmDbFile` constant (default unchanged), following the `Paths` pattern from `DataProductsConfig`. Updates the two related doc pages.

This exposed a second issue: `FileHandlingConfig` was registered `INTERFACE` in CMake, so the FPP-generated `FppConstantsAc.cpp` — which defines the string constant out of line — was never compiled and Ref failed to link. Dropped `INTERFACE`, matching the other subtopology config modules that carry string constants (DataProducts, ComFprime, ComCcsds, DpCompression).

## Rationale

Resolves #4425: the file name is project configuration and belongs in config. The default is unchanged, so existing deployments keep writing `PrmDb.dat`. Note for projects overriding `FileHandlingConfig.fpp` via `CONFIGURATION_OVERRIDES`: the new constant must be added to the override copy, as with any constant added to a config module.

- Built and ran the Ref deployment (instantiates this subtopology): generated topology code passes `FileHandlingConfig::Paths::prmDbFile` to `prmDb.configure()`, with `"PrmDb.dat"` in the binary as before.
- The pre-fix `INTERFACE` registration reproduces the undefined-reference link failure; dropping it fixes the link.
- Override check: setting the constant to a different name and rebuilding shows the new name in the binary; reverting restores the default.

## Future Work

None.

Claude (AI agent) drafted the change and ran the Ref build and override checks, under human direction and review.

IAMAI

### DIFF
--- a/Svc/Subtopologies/FileHandling/FileHandling.fpp
+++ b/Svc/Subtopologies/FileHandling/FileHandling.fpp
@@ -33,7 +33,7 @@ module FileHandling {
         priority FileHandlingConfig.Priorities.prmDb \
     {
         phase Fpp.ToCpp.Phases.configComponents """
-            FileHandling::prmDb.configure("PrmDb.dat");
+            FileHandling::prmDb.configure(FileHandlingConfig::Paths::prmDbFile);
         """
         phase Fpp.ToCpp.Phases.readParameters """
             FileHandling::prmDb.readParamFile();
--- a/Svc/Subtopologies/FileHandling/FileHandlingConfig/CMakeLists.txt
+++ b/Svc/Subtopologies/FileHandling/FileHandlingConfig/CMakeLists.txt
@@ -2,5 +2,4 @@ register_fprime_config(
     EXCLUDE_FROM_ALL
     AUTOCODER_INPUTS
         "${CMAKE_CURRENT_LIST_DIR}/FileHandlingConfig.fpp"
-    INTERFACE
 )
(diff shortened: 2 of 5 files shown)


LeStarch:
This isn't following the recommendation in the issue.  The recommendation is to simply remove that line and require it be called from within the using topology.

rvaccone:
Totally understand. I read the move to config as the config module, not the project configuration. I'll submit the removal version instead.

PR #5505, opened 2026-07-25
Add SDD for Drv::LinuxGpioDriver

### LINKED ISSUE #5346: GPIO driver does not have an SDD
| | |
|:---|:---|
|**_F´ Version_**| devel |
|**_Affected Component_**|  docs |
---
## Problem Description

There is no system design document for the GPIO driver.

## Context / Environment

https://fprime.jpl.nasa.gov/devel/docs/reference/

## How to Reproduce

1. Go to https://fprime.jpl.nasa.gov/devel/docs/reference/
2. Click Software Design Documents -> Drv
3. There is no GPIO driver tab

## Expected Behavior

There should be a GPIO driver tab.

Moferanoluwa:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5346 |
|**_Has Unit Tests (y/n)_**| n (documentation only) |

---
## Change Description

Adds `Drv/LinuxGpioDriver/docs/sdd.md`, the software design document for `Drv::LinuxGpioDriver`. It covers the `Drv.Gpio` interface implementation and its ports, the five pin configurations, the `open()` sequence against the Linux GPIO character device, the interrupt polling thread and its `start()`/`stop()`/`join()` lifecycle, the stubbed-driver build, and the component's events and status codes.

Also updates `docs/reference/system-functional/hardware-drivers.md` to link the new SDD rather than the bare component directory, matching how the drivers that have SDDs are listed there.

No `.nav.yml` change is needed: the Drv section is built from the `"Drv/**/docs/sdd.md"` glob, so adding the file is sufficient for the tab to appear.

## Rationale

Fixes #5346 — `Drv::LinuxGpioDriver` had no SDD, so it had no tab under Software Design Documents -> Drv.

Two behaviors seemed worth stating explicitly, as neither is apparent from the component interface alone:

- `gpioRead` and `gpioWrite` return `INVALID_MODE` unless the pin is configured as `GPIO_INPUT` or `GPIO_OUTPUT` respectively, so the interrupt configurations do not support reads.
- `start()` returns `INVALID_MODE` and starts no thread outside the three interrupt configurations.

Documentation only; no source is modified.

The useful review is checking the SDD's behavioral claims against `LinuxGpioDriver.cpp`, `LinuxGpioDriverCommon.cpp`, and `LinuxGpioDriverStub.cpp`. I verified that the three relative links resolve from the document's location, and that the `CycleIn` in the topology snippet matches the `Svc.Cycle` input port on `Svc::ActiveRateGroup`.

## Future Work

While reading `open()` for this write-up, the line-existence check at `LinuxGpioDriver.cpp:213` appears to log `OpenPinError` and then return the preceding status, which is `OP_OK` when the chip queries succeeded. I have kept it out of this PR since fixing it is a behavior change, and the SDD does not describe a returned status for that case. Happy to open a separate issue if that is in fact a bug.

`Drv::LinuxI2cDriver` and `Drv::LinuxSpiDriver` are also missing SDDs. I kept those out of scope here and can follow up separately.

- **Type**: Source comprehension and drafting the SDD prose
- **Scope**: All of `Drv/LinuxGpioDriver/docs/sdd.md`, plus the one-line link change in `hardware-drivers.md`. No framework source was modified.
- **Level of modification**: Drafted from the component sources, then reviewed and corrected before submission. Corrections made during that review: replaced a `gpioSet` port name that came from the external LED Blinker tutorial rather than this repository; changed the thread parameter types to `FwSizeType` to match `LinuxGpioDriver.hpp`, rather than the `Os::Task::ParamType` used in the LinuxUartDriver SDD; and verified the relative links and the `CycleIn` port name against the tree.

The document follows the section layout of `Drv/LinuxUartDriver/docs/sdd.md` for consistency.

### DIFF
--- a/Drv/LinuxGpioDriver/docs/sdd.md
+++ b/Drv/LinuxGpioDriver/docs/sdd.md
@@ -65,7 +65,7 @@ A line is configured once through the `open()` method, which selects one of five
 4. Requests a line handle (`GPIO_GET_LINEHANDLE_IOCTL`) for the input and output modes, or a line event (`GPIO_GET_LINEEVENT_IOCTL`) for the interrupt modes
 5. Retains the resulting file descriptor and configuration on success
 
-The component name is passed to the kernel as the consumer label for the line, so the owning component is identifiable in tools such as `gpioinfo`.
+The component name is passed to the kernel as the consumer label for the line (when object names are enabled via `FW_OBJECT_NAMES`), so the owning component is identifiable in tools such as `gpioinfo`.
 
 ### 3.4 Read and Write Operations
 
@@ -86,15 +86,15 @@ The polling loop performs the following:
 2. Reads the kernel event record when the descriptor becomes ready
 3. Captures a timestamp with `Os::RawTime`
 4. Invokes the `gpioInterrupt` output port with that timestamp
-5. Reports short reads, polling failures, and timestamp failures via events
+5. Reports short reads, polling failures, and timestamp failures via events; on a timestamp failure the interrupt is still emitted with the (possibly invalid) timestamp
 
-Because `gpioInterrupt` is invoked directly from the polling thread, the receiving component is responsible for any thread-safety and execution-time constraints on that path. Connecting the port to an active component, such as `Svc::ActiveRateGroup`, keeps work off the polling thread.
+Because `gpioInterrupt` is invoked directly from the polling thread, the receiving component is responsible for any thread-safety and execution-time constraints on that path. Connecting the port to an `async` input port, such as `CycleIn` on `Svc::ActiveRateGroup`, keeps work off the polling thread.
 
 ### 3.6 Threading Model
 
 The interrupt thread is controlled by three methods:
 
-- `start()`: sets the running flag and starts the polling task
+- `start()`: sets the running flag and starts the polling task; returns `Drv::GpioStatus::UNKNOWN_ERROR` if the task fails to start
 - `stop()`: clears the running flag, requesting shutdown
 - `join()`: blocks until the polling task exits
 
@@ -198,7 +198,7 @@ Port handlers return `Drv::GpioStatus`:
 | INVALID_MODE | Operation not permitted with the current configuration |
 | UNKNOWN_ERROR | An unknown error occurred |
 
-`open()` returns `Os::File::Status`, translated from the `errno` reported by the underlying system call.
+`open()` returns `Os::File::Status`, translated from the `errno` reported by the underlying system call. Note: in the current implementation, a request for a line number beyond the chip's line count logs `OpenPinError` but returns the preceding status (`OP_OK`); callers should not rely on `open()` returning an error in that case.
 
 ### 5.4 Events
 
--- a/Drv/LinuxGpioDriver/docs/sdd.md
+++ b/Drv/LinuxGpioDriver/docs/sdd.md
@@ -1,0 +1,214 @@
+# Drv::LinuxGpioDriver
+
+## 1. Introduction
+
+The LinuxGpioDriver component provides a Linux-specific implementation of a General Purpose Input/Output (GPIO) pin driver. It implements the [`Drv.Gpio`](../../Interfaces/Gpio.fpp) interface to enable reading, writing, and interrupt detection on a single GPIO line.
+
+The component wraps the Linux GPIO character device ABI (`/dev/gpiochip*`), configuring and accessing a line through `ioctl` requests rather than the deprecated sysfs interface. Each component instance manages exactly one GPIO line, and that line is fixed to a single mode chosen at configuration time.
+
+For more information on the driver interfaces see: [`Drv FPP Interfaces`](../../Interfaces/docs/sdd.md).
+
+## 2. Requirements
+
+| Name | Description | Validation |
+|---|---|---|
+| LINUX-GPIO-COMP-001 | The LinuxGpioDriver component shall implement the Drv.Gpio interface | inspection |
+| LINUX-GPIO-COMP-002 | The LinuxGpioDriver component shall configure a GPIO line as input, output, or interrupt using the Linux GPIO character device | inspection |
+| LINUX-GPIO-COMP-003 | The LinuxGpioDriver component shall set a caller-supplied default state when a line is configured as output | inspection |
+| LINUX-GPIO-COMP-004 | The LinuxGpioDriver component shall reject read and write requests that do not match the configured mode | inspection |
+| LINUX-GPIO-COMP-005 | The LinuxGpioDriver component shall detect line transitions on a rising edge, a falling edge, or both edges as configured | inspection |
+| LINUX-GPIO-COMP-006 | The LinuxGpioDriver component shall provide a dedicated thread for interrupt detection | inspection |
+| LINUX-GPIO-COMP-007 | The LinuxGpioDriver component shall emit a timestamped interrupt on the gpioInterrupt output port when a configured transition occurs | inspection |
+| LINUX-GPIO-COMP-008 | The LinuxGpioDriver component shall report configuration and runtime errors via events | inspection |
+
+## 3. Design
+
+LinuxGpioDriver is a passive component implementing the design specified by the [`Drv.Gpio`](../../Interfaces/Gpio.fpp) interface.
+
+### 3.1 Port Descriptions
+
+| Name | Type | Kind | Description |
+|---|---|---|---|
+| gpioRead | [`Drv.GpioRead`](../../Ports/GpioDriverPorts.fpp) | sync input | Reads the current logic state of the line |
+| gpioWrite | [`Drv.GpioWrite`](../../Ports/GpioDriverPorts.fpp) | sync input | Sets the logic state of the line |
+| gpioInterrupt | `Svc.Cycle` | output | Emits a timestamp when a configured line transition is detected |
+
+The component declares no commands, telemetry channels, or parameters. Since the component is passive, `gpioRead` and `gpioWrite` execute on the caller's thread.
+
+### 3.2 Architecture
+
+The component consists of the following key elements:
+
+- **Chip and Line Configuration**: Opens the `/dev/gpiochip*` device, validates the requested line, and requests either a line handle or a line event from the kernel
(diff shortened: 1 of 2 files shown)


Drv/LinuxGpioDriver/docs/sdd.md
```diff
@@ -0,0 +1,214 @@
...
+
+The polling loop performs the following:
+
+1. Polls the line event descriptor for readable data, using a fixed `GPIO_POLL_TIMEOUT` of 500 milliseconds
+2. Reads the kernel event record when the descriptor becomes ready
+3. Captures a timestamp with `Os::RawTime`
+4. Invokes the `gpioInterrupt` output port with that timestamp
+5. Reports short reads, polling failures, and timestamp failures via events
+
+Because `gpioInterrupt` is invoked directly from the polling thread, the receiving component is responsible for any thread-safety and execution-time constraints on that path. Connecting the port to an active component, such as `Svc::ActiveRateGroup`, keeps work off the polling thread.
```
lestarch-autobot: [Architecture] **suggestion** “Active component” alone does not keep work off the polling thread — the input port kind does.

`gpioInterrupt` runs handlers in the polling thread whenever the receiving input port is `sync` or `guarded`, even on an active component. The decoupling here comes from `Svc::ActiveRateGroup`'s `CycleIn` being an `async` input (see `docs/user-manual/framework/component-and-port-selection.md`). Stating the port-kind condition avoids readers wiring a sync input on an active component and assuming thread isolation.

```suggestion
Because `gpioInterrupt` is invoked directly from the polling thread, the receiving component is responsible for any thread-safety and execution-time constraints on that path. Connecting the port to an `async` input port, such as `CycleIn` on `Svc::ActiveRateGroup`, keeps work off the polling thread.
```

docs/reference/system-functional/hardware-drivers.md
```diff
@@ -12,7 +12,7 @@
 - [F Prime Udp SDD](https://github.com/nasa/fprime/blob/devel/Drv/Udp/docs/sdd.md)
 - [F Prime Ip SDD](https://github.com/nasa/fprime/blob/devel/Drv/Ip/docs/sdd.md)
 - [F Prime LinuxUartDriver SDD](https://github.com/nasa/fprime/blob/devel/Drv/LinuxUartDriver/docs/sdd.md)
-- [F Prime LinuxGpioDriver](https://github.com/nasa/fprime/blob/devel/Drv/LinuxGpioDriver)
+- [F Prime LinuxGpioDriver SDD](https://github.com/nasa/fprime/blob/devel/Drv/LinuxGpioDriver/docs/sdd.md)
 - [F Prime LinuxI2cDriver](https://github.com/nasa/fprime/blob/devel/Drv/LinuxI2cDriver)
```
lestarch-autobot, line 16: [Documentation] **future work** `Drv::LinuxI2cDriver` and `Drv::LinuxSpiDriver` still lack SDDs and are listed here as bare directory links.

Preexisting gap the PR author explicitly deferred; recording it so the follow-up (per the PR description's Future Work section) isn't lost.

Drv/LinuxGpioDriver/docs/sdd.md
```diff
@@ -0,0 +1,214 @@
...
+
+### 3.6 Threading Model
+
+The interrupt thread is controlled by three methods:
+
+- `start()`: sets the running flag and starts the polling task
```
lestarch-autobot: [Documentation] **suggestion** `start()` can also return `UNKNOWN_ERROR` in a valid interrupt mode when the underlying task fails to start (`LinuxGpioDriverCommon.cpp:40-43`); the SDD only documents the `INVALID_MODE` failure path.

```suggestion
- `start()`: sets the running flag and starts the polling task; returns `Drv::GpioStatus::UNKNOWN_ERROR` if the task fails to start
```

Drv/LinuxGpioDriver/docs/sdd.md
```diff
@@ -0,0 +1,214 @@
...
+
+`open()` performs the following steps:
+
+1. Opens the GPIO chip device and retrieves chip information
+2. Verifies that the requested line number exists on that chip
+3. Retrieves line information, including the name and any current consumer
+4. Requests a line handle (`GPIO_GET_LINEHANDLE_IOCTL`) for the input and output modes, or a line event (`GPIO_GET_LINEEVENT_IOCTL`) for the interrupt modes
+5. Retains the resulting file descriptor and configuration on success
+
+The component name is passed to the kernel as the consumer label for the line, so the owning component is identifiable in tools such as `gpioinfo`.
```
lestarch-autobot: [Documentation] **could fix** The consumer label is the component name only when object names are compiled in: the code uses `FW_OPTIONAL_NAME(this->getObjName())` (`LinuxGpioDriver.cpp:145`), which substitutes an empty string when `FW_OBJECT_NAMES` is disabled.

```suggestion
The component name is passed to the kernel as the consumer label for the line (when object names are enabled via `FW_OBJECT_NAMES`), so the owning component is identifiable in tools such as `gpioinfo`.
```

Drv/LinuxGpioDriver/docs/sdd.md
```diff
@@ -0,0 +1,214 @@
...
+
+Interrupt detection runs on a dedicated thread started by `start()`. The thread is only available in the three interrupt configurations; calling `start()` in any other mode returns `Drv::GpioStatus::INVALID_MODE` and starts no thread.
+
+The polling loop performs the following:
+
+1. Polls the line event descriptor for readable data, using a fixed `GPIO_POLL_TIMEOUT` of 500 milliseconds
+2. Reads the kernel event record when the descriptor becomes ready
+3. Captures a timestamp with `Os::RawTime`
+4. Invokes the `gpioInterrupt` output port with that timestamp
+5. Reports short reads, polling failures, and timestamp failures via events
```
lestarch-autobot: [Documentation] **could fix** On a timestamp failure the code logs `InterruptTimeError` but still invokes `gpioInterrupt` with the unpopulated timestamp (`LinuxGpioDriver.cpp:318-324`); the SDD reads as if the failure suppresses the interrupt.

```suggestion
5. Reports short reads, polling failures, and timestamp failures via events; on a timestamp failure the interrupt is still emitted with the unpopulated timestamp
```

Drv/LinuxGpioDriver/docs/sdd.md
```diff
@@ -0,0 +1,214 @@
...
+Port handlers return `Drv::GpioStatus`:
+
+| Status | Meaning |
+|---|---|
+| OP_OK | Operation succeeded |
+| NOT_OPENED | Pin was never opened |
+| INVALID_MODE | Operation not permitted with the current configuration |
+| UNKNOWN_ERROR | An unknown error occurred |
+
+`open()` returns `Os::File::Status`, translated from the `errno` reported by the underlying system call.
```
lestarch-autobot: [Design] **suggestion** SDD states `open()` returns an errno-derived error status, but the line-existence failure path returns `OP_OK`.

In `LinuxGpioDriver.cpp` (`open()`), when `gpio >= chip_info.lines` the code logs `OpenPinError` and returns the preceding status, which is `OP_OK` — no failing syscall sets `errno`. As written, the SDD would lead a reader to expect an error return in that case. Since the code bug is deliberately out of scope (per the PR description), the SDD should state the actual behavior.

```suggestion
`open()` returns `Os::File::Status`, translated from the `errno` reported by the underlying system call. Note: in the current implementation, a request for a line number beyond the chip's line count logs `OpenPinError` but returns the preceding status (`OP_OK`); callers should not rely on `open()` returning an error in that case.
```

Drv/LinuxGpioDriver/docs/sdd.md
lestarch-autobot: [Design] **suggestion** The step list can be read as a timestamp failure suppressing the interrupt; the code still invokes `gpioInterrupt` after logging `InterruptTimeError`.

In `pollLoop()` (`LinuxGpioDriver.cpp`), `gpioInterrupt_out` is called unconditionally after the timestamp read, so a failed `Os::RawTime::now()` still emits the interrupt with the (possibly invalid) timestamp. Worth stating so consumers know the event stream is not gated on timestamp validity.

```suggestion
5. Reports short reads, polling failures, and timestamp failures via events; on a timestamp failure the interrupt is still emitted with the (possibly invalid) timestamp
```

lestarch-autobot (approved):

### Per-agent results

| Documentation Currency | 0 | 1 | 2 | 1 | 4 | Go |
| Design | 0 | 2 | 0 | 0 | 2 | Go |
| Architecture | 0 | 1 | 0 | 0 | 1 | Go |
| **Totals** | 0 | 4 | 2 | 1 | 7 | **Go** |

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

### Merge readiness
**Merge readiness: Go** — all seven reviewers completed with zero outstanding must-fix findings.

- None — all registered reviewers ran.

---

A clean docs run — this SDD is cleared for docking. Nice work.

Moferanoluwa:
Thanks for the review — all five actionable findings applied in 1ba0a39.

The port-kind one was a genuine misunderstanding on my part rather than just loose wording: I had attributed the decoupling to the component being active, when it comes from `CycleIn` being an `async` input. A `sync` or `guarded` input on an active component would still run in the polling thread, so the original text could have led someone to wire exactly that and assume isolation. Reworded to state the port-kind condition.

The other four:

- `start()` now documents the `UNKNOWN_ERROR` path when the polling task fails to start, not just `INVALID_MODE`.
- The consumer label is qualified with `FW_OBJECT_NAMES`, since `FW_OPTIONAL_NAME` substitutes an empty string when object names are compiled out.
- Both timestamp comments folded into one line: the interrupt is still emitted with the unpopulated timestamp, so consumers know the event stream is not gated on timestamp validity.
- `open()` now states that a line number beyond the chip's line count logs `OpenPinError` and returns `OP_OK`, and that callers should not rely on an error return there. This is the behavior I flagged under Future Work; documenting the actual behavior rather than the intended one seems right while the code fix is out of scope.

On the `Drv::LinuxI2cDriver` / `Drv::LinuxSpiDriver` future-work note: still happy to pick those up as separate issue-backed PRs once this one settles.

LeStarch (approved):

PR #5506, opened 2026-07-25
Add other references and CI badges to the supported platforms page

### LINKED ISSUE #5428: Add all references and CI badges to platform page
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

https://fprime.jpl.nasa.gov/platforms lists all supported platforms and their reference repos, which is really neat.

We have many other references, which while they aren't necessarily tied to a specific platform, are extremely useful.

Example:
- all ext-*.yml in our CI: https://github.com/nasa/fprime/tree/devel/.github/workflows
- all *reference* repos under https://github.com/fprime-community

These references should be added to a new "other reference" table in the platforms document.

Moferanoluwa:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5428 |
|**_Has Unit Tests (y/n)_**| n (documentation only) |

---
## Change Description

Adds an "Other References" section to `docs/user-manual/framework/supported-platforms.md` with two tables:

- **Other Reference Projects** — the 11 reference repositories under `fprime-community` that are not already covered by the supported or targeted platform tables.
- **External Repository CI** — all 17 `ext-*.yml` workflows, each with the project it exercises, its trigger, and its build-status badge.

The CI table includes a Trigger column because several soak workflows run on a schedule, on release, or on manual dispatch only. Without that context their badges read as stale or broken.

## Rationale

Fixes #5428. The page listed only reference projects tied to a specific hardware and OS pair, so the platform-independent references were not linked anywhere on it, and the external CI that keeps those repositories current was not surfaced.

Documentation only; no source is modified. I verified mechanically rather than by eye:

- All 17 `ext-*.yml` files on disk appear in the table, with no extra or missing entries. Workflow names, target repositories, and triggers were extracted from the workflow files rather than inferred from filenames.
- All 18 linked repositories resolve via the GitHub API and are public and non-archived.
- Table column counts are consistent, there are no duplicate rows, the `CONTRIBUTING.md` anchor resolves, and `git diff --check` is clean.

Three judgment calls worth a maintainer's opinion:

1. I excluded the two archived repositories, `fprime-system-reference` and `fprime-scales-reference-archive`. Happy to add them with an "archived" note if you would rather the list be exhaustive.
2. I excluded reference repos already named in the platform tables to avoid duplicating them.
3. For `ext-cookiecutters-test.yml` I listed both `fprime-bootstrap` and `fprime-tools`, since the workflow exercises templates from both.

## Future Work

The Supported Platforms table lists PyCubed / Zephyr against `fprime-zephyr-reference`, but a dedicated `fprime-pycubed-zephyr-reference` repository also exists. I did not change that row, as I could not tell from the repositories alone which is intended. Glad to follow up if it should point at the PyCubed-specific reference.

- **Type**: Gathering the workflow and repository data, and drafting the section
- **Scope**: The added section in `docs/user-manual/framework/supported-platforms.md`. No source is modified.
- **Level of modification**: The CI table rows were generated from the `ext-*.yml` files themselves rather than typed by hand, to avoid transcription errors in the badge URLs, then reviewed against the workflow files. Corrections made during that review: a first pass missed `ext-cookiecutters-test.yml` because that file indents its `on:` block by one space, and the `fprime-tools` entry for that same row was added after re-reading what the workflow actually tests. Descriptions were rewritten rather than copied from the repository blurbs.

### DIFF
--- a/docs/user-manual/framework/supported-platforms.md
+++ b/docs/user-manual/framework/supported-platforms.md
@@ -33,6 +33,50 @@ These are platforms that the F Prime community or core team is actively working 
 | Vorago | Baremetal | ARM | In development | |TBD|
 | VOXL2 | Linux | ARM64 | In development | | TBD |
 
+## Other References
+
+The tables above list reference projects tied to a specific hardware and operating system pair. F Prime maintains further reference projects, tutorials, and examples that are not specific to a single platform.
+
+### Other Reference Projects
+
+| Reference Project | Description |
+| ----------------- | ----------- |
+| [`fprime_cfs_reference`](https://github.com/fprime-community/fprime_cfs_reference) | F Prime applications running under the Core Flight System (cFS) |
+| [`fprime-amsat-main-board-reference`](https://github.com/fprime-community/fprime-amsat-main-board-reference) | Port to the Pico main sensor board for the AMSAT CubeSat Simulator |
+| [`fprime-amsat-reference`](https://github.com/fprime-community/fprime-amsat-reference) | Port to the C&DH Pi Zero 2 board for the AMSAT CubeSat Simulator |
+| [`fprime-generic-hub-reference`](https://github.com/fprime-community/fprime-generic-hub-reference) | Applications built around the Hub pattern |
+| [`fprime-phased-deployment-reference`](https://github.com/fprime-community/fprime-phased-deployment-reference) | Deployment run entirely through phases |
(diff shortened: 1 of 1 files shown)


LeStarch:
As stated in the last PR that made this same change, our goal would be to move references to another page (e.g. "References") and link to that from supported platforms.

Issue #5507, opened 2026-07-25
Add unit tests for Fw::CmdPacket deserialization


Moferanoluwa:
| | |
|:---|:---|
|**_F´ Version_**| devel (35d4c0e881) |
|**_Affected Component_**| Fw/Cmd |
---
## Feature Description

Add unit tests for `Fw::CmdPacket`, which currently has none. `Fw/Cmd` registers `CmdPacket.cpp` via `register_fprime_module` but has no `register_fprime_ut` block, so `CmdPacket::deserializeFrom` is not exercised by any test in the repository.

Proposed coverage for `deserializeFrom`, all of which is reachable without hardware:

- A well-formed command packet round-trips to the expected opcode and argument bytes.
- A packet whose descriptor is not `FW_PACKET_COMMAND` returns `FW_DESERIALIZE_TYPE_MISMATCH`.
- A buffer too short for the packet descriptor propagates the failure from `ComPacket::deserializeBase`.
- A buffer that ends after the descriptor, before the opcode, propagates the opcode deserialization failure.
- A command with no arguments succeeds and yields an empty argument buffer, exercising the `getDeserializeSizeLeft()` branch that is skipped in that case.
- Arguments that exactly fill `FW_CMD_ARG_BUFFER_MAX_SIZE`, and arguments that exceed it, to pin the `copyRaw` boundary.

If this is in scope I would also like to check whether `serializeTo` should be covered. It currently asserts unconditionally with the comment "no use case for serializing CmdPackets in FSW (currently)", so a death test would pin that contract, but I did not want to assume that is desired.

## Rationale

`CmdPacket::deserializeFrom` is on the path of every command uplinked to a deployment, and its failure modes are the ones that matter for malformed or truncated uplink. It is currently untested.

This is also the module from #4896, where `CmdArgBuffer`'s serialized size was computed with a hardcoded `I32` instead of `FwSizeStoreType`. That defect was benign in practice and has since been fixed by the move to `LinearBufferTemplate`, but it went unnoticed partly because nothing in the test suite exercises this area. Tests here would pin the argument-buffer boundary behavior against future changes.

Scope note: I looked at what else lacks a `register_fprime_ut` block, and most of the remaining modules are port or interface definitions with no compiled logic to test. `Fw/Cmd` is one of the few with real behavior and no coverage, which is why I am proposing it on its own rather than as a sweep.

I am happy to implement this if the CCB considers it in scope, and to adjust the case list based on what maintainers want pinned.

---
[...]

PR #5517, opened 2026-07-28
fix(Fw/Cmd): clear stale command arguments, with deserialization tests

tekinertekin:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5507 |

---
## Change Description

Fw/Cmd was without any unit tests and thus CmdPacket::deserializeFrom was not covered. Add Fw/Cmd/test/ut/CmdPacketTest.cpp and register it in Fw/Cmd/CMakeLists.txt.

## Rationale

Refs #5507.

Coverage:
- well-formed packet deserializes into the correct opcode and args
- non-command descriptor causes FW_DESERIALIZE_TYPE_MISMATCH error
- buffer shorter than descriptor is passed back via deserializeBase
- buffer cut short of the opcode fails deserialization of the opcode
- no-args command results in empty args buffer (copyRaw path not taken)
- args equal to FW_CMD_ARG_BUFFER_MAX_SIZE pass
- args one byte beyond the max size fail deserialization at the copyRaw step

## Future Work

None.

I used an AI tool to help write this change: the unit test and its CMake registration were AI-generated. I built and ran the Fw/Cmd unit test suite locally and all cases pass. I have reviewed, understand, and verify this change and am responsible for it.

### DIFF
--- a/Fw/Cmd/test/ut/CmdPacketTest.cpp
+++ b/Fw/Cmd/test/ut/CmdPacketTest.cpp
@@ -1,0 +1,123 @@
+// ======================================================================
+// @file   CmdPacketTest.cpp
+// @brief  Unit tests for Fw::CmdPacket::deserializeFrom
+//
+// CmdPacket has no serializeTo path in FSW (it asserts), so each test builds
+// the wire bytes by hand -- [packet descriptor][opcode][args] -- and exercises
+// deserializeFrom. All cases are reachable without hardware.
+// ======================================================================
+
+#include <gtest/gtest.h>
+#include <cstring>
+
+#include <Fw/Cmd/CmdPacket.hpp>
+#include <Fw/Com/ComBuffer.hpp>
+#include <Fw/Types/Serializable.hpp>
+
+namespace {
+
+// Serialize a packet descriptor into an already-reset buffer.
+Fw::SerializeStatus putDescriptor(Fw::SerialBufferBase& buff, Fw::ComPacketType type) {
+    return buff.serializeFrom(static_cast<FwPacketDescriptorType>(type));
+}
+
+}  // namespace
+
+// 1. A well-formed command packet round-trips to the expected opcode and args.
+TEST(FwCmdPacketTest, DeserializeWellFormed) {
+    const FwOpcodeType opcode = 0x1234;
+    const U8 args[] = {0xDE, 0xAD, 0xBE, 0xEF};
+
+    Fw::ComBuffer buff;
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, putDescriptor(buff, Fw::ComPacketType::FW_PACKET_COMMAND));
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, buff.serializeFrom(opcode));
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, buff.serializeFrom(args, static_cast<FwSizeType>(sizeof(args)), Fw::Serialization::OMIT_LENGTH));
+
+    Fw::CmdPacket pkt;
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, pkt.deserializeFrom(buff));
+    ASSERT_EQ(opcode, pkt.getOpCode());
+    ASSERT_EQ(static_cast<FwSizeType>(sizeof(args)), pkt.getArgBuffer().getSize());
+    ASSERT_EQ(0, memcmp(pkt.getArgBuffer().getBuffAddr(), args, sizeof(args)));
+}
+
+// 2. A packet whose descriptor is not FW_PACKET_COMMAND is rejected.
+TEST(FwCmdPacketTest, DeserializeWrongDescriptor) {
+    Fw::ComBuffer buff;
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, putDescriptor(buff, Fw::ComPacketType::FW_PACKET_TELEM));
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, buff.serializeFrom(static_cast<FwOpcodeType>(0x1234)));
+
+    Fw::CmdPacket pkt;
+    ASSERT_EQ(Fw::FW_DESERIALIZE_TYPE_MISMATCH, pkt.deserializeFrom(buff));
+}
+
+// 3. A buffer too short for even the descriptor propagates the failure from
+//    ComPacket::deserializeBase.
+TEST(FwCmdPacketTest, DeserializeEmptyBuffer) {
+    Fw::ComBuffer buff;  // nothing serialized
+
+    Fw::CmdPacket pkt;
+    ASSERT_EQ(Fw::FW_DESERIALIZE_BUFFER_EMPTY, pkt.deserializeFrom(buff));
+}
+
+// 4. A buffer that ends after the descriptor, before the opcode, propagates the
+//    opcode deserialization failure.
+TEST(FwCmdPacketTest, DeserializeTruncatedBeforeOpcode) {
+    Fw::ComBuffer buff;
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, putDescriptor(buff, Fw::ComPacketType::FW_PACKET_COMMAND));
+
+    Fw::CmdPacket pkt;
+    ASSERT_EQ(Fw::FW_DESERIALIZE_BUFFER_EMPTY, pkt.deserializeFrom(buff));
+}
+
+// 5. A command with no arguments succeeds and yields an empty argument buffer,
+//    exercising the getDeserializeSizeLeft() == 0 branch that skips copyRaw.
+TEST(FwCmdPacketTest, DeserializeNoArgs) {
+    const FwOpcodeType opcode = 0x42;
+
+    Fw::ComBuffer buff;
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, putDescriptor(buff, Fw::ComPacketType::FW_PACKET_COMMAND));
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, buff.serializeFrom(opcode));
+
+    Fw::CmdPacket pkt;
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, pkt.deserializeFrom(buff));
+    ASSERT_EQ(opcode, pkt.getOpCode());
+    ASSERT_EQ(static_cast<FwSizeType>(0), pkt.getArgBuffer().getSize());
+}
+
+// 6a. Arguments that exactly fill FW_CMD_ARG_BUFFER_MAX_SIZE succeed (a ComBuffer
+//     holds exactly descriptor + opcode + FW_CMD_ARG_BUFFER_MAX_SIZE bytes).
+TEST(FwCmdPacketTest, DeserializeArgsExactlyFill) {
+    U8 args[FW_CMD_ARG_BUFFER_MAX_SIZE];
+    memset(args, 0xAB, sizeof(args));
+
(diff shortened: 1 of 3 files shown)


tekinertekin:
Hi @Moferanoluwa, thanks for opening #5507 and laying out the coverage so clearly. I followed your outline for these tests. I opened this PR since the issue was unassigned, but if you were already working on it or would rather take it yourself, I'm happy to close this and step back, or to collaborate on it. Whatever works best for you and the maintainers.

Moferanoluwa:
Hi @tekinertekin — please keep it, and thanks for asking rather than just landing it.

To be straight about the timing: I hadn't started. I filed #5507 as a scope proposal and was waiting to hear whether the CCB considered it in scope before writing anything, so there's nothing of mine for this to collide with. You did the work.

I did build and run it rather than just read it, so this is a verified review rather than a drive-by:

```
Ubuntu 24.04, gcc 13.3, ninja, requirements.txt pinned venv
fprime-util purge && fprime-util generate --ut && cd Fw/Cmd && fprime-util check

[==========] 7 tests from 1 test suite ran. (16 ms total)
[  PASSED  ] 7 tests.
100% tests passed, 0 tests failed out of 1
```

Notes from reading it:

- The `register_fprime_ut` block matches the existing `Fw/Tlm` and `Fw/Log` pattern (SOURCES + DEPENDS, no `AUTOCODER_INPUTS`), so it is consistent with how the rest of `Fw` registers tests.
- Splitting the boundary case into `DeserializeArgsExactlyFill` and `DeserializeArgsExceedMax` is better than what I proposed. Using `ExternalSerializeBuffer` for the oversized case is a neat way around `ComBuffer` not being able to hold the input, and the comment explains why it is there.
- `DeserializeTruncatedBeforeOpcode` is the case I most wanted covered, since that is the one that would regress silently.

One open item from the issue you may want a maintainer's call on: `serializeTo` currently asserts unconditionally with the comment "no use case for serializing CmdPackets in FSW (currently)". I raised whether a death test should pin that contract but deliberately did not assume it was wanted. Your PR leaves it uncovered, which I think is the right default — worth flagging so the decision is explicit rather than an oversight.

Happy to keep reviewing if it is useful.

tekinertekin:
> Hi @tekinertekin — please keep it, and thanks for asking rather than just landing it.
> 
> To be straight about the timing: I hadn't started. I filed #5507 as a scope proposal and was waiting to hear whether the CCB considered it in scope before writing anything, so there's nothing of mine for this to collide with. You did the work.
> 
> I did build and run it rather than just read it, so this is a verified review rather than a drive-by:
> 
> ```
> Ubuntu 24.04, gcc 13.3, ninja, requirements.txt pinned venv
> fprime-util purge && fprime-util generate --ut && cd Fw/Cmd && fprime-util check
> 
> [==========] 7 tests from 1 test suite ran. (16 ms total)
> [  PASSED  ] 7 tests.
> 100% tests passed, 0 tests failed out of 1
> ```
> 
> Notes from reading it:
> 
> * The `register_fprime_ut` block matches the existing `Fw/Tlm` and `Fw/Log` pattern (SOURCES + DEPENDS, no `AUTOCODER_INPUTS`), so it is consistent with how the rest of `Fw` registers tests.
> * Splitting the boundary case into `DeserializeArgsExactlyFill` and `DeserializeArgsExceedMax` is better than what I proposed. Using `ExternalSerializeBuffer` for the oversized case is a neat way around `ComBuffer` not being able to hold the input, and the comment explains why it is there.
> * `DeserializeTruncatedBeforeOpcode` is the case I most wanted covered, since that is the one that would regress silently.
> 
> One open item from the issue you may want a maintainer's call on: `serializeTo` currently asserts unconditionally with the comment "no use case for serializing CmdPackets in FSW (currently)". I raised whether a death test should pin that contract but deliberately did not assume it was wanted. Your PR leaves it uncovered, which I think is the right default — worth flagging so the decision is explicit rather than an oversight.
> 
> Happy to keep reviewing if it is useful.

Thank you @Moferanoluwa, it is great to see that you not only read but also built and executed it, and the detailed notes are much appreciated. Good to hear there are no collisions.

On serializeTo(): I agree that leaving it uncovered is the right default. I did so intentionally, since it is an unconditional FW_ASSERT(false) with the "no use case in FSW" comment, so a death test would really only prove "it asserts". That feels more like a CCB/maintainer call.

Fw/Cmd/test/ut/CmdPacketTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, buff.serializeFrom(static_cast<FwOpcodeType>(0x1234)));
+
+    Fw::CmdPacket pkt;
+    ASSERT_EQ(Fw::FW_DESERIALIZE_TYPE_MISMATCH, pkt.deserializeFrom(buff));
+}
+
+// 3. A buffer too short for even the descriptor propagates the failure from
+//    ComPacket::deserializeBase.
+TEST(FwCmdPacketTest, DeserializeEmptyBuffer) {
+    Fw::ComBuffer buff;  // nothing serialized
```
lntutor, line 57: The claimed truncated-field coverage currently exercises only a completely empty read. With the default configuration `FwPacketDescriptorType` is `U16`, so a one-byte descriptor takes the distinct `FW_DESERIALIZE_SIZE_MISMATCH` branch in `LinearBufferBase::deserializeTo`; similarly, `DeserializeTruncatedBeforeOpcode` leaves zero opcode bytes and never covers a partially present `FwOpcodeType` (`U32` by default). Please add at least one partial descriptor and partial opcode case and assert `FW_DESERIALIZE_SIZE_MISMATCH`, otherwise the malformed-uplink path highlighted in #5507 remains unpinned.

Fw/Cmd/test/ut/CmdPacketTest.cpp
```diff
@@ -0,0 +1,123 @@
...
+    const FwOpcodeType opcode = 0x42;
+
+    Fw::ComBuffer buff;
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, putDescriptor(buff, Fw::ComPacketType::FW_PACKET_COMMAND));
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, buff.serializeFrom(opcode));
+
+    Fw::CmdPacket pkt;
+    ASSERT_EQ(Fw::FW_SERIALIZE_OK, pkt.deserializeFrom(buff));
+    ASSERT_EQ(opcode, pkt.getOpCode());
+    ASSERT_EQ(static_cast<FwSizeType>(0), pkt.getArgBuffer().getSize());
```
lntutor, line 123: This only proves an empty argument buffer on a freshly constructed packet. `CmdPacket::deserializeFrom` skips `copyRaw` when no bytes remain and never resets `m_argBuffer`, so reusing one `CmdPacket` to parse an argument-bearing command and then this no-args command returns `FW_SERIALIZE_OK` while retaining the old argument bytes. A reuse test here would expose that stale-state bug; either clear the argument buffer during deserialization or explicitly establish and test a non-reusable object contract.

lntutor (commented):
The proposed suite covers the main success, type, and capacity boundaries, but two advertised contracts are still masked by fresh/empty test fixtures: partial-field truncation and replacement of prior argument state.

Fw/Cmd/test/ut/CmdPacketTest.cpp
tekinertekin: You're right — both existing truncation cases land on the BUFFER_EMPTY branch, so
SIZE_MISMATCH was never reached. Added in 6db9d98:

- `DeserializePartialDescriptor`: one byte of the U16 descriptor
- `DeserializePartialOpcode`: complete descriptor, then `sizeof(FwOpcodeType) - 1` bytes

Both assert `FW_DESERIALIZE_SIZE_MISMATCH`. I sized the partial arrays from
`sizeof(...)` with static_asserts so the cases stay meaningful if those config
types ever change width. Suite is 9/9 locally via `fprime-util check` in Fw/Cmd.

Fw/Cmd/test/ut/CmdPacketTest.cpp
tekinertekin: Good catch — and it is a code bug rather than only a test gap. I checked with a
throwaway reuse test: parse an argument-bearing command, then a no-args command
through the same `CmdPacket`. `deserializeFrom` returns `FW_SERIALIZE_OK` and
`getOpCode()` is correctly the new opcode, but `getArgBuffer().getSize()` is
still 4 — the previous command's argument bytes.

Since that is a behaviour change rather than test coverage, I would rather not
fold it into this test-only PR. I have the reproducer ready and can put the fix in
a follow-up PR, but the resolution is a design call I do not want to assume:
clear `m_argBuffer` during deserialization, or state that a `CmdPacket` is not
reusable and pin that contract instead. @LeStarch which would you prefer?

lntutor: Rechecked on current head `6db9d98c5f5633b7f6bef57d0f332b43e2919105`. The new partial-descriptor/opcode cases address the first review finding, but this reuse case still reproduces deterministically.

I added a one-off test that deserializes `{0xAA, 0x55}` into a `CmdPacket`, then deserializes a valid no-argument command into the same object. Both calls return `FW_SERIALIZE_OK`, but after the second call `getArgBuffer().getSize()` is still 2 rather than 0. Focused `Fw/Cmd` result: the nine PR tests pass and this test fails.

The root cause is the branch in `CmdPacket::deserializeFrom`: `copyRaw` replaces `m_argBuffer` only when source bytes remain, while the empty case never resets its serialization length. Please either clear `m_argBuffer` before the conditional copy and keep the reuse test, or make a non-reusable `CmdPacket` contract explicit and test that contract. Returning success with arguments from the previous command is not safe to leave implicit.

tekinertekin: We crossed — I posted the same conclusion a few seconds before your update, so we
agree on both the root cause and the two options.

The fix is now pushed as 1f0f8fd: `m_argBuffer.resetSer()` on the empty-argument
branch, plus a `ReuseClearsPreviousArgs` test. With the fix the suite is 10/10;
with the fix reverted that one test fails 0 vs 4 while the other nine pass, so it
isolates the change rather than the surrounding code.

I kept it as a separate commit from the test work so it can be split into its own
PR if a maintainer would rather it landed on its own. And if the preferred
resolution is the non-reusable-contract route instead, I will swap it.

lntutor (approved):
Rechecked exact head 1f0f8fd5d1d2f11cc5af1bfeaaa94bf753e02a85. The new no-argument branch reset removes the stale argument state, and the added reuse test directly covers the previously demonstrated failure. I built and ran the focused Fw_Cmd suite locally: all 10 tests passed, including ReuseClearsPreviousArgs. The earlier partial descriptor and partial opcode cases remain present, so both findings from my prior review are addressed.

tekinertekin:
> Rechecked exact head [1f0f8fd](https://github.com/nasa/fprime/commit/1f0f8fd5d1d2f11cc5af1bfeaaa94bf753e02a85). The new no-argument branch reset removes the stale argument state, and the added reuse test directly covers the previously demonstrated failure. I built and ran the focused Fw_Cmd suite locally: all 10 tests passed, including ReuseClearsPreviousArgs. The earlier partial descriptor and partial opcode cases remain present, so both findings from my prior review are addressed.

Thanks a lot for building and re-running it, appreciated.

LeStarch (dismissed):

LeStarch (dismissed):
I made minor edits in that we should not static_assert by rather skip tests if the configured type size is too small.

LeStarch (approved):

PR #5529, opened 2026-07-29
FpySequencer add FFLOOR, FABS and IABS directives, math func consistency

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

* Adds the FFLOOR directive, which has the IEEE 754 `roundToIntegralTowardNegative` behavior
* Adds IABS, which has the same behavior as Rust's `abs` func
* Adds FABS, which has the IEEE 754 `abs` behavior
* Also, bring EXIT directive up to date with Fpy, which now uses I32 instead of U8 for exit code
* FPTOSI and FPTOUI saturate if OOB of their int types, and map NaN to 0 (this avoids UB and is how Rust does it)
* Fix DOMAIN_ERROR which should have been ARITHMETIC_OVERFLOW in sdiv
* Fix sign of zero in FMOD to match python
* Make the docs much more explicit about the various arithmetic edge cases
* Bump schema to 7
* Update fprime-fpy requirement in requirements.txt

## Rationale

* Fpy already has code for all of these things, but the code doesn't implement standard IEEE behavior (for FFLOOR or FABS), or for IABS doesn't properly behave at I64 MIN. 
* Technically, we could just generate code to do all of these, but I think we should just add math functions to the directives so that they can link very clearly and directly to standard math functions. 
* Later there will be trig functions in the bytecode, so this is a precedent that will be repeated.

AI wrote the code.


Svc/FpySequencer/FpySequencerTypes.fpp
```diff
@@ -1,7 +1,7 @@
 module Svc {
     module Fpy {
         @ the current schema version (must be representable in U8)
-        constant SCHEMA_VERSION = 6;
+        constant SCHEMA_VERSION = 7;
```
lntutor, line 4: [P1] Please update the compiler dependency in tandem with this schema bump. FPrime's root `requirements.txt` still pins `fprime-fpy==0.4.0`, and I checked the latest published `0.4.1` wheel as well: both use `SCHEMA_VERSION = 6`, omit directive IDs 79–81, and emit/pop a one-byte `U8` for `EXIT`. After this line changes the runtime to schema 7, sequences produced by the standard pinned installation are rejected by `WrongSchemaVersion` before execution; the new directives and four-byte signed `EXIT` cannot interoperate with a published compiler either. Please land and release the compatible Fpy protocol changes, update `requirements.txt` to that release, and ideally add a smoke test that compiles a sequence with the pinned installed compiler and validates/runs it against this runtime. The current head of fprime-community/fpy#140 contains the matching schema 7/opcode/`I32` behavior, but it is not a released package.

lntutor (commented):
Compatibility blocker: the runtime schema and compiler package need to move in tandem. The companion Fpy work matches this protocol on its current head, but no compatible Fpy release is published yet. The failing fprime-python-reference check appears unrelated: its FSW aborts while opening `/dev/video0`.

LeStarch (approved):

PR #5533, opened 2026-07-29
Fix undefined behavior in FpySequencer op_mul overflow guard

### LINKED ISSUE #128: Flight VM op_mul overflow guard has signed-overflow UB and misses I64_MIN * negative
The overflow guard in the flight VM's `op_mul` (`fprime/Svc/FpySequencer/FpySequencerDirectives.cpp:749`, fprime devel `d0aacb8e7`) evaluates `-1 * lhs` / `-1 * rhs` in the both-negative branch:

```cpp
else if ((rhs < 0) && (lhs < 0) && ((std::numeric_limits<I64>::max() / (-1 * rhs)) < (-1 * lhs))) {
```

For `lhs == I64_MIN`, `-1 * lhs` is itself signed-overflow **undefined behavior**, and with ordinary codegen it wraps back to `I64_MIN` (negative), making the threshold comparison false — so the guard misses, and the subsequent `lhs * rhs` executes a *second* UB overflow and silently pushes a wrapped value. Observed on x86-64 g++ (the model traps correctly on all of these):

| expression | Python model | flight VM (x86 g++) |
|---|---|---|
| `I64_MIN * -1` | ARITHMETIC_OVERFLOW | silently `I64_MIN` |
| `I64_MIN * -2` | ARITHMETIC_OVERFLOW | silently `0` |
| `I64_MIN * I64_MIN` | ARITHMETIC_OVERFLOW | silently `0` |
| `-1 * I64_MIN` | ARITHMETIC_OVERFLOW | ARITHMETIC_OVERFLOW (wrap happens to trip the guard) |

Note the behavior is operand-order-dependent, and being UB it can change arbitrarily with compiler/flags — this is exactly the class of thing a flight VM must not do.

Fix: restructure the both-negative check without negation, e.g. treat `I64_MIN` operands explicitly (any `I64_MIN * x` with `x < 0`, or `x > 1`, overflows), or use `__builtin_mul_overflow` where toolchains allow. The Python model's `_trunc_div`-based guards handle these cases correctly and can serve as the reference.

Found by differential-testing `i64.wast` mul vectors through the model and the verbatim C++ op expressions.

rvaccone:
|                                                         |                                |
| :------------------------------------------------------ | :----------------------------- |
| **_Related Issue(s)_**                                  | Fixes fprime-community/fpy#128 |
| **_Has Unit Tests (y/n)_**                              | y                              |
| **_Documentation Included (y/n)_**                      | n                              |
| **_Generative AI was used in this contribution (y/n)_** | y                              |

---

## Change Description

The both-negative overflow check in `op_mul` negated its operands via `-1 * rhs` and `-1 * lhs`. Negating `I64` min is undefined behavior, so the guard missed the overflow and pushed a silently wrapped product (`I64_MIN * -2` pushed `0`). The check is now `lhs < std::numeric_limits<I64>::max() / rhs`, which evaluates no negation and is algebraically identical to the fpy model's reference guard. Tests add the issue's vectors in both operand orders plus boundary tightness checks.

## Rationale

A flight sequence VM must not contain arithmetic undefined behavior: results were operand order dependent, varied with compiler flags (aborting outright under `-ftrapv`), and diverged from the ground model, which traps correctly on all of these cases.

Run `fprime-util check` in `Svc/FpySequencer`; the new tests fail on the previous guard and the UBSan build flags the negation. The fix was also differentially verified against `__int128` ground truth and the fpy model guard over millions of vectors with zero mismatches.

## Future Work

fpy#127 (`op_sdiv` rounding) will be proposed separately.

AI assisted (Claude Code): defect analysis, the guard rewrite, unit tests, and verification harnesses. All changes reviewed by the submitter, who takes responsibility for every line.

### DIFF
--- a/Svc/FpySequencer/FpySequencerDirectives.cpp
+++ b/Svc/FpySequencer/FpySequencerDirectives.cpp
@@ -751,8 +751,8 @@ DirectiveError FpySequencer::op_mul() {
     if ((rhs > 0) && (lhs > 0) && ((std::numeric_limits<I64>::max() / rhs) < lhs)) {
         return DirectiveError::ARITHMETIC_OVERFLOW;
     }
-    // Check the both negative case
-    else if ((rhs < 0) && (lhs < 0) && ((std::numeric_limits<I64>::max() / (-1 * rhs)) < (-1 * lhs))) {
+    // Check the both negative case. Compare without negation: negating a value of min is undefined behavior
+    else if ((rhs < 0) && (lhs < 0) && (lhs < (std::numeric_limits<I64>::max() / rhs))) {
         return DirectiveError::ARITHMETIC_OVERFLOW;
     }
     // Underflow can occur with operands of differing signs and occurs when one operand is less than the minimum value
--- a/Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
+++ b/Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
@@ -1210,6 +1210,51 @@ TEST_F(FpySequencerTester, mul_overflow_neg_neg) {
     tester_push<I64>(-3);
     ASSERT_EQ(tester_op_mul(), DirectiveError::ARITHMETIC_OVERFLOW);
 }
+// I64 min has no positive counterpart, so any negation-based guard misses these cases
+TEST_F(FpySequencerTester, mul_overflow_min_neg_one) {
+    tester_push<I64>(std::numeric_limits<I64>::min());
+    tester_push<I64>(-1);
+    ASSERT_EQ(tester_op_mul(), DirectiveError::ARITHMETIC_OVERFLOW);
+}
+TEST_F(FpySequencerTester, mul_overflow_neg_one_min) {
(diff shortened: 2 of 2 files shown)


lntutor (approved):
Reviewed the signed-multiplication guard across all four sign quadrants and the exact I64 boundaries. For two negative operands, lhs < max / rhs is safe under C++ truncating division and correctly handles either operand being I64_MIN without evaluating negation.

Fresh local validation on head 1a9b09b8610880d5706e153fc60b804cea251dc1:
- Component and unit-test targets compiled successfully.
- 279/279 FpySequencer GoogleTests passed.
- 1/1 CTest target passed.

I also compiled the target with AddressSanitizer and UndefinedBehaviorSanitizer enabled. On macOS 26.5.2 arm64 the test binary stalls inside the known ASan InitializeShadowMemory startup issue before main, so the runtime count above used the sanitizer-disabled cache. I found no PR-specific correctness issue.

LeStarch (approved):

zimri-leisher (approved):

PR #5534, opened 2026-07-30
Fix type-limits warning in Svc/Ccsds/TestUtils/TestUtils.cpp

bevinduckett:
| | |
|:---|:---|
|**_Related Issue(s)_**| No  |
|**_Has Unit Tests (y/n)_**| Yes |
|**_Documentation Included (y/n)_**| N/A  |

---
## Change Description

The flags we're using when compiling unit tests generate an warning message (treated as an error) when compiling Svc/Ccsds/TestUtils/TestUtils.cpp - so the loop was modified 
```
/workspaces/<project>/lib/fprime/Svc/Ccsds/TestUtils/TestUtils.cpp:30:54: error: comparison is always true due to limited range of data type [-Werror=type-limits]
     for (SerialType candidateApid = 0; candidateApid <= bound; candidateApid++) {
                                        ~~~~~~~~~~~~~~^~~~~~~~
```

## Rationale

Our project uses U8 integers for APID (because we're not using CCSDS) and the maximum APID value is 0xFF , so a type-limits is generated when Svc/Ccsds/TestUtils/TestUtils.cpp is compiled (which it is when `fprime-utils build --ut` is run, despite `FPRIME_ENABLE_FRAMEWORK_UTS=OFF` being in settings.ini)
```

-Wtype-limits

    Warn if a comparison is always true or always false due to the limited range of the data type, but do not warn for constant expressions. For example, warn if an unsigned variable is compared against zero with < or >=. This warning is also enabled by -Wextra.
```

Compile UTs & verify they pass 

## Future Work

None

Code review / spelling

### DIFF
--- a/Svc/Ccsds/TestUtils/TestUtils.cpp
+++ b/Svc/Ccsds/TestUtils/TestUtils.cpp
@@ -27,7 +27,9 @@ ApidOption getRandomApid() {
     // selected index, or we run out of numbers
     SerialType idx = 0;
     SerialType apid = 0;
-    for (SerialType candidateApid = 0; candidateApid <= bound; candidateApid++) {
+    // If bound is the type max then text candidate <= bound causes a type-limits warning
+    // So bound is checked within in the loop
+    for (SerialType candidateApid = 0; bound > 0; candidateApid++) {
         if (ComCfg::Apid::isValid(candidateApid)) {
             // Found a valid APID: store it
             apid = candidateApid;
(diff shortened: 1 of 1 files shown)


lntutor (commented):
The warning workaround still leaves the loop condition true for every representable value when `bound == std::numeric_limits<SerialType>::max()`: each value is either `< bound` or `== bound`. If the selected valid APID is not found, incrementing `candidateApid` at the maximum wraps it to zero and the scan repeats forever.

The usual U8 configuration will generally exit early because its configured values are necessarily within the 11-bit range, but termination should not depend on that invariant—especially because this helper explicitly documents returning `NONE` when no configured APID fits. Please keep the explicit terminal-value check from the prior revision: run the body, then `break` when `candidateApid == bound`, and only increment otherwise. A max-bound exhaustion test would protect both the warning fix and termination behavior.

LeStarch (approved):

PR #5539, opened 2026-07-30
split Verson's config function so events and tlm can be sent later in the startup process

csmith608:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Deployments that require topology to be configured before `time` can be used could not use CdhCore subtopology because Version's config function sent events and telemetry. Sending events and telemetry involves a time call.

## Rationale

This PR splits Version's config function in to `config()`, which is called in `configComponents` and `start()`, which can be called later in the startup process to output events and tlm, `startTasks` in this case

Tested on pfsoc video kit hardware

### DIFF
--- a/Svc/Subtopologies/CdhCore/CdhCore.fpp
+++ b/Svc/Subtopologies/CdhCore/CdhCore.fpp
@@ -42,6 +42,10 @@ module CdhCore {
         // Startup TLM and Config verbosity for Versions
         CdhCore::version.config(true);
         """
+        phase Fpp.ToCpp.Phases.startTasks """
+        // Startup TLM and EVENTS
+        CdhCore::version.start();
+        """
     }
 
     instance textLogger: Svc.PassiveTextLogger base id CdhCoreConfig.BASE_ID + 0x004000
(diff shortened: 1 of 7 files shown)


lntutor (commented):
The lifecycle split makes sense, but two gaps currently hide regressions and leave the public API contract stale.

First, the newly enabled startup test asserts the `ProjectVersion` event against `Project::Version::FRAMEWORK_VERSION`, while `projectVersion_tlm()` emits `PROJECT_VERSION`. The test fixture defines both strings identically, so this passes even if the implementation reports the wrong source. Please give the framework and project fixtures distinct values and assert `PROJECT_VERSION` for project telemetry/events (including the existing command-path assertions). It would also strengthen the split to verify that `config()` leaves all startup event/TLM histories empty before `start()` populates them.

Second, `Svc/Version/docs/sdd.md` still says startup reporting is triggered by `version.config(true)`, and the CdhCore SDD describes only that configuration hook. Since direct users must now call `start()` at an appropriate later phase or silently lose startup reporting, please update those integration instructions to document the new two-step lifecycle and the CdhCore `startTasks` call.

LeStarch (approved):

lntutor (approved):
Rechecked exact head 3b8ba9c. The updated tests now distinguish project and framework version strings, assert the project source consistently, and verify config() emits no startup telemetry before start(). The Version and CdhCore SDDs now document the two-step lifecycle and the startTasks call. Local verification: Svc_Version built and all 3 focused tests passed; CdhCore subtopology autocode generated successfully; git diff --check passed. My earlier findings are resolved.

PR #5540, opened 2026-07-30
remove platform restriction

csmith608:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Removed platform restriction from BufferAccumulator

## Rationale

BufferAccumulator was added to the DataProducts subtopology, but meant that that subtop was restricted to Linux and Darwin because BufferAccumulator included `#include <sys/time.h>` but that include is now unused. 

Tested on pfsoc video kit running VxWorks

### DIFF
--- a/Svc/BufferAccumulator/BufferAccumulator.cpp
+++ b/Svc/BufferAccumulator/BufferAccumulator.cpp
@@ -12,8 +12,6 @@
 
 #include "Svc/BufferAccumulator/BufferAccumulator.hpp"
 
-#include <sys/time.h>
-
 #include <limits>
 #include "Fw/Types/BasicTypes.hpp"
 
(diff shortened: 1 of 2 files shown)


lntutor (approved):
Verified at 7c441521: the change removes only the unused POSIX header and the corresponding Linux/Darwin CMake restriction. The complete BufferAccumulator implementation and its FIFO helper use standard C++ plus F Prime/Os abstractions; I found no remaining POSIX API or platform-specific dependency. Core Linux/macOS/RHEL builds, unit tests, TSan, CodeQL, cppcheck, and FppTest are also passing, consistent with the reported VxWorks validation.

lntutor (approved):
Reviewed exact head 7c441521da769641854f3b45453e5d5cb4461dc8. The production BufferAccumulator module no longer contains any POSIX-only header or API use; the only remaining sys/time.h reference is confined to test/ut/Drain.cpp. Removing restrict_platforms(Linux Darwin) therefore matches the production dependency surface, and the full CI/cross-compilation matrix is green. The reported VxWorks hardware build provides the target-platform coverage that CI does not.

LeStarch (approved):

PR #5556, opened 2026-08-02
fix(FileDownlink): set cancel packet buffer to exact serialized size

### LINKED ISSUE #5347: FileDownlink cancel packet buffer is set to wrong size

philphauler:
## Overview
Cancel packets from `Svc::FileDownlink` were leaving the outbound buffer sized at the full internal buffer (`FILEDOWNLINK_INTERNAL_BUFFER_SIZE`) instead of the exact serialized packet length. Start/Data/End already shrink via `sendFilePacket()`; Cancel did not.

Strict consumers that call `Fw::FilePacket::fromBuffer` on the cancel buffer then fail with `FW_DESERIALIZE_SIZE_MISMATCH`, because `CancelPacket::fromSerialBuffer` requires **zero leftover bytes** after the header (cancel is header-only).

## Changes
- After serializing the cancel packet, call `buffer.setSize(filePacket.bufferSize() + sizeof(FwPacketDescriptorType))` — same contract as `sendFilePacket()`.
- Unit test: `validateCancelPacket` asserts the buffer advertises that exact size.

## Testing
- FileDownlink unit tests (`CancelDownlink` path) expected to cover this via `validateCancelPacket`.
- Logic verified against `CancelPacket.cpp` leftover check and side-by-side with `sendFilePacket()`.

## Related issue
Fixes #5347

## Notes
Small, isolated fix. No API changes. Helps any component that deserializes FileDownlink cancel buffers using the official `fromBuffer` path.

### DIFF
--- a/Svc/FileDownlink/FileDownlink.cpp
+++ b/Svc/FileDownlink/FileDownlink.cpp
@@ -396,6 +396,15 @@ void FileDownlink ::sendCancelPacket() {
     // Serialize the filePacket content into the buffer
     status = filePacket.toBuffer(offsetBuffer);
     FW_ASSERT(status == Fw::FW_SERIALIZE_OK);
+    // Match sendFilePacket(): expose the exact serialized size on the wire.
+    // getBuffer() leaves size at FILEDOWNLINK_INTERNAL_BUFFER_SIZE; leaving that
+    // full size makes receivers that call filePacket.fromBuffer(fwBuffer) fail with
+    // FW_DESERIALIZE_SIZE_MISMATCH on Cancel (Start/Data/End already set size).
+    // See nasa/fprime#5347.
+    const U32 bufferSize = filePacket.bufferSize() + static_cast<U32>(sizeof(FwPacketDescriptorType));
+    FW_ASSERT(buffer.getSize() >= bufferSize, static_cast<FwAssertArgType>(buffer.getSize()),
+              static_cast<FwAssertArgType>(bufferSize));
+    buffer.setSize(bufferSize);
     this->bufferSendOut_out(0, buffer);
     this->m_packetsSent.packetSent();
 }
--- a/Svc/FileDownlink/FileDownlink.cpp
+++ b/Svc/FileDownlink/FileDownlink.cpp
@@ -396,11 +396,6 @@ void FileDownlink ::sendCancelPacket() {
     // Serialize the filePacket content into the buffer
     status = filePacket.toBuffer(offsetBuffer);
     FW_ASSERT(status == Fw::FW_SERIALIZE_OK);
-    // Match sendFilePacket(): expose the exact serialized size on the wire.
-    // getBuffer() leaves size at FILEDOWNLINK_INTERNAL_BUFFER_SIZE; leaving that
-    // full size makes receivers that call filePacket.fromBuffer(fwBuffer) fail with
-    // FW_DESERIALIZE_SIZE_MISMATCH on Cancel (Start/Data/End already set size).
-    // See nasa/fprime#5347.
     const U32 bufferSize = filePacket.bufferSize() + static_cast<U32>(sizeof(FwPacketDescriptorType));
     FW_ASSERT(buffer.getSize() >= bufferSize, static_cast<FwAssertArgType>(buffer.getSize()),
               static_cast<FwAssertArgType>(bufferSize));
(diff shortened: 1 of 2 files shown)


Svc/FileDownlink/FileDownlink.cpp
```diff
@@ -396,6 +396,15 @@ void FileDownlink ::sendCancelPacket() {
...
+    // getBuffer() leaves size at FILEDOWNLINK_INTERNAL_BUFFER_SIZE; leaving that
+    // full size makes receivers that call filePacket.fromBuffer(fwBuffer) fail with
+    // FW_DESERIALIZE_SIZE_MISMATCH on Cancel (Start/Data/End already set size).
+    // See nasa/fprime#5347.
+    const U32 bufferSize = filePacket.bufferSize() + static_cast<U32>(sizeof(FwPacketDescriptorType));
+    FW_ASSERT(buffer.getSize() >= bufferSize, static_cast<FwAssertArgType>(buffer.getSize()),
+              static_cast<FwAssertArgType>(bufferSize));
+    buffer.setSize(bufferSize);
     this->bufferSendOut_out(0, buffer);
     this->m_packetsSent.packetSent();
```
LeStarch, line 404: You forgot to reset the size after the send out.

Svc/FileDownlink/FileDownlink.cpp
```diff
@@ -396,6 +396,15 @@ void FileDownlink ::sendCancelPacket() {
     // Serialize the filePacket content into the buffer
     status = filePacket.toBuffer(offsetBuffer);
     FW_ASSERT(status == Fw::FW_SERIALIZE_OK);
+    // Match sendFilePacket(): expose the exact serialized size on the wire.
```
LeStarch: No need for this large comment if it mirrors existing code.

LeStarch (changes requested):

LeStarch (dismissed):

LeStarch (approved):

PR #5557, opened 2026-08-02
Add unit tests for Fw::CmdPacket deserialization

### LINKED ISSUE #5507: Add unit tests for Fw::CmdPacket deserialization

durasi:
## Change Description
Adds a unit test suite for `Fw::CmdPacket`, which previously had no tests. Covers all return paths of `deserializeFrom`: nominal (with/without args), type mismatch, empty buffer, and truncated buffer. `serializeTo` is intentionally not tested as it is `FW_ASSERT(false)` by design.

## Rationale
Closes #5507. CmdPacket is on the critical path of every uplinked command; this adds coverage for its deserialization logic including all error branches.

Run `fprime-util check` in `Fw/Cmd`. All 5 tests pass locally on macOS (AppleClang 21, ASan+UBSan enabled).

### DIFF
--- a/Fw/Cmd/CMakeLists.txt
+++ b/Fw/Cmd/CMakeLists.txt
@@ -17,3 +17,13 @@ register_fprime_module(
   DEPENDS
     Fw_Com
 )
+
+### UTs ###
+register_fprime_ut(
+  SOURCES
+    "${CMAKE_CURRENT_LIST_DIR}/test/ut/CmdPacketTest.cpp"
+  DEPENDS
+    Fw_Cmd
+    Fw_Com
+    Fw_Types
+)
--- a/Fw/Cmd/test/ut/CmdPacketTest.cpp
+++ b/Fw/Cmd/test/ut/CmdPacketTest.cpp
@@ -1,0 +1,79 @@
+#include <gtest/gtest.h>
+#include <Fw/Cmd/CmdPacket.hpp>
+#include <Fw/Com/ComBuffer.hpp>
+
+// Helper: build a ComBuffer containing [descriptor][opcode][args...]
+static Fw::ComBuffer buildPacketBuffer(FwPacketDescriptorType descriptor,
+                                       FwOpcodeType opcode,
+                                       const U8* args = nullptr,
+                                       FwSizeType argsSize = 0) {
+    Fw::ComBuffer buffer;
+    EXPECT_EQ(Fw::FW_SERIALIZE_OK, buffer.serializeFrom(descriptor));
+    EXPECT_EQ(Fw::FW_SERIALIZE_OK, buffer.serializeFrom(opcode));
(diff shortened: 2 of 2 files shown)


LeStarch:
Seems like a duplicate of: https://github.com/nasa/fprime/pull/5517

Issue #5142, opened 2026-05-12
TlmChan Susceptible to a Double-Buffered Table Race Condition


bitWarrior:
| | |
|:---|:---|
|**_F´ Version_**| 4.2.1 |
|**_Affected Component_**|  TlmChan  |
---
## Problem Description

The TlmChan component uses a double-buffering strategy (m_tlmEntries[2]) to allow simultaneous telemetry updates and packetization with two defined guarded ports called TlmRecv and TlmGet, the standard Run_handler port.

However, it's the async Run_handler that performs the locking and unlocking when swapping the index (m_activeBuffer) between the buffers. During that iteration window when the Run_handler is locking and unlocking, a concurrent TlmGet_handler call can read a partially-swapped buffer. There will be no memory ordering guarantee if this occurs. In C++, this is known as a "data race" condition and is undefined behavior. 

There is a related issue downstream from this in the TlmGet port handler. If timestamps are incomparable, the code returns the inactive entry, the updated flag is not set correctly and stale active entry can be returned. This will send stale data out as telemetry data.

## Expected Behavior

My recommendation is to strengthen the thread-safety of TlmChan during the buffer swap and ensuring the TlmGet_handler cannot access the "inactive" buffer while it is being updated in the Run_handler.

    [In Run_Handler]

    // lock mutex long enough to modify active telemetry buffer
    // so the data can be read without worrying about updates
    this->lock();
    **[NEW]** U32 bufferToProcess = this->m_activeBuffer;
    this->m_activeBuffer = 1 - this->m_activeBuffer;
    // set activeBuffer to not updated
    for (U32 entry = 0; entry < TLMCHAN_HASH_BUCKETS; entry++) {
        this->m_tlmEntries[this->m_activeBuffer].buckets[entry].updated = false;
    }
    this->unLock();
    ..
    ..
    ** // Use bufferToProcess since it's set in the mutex lock
    for (U32 entry = 0; entry < TLMCHAN_HASH_BUCKETS; entry++) {
        TlmEntry* p_entry = &this->m_tlmEntries[bufferToProcess].buckets[entry];
        if ((p_entry->updated) && (p_entry->used)) {
             ..
             ..    

    [In TlmGet_handler]

    if (activeEntry && inactiveEntry) {
             ..
             ..
     [line 108] 
     ** UPDATED ELSE BLOCK **
    } else {
            // Use the updated flag to determine the newest telemetry channel value:
            //   - inactive updated, active not  -> prefer inactive
[...]

PR #5560, opened 2026-08-03
Svc: Guard TlmChan updated flag clears

ChimdumebiNebolisa:
Fixes the proven race in #5142 where Run_handler clears an inactive entry's updated flag without the guarded-port mutex while TlmGet_handler can read it. Adds a bounded deterministic regression test. Selector handling and incomparable-timestamp semantics are unchanged.


bitWarrior (approved):
Fixes race condition.

LeStarch (approved):

PR #5561, opened 2026-08-03
Add GenerateDp command to FileManager

### LINKED ISSUE #5515: Add command to generate data product from file to Svc/FileManager
| | |
|:---|:---|
|**_F´ Version_**| 4.x |
|**_Affected Component_**| Svc/FileManager |
---
## Feature Description

Add a command to the `Svc/FileManager` component to generate a data procuct from a file specified in the command.

## Rationale

This would allow making a DP and prioritizing the downlink of files that are not themselves DPs.

durasi:
## Change Description
Adds a `GenerateDp` command to `Svc::FileManager` that packages a file into data products.

Following the design described in #5515, the command takes a `fileName` and a `chunkSize`, and each chunk is emitted as a pair of records:
- `FileChunkHeaderRecord`: a struct carrying the source file name, the offset of the chunk within the file, and the number of data bytes
- `FileChunkDataRecord`: a U8 array record with the chunk bytes

Ground tools can reassemble the original file from the chunks, and the command works regardless of the size of the buffers allocated to data products.

Implementation notes:
- Chunks are paced by the rate group using the same internal port pattern as `ListDirectory`, one chunk per tick by default (`FileManagerConfig::CHUNKS_PER_RATE_TICK`), so per tick work stays bounded
- The command response is deferred until the last chunk is sent, matching the `ListDirectory` pattern
- The requested chunk size is clamped to `FileManagerConfig::GENERATE_DP_MAX_CHUNK_SIZE`, which bounds the component member read buffer so no dynamic allocation is needed
- A pair of records per chunk was chosen over a single fixed size struct record because FPP struct members are fixed size; with array records the serialized size tracks the real chunk size instead of always paying for the maximum

## Rationale
Closes #5515.

Run `fprime-util check` in `Svc/FileManager`. 21 tests pass locally on macOS (AppleClang 21, ASan and UBSan enabled), including four new cases: nominal chunking of a 100 byte file into 4 chunks with a 32 byte chunk size, a missing file, an empty file, and a chunk size larger than the configured maximum.

Note: `runRateGroupCycles` in the tester previously stopped cycling as soon as directory listing was idle, so it was generalised to keep cycling while either directory listing or data product generation is active.

### DIFF
--- a/Svc/FileManager/Commands.fppi
+++ b/Svc/FileManager/Commands.fppi
@@ -49,3 +49,13 @@ async command CalculateCrc(
                               filename: string size FileNameStringSize @< The file to CRC
                             ) \
   opcode 0x08
+
+@ Generate a data product from a file, split into chunks.
+@ Each chunk is emitted as a FileChunkHeaderRecord (file name, offset, data size)
+@ followed by a FileChunkDataRecord containing the chunk bytes, so ground tools
+@ can reassemble the original file regardless of DP buffer sizing.
+async command GenerateDp(
+                          fileName: string size FileNameStringSize @< The file to package as a data product
+                          chunkSize: U32 @< The maximum number of file bytes per chunk
+                        ) \
+  opcode 0x09
--- a/Svc/FileManager/Commands.fppi
+++ b/Svc/FileManager/Commands.fppi
@@ -59,5 +59,7 @@ async command GenerateDp(
                           chunkSize: U32 @< The maximum number of file bytes per chunk
                           beginOffset: U64 @< The offset in the file at which to start
                           endOffset: U64 @< The offset in the file at which to stop, exclusive; 0 means the end of the file
+                          $priority: U32 @< The container priority; 0 means the configured default
+                          mode: Svc.FileManager.GenerateDpMode @< Whether to pace the chunks with the rate group or emit them immediately
                         ) \
   opcode 0x09
--- a/Svc/FileManager/Commands.fppi
+++ b/Svc/FileManager/Commands.fppi
@@ -57,5 +57,7 @@ async command CalculateCrc(
 async command GenerateDp(
                           fileName: string size FileNameStringSize @< The file to package as a data product
                           chunkSize: U32 @< The maximum number of file bytes per chunk
+                          beginOffset: U64 @< The offset in the file at which to start
+                          endOffset: U64 @< The offset in the file at which to stop, exclusive; 0 means the end of the file
                         ) \
   opcode 0x09
--- a/Svc/FileManager/Events.fppi
+++ b/Svc/FileManager/Events.fppi
@@ -233,7 +233,7 @@ event CalculateCrcSucceeded(fileName: string size FileNameStringSize @< The name
   format "{} has CRC value 0x{x}"
 @ File data product generation started
 event GenerateDpStarted(fileName: string size FileNameStringSize @< The file being packaged
-                        fileSize: U64 @< The size of the file in bytes
+                        bytesToWrite: U64 @< The number of file bytes that will be written
                        ) \
   severity activity high \
   id 0x1e \
@@ -249,11 +249,12 @@ event GenerateDpComplete(fileName: string size FileNameStringSize @< The file th
 
 @ File data product generation failed
 event GenerateDpFailed(fileName: string size FileNameStringSize @< The file being packaged
+                       stage: Svc.FileManager.GenerateDpStage @< Where the failure occurred
                        status: U32 @< The error status
                       ) \
   severity warning high \
   id 0x20 \
-  format "Data product generation failed for file {}, status {}"
+  format "Data product generation failed for file {} at stage {}, status {}"
 
 @ Data product container allocation failed
 event GenerateDpBufferFailed(fileName: string size FileNameStringSize @< The file being packaged
--- a/Svc/FileManager/Events.fppi
+++ b/Svc/FileManager/Events.fppi
@@ -261,3 +261,13 @@ event GenerateDpBufferFailed(fileName: string size FileNameStringSize @< The fil
   severity warning high \
   id 0x21 \
   format "Failed to allocate data product buffer for file {}"
+
+@ Invalid offset range requested for data product generation
+event GenerateDpInvalidRange(fileName: string size FileNameStringSize @< The file being packaged
+                             beginOffset: U64 @< The requested begin offset
+                             endOffset: U64 @< The requested end offset
+                             fileSize: U64 @< The size of the file
+                            ) \
+  severity warning high \
+  id 0x22 \
+  format "Invalid offset range requested for file {}: [{}, {}) with file size {}"
--- a/Svc/FileManager/Events.fppi
+++ b/Svc/FileManager/Events.fppi
@@ -231,3 +231,33 @@ event CalculateCrcSucceeded(fileName: string size FileNameStringSize @< The name
   severity activity high \
   id 0x1c \
   format "{} has CRC value 0x{x}"
+@ File data product generation started
+event GenerateDpStarted(fileName: string size FileNameStringSize @< The file being packaged
+                        fileSize: U64 @< The size of the file in bytes
+                       ) \
+  severity activity high \
+  id 0x1e \
+  format "Started data product generation for file {} ({} bytes)"
+
+@ File data product generation completed
+event GenerateDpComplete(fileName: string size FileNameStringSize @< The file that was packaged
+                         chunks: U32 @< The number of chunks emitted
+                        ) \
+  severity activity high \
+  id 0x1f \
+  format "Completed data product generation for file {} in {} chunks"
+
+@ File data product generation failed
+event GenerateDpFailed(fileName: string size FileNameStringSize @< The file being packaged
+                       status: U32 @< The error status
+                      ) \
+  severity warning high \
+  id 0x20 \
+  format "Data product generation failed for file {}, status {}"
+
+@ Data product container allocation failed
+event GenerateDpBufferFailed(fileName: string size FileNameStringSize @< The file being packaged
+                            ) \
+  severity warning high \
+  id 0x21 \
+  format "Failed to allocate data product buffer for file {}"
--- a/Svc/FileManager/FileManager.cpp
+++ b/Svc/FileManager/FileManager.cpp
@@ -223,14 +223,14 @@ void FileManager ::GenerateDp_cmdHandler(FwOpcodeType opCode,
     // Reject a second request while one is already running
     if (this->m_dpState != DP_IDLE) {
         this->log_WARNING_HI_GenerateDpFailed(logFileName, 0);
-        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::OK);
         return;
     }
 
     // Data products must be available
     if (!this->isConnected_productGetOut_OutputPort(0) || !this->isConnected_productSendOut_OutputPort(0)) {
         this->log_WARNING_HI_GenerateDpBufferFailed(logFileName);
-        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::OK);
         return;
     }
 
@@ -243,7 +243,7 @@ void FileManager ::GenerateDp_cmdHandler(FwOpcodeType opCode,
     Os::File::Status status = this->m_dpFile.open(fileName.toChar(), Os::File::OPEN_READ);
     if (status != Os::File::OP_OK) {
         this->log_WARNING_HI_GenerateDpFailed(logFileName, static_cast<U32>(status));
-        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::OK);
         return;
     }
 
@@ -252,7 +252,7 @@ void FileManager ::GenerateDp_cmdHandler(FwOpcodeType opCode,
     if (status != Os::File::OP_OK) {
         this->m_dpFile.close();
         this->log_WARNING_HI_GenerateDpFailed(logFileName, static_cast<U32>(status));
-        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::OK);
         return;
     }
 
@@ -271,7 +271,7 @@ void FileManager ::GenerateDp_cmdHandler(FwOpcodeType opCode,
         this->m_dpFile.close();
         this->log_WARNING_HI_GenerateDpInvalidRange(logFileName, beginOffset, endOffset,
                                                     static_cast<U64>(fileSize));
-        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::VALIDATION_ERROR);
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::OK);
         return;
     }
 
@@ -281,7 +281,7 @@ void FileManager ::GenerateDp_cmdHandler(FwOpcodeType opCode,
         if (status != Os::File::OP_OK) {
             this->m_dpFile.close();
             this->log_WARNING_HI_GenerateDpFailed(logFileName, static_cast<U32>(status));
-            this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
+            this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::OK);
             return;
         }
     }
@@ -301,7 +301,7 @@ void FileManager ::GenerateDp_cmdHandler(FwOpcodeType opCode,
     // An empty range produces no chunks, so complete immediately
     if (this->m_dpOffset >= this->m_dpEndOffset) {
         this->log_ACTIVITY_HI_GenerateDpComplete(logFileName, this->m_dpChunkCount);
-        this->finishDpGeneration(Fw::CmdResponse::OK);
+        this->finishDpGeneration();
     }
 
     // Otherwise the response is deferred until the last chunk is sent
@@ -319,7 +319,7 @@ void FileManager ::processDpChunks() {
         const FwSizeType remaining = static_cast<FwSizeType>(this->m_dpEndOffset - this->m_dpOffset);
         if (remaining == 0) {
             this->log_ACTIVITY_HI_GenerateDpComplete(logFileName, this->m_dpChunkCount);
-            this->finishDpGeneration(Fw::CmdResponse::OK);
+            this->finishDpGeneration();
             return;
         }
 
@@ -330,7 +330,7 @@ void FileManager ::processDpChunks() {
         Os::File::Status status = this->m_dpFile.read(this->m_dpBuffer, readSize);
         if ((status != Os::File::OP_OK) || (readSize == 0)) {
             this->log_WARNING_HI_GenerateDpFailed(logFileName, static_cast<U32>(status));
-            this->finishDpGeneration(Fw::CmdResponse::EXECUTION_ERROR);
+            this->finishDpGeneration();
             return;
         }
 
@@ -341,7 +341,7 @@ void FileManager ::processDpChunks() {
         const Fw::Success::T dpStatus = this->dpGet_FileDpContainer(dpSize, container);
         if (dpStatus != Fw::Success::SUCCESS) {
             this->log_WARNING_HI_GenerateDpBufferFailed(logFileName);
-            this->finishDpGeneration(Fw::CmdResponse::EXECUTION_ERROR);
+            this->finishDpGeneration();
             return;
         }
 
@@ -355,7 +355,7 @@ void FileManager ::processDpChunks() {
         }
         if (serializeStatus != Fw::FW_SERIALIZE_OK) {
             this->log_WARNING_HI_GenerateDpFailed(logFileName, static_cast<U32>(serializeStatus));
-            this->finishDpGeneration(Fw::CmdResponse::EXECUTION_ERROR);
+            this->finishDpGeneration();
             return;
         }
 
@@ -367,19 +367,21 @@ void FileManager ::processDpChunks() {
         // Last chunk of the requested range
         if (this->m_dpOffset >= this->m_dpEndOffset) {
             this->log_ACTIVITY_HI_GenerateDpComplete(logFileName, this->m_dpChunkCount);
-            this->finishDpGeneration(Fw::CmdResponse::OK);
+            this->finishDpGeneration();
             return;
         }
     }
 }
 
-void FileManager ::finishDpGeneration(Fw::CmdResponse::T response) {
+void FileManager ::finishDpGeneration() {
     this->m_dpFile.close();
     this->m_dpState = DP_IDLE;
     this->m_dpOffset = 0;
     this->m_dpEndOffset = 0;
     this->m_dpFileSize = 0;
-    this->cmdResponse_out(this->m_dpOpCode, this->m_dpCmdSeq, response);
+    // Failures emit a warning event but still respond with OK, so that a bad
+    // file name or a transient resource problem does not stop a whole sequence
+    this->cmdResponse_out(this->m_dpOpCode, this->m_dpCmdSeq, Fw::CmdResponse::OK);
 }
 
 void FileManager ::pingIn_handler(const FwIndexType portNum, U32 key) {
--- a/Svc/FileManager/FileManager.cpp
+++ b/Svc/FileManager/FileManager.cpp
@@ -35,7 +35,15 @@ FileManager ::FileManager(const char* const compName  //!< The component name
       m_totalEntries(0),
       m_currentOpCode(0),
       m_currentCmdSeq(0),
-      m_runQueued(false) {}
+      m_runQueued(false),
+      m_dpState(DP_IDLE),
+      m_dpFileSize(0),
+      m_dpOffset(0),
+      m_dpChunkSize(0),
+      m_dpChunkCount(0),
+      m_dpOpCode(0),
+      m_dpCmdSeq(0),
+      m_dpBuffer{} {}
 
 FileManager ::~FileManager() {}
 
@@ -203,6 +211,141 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
     file.close();
 }
 
+void FileManager ::GenerateDp_cmdHandler(FwOpcodeType opCode,
+                                         U32 cmdSeq,
+                                         const Fw::CmdStringArg& fileName,
+                                         U32 chunkSize) {
+    Fw::LogStringArg logFileName(fileName.toChar());
+
+    // Reject a second request while one is already running
+    if (this->m_dpState != DP_IDLE) {
+        this->log_WARNING_HI_GenerateDpFailed(logFileName, 0);
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
+        return;
+    }
+
+    // Data products must be available
+    if (!this->isConnected_productGetOut_OutputPort(0) || !this->isConnected_productSendOut_OutputPort(0)) {
+        this->log_WARNING_HI_GenerateDpBufferFailed(logFileName);
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
+        return;
+    }
+
+    // Clamp the requested chunk size to the configured read buffer
+    U32 effectiveChunkSize = chunkSize;
+    if ((effectiveChunkSize == 0) || (effectiveChunkSize > FileManagerConfig::GENERATE_DP_MAX_CHUNK_SIZE)) {
+        effectiveChunkSize = FileManagerConfig::GENERATE_DP_MAX_CHUNK_SIZE;
+    }
+
+    Os::File::Status status = this->m_dpFile.open(fileName.toChar(), Os::File::OPEN_READ);
+    if (status != Os::File::OP_OK) {
+        this->log_WARNING_HI_GenerateDpFailed(logFileName, static_cast<U32>(status));
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
+        return;
+    }
+
+    FwSizeType fileSize = 0;
+    status = this->m_dpFile.size(fileSize);
+    if (status != Os::File::OP_OK) {
+        this->m_dpFile.close();
+        this->log_WARNING_HI_GenerateDpFailed(logFileName, static_cast<U32>(status));
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
+        return;
+    }
+
+    this->m_dpFileName = Fw::String(fileName.toChar());
+    this->m_dpFileSize = fileSize;
+    this->m_dpOffset = 0;
+    this->m_dpChunkSize = effectiveChunkSize;
+    this->m_dpChunkCount = 0;
+    this->m_dpOpCode = opCode;
+    this->m_dpCmdSeq = cmdSeq;
+    this->m_dpState = DP_IN_PROGRESS;
+
+    this->log_ACTIVITY_HI_GenerateDpStarted(logFileName, static_cast<U64>(fileSize));
+
+    // An empty file produces no chunks, so complete immediately
+    if (fileSize == 0) {
+        this->finishDpGeneration(Fw::CmdResponse::OK);
+    }
+
+    // Otherwise the response is deferred until the last chunk is sent
+}
+
+void FileManager ::processDpChunks() {
+    Fw::LogStringArg logFileName(this->m_dpFileName.toChar());
+
+    for (U32 chunk = 0; chunk < FileManagerConfig::CHUNKS_PER_RATE_TICK; chunk++) {
+        if (this->m_dpState != DP_IN_PROGRESS) {
+            return;
+        }
+
+        // Number of bytes remaining in the file
+        const FwSizeType remaining = this->m_dpFileSize - static_cast<FwSizeType>(this->m_dpOffset);
+        if (remaining == 0) {
+            this->log_ACTIVITY_HI_GenerateDpComplete(logFileName, this->m_dpChunkCount);
+            this->finishDpGeneration(Fw::CmdResponse::OK);
+            return;
+        }
+
+        FwSizeType readSize = (remaining < static_cast<FwSizeType>(this->m_dpChunkSize))
+                                  ? remaining
+                                  : static_cast<FwSizeType>(this->m_dpChunkSize);
+
+        Os::File::Status status = this->m_dpFile.read(this->m_dpBuffer, readSize);
+        if ((status != Os::File::OP_OK) || (readSize == 0)) {
+            this->log_WARNING_HI_GenerateDpFailed(logFileName, static_cast<U32>(status));
+            this->finishDpGeneration(Fw::CmdResponse::EXECUTION_ERROR);
+            return;
+        }
+
+        // Request a container large enough for this chunk's header and data
+        const FwSizeType dpSize = SIZE_OF_FileChunkHeaderRecord_RECORD +
+                                  SIZE_OF_FileChunkDataRecord_RECORD(readSize);
+        DpContainer container;
+        const Fw::Success::T dpStatus = this->dpGet_FileDpContainer(dpSize, container);
+        if (dpStatus != Fw::Success::SUCCESS) {
+            this->log_WARNING_HI_GenerateDpBufferFailed(logFileName);
+            this->finishDpGeneration(Fw::CmdResponse::EXECUTION_ERROR);
+            return;
+        }
+
+        // Each chunk is a metadata record followed by a data record, so that
+        // ground tools can reassemble the file from any number of containers
+        const FileManager_FileChunkHeader header(this->m_dpFileName, this->m_dpOffset,
+                                                 static_cast<U32>(readSize));
+        Fw::SerializeStatus serializeStatus = container.serializeRecord_FileChunkHeaderRecord(header);
+        if (serializeStatus == Fw::FW_SERIALIZE_OK) {
+            serializeStatus = container.serializeRecord_FileChunkDataRecord(this->m_dpBuffer, readSize);
+        }
+        if (serializeStatus != Fw::FW_SERIALIZE_OK) {
+            this->log_WARNING_HI_GenerateDpFailed(logFileName, static_cast<U32>(serializeStatus));
+            this->finishDpGeneration(Fw::CmdResponse::EXECUTION_ERROR);
+            return;
+        }
+
+        this->dpSend(container);
+
+        this->m_dpOffset += static_cast<U64>(readSize);
+        this->m_dpChunkCount++;
+
+        // Last chunk of the file
+        if (static_cast<FwSizeType>(this->m_dpOffset) >= this->m_dpFileSize) {
+            this->log_ACTIVITY_HI_GenerateDpComplete(logFileName, this->m_dpChunkCount);
+            this->finishDpGeneration(Fw::CmdResponse::OK);
+            return;
+        }
+    }
+}
+
+void FileManager ::finishDpGeneration(Fw::CmdResponse::T response) {
+    this->m_dpFile.close();
+    this->m_dpState = DP_IDLE;
+    this->m_dpOffset = 0;
+    this->m_dpFileSize = 0;
+    this->cmdResponse_out(this->m_dpOpCode, this->m_dpCmdSeq, response);
+}
+
 void FileManager ::pingIn_handler(const FwIndexType portNum, U32 key) {
     // return key
     this->pingOut_out(0, key);
@@ -222,6 +365,11 @@ void FileManager ::schedIn_handler(const FwIndexType portNum, U32 context) {
 void FileManager ::run_internalInterfaceHandler() {
     FW_ASSERT(this->m_runQueued);
     this->m_runQueued = false;  // Run is not queued anymore (we are running)
+    // Data product generation is paced the same way as directory listing
+    if (this->m_dpState == DP_IN_PROGRESS) {
+        this->processDpChunks();
+    }
+
     // Only process if we're in the middle of a directory listing
     if (m_listState == LISTING_IN_PROGRESS) {
         // Process multiple files per rate tick based on configuration
--- a/Svc/FileManager/FileManager.cpp
+++ b/Svc/FileManager/FileManager.cpp
@@ -40,6 +40,7 @@ FileManager ::FileManager(const char* const compName  //!< The component name
       m_dpFileSize(0),
       m_dpOffset(0),
       m_dpChunkSize(0),
+      m_dpEndOffset(0),
       m_dpChunkCount(0),
       m_dpOpCode(0),
       m_dpCmdSeq(0),
@@ -214,7 +215,9 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
 void FileManager ::GenerateDp_cmdHandler(FwOpcodeType opCode,
                                          U32 cmdSeq,
                                          const Fw::CmdStringArg& fileName,
-                                         U32 chunkSize) {
+                                         U32 chunkSize,
+                                         U64 beginOffset,
+                                         U64 endOffset) {
     Fw::LogStringArg logFileName(fileName.toChar());
 
     // Reject a second request while one is already running
@@ -253,9 +256,40 @@ void FileManager ::GenerateDp_cmdHandler(FwOpcodeType opCode,
         return;
     }
 
+    // An end offset of zero, or one past the end of the file, means the end of
+    // the file. Ranges let an operator retransmit part of a file or spread the
+    // downlink over several commands.
+    U64 effectiveEnd = endOffset;
+    if ((effectiveEnd == 0) || (effectiveEnd > static_cast<U64>(fileSize))) {
+        effectiveEnd = static_cast<U64>(fileSize);
+    }
+
+    const bool emptyFile = (fileSize == 0);
+    const bool badRange = (beginOffset > static_cast<U64>(fileSize)) ||
+                          (!emptyFile && (beginOffset >= effectiveEnd));
+    if (badRange) {
+        this->m_dpFile.close();
+        this->log_WARNING_HI_GenerateDpInvalidRange(logFileName, beginOffset, endOffset,
+                                                    static_cast<U64>(fileSize));
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::VALIDATION_ERROR);
+        return;
+    }
+
+    // Position the file at the start of the requested range
+    if (beginOffset > 0) {
+        status = this->m_dpFile.seek(static_cast<FwSignedSizeType>(beginOffset), Os::File::SeekType::ABSOLUTE);
+        if (status != Os::File::OP_OK) {
+            this->m_dpFile.close();
+            this->log_WARNING_HI_GenerateDpFailed(logFileName, static_cast<U32>(status));
+            this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::EXECUTION_ERROR);
+            return;
+        }
+    }
+
     this->m_dpFileName = Fw::String(fileName.toChar());
     this->m_dpFileSize = fileSize;
-    this->m_dpOffset = 0;
+    this->m_dpOffset = beginOffset;
+    this->m_dpEndOffset = effectiveEnd;
     this->m_dpChunkSize = effectiveChunkSize;
     this->m_dpChunkCount = 0;
     this->m_dpOpCode = opCode;
@@ -264,8 +298,9 @@ void FileManager ::GenerateDp_cmdHandler(FwOpcodeType opCode,
 
     this->log_ACTIVITY_HI_GenerateDpStarted(logFileName, static_cast<U64>(fileSize));
 
-    // An empty file produces no chunks, so complete immediately
-    if (fileSize == 0) {
+    // An empty range produces no chunks, so complete immediately
+    if (this->m_dpOffset >= this->m_dpEndOffset) {
+        this->log_ACTIVITY_HI_GenerateDpComplete(logFileName, this->m_dpChunkCount);
         this->finishDpGeneration(Fw::CmdResponse::OK);
     }
 
@@ -280,8 +315,8 @@ void FileManager ::processDpChunks() {
             return;
         }
 
-        // Number of bytes remaining in the file
-        const FwSizeType remaining = this->m_dpFileSize - static_cast<FwSizeType>(this->m_dpOffset);
+        // Number of bytes remaining in the requested range
+        const FwSizeType remaining = static_cast<FwSizeType>(this->m_dpEndOffset - this->m_dpOffset);
         if (remaining == 0) {
             this->log_ACTIVITY_HI_GenerateDpComplete(logFileName, this->m_dpChunkCount);
             this->finishDpGeneration(Fw::CmdResponse::OK);
@@ -329,8 +364,8 @@ void FileManager ::processDpChunks() {
         this->m_dpOffset += static_cast<U64>(readSize);
         this->m_dpChunkCount++;
 
-        // Last chunk of the file
-        if (static_cast<FwSizeType>(this->m_dpOffset) >= this->m_dpFileSize) {
+        // Last chunk of the requested range
+        if (this->m_dpOffset >= this->m_dpEndOffset) {
             this->log_ACTIVITY_HI_GenerateDpComplete(logFileName, this->m_dpChunkCount);
             this->finishDpGeneration(Fw::CmdResponse::OK);
             return;
@@ -342,6 +377,7 @@ void FileManager ::finishDpGeneration(Fw::CmdResponse::T response) {
     this->m_dpFile.close();
     this->m_dpState = DP_IDLE;
     this->m_dpOffset = 0;
+    this->m_dpEndOffset = 0;
     this->m_dpFileSize = 0;
     this->cmdResponse_out(this->m_dpOpCode, this->m_dpCmdSeq, response);
 }
--- a/Svc/FileManager/FileManager.cpp
+++ b/Svc/FileManager/FileManager.cpp
@@ -41,6 +41,7 @@ FileManager ::FileManager(const char* const compName  //!< The component name
       m_dpOffset(0),
       m_dpChunkSize(0),
       m_dpEndOffset(0),
+      m_dpPriority(0),
       m_dpChunkCount(0),
       m_dpOpCode(0),
       m_dpCmdSeq(0),
@@ -217,12 +218,14 @@ void FileManager ::GenerateDp_cmdHandler(FwOpcodeType opCode,
                                          const Fw::CmdStringArg& fileName,
                                          U32 chunkSize,
                                          U64 beginOffset,
-                                         U64 endOffset) {
+                                         U64 endOffset,
+                                         U32 priority,
+                                         const FileManager_GenerateDpMode& mode) {
     Fw::LogStringArg logFileName(fileName.toChar());
 
     // Reject a second request while one is already running
     if (this->m_dpState != DP_IDLE) {
-        this->log_WARNING_HI_GenerateDpFailed(logFileName, 0);
+        this->log_WARNING_HI_GenerateDpFailed(logFileName, FileManager_GenerateDpStage::BUSY, 0);
         this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::OK);
         return;
     }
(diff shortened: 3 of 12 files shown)


durasi:
Added the SDD section for the command.

Two notes on scope:

The data product ports are left unconnected in `Svc/Subtopologies/FileHandling`, matching how `fileManager.schedIn` is exposed for the deployment to wire up. The command checks `isConnected` and fails with an event if the ports are not connected, so nothing breaks when they are absent. Happy to wire them into the FileHandling subtopology if you would prefer that here.

On the ground side, reassembling the chunks into the original file would need a small tool in fprime-gds that reads the `FileChunkHeaderRecord` and `FileChunkDataRecord` pairs and writes them out by offset. That felt like a separate change, but I am glad to open an issue for it or take it on if useful.

Full build (`fprime-util build --all`, 763 targets) and the FileManager unit tests (21 passing) are clean locally.

durasi:
Added `beginOffset` and `endOffset` per the discussion in #5515.

Semantics as described there: an `endOffset` of `0` means the end of the file, so `GenerateDp(file, chunkSize, 0, 0)` still packages the whole file; an `endOffset` past the end of the file is clamped to the file size; and a `beginOffset` at or past the `endOffset` is rejected with a `GenerateDpInvalidRange` event and a validation error response. Since each chunk header carries the absolute offset within the source file, chunks produced by separate commands reassemble on the ground without extra bookkeeping.

Two more tests, so 23 passing locally: bytes [20, 60) of a 100 byte file with a 16 byte chunk size produces 3 chunks, and an inverted range is rejected. Full build is clean. SDD updated.

durasi:
Applied the sequence-friendly failure handling from #5515: every failure path now emits its warning event and still responds with OK, so a bad file name or a transient resource problem will not halt a sequence that contains the command. That covers file open and read errors, the size query, an invalid offset range, container allocation, serialization, and a request arriving while another is in progress. The events are unchanged, so operators still see exactly what went wrong.

23 tests still passing locally and the full build is clean. SDD notes the behaviour.

durasi:
Addressed the coverage regression flagged by the bot.

Added three tests covering the failure paths that were not exercised: a container allocation failure, a request arriving while another generation is in progress, and a serialization failure (the tester can now hand out an undersized buffer for that one).

I also removed two guards in `processDpChunks` that turned out to be unreachable: the loop returns as soon as the requested range is exhausted, so neither the idle-state check at the top nor the zero-remaining check could ever be hit.

`Svc/FileManager` line coverage is now 88.4% locally, up from 84.4% and above the 88.02% baseline. 26 tests passing.

On the failing YAMCS Reference job: it looks unrelated to this change, the log shows "YAMCS failed to start" before any test runs, and the same job is failing on other PRs against devel. Happy to look into it if that is not the case.

Svc/FileManager/Commands.fppi
```diff
@@ -49,3 +49,15 @@ async command CalculateCrc(
                               filename: string size FileNameStringSize @< The file to CRC
                             ) \
   opcode 0x08
+
+@ Generate a data product from a file, split into chunks.
+@ Each chunk is emitted as a FileChunkHeaderRecord (file name, offset, data size)
+@ followed by a FileChunkDataRecord containing the chunk bytes, so ground tools
+@ can reassemble the original file regardless of DP buffer sizing.
+async command GenerateDp(
```
timcanham, line 57: Add a priority argument to this function or projects will be stuck with 10, which is pretty high for a file. It could starve more important items. The convention is that a priority of zero reverts to the default, while > 0 is the new priority.

Svc/FileManager/FileManager.fpp
```diff
@@ -19,6 +19,29 @@ module Svc {
...
+    }
+
+    @ Chunk metadata record; each instance is followed by a FileChunkDataRecord
+    product record FileChunkHeaderRecord: FileChunkHeader id 0
+
+    @ Chunk data record; carries the actual file bytes for the preceding header
+    product record FileChunkDataRecord: U8 array id 1
+
+    @ Container for file data products
+    product container FileDpContainer id 0 default priority 10
```
timcanham: Move default priority to a config FPP file so it can be adjusted by projects.

Svc/FileManager/Events.fppi
```diff
@@ -230,4 +230,44 @@ event CalculateCrcSucceeded(fileName: string size FileNameStringSize @< The name
                       ) \
   severity activity high \
   id 0x1c \
-  format "{} has CRC value 0x{x}"
\ No newline at end of file
+  format "{} has CRC value 0x{x}"
+@ File data product generation started
+event GenerateDpStarted(fileName: string size FileNameStringSize @< The file being packaged
+                        fileSize: U64 @< The size of the file in bytes
```
timcanham: This may be answered later in the code, but it should reflect the size written if a portion of a file was written instead of the whole file.

Svc/FileManager/Events.fppi
```diff
@@ -230,4 +230,44 @@ event CalculateCrcSucceeded(fileName: string size FileNameStringSize @< The name
...
+@ File data product generation completed
+event GenerateDpComplete(fileName: string size FileNameStringSize @< The file that was packaged
+                         chunks: U32 @< The number of chunks emitted
+                        ) \
+  severity activity high \
+  id 0x1f \
+  format "Completed data product generation for file {} in {} chunks"
+
+@ File data product generation failed
+event GenerateDpFailed(fileName: string size FileNameStringSize @< The file being packaged
```
timcanham, line 251: This is used in a number of places, so it might be difficult to disambiguate. If you want to reuse an event, a common practice is to add an enum argument so the location of the error can be uniquely found.

Svc/FileManager/FileManager.cpp
```diff
@@ -203,6 +212,166 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
...
+    this->m_dpState = DP_IN_PROGRESS;
+
+    this->log_ACTIVITY_HI_GenerateDpStarted(logFileName, static_cast<U64>(fileSize));
+
+    // An empty range produces no chunks, so complete immediately
+    if (this->m_dpOffset >= this->m_dpEndOffset) {
+        this->log_ACTIVITY_HI_GenerateDpComplete(logFileName, this->m_dpChunkCount);
+        this->finishDpGeneration();
+    }
+
```
timcanham, line 314: Using the rate group to write chunks could be very slow (a small piece per cycle) and slow down the rate group too much. It's not a bad idea if projects want to mete out the file, but others could want it to go quickly for power reasons, etc. Let's make a "mode" option where it could use the rate group OR do it all at once.

Svc/FileManager/FileManager.cpp
```diff
@@ -203,6 +212,166 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
...
+        // Number of bytes remaining in the requested range. The loop returns as
+        // soon as the range is exhausted, so this is always non-zero here.
+        const FwSizeType remaining = static_cast<FwSizeType>(this->m_dpEndOffset - this->m_dpOffset);
+
+        FwSizeType readSize = (remaining < static_cast<FwSizeType>(this->m_dpChunkSize))
+                                  ? remaining
+                                  : static_cast<FwSizeType>(this->m_dpChunkSize);
+
+        Os::File::Status status = this->m_dpFile.read(this->m_dpBuffer, readSize);
+        if ((status != Os::File::OP_OK) || (readSize == 0)) {
```
timcanham: If the file size math was right, `readSize` should always come back as the requested size, so I would check for that instead of `readSize == 0`. You can stash the computed size in another variable and check it against `readSize` after the read.

timcanham (changes requested):
Good work! One moderate and a few small changes requested.

timcanham:
@durasi Did you try running this and decoding with the ground tools?

durasi:
Thanks for the review. All six points are addressed in the latest commit, and the branch is rebased on current devel.

**Priority argument.** The command now takes a `priority`, with zero reverting to the configured default and a non-zero value applying to the containers of that request.

**Default priority in config.** Moved to a new `default/config/FileManagerCfg.fpp` as `Svc::FileManagerCfg::DEFAULT_DP_PRIORITY`.

**Mode option.** Added a `GenerateDpMode` enum with `PACED` and `IMMEDIATE`. In paced mode the rate group meters the chunks out as before; in immediate mode the whole range is written in the command handler, so a project that wants the file out quickly is not limited by the rate group.

**GenerateDpStarted size.** Now reports the number of bytes that will be written, so a partial range is reflected correctly rather than the size of the whole file.

**Failure disambiguation.** Added a `GenerateDpStage` enum argument to `GenerateDpFailed` covering OPEN, SIZE, SEEK, READ, SERIALIZE and BUSY.

**Read size check.** The requested size is now stashed and compared against the returned size, instead of only checking for zero.

Two more tests, so 28 passing locally, and line coverage is 88.1%. The priority test recovers the value from the serialized container header, so it checks what actually goes out rather than just the member.

On decoding with the ground tools: no, I have only exercised this through unit tests so far. `Svc/FileManager` does not have its data product ports wired in the `FileHandling` subtopology, so a real run needs that wiring first, and I did not want to add it without asking. If you would like the end-to-end check, I am happy to wire the ports through the subtopology, run Ref, generate a data product from a file and decode it with `fprime-gds data-products decode`, and report back.

Svc/FileManager/FileManager.fpp
```diff
@@ -19,6 +19,53 @@ module Svc {
...
+      @ A request arrived while another was in progress
+      BUSY = 5
+    }
+
+    @ How the chunks of a file are emitted
+    enum GenerateDpMode {
+      @ Emit one chunk per rate group tick, spreading the work over time
+      PACED = 0
+      @ Emit all chunks in the command handler, completing immediately
+      IMMEDIATE = 1
```
lestarch-autobot, line 50: [Architecture] **suggestion** `IMMEDIATE` mode performs unbounded work in an async command handler, departing from this component's bounded-per-tick pacing pattern.

The baseline FileManager (active, all-async commands, sync `schedIn` delegating via the internal `run` port) deliberately bounds per-dispatch work (`FILES_PER_RATE_TICK`, `CHUNKS_PER_RATE_TICK`). `processDpChunks(0)` in the handler loops over the whole range, blocking the queue (other commands, `pingIn` health pings) for a duration proportional to file size. The PR description documents pacing but not this departure; see "Cyclic Notification Pattern" in `docs/user-manual/framework/component-and-port-selection.md`. Please add the rationale to the PR description and warn operators in the model doc.

```suggestion
      @ Emit all chunks in the command handler, completing immediately; runtime is
      @ unbounded by range size and blocks the component queue (commands, pings)
      IMMEDIATE = 1
```

Svc/FileManager/FileManager.cpp
```diff
@@ -222,6 +414,11 @@ void FileManager ::schedIn_handler(const FwIndexType portNum, U32 context) {
 void FileManager ::run_internalInterfaceHandler() {
     FW_ASSERT(this->m_runQueued);
     this->m_runQueued = false;  // Run is not queued anymore (we are running)
+    // Data product generation is paced the same way as directory listing
+    if (this->m_dpState == DP_IN_PROGRESS) {
+        this->processDpChunks(FileManagerConfig::CHUNKS_PER_RATE_TICK);
```
lestarch-autobot, line 419: [Architecture] **future work** DP pacing now depends on the preexisting `run drop` internal-port latch, which can stall permanently if the queue is ever full.

`schedIn_handler` sets `m_runQueued` true before `run_internalInterfaceInvoke()`; the internal port is declared `drop`, so on a full queue the message is discarded and `m_runQueued` never resets, silencing all future ticks. Preexisting (lines unchanged by this PR), but a stall now also strands an in-progress `GenerateDp`: `m_dpFile` stays open and the deferred command response is never sent.

Svc/FileManager/FileManager.cpp
```diff
@@ -203,6 +213,188 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
...
+    // Reject a second request while one is already running
+    if (this->m_dpState != DP_IDLE) {
+        this->log_WARNING_HI_GenerateDpFailed(logFileName, FileManager_GenerateDpStage::BUSY, 0);
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::OK);
+        return;
+    }
+
+    // Data products must be available
+    if (!this->isConnected_productGetOut_OutputPort(0) || !this->isConnected_productSendOut_OutputPort(0)) {
+        this->log_WARNING_HI_GenerateDpBufferFailed(logFileName);
```
lestarch-autobot, line 235: [Maintainability] **could fix** `maint-unclear-naming`: `GenerateDpBufferFailed` fired for unconnected ports.

The event's format string says "Failed to allocate data product buffer", but this path fires when the DP ports are not connected — an operator debugging from the event log will chase a buffer-pool problem instead of a topology problem. A dedicated `GenerateDpStage::NOT_CONNECTED` value on the existing `GenerateDpFailed` event would name the actual cause.

Svc/FileManager/FileManager.cpp
```diff
@@ -203,6 +213,188 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
...
+        return;
+    }
+
+    FwSizeType fileSize = 0;
+    status = this->m_dpFile.size(fileSize);
+    if (status != Os::File::OP_OK) {
+        this->m_dpFile.close();
+        this->log_WARNING_HI_GenerateDpFailed(logFileName, FileManager_GenerateDpStage::SIZE, static_cast<U32>(status));
+        this->cmdResponse_out(opCode, cmdSeq, Fw::CmdResponse::OK);
+        return;
```
lestarch-autobot, line 259: [Maintainability] **could fix** `maint-code-duplication`: the close + log + respond + return exit block is repeated on four setup-failure paths (OPEN, SIZE, bad range, SEEK), three of which must remember `m_dpFile.close()`.

A fix to this exit sequence must be applied in four places, and the next guard clause added between `open()` and the state hand-off will silently leak the file handle. A small helper, e.g. `failDpSetup(logFileName, stage, status)` that closes, logs, and responds, makes the cleanup structural.

Svc/FileManager/FileManager.hpp
```diff
@@ -189,6 +203,63 @@ class FileManager final : public FileManagerComponentBase {
...
+    GenerateDpState m_dpState;
+
+    //! File handle for the file being packaged
+    Os::File m_dpFile;
+
+    //! Name of the file being packaged (stored for records and events)
+    Fw::String m_dpFileName;
+
+    //! Total size of the file being packaged
+    FwSizeType m_dpFileSize;
```
lestarch-autobot, line 231: [Maintainability] **could fix** `maint-dead-code`: `m_dpFileSize` is written (ctor, `GenerateDp_cmdHandler`, `finishDpGeneration`) but never read.

It is a private member of a `final` class, so nothing outside this component can reach it; the range logic already uses the local `fileSize`. Dead state invites future readers to assume it is load-bearing. If removed, also drop the ctor initializer and the two assignments in `FileManager.cpp`.

```suggestion
```

Svc/FileManager/FileManager.cpp
```diff
@@ -203,6 +213,188 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
...
+    if (mode == FileManager_GenerateDpMode::IMMEDIATE) {
+        this->processDpChunks(0);
+    }
+}
+
+void FileManager ::processDpChunks(U32 chunkLimit) {
+    Fw::LogStringArg logFileName(this->m_dpFileName.toChar());
+
+    // A limit of zero means emit the whole remaining range in this call
+    const bool paced = (chunkLimit > 0);
```
lestarch-autobot, line 327: [Maintainability] **could fix** `maint-unclear-parameters`: `chunkLimit == 0` is a sentinel meaning "emit everything".

At the call site `processDpChunks(0)` reads as "process zero chunks" — the opposite of what happens — and a future caller passing a computed limit that can reach 0 flips pacing into a full drain silently. Passing the `GenerateDpMode` (or a named constant such as `ALL_REMAINING_CHUNKS`) makes the meaning visible at the call site.

Svc/FileManager/Events.fppi
```diff
@@ -230,4 +230,45 @@ event CalculateCrcSucceeded(fileName: string size FileNameStringSize @< The name
                       ) \
   severity activity high \
   id 0x1c \
-  format "{} has CRC value 0x{x}"
\ No newline at end of file
+  format "{} has CRC value 0x{x}"
```
lestarch-autobot, line 233: [Maintainability] **suggestion** `maint-inconsistent-local-convention`: missing blank line between `CalculateCrcSucceeded` and the new `GenerateDpStarted` event.

Every other event in this file is separated by a blank line; the missing separator makes the new block read as part of the previous event when scanning.

```suggestion
  format "{} has CRC value 0x{x}"

```

Svc/FileManager/test/ut/FileManagerTester.hpp
```diff
@@ -161,11 +194,50 @@ class FileManagerTester : public FileManagerGTestBase {
...
+    U8 m_dpContainerData[FW_COM_BUFFER_MAX_SIZE];
+
+    //! Buffer wrapping m_dpContainerData
+    Fw::Buffer m_dpContainerBuffer;
+
+    //! Number of containers sent by the component
+    U32 m_dpSendCount;
+
+    //! Total number of data bytes observed across sent containers
+    FwSizeType m_dpBytesSent;
```
lestarch-autobot: [Maintainability] **could fix** `maint-dead-code`: `m_dpBytesSent` is accumulated in `productSend_handler` but never asserted or read.

It reads as if a byte-count check exists somewhere, so a future test author will trust coverage that is not there. Either assert it in the chunking tests (it would strengthen `generateDpSucceed`) or remove the member and its two writes in `FileManagerTester.cpp`.

Svc/FileManager/docs/sdd.md
```diff
@@ -19,8 +19,49 @@ telemetry, and command responses.
 - AppendFile
 - FileSize
 - CalculateCrc
+- GenerateDp
```
lestarch-autobot, line 22: [Documentation] **could fix** `docs/reference/system-functional/file-management.md` does not list the new `GenerateDp` capability.

The "File System Operations" bullet list in that reference catalogs the File Manager's ground commands; this PR adds data-product generation without a corresponding entry. (Anchored here because the reference file is not in the PR diff.) Suggested addition: `- Package a file into data products (GenerateDp)`.

Svc/FileManager/Commands.fppi
```diff
@@ -49,3 +49,17 @@ async command CalculateCrc(
...
```
lestarch-autobot, line 57: [Documentation] **could fix** `docs/reference/system-functional/subtopology-file-handling.md` capability list omits the new command.

That reference describes File Manager as "(list, remove, create directory, move, copy, concatenate, hash)"; this PR adds data-product generation, so the catalog line is now stale. (Anchored here because the reference file is not in the PR diff.)

Svc/FileManager/docs/sdd.md
```diff
@@ -19,8 +19,49 @@ telemetry, and command responses.
...
+A priority of zero uses `Svc::FileManagerCfg::DEFAULT_DP_PRIORITY`, which
+projects can adjust, while a non-zero priority applies to the containers of
+that request only. The requested chunk size is clamped to
+`FileManagerConfig::GENERATE_DP_MAX_CHUNK_SIZE`, which bounds the read buffer
+held by the component.
+
+Failures during data product generation emit a warning event but still return a
+successful command response. A bad file name or a transient resource problem
+therefore does not stop a command sequence that happens to contain the command,
+while operators still see exactly what went wrong in the event log.
```
lestarch-autobot, line 58: [Documentation] **suggestion** The SDD does not document the busy-rejection behavior of `GenerateDp`.

`GenerateDp_cmdHandler` rejects a request while another is in progress (`GenerateDpFailed` at stage `BUSY`, response still `OK`). Operators sequencing multiple `GenerateDp` commands need this in the SDD.

```suggestion
Failures during data product generation emit a warning event but still return a
successful command response. A bad file name or a transient resource problem
therefore does not stop a command sequence that happens to contain the command,
while operators still see exactly what went wrong in the event log. Only one
request runs at a time: a `GenerateDp` command received while another is in
progress is rejected with a `GenerateDpFailed` event at the `BUSY` stage.
```

Svc/FileManager/docs/sdd.md
```diff
@@ -19,8 +19,49 @@ telemetry, and command responses.
...
+per tick by default, and the command response is deferred until the last chunk
+has been sent. In `IMMEDIATE` mode the whole range is written in the command
+handler, which suits projects that would rather finish quickly than spread the
+work over time.
+
+A priority of zero uses `Svc::FileManagerCfg::DEFAULT_DP_PRIORITY`, which
+projects can adjust, while a non-zero priority applies to the containers of
+that request only. The requested chunk size is clamped to
+`FileManagerConfig::GENERATE_DP_MAX_CHUNK_SIZE`, which bounds the read buffer
+held by the component.
```
lestarch-autobot, line 53: [Documentation] **suggestion** The SDD does not say that a chunk size of zero selects the configured maximum.

The handler maps `chunkSize == 0` to `GENERATE_DP_MAX_CHUNK_SIZE`, but the SDD only describes the clamp for oversized values.

```suggestion
that request only. The requested chunk size is clamped to
`FileManagerConfig::GENERATE_DP_MAX_CHUNK_SIZE`, which bounds the read buffer
held by the component; a chunk size of zero also selects this maximum.
```

Svc/FileManager/docs/sdd.md
```diff
@@ -19,8 +19,49 @@ telemetry, and command responses.
...
 For each command, the component returns success or failure and emits status
 information for operators.
 
+### GenerateDp
+
+`GenerateDp` packages a file into data products. The command takes the file
+name, a chunk size, a begin and end offset, a container priority and an
+emission mode, and the requested range of the file is emitted one chunk at a
+time. An end offset of zero means the end of the
+file, so a begin and end offset of zero packages the whole file. Ranges let an
```
lestarch-autobot, line 33: [Documentation] **suggestion** The SDD documents only `endOffset == 0` as "end of file", but the handler also clamps `endOffset > fileSize` to the file size.

```suggestion
time. An end offset of zero, or one beyond the end of the file, means the end of the
file, so a begin and end offset of zero packages the whole file. Ranges let an
```

Svc/FileManager/FileManager.fpp
```diff
@@ -19,6 +19,53 @@ module Svc {
...
+      fileName: string size FileNameStringSize @< The name of the source file
+      offset: U64 @< The offset of this chunk within the source file
+      dataSize: U32 @< The number of data bytes in this chunk
+    }
+
+    @ Chunk metadata record; each instance is followed by a FileChunkDataRecord
+    product record FileChunkHeaderRecord: FileChunkHeader id 0
+
+    @ Chunk data record; carries the actual file bytes for the preceding header
+    product record FileChunkDataRecord: U8 array id 1
```
lestarch-autobot, line 64: [Documentation] **could fix** The chunk-record format is a ground contract, but no ground-side (fprime-gds / GDS user-doc) reassembly documentation or tooling reference exists yet.

The SDD states "Ground tools reassemble the original file from these records"; the external doc gap (fprime-gds docs) should be tracked so the reassembly contract (header record followed by data record, absolute offsets) is documented where ground operators look.

Svc/FileManager/FileManager.cpp
```diff
@@ -203,6 +213,188 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
...
+        this->log_ACTIVITY_HI_GenerateDpComplete(logFileName, this->m_dpChunkCount);
+        this->finishDpGeneration();
+        return;
+    }
+
+    // In immediate mode the whole range is emitted here, so that a project that
+    // wants the file out quickly is not limited by the rate group. In paced
+    // mode the rate group meters the work out and the response is deferred.
+    if (mode == FileManager_GenerateDpMode::IMMEDIATE) {
+        this->processDpChunks(0);
```
lestarch-autobot, line 319: [Security] **must fix** `IMMEDIATE` mode performs ground-controlled unbounded work in the command handler (`ground-overflow-or-ddos`).

A single `GenerateDp` with `mode = IMMEDIATE` on an arbitrarily large file reads and emits the whole range inside the async command handler. This active component's queue (subsequent commands, `pingIn`) is blocked for a duration proportional to the ground-controlled file size — on deployments where `Svc::Health` pings FileManager, a large file can starve ping responses and trigger a FATAL, a ground-reachable DoS. Each iteration also requests a DP container, so one command can drain the DP buffer pool in a burst. Bound the per-call work (e.g. cap the chunks emitted per invocation via a config constant) or remove `IMMEDIATE` mode.

```suggestion
    if (mode == FileManager_GenerateDpMode::IMMEDIATE) {
        this->processDpChunks(FileManagerConfig::GENERATE_DP_MAX_IMMEDIATE_CHUNKS);
    }
```

(best-effort fix; verify before applying — requires adding `GENERATE_DP_MAX_IMMEDIATE_CHUNKS` to `FileManagerConfig.hpp`)

Svc/FileManager/test/ut/FileManagerTester.cpp
```diff
@@ -355,6 +355,336 @@ void FileManagerTester ::calculateCrcSucceed() {
...
+
+    this->resetDpState();
+
+    Fw::CmdStringArg cmdStringFile(fileName);
+    // 32 byte chunks over a 100 byte file yields 4 chunks (32 + 32 + 32 + 4)
+    this->sendCmd_GenerateDp(INSTANCE, CMD_SEQ, cmdStringFile, 32, 0, 0, 0, FileManager_GenerateDpMode::PACED);
+    this->component.doDispatch();
+
+    // Only the start event so far; the response is deferred
+    ASSERT_EVENTS_GenerateDpStarted_SIZE(1);
```
lestarch-autobot, line 419: [Test Quality] **must fix** `GenerateDpStarted` is asserted by count only; its payload (fileName, bytesToWrite) is never checked.

A new event with declared payload needs a payload assertion — the test passes even if the component reports the wrong file name or byte count (100 bytes here).

```suggestion
    ASSERT_EVENTS_GenerateDpStarted_SIZE(1);
    ASSERT_EVENTS_GenerateDpStarted(0, fileName, 100);
```

Svc/FileManager/test/ut/FileManagerTester.cpp
```diff
@@ -355,6 +355,336 @@ void FileManagerTester ::calculateCrcSucceed() {
...
+void FileManagerTester ::generateDpFileNotFound() {
+    this->resetDpState();
+
+    Fw::CmdStringArg cmdStringFile("no_such_dp_file");
+    this->sendCmd_GenerateDp(INSTANCE, CMD_SEQ, cmdStringFile, 32, 0, 0, 0, FileManager_GenerateDpMode::PACED);
+    this->component.doDispatch();
+
+    ASSERT_CMD_RESPONSE_SIZE(1);
+    ASSERT_CMD_RESPONSE(0, FileManager::OPCODE_GENERATEDP, CMD_SEQ, Fw::CmdResponse::OK);
+    ASSERT_EVENTS_GenerateDpFailed_SIZE(1);
```
lestarch-autobot, line 461: [Test Quality] **must fix** `GenerateDpFailed` is asserted by count only; the stage/status payload is never checked.

Without asserting `stage == OPEN`, this test would pass if the failure occurred at any other stage for any other reason. (best-effort fix; verify the expected status value before applying)

```suggestion
    ASSERT_EVENTS_GenerateDpFailed_SIZE(1);
    ASSERT_EVENTS_GenerateDpFailed(0, "no_such_dp_file", FileManager_GenerateDpStage::OPEN, static_cast<U32>(Os::File::DOESNT_EXIST));
```

Svc/FileManager/test/ut/FileManagerTester.cpp
```diff
@@ -355,6 +355,336 @@ void FileManagerTester ::calculateCrcSucceed() {
...
+    this->resetDpState();
+
+    Fw::CmdStringArg cmdStringFile(fileName);
+    // A begin offset at or past the end offset has nothing to package
+    this->sendCmd_GenerateDp(INSTANCE, CMD_SEQ, cmdStringFile, 16, 60, 20, 0, FileManager_GenerateDpMode::PACED);
+    this->component.doDispatch();
+
+    ASSERT_CMD_RESPONSE_SIZE(1);
+    ASSERT_CMD_RESPONSE(0, FileManager::OPCODE_GENERATEDP, CMD_SEQ, Fw::CmdResponse::OK);
+    ASSERT_EVENTS_GenerateDpInvalidRange_SIZE(1);
```
lestarch-autobot, line 567: [Test Quality] **must fix** `GenerateDpInvalidRange` is asserted by count only; the offset/size payload is never checked.

The event carries beginOffset, endOffset, and fileSize; asserting them verifies the rejection was for the requested [60, 20) range on the 100-byte file.

```suggestion
    ASSERT_EVENTS_GenerateDpInvalidRange_SIZE(1);
    ASSERT_EVENTS_GenerateDpInvalidRange(0, fileName, 60, 20, 100);
```

Svc/FileManager/test/ut/FileManagerTester.cpp
```diff
@@ -355,6 +355,336 @@ void FileManagerTester ::calculateCrcSucceed() {
...
+
+    Fw::CmdStringArg cmdStringFile(fileName);
+    this->sendCmd_GenerateDp(INSTANCE, CMD_SEQ, cmdStringFile, 16, 0, 0, 0, FileManager_GenerateDpMode::PACED);
+    this->component.doDispatch();
+    this->runRateGroupCycles(5);
+
+    // A container allocation failure is reported but does not fail the command
+    ASSERT_CMD_RESPONSE_SIZE(1);
+    ASSERT_CMD_RESPONSE(0, FileManager::OPCODE_GENERATEDP, CMD_SEQ, Fw::CmdResponse::OK);
+    ASSERT_EVENTS_GenerateDpBufferFailed_SIZE(1);
```
lestarch-autobot, line 596: [Test Quality] **must fix** `GenerateDpBufferFailed` is asserted by count only; the fileName payload is never checked.

One-line payload assertion closes the gap.

```suggestion
    ASSERT_EVENTS_GenerateDpBufferFailed_SIZE(1);
    ASSERT_EVENTS_GenerateDpBufferFailed(0, fileName);
```

Svc/FileManager/test/ut/FileManagerTester.cpp
```diff
@@ -355,6 +355,336 @@ void FileManagerTester ::calculateCrcSucceed() {
...
+    this->component.doDispatch();
+    ASSERT_CMD_RESPONSE_SIZE(0);
+
+    // A second request while one is in progress is reported and rejected
+    this->sendCmd_GenerateDp(INSTANCE, CMD_SEQ, cmdStringFile, 16, 0, 0, 0, FileManager_GenerateDpMode::PACED);
+    this->component.doDispatch();
+
+    ASSERT_CMD_RESPONSE_SIZE(1);
+    ASSERT_CMD_RESPONSE(0, FileManager::OPCODE_GENERATEDP, CMD_SEQ, Fw::CmdResponse::OK);
+    ASSERT_EVENTS_GenerateDpFailed_SIZE(1);
```
lestarch-autobot, line 628: [Test Quality] **must fix** The busy rejection asserts `GenerateDpFailed` by count only; `stage == BUSY` is never checked.

Without the payload assertion, this test also passes if the second command failed for an unrelated reason (e.g., OPEN).

```suggestion
    ASSERT_EVENTS_GenerateDpFailed_SIZE(1);
    ASSERT_EVENTS_GenerateDpFailed(0, fileName, FileManager_GenerateDpStage::BUSY, 0);
```

Svc/FileManager/test/ut/FileManagerTester.cpp
```diff
@@ -355,6 +355,336 @@ void FileManagerTester ::calculateCrcSucceed() {
...
+
+    Fw::CmdStringArg cmdStringFile(fileName);
+    this->sendCmd_GenerateDp(INSTANCE, CMD_SEQ, cmdStringFile, 16, 0, 0, 0, FileManager_GenerateDpMode::PACED);
+    this->component.doDispatch();
+    this->runRateGroupCycles(5);
+
+    // A serialization failure is reported but does not fail the command
+    ASSERT_CMD_RESPONSE_SIZE(1);
+    ASSERT_CMD_RESPONSE(0, FileManager::OPCODE_GENERATEDP, CMD_SEQ, Fw::CmdResponse::OK);
+    ASSERT_EVENTS_GenerateDpFailed_SIZE(1);
```
lestarch-autobot, line 665: [Test Quality] **must fix** The serialization-failure test asserts `GenerateDpFailed` by count only; `stage == SERIALIZE` is never checked.

Asserting the stage confirms the undersized buffer actually drove the SERIALIZE path rather than some other failure. (best-effort fix; verify the expected serialize status before applying)

```suggestion
    ASSERT_EVENTS_GenerateDpFailed_SIZE(1);
    ASSERT_EVENTS_GenerateDpFailed(0, fileName, FileManager_GenerateDpStage::SERIALIZE, static_cast<U32>(Fw::FW_SERIALIZE_NO_ROOM_LEFT));
```

Svc/FileManager/test/ut/FileManagerTester.cpp
```diff
@@ -355,6 +355,336 @@ void FileManagerTester ::calculateCrcSucceed() {
...
+    // An undersized buffer lets the serialization failure path be exercised
+    this->m_dpContainerBuffer.setSize(this->m_dpGetUndersizedBuffer ? Fw::DpContainer::MIN_PACKET_SIZE
+                                                                    : sizeof this->m_dpContainerData);
+    buffer = this->m_dpContainerBuffer;
+    return Fw::Success::SUCCESS;
+}
+
+void FileManagerTester ::productSend_handler(FwDpIdType id, const Fw::Buffer& buffer) {
+    this->m_dpSendCount++;
+    this->m_dpBytesSent += buffer.getSize();
```
lestarch-autobot, line 384: [Test Quality] **must fix** No test verifies the content of the emitted `FileChunkHeaderRecord`/`FileChunkDataRecord` records.

Tests assert only container counts; a bug writing wrong header offsets/dataSize or wrong chunk bytes would pass, defeating ground reassembly — the core contract of `GenerateDp`. `m_dpBytesSent` is accumulated here but never asserted anywhere. Suggest deserializing the records in this handler (or in `generateDpSucceed`) and asserting the header fields and data bytes against the known file content.

Svc/FileManager/test/ut/FileManagerTester.cpp
```diff
@@ -355,6 +355,336 @@ void FileManagerTester ::calculateCrcSucceed() {
...
+    this->resetDpState();
+
+    Fw::CmdStringArg cmdStringFile(fileName);
+    this->sendCmd_GenerateDp(INSTANCE, CMD_SEQ, cmdStringFile, 32, 0, 0, 0, FileManager_GenerateDpMode::PACED);
+    this->component.doDispatch();
+
+    // An empty file completes immediately with no chunks
+    ASSERT_CMD_RESPONSE_SIZE(1);
+    ASSERT_CMD_RESPONSE(0, FileManager::OPCODE_GENERATEDP, CMD_SEQ, Fw::CmdResponse::OK);
+    ASSERT_EVENTS_GenerateDpStarted_SIZE(1);
```
lestarch-autobot, line 484: [Test Quality] **suggestion** The empty-file case never asserts `GenerateDpComplete`, which the handler emits with chunks == 0.

Asserting completion (with payload) pins down the immediate-complete behavior for an empty range.

```suggestion
    ASSERT_EVENTS_GenerateDpStarted_SIZE(1);
    ASSERT_EVENTS_GenerateDpComplete_SIZE(1);
    ASSERT_EVENTS_GenerateDpComplete(0, fileName, 0);
```

Svc/FileManager/test/ut/FileManagerTester.cpp
```diff
@@ -355,6 +355,336 @@ void FileManagerTester ::calculateCrcSucceed() {
...
+    // A second request while one is in progress is reported and rejected
+    this->sendCmd_GenerateDp(INSTANCE, CMD_SEQ, cmdStringFile, 16, 0, 0, 0, FileManager_GenerateDpMode::PACED);
+    this->component.doDispatch();
+
+    ASSERT_CMD_RESPONSE_SIZE(1);
+    ASSERT_CMD_RESPONSE(0, FileManager::OPCODE_GENERATEDP, CMD_SEQ, Fw::CmdResponse::OK);
+    ASSERT_EVENTS_GenerateDpFailed_SIZE(1);
+
+    // Let the first generation finish so the component returns to idle
+    this->runRateGroupCycles(10);
```
lestarch-autobot, line 633: [Test Quality] **suggestion** The final `runRateGroupCycles(10)` drives the first generation to completion with no assertions afterward.

Asserting the second command response and the completion event verifies the busy rejection did not corrupt the in-progress generation.

```suggestion
    this->runRateGroupCycles(10);
    ASSERT_CMD_RESPONSE_SIZE(2);
    ASSERT_EVENTS_GenerateDpComplete_SIZE(1);
    ASSERT_EVENTS_GenerateDpComplete(0, fileName, 7);
```

Svc/FileManager/FileManager.cpp
```diff
@@ -203,6 +213,188 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
...
+
+        const FwSizeType requestedSize = (remaining < static_cast<FwSizeType>(this->m_dpChunkSize))
+                                             ? remaining
+                                             : static_cast<FwSizeType>(this->m_dpChunkSize);
+
+        // The file size is known, so a short read means the file changed underneath us
+        FwSizeType readSize = requestedSize;
+        const Os::File::Status status = this->m_dpFile.read(this->m_dpBuffer, readSize);
+        if ((status != Os::File::OP_OK) || (readSize != requestedSize)) {
+            this->log_WARNING_HI_GenerateDpFailed(logFileName, FileManager_GenerateDpStage::READ,
```
lestarch-autobot, line 342: [Test Quality] **could fix** The SIZE, SEEK, and READ failure stages, and the disconnected-DP-port path, have no test coverage.

OPEN, BUSY, SERIALIZE, and buffer-allocation failures are tested, but a short/failed read (file changed underneath), size/seek failures, and the `isConnected_productGetOut` guard are not. These need fault injection or an unconnected-port tester variant.

Svc/FileManager/test/ut/FileManagerTester.cpp
```diff
@@ -355,6 +355,336 @@ void FileManagerTester ::calculateCrcSucceed() {
...
+    this->m_dpBytesSent += buffer.getSize();
+
+    // Recover the priority from the serialized container header
+    Fw::DpContainer container(id, buffer);
+    if (container.deserializeHeader() == Fw::FW_SERIALIZE_OK) {
+        this->m_dpLastPriority = container.getPriority();
+    }
+}
+
+void FileManagerTester ::generateDpSucceed() {
```
lestarch-autobot, line 401: [Test Quality] **could fix** Eleven `generateDp*` test methods repeat an identical create-file / send / dispatch / cycle / assert / rm template.

Factoring a parameterized helper (fileName, size, chunkSize, offsets, priority, mode, expected chunks/events) would make each case a one-liner and reduce maintenance risk. See the repetitive-structure guidance in the F Prime test docs (RBT or parameterized helpers).

Svc/FileManager/test/ut/FileManagerTester.cpp
```diff
@@ -355,6 +355,336 @@ void FileManagerTester ::calculateCrcSucceed() {
...
+#if defined TGT_OS_TYPE_LINUX || TGT_OS_TYPE_DARWIN
+    const char* const fileName = "dp_test_file";
+    this->system("rm -f dp_test_file");
+    // 100 bytes of deterministic content
+    this->system("printf '0123456789%.0s' $(seq 1 10) > dp_test_file");
+#else
+    SKIP();  // Commands not implemented for this OS
```
lestarch-autobot, line 408: [Test Quality] **suggestion** New `SKIP()` calls reference a macro with no definition in the repository (gtest provides `GTEST_SKIP()`).

This branch only compiles on non-Linux/Darwin targets, where it will fail to build. The pattern repeats at the 9 other new `generateDp*` sites (and pre-exists at two `listDirectory*` sites). (best-effort fix; verify before applying)

```suggestion
    GTEST_SKIP();  // Commands not implemented for this OS
```

Svc/FileManager/test/ut/FileManagerTester.cpp
```diff
@@ -355,6 +355,336 @@ void FileManagerTester ::calculateCrcSucceed() {
...
+    ASSERT_CMD_RESPONSE_SIZE(0);
+
+    // Drive the rate group until the file is fully packaged
+    this->runRateGroupCycles(10);
+
+    ASSERT_CMD_RESPONSE_SIZE(1);
+    ASSERT_CMD_RESPONSE(0, FileManager::OPCODE_GENERATEDP, CMD_SEQ, Fw::CmdResponse::OK);
+    ASSERT_EVENTS_GenerateDpComplete_SIZE(1);
+    ASSERT_EVENTS_GenerateDpComplete(0, fileName, 4);
+    ASSERT_EQ(4u, this->m_dpSendCount);
```
lestarch-autobot, line 430: [Test Quality] **suggestion** The default container priority (priority arg == 0) is exercised but its value is never asserted.

`generateDpCustomPriority` checks the override (42); no test checks that priority 0 maps to `FileManagerCfg.DEFAULT_DP_PRIORITY`. (best-effort fix; verify the generated constant name before applying)

```suggestion
    ASSERT_EQ(4u, this->m_dpSendCount);
    ASSERT_EQ(static_cast<FwDpPriorityType>(FileManagerCfg::DEFAULT_DP_PRIORITY), this->m_dpLastPriority);
```

Svc/FileManager/FileManager.cpp
```diff
@@ -203,6 +213,188 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
...
+        container.setPriority(this->m_dpPriority);
+
+        // Each chunk is a metadata record followed by a data record, so that
+        // ground tools can reassemble the file from any number of containers
+        const FileManager_FileChunkHeader header(this->m_dpFileName, this->m_dpOffset, static_cast<U32>(readSize));
+        Fw::SerializeStatus serializeStatus = container.serializeRecord_FileChunkHeaderRecord(header);
+        if (serializeStatus == Fw::FW_SERIALIZE_OK) {
+            serializeStatus = container.serializeRecord_FileChunkDataRecord(this->m_dpBuffer, readSize);
+        }
+        if (serializeStatus != Fw::FW_SERIALIZE_OK) {
```
lestarch-autobot, line 366: [Correctness] **could fix** `correctness-resource-leak`: on serialization failure, the container obtained from `dpGet_FileDpContainer` is dropped without `dpSend`, leaking the underlying `BufferManager` allocation.

Trigger: `productGetOut` returns a valid buffer smaller than the requested `dpSize` (the port contract does not guarantee the size). Consequence: the buffer is never returned (`Fw.DataProductSync` has no return port), so repeated failures monotonically exhaust the buffer pool. Filed as could-fix because stock `Svc.DpManager` + `Svc.BufferManager` never hand out an undersized valid buffer, so the path is guarded upstream in standard topologies.

Svc/FileManager/test/ut/FileManagerTester.cpp
lestarch-autobot, line 408: [Correctness] **could fix** `correctness-other`: this `#else` branch does not compile on targets other than Linux/Darwin.

`SKIP()` is not defined anywhere in this test (gtest's macro is `GTEST_SKIP()`), and `fileName` is declared only inside the `#if` block but used after the `#endif`. The same pattern appears in all 10 new `generateDp*` tests. Consequence: the UT translation unit fails to compile on any other OS. The preexisting tests avoid this by using `FAIL()` and keeping declarations outside the conditional; declaring `fileName` before the `#if` and using `GTEST_SKIP()` would fix both.

Svc/FileManager/FileManager.cpp
```diff
@@ -203,6 +213,188 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
...
+
+void FileManager ::finishDpGeneration() {
+    this->m_dpFile.close();
+    this->m_dpState = DP_IDLE;
+    this->m_dpOffset = 0;
+    this->m_dpEndOffset = 0;
+    this->m_dpFileSize = 0;
+    // Failures emit a warning event but still respond with OK, so that a bad
+    // file name or a transient resource problem does not stop a whole sequence
+    this->cmdResponse_out(this->m_dpOpCode, this->m_dpCmdSeq, Fw::CmdResponse::OK);
```
lestarch-autobot, line 395: [Correctness] **could fix** `correctness-other`: `GenerateDp` never calls `emitTelemetry`, so the `CommandsExecuted`/`Errors` channels do not reflect this command on any path (success or failure).

Trigger: any `GenerateDp` execution. Consequence: operator-facing command/error counters are wrong for this command, unlike the other FileManager commands (only `CalculateCrc` shares this preexisting inconsistency). If intentional, a note in the SDD would help.

Svc/FileManager/FileManager.cpp
lestarch-autobot, line 419: [Correctness] **could fix** `correctness-predicate`: `processDpChunks` treats a limit of 0 as "emit all remaining chunks", so a project that configures `CHUNKS_PER_RATE_TICK = 0` silently gets the whole file emitted in one tick — the opposite of the pacing intent.

Trigger: `FileManagerConfig::CHUNKS_PER_RATE_TICK` set to 0 (a plausible "pause" configuration). Consequence: unbounded work in a single rate-group tick. A compile-time guard removes the ambiguity (best-effort fix; verify before applying).

```suggestion
        static_assert(FileManagerConfig::CHUNKS_PER_RATE_TICK > 0, "DP chunk pacing must be positive");
        this->processDpChunks(FileManagerConfig::CHUNKS_PER_RATE_TICK);
```

Svc/FileManager/FileManager.cpp
```diff
@@ -222,6 +414,11 @@ void FileManager ::schedIn_handler(const FwIndexType portNum, U32 context) {
 void FileManager ::run_internalInterfaceHandler() {
     FW_ASSERT(this->m_runQueued);
     this->m_runQueued = false;  // Run is not queued anymore (we are running)
+    // Data product generation is paced the same way as directory listing
+    if (this->m_dpState == DP_IN_PROGRESS) {
```
lestarch-autobot, line 418: [Correctness] **future work** `correctness-state-machine` (preexisting): the internal `run` port is declared `drop`; if a `run` message is dropped on a full queue, `m_runQueued` stays `true` forever and `schedIn_handler`'s compare-exchange never re-queues, permanently wedging paced operations.

Trigger: component queue full at the moment `run_internalInterfaceInvoke` is called. Consequence: directory listing — and now paced `GenerateDp` — stalls forever and the deferred command response is never sent. Preexisting mechanism, but this PR adds a second waited command that depends on it.

Svc/FileManager/FileManager.hpp
```diff
@@ -189,6 +203,63 @@ class FileManager final : public FileManagerComponentBase {
...
+    };
+
+    //! Current state of data product generation
+    GenerateDpState m_dpState;
+
+    //! File handle for the file being packaged
+    Os::File m_dpFile;
+
+    //! Name of the file being packaged (stored for records and events)
+    Fw::String m_dpFileName;
```
lestarch-autobot, line 228: [C++ Design] **suggestion** CPP-31 (silent truncation): `Fw::String` capacity is independent of `FileNameStringSize`; the stored name can silently truncate before it is written into `FileChunkHeaderRecord`.

The command arg is sized `FileNameStringSize` (240) but `Fw::String` is sized `FW_FIXED_LENGTH_STRING_SIZE`. Both are project-configurable; a project that shrinks the latter below the former silently truncates the file name that ground tools use to reassemble chunks. `Fw::FileNameString` (Fw/Types/FileNameString.hpp) is sized exactly to `FileNameStringSize`, making truncation impossible.

```suggestion
    Fw::FileNameString m_dpFileName;
```

Svc/FileManager/test/ut/FileManagerTester.hpp
```diff
@@ -161,11 +194,50 @@ class FileManagerTester : public FileManagerGTestBase {
...
+
+    //! Whether the get port should fail, for testing the failure path
+    bool m_dpGetShouldFail;
+
+    //! Whether the get port should hand out a buffer too small for a chunk
+    bool m_dpGetUndersizedBuffer;
+
+    //! Priority carried by the most recently sent container
+    FwDpPriorityType m_dpLastPriority;
```
lestarch-autobot: [C++ Design] **must fix** CPP-19 (initialize all variables): the new data-product bookkeeping members are uninitialized until `resetDpState()` runs.

The tester constructor does not initialize these members and does not call `resetDpState()`; any future test (or handler call) that touches them before `resetDpState()` reads indeterminate values. Default member initializers make the fixture safe by construction.

```suggestion
    U8 m_dpContainerData[FW_COM_BUFFER_MAX_SIZE] = {};

    //! Buffer wrapping m_dpContainerData
    Fw::Buffer m_dpContainerBuffer;

    //! Number of containers sent by the component
    U32 m_dpSendCount = 0;

    //! Total number of data bytes observed across sent containers
    FwSizeType m_dpBytesSent = 0;

    //! Whether the get port should fail, for testing the failure path
    bool m_dpGetShouldFail = false;

    //! Whether the get port should hand out a buffer too small for a chunk
    bool m_dpGetUndersizedBuffer = false;

    //! Priority carried by the most recently sent container
    FwDpPriorityType m_dpLastPriority = 0;
```

Svc/FileManager/FileManager.cpp
```diff
@@ -203,6 +213,188 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
...
+    }
+}
+
+void FileManager ::processDpChunks(U32 chunkLimit) {
+    Fw::LogStringArg logFileName(this->m_dpFileName.toChar());
+
+    // A limit of zero means emit the whole remaining range in this call
+    const bool paced = (chunkLimit > 0);
+
+    for (U32 chunk = 0; !paced || (chunk < chunkLimit); chunk++) {
```
lestarch-autobot, line 329: [C++ Design] **could fix** CPP-34 (bounded loops): in IMMEDIATE mode (`chunkLimit == 0`) the loop condition is always true, so the bound is not visible in the loop header.

Termination relies on the early returns (range exhausted, read/serialize/get failure) plus monotonic `m_dpOffset` progress, which holds because `readSize == requestedSize > 0` is enforced. Per the JPL loop-bound rule (CPP-27/CPP-34), prefer an explicit bound in the header, e.g. compute `maxChunks = (remaining range + chunkSize - 1) / chunkSize` up front and loop `chunk < maxChunks`.

Svc/FileManager/FileManager.cpp
```diff
@@ -203,6 +213,188 @@ void FileManager ::CalculateCrc_cmdHandler(FwOpcodeType opCode, U32 cmdSeq, cons
...
+    if (this->m_dpOffset >= this->m_dpEndOffset) {
+        this->log_ACTIVITY_HI_GenerateDpComplete(logFileName, this->m_dpChunkCount);
+        this->finishDpGeneration();
+        return;
+    }
+
+    // In immediate mode the whole range is emitted here, so that a project that
+    // wants the file out quickly is not limited by the rate group. In paced
+    // mode the rate group meters the work out and the response is deferred.
+    if (mode == FileManager_GenerateDpMode::IMMEDIATE) {
```
lestarch-autobot, line 318: [Design] **must fix** `design-fprime-anti-pattern`: `IMMEDIATE` mode performs unbounded ground-commanded work inside an async command handler.

For a large file, the active component's queue (including async `pingIn`) is blocked for the entire read + DP emission, risking health ping timeouts. Per `docs/user-manual/framework/component-and-port-selection.md`, event-driven work should stay bounded per dispatch; a configurable burst size (e.g. a per-request chunks-per-tick) would deliver the requested speed without starving the queue. Noting the mode was requested in PR review, so this is a deliberate trade-off needing confirmation.

cc @LeStarch @thomas-bc @timcanham — low-confidence finding, please confirm.

Svc/FileManager/FileManager.cpp
lestarch-autobot, line 395: [Design] **could fix** `design-code-mismatch`: `GenerateDp` never updates the component's `CommandsExecuted`/`Errors` telemetry counters.

The SDD states each command "returns success or failure and emits status information", and the other handlers call `emitTelemetry()`; `GenerateDp` (like `CalculateCrc`) bypasses it, so operators tracking the counters will not see these commands. Consider incrementing the counters on completion/failure for consistency.

lestarch-autobot (changes requested):

### Per-agent results

| Security Vulnerabilities | 1 | 0 | 0 | 0 | 1 | No-Go |
| F Prime C/C++ Design | 1 | 1 | 1 | 0 | 3 | No-Go |
| Documentation Currency | 0 | 3 | 3 | 0 | 6 | Go |
| Architecture | 0 | 1 | 0 | 1 | 2 | Go |
| Test Quality | 7 | 4 | 2 | 0 | 13 | No-Go |
| Correctness | 0 | 0 | 4 | 1 | 5 | Go |
| Maintainability | 0 | 1 | 5 | 0 | 6 | Go |
| **Totals** | 10 | 10 | 16 | 2 | 38 | **No-Go** |

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>
<summary>Outstanding must-fix items (10)</summary>

**Security Vulnerabilities**
- `IMMEDIATE` mode performs ground-controlled unbounded work in the command handler (`ground-overflow-or-ddos`) — https://github.com/nasa/fprime/pull/5561#discussion_r3751941372

**F Prime C/C++ Design**
- CPP-19: new data-product bookkeeping members are uninitialized until `resetDpState()` runs — https://github.com/nasa/fprime/pull/5561#discussion_r3751965453

**Design**
- `design-fprime-anti-pattern`: `IMMEDIATE` mode performs unbounded ground-commanded work inside an async command handler — https://github.com/nasa/fprime/pull/5561#discussion_r3751979816

**Test Quality**
- `GenerateDpStarted` asserted by count only; payload (fileName, bytesToWrite) never checked — https://github.com/nasa/fprime/pull/5561#discussion_r3751941461
- `GenerateDpFailed` asserted by count only; stage/status payload never checked — https://github.com/nasa/fprime/pull/5561#discussion_r3751941470
- `GenerateDpInvalidRange` asserted by count only; offset/size payload never checked — https://github.com/nasa/fprime/pull/5561#discussion_r3751941481
- `GenerateDpBufferFailed` asserted by count only; fileName payload never checked — https://github.com/nasa/fprime/pull/5561#discussion_r3751941488
- Busy rejection asserts `GenerateDpFailed` by count only; `stage == BUSY` never checked — https://github.com/nasa/fprime/pull/5561#discussion_r3751941498
- Serialization-failure test asserts `GenerateDpFailed` by count only; `stage == SERIALIZE` never checked — https://github.com/nasa/fprime/pull/5561#discussion_r3751941507
- No test verifies content of emitted `FileChunkHeaderRecord`/`FileChunkDataRecord` records — https://github.com/nasa/fprime/pull/5561#discussion_r3751941513

</details>

### Merge readiness
**Merge readiness: No-Go** — 10 outstanding must-fix items across Security Vulnerabilities (1), F Prime C/C++ Design (1), Design (1), and Test Quality (7).

- None — all nine registered reviewers completed.

---

A solid data-product downlink capability taking shape — clear the flagged items and this one is go for launch.

timcanham (commented):
The changes I requested were made.

durasi:
Thanks for the positive review.
Regarding the automated review from lestarch-autobot, I'm on it. I will address all the "must fix" items and push the updates as soon as possible.

durasi:
I've addressed the eight test quality and C++ design must-fix items in e8e6b3f7:

- Every GenerateDp event now asserts its full payload, not just the count: GenerateDpStarted (name + bytes), GenerateDpFailed at the OPEN, BUSY and SERIALIZE stages, GenerateDpInvalidRange offsets, and GenerateDpBufferFailed name. The open status and serialize status are recovered from the framework rather than hardcoded, so the assertions track the real values.

- The busy rejection test now drives the first generation to completion and asserts its response and GenerateDpComplete, confirming the rejection left the inprogress run intact.

- productSend_handler accumulates each container's data size, and generateDpSucceed asserts the totals, so the chunks are verified to carry real header-plus-data payloads.

- The tester's data product members have default initializers (CPP-19)

28 tests pass locally, line coverage is 88.2% (above the 88.02% baseline), and clangformat is clean.

On the two remaining must fix items (Security and Design), both concern IMMEDIATE mode doing unbounded work in the command handler. Since IMMEDIATE was added at your request for projects that want the file out quickly, I'd rather confirm the direction before changing it. Two options as I see them:

1) Keep IMMEDIATE but bound it with a per request burst cap, which keeps the speed intent while protecting the queue and health pings.

2) Drop IMMEDIATE and keep only the paced mode.

Which would you prefer, happy to implement either

LeStarch (dismissed):

LeStarch:
@durasi this looks to be an actual failure in your new UT.  Please resolve.

durasi:
Resolved in df7186f, and the branch is rebased onto current devel.

The abort came from Fw::Buffer::setData, which now advances within an existing allocation and asserts on a null backing pointer, so it can no longer initialize a default-constructed buffer. The tester used setData plus setSize to set up the DP container buffer valid when the earlier commits were written but no longer so as devel moved. Rebasing and purging the build cache reproduced the abort locally which made the cause clear.

Both call sites now build the buffer with the Fw::Buffer constructor, keeping the undersized-buffer path the serialization-failure test relies on.

Locally: 28 FileManager tests pass, the full framework suite passes (127/127), line coverage is 88.2%, and clangformat is clean. The Dp-adjacent modules (DpManager, DpWriter, DpCatalog, Fw/Dp, Fw/Buffer) pass as well.

The force push dismissed your approval sorry about that whenever you have a moment. Could you take another look?

timcanham (approved):
Changes I requested were made, but overall other actions need to be finished.

LeStarch (approved):

PR #5567, opened 2026-08-04
Update in support of per-component duty cycle high-water marks.

bdshenker:
| | |
|:---|:---|
|**_Related Issue(s)_**|5293, 5370,  5371 |

---
## Change Description

    We are submitting the following updates to the pull requests https://github.com/nasa/fprime/pull/5370, https://github.com/nasa/fprime/pull/5371:
    1) Declare m_portMaxTimes,
    2) Remove portTimes,
    3) Write the HWMs into m_portMaxTimes,
    4) Directly telemeter m_portDurationHighWaterMarks,
    5) Add command to clear HWMs: CLEAR_PORT_HWM.

## Rationale

We need to keep track of per-component duty cycle high-water marks, not running timing counts.
In addition, we would like to be able to reset high water marks on command.

The updates are minor.
They were tested on the test beds.

## Future Work

It is not directly related to this issue. As soon as these updates are approved and integrated, we will submit next pull request for issue https://github.com/nasa/fprime/issues/5243. It is currently being backlogged. 

IMPORTANT: I could not find in documentation if _update on change_ modifier works with array 
  telemetry. If it works on arrays or if there is a plan to make it work in the future, then it should apply here. 

AI (Claude) was used to add an array typed channel PortCycleTime. In addition, it was used in debugging of deployment configuration. Specifically, it helped to allocate sufficient memory in support of PortCycleTime addition.

### DIFF
--- a/Svc/PassiveRateGroup/PassiveRateGroup.cpp
+++ b/Svc/PassiveRateGroup/PassiveRateGroup.cpp
@@ -20,12 +20,11 @@
 
 namespace Svc {
 PassiveRateGroup::PassiveRateGroup(const char* compName)
-    : PassiveRateGroupComponentBase(compName), m_cycles(0), m_maxTime(0), m_numContexts(0) {
-    // Initialize port duration high water marks to zero
-    for (FwIndexType port = 0; port < NUM_RATEGROUPMEMBEROUT_OUTPUT_PORTS; port++) {
-        this->m_portDurationHighWaterMarks[port] = 0;
-    }
-}
+    : PassiveRateGroupComponentBase(compName),
+      m_cycles(0),
+      m_maxTime(0),
+      m_portDurationHighWaterMarks{},
+      m_numContexts(0) {}
 
 PassiveRateGroup::~PassiveRateGroup() {}
 
@@ -59,6 +58,7 @@ void PassiveRateGroup::CycleIn_handler(FwIndexType portNum, Os::RawTime& cycleSt
     // Pre-allocate RawTime objects outside loop to avoid repeated constructor calls
     Os::RawTime portStart;
     Os::RawTime portEnd;
+    PassiveRateGroup_CycleTime portTimes;
 
     // invoke any members of the rate group
     for (FwIndexType port = 0; port < this->getNum_RateGroupMemberOut_OutputPorts(); port++) {
@@ -73,9 +73,10 @@ void PassiveRateGroup::CycleIn_handler(FwIndexType portNum, Os::RawTime& cycleSt
                 (void)portEnd.now();
                 U32 cycleTime;
                 (void)portEnd.getDiffUsec(portStart, cycleTime);
+                portTimes[static_cast<FwSizeType>(port)] = cycleTime;
                 // Update high water mark if current cycle time exceeds it
-                if (cycleTime > this->m_portDurationHighWaterMarks[port]) {
-                    this->m_portDurationHighWaterMarks[port] = cycleTime;
+                if (cycleTime > this->m_portDurationHighWaterMarks[static_cast<FwSizeType>(port)]) {
+                    this->m_portDurationHighWaterMarks[static_cast<FwSizeType>(port)] = cycleTime;
                 }
             }
         }
@@ -95,12 +96,8 @@ void PassiveRateGroup::CycleIn_handler(FwIndexType portNum, Os::RawTime& cycleSt
     }
 
     if (Svc::PassiveRateGroupCfg::PortCycleTime) {
-        // Create array from HWM data for telemetry
-        PassiveRateGroup_CycleTime hwmArray;
-        for (FwSizeType i = 0; i < NUM_RATEGROUPMEMBEROUT_OUTPUT_PORTS; i++) {
-            hwmArray[i] = this->m_portDurationHighWaterMarks[i];
-        }
-        this->tlmWrite_PortCycleTime(hwmArray);
+        this->tlmWrite_PortCycleTimeLast(portTimes);
+        this->tlmWrite_PortCycleTimeHWM(this->m_portDurationHighWaterMarks);
     }
 
     this->tlmWrite_MaxCycleTime(this->m_maxTime);
@@ -108,7 +105,7 @@ void PassiveRateGroup::CycleIn_handler(FwIndexType portNum, Os::RawTime& cycleSt
     this->tlmWrite_CycleCount(++this->m_cycles);
 }
 
-void PassiveRateGroup::CLEAR_PORT_HWM_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
+void PassiveRateGroup::CLEAR_STATISTICS_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
     // Clear all port duration high water marks
     for (FwIndexType port = 0; port < NUM_RATEGROUPMEMBEROUT_OUTPUT_PORTS; port++) {
         this->m_portDurationHighWaterMarks[port] = 0;
--- a/Svc/PassiveRateGroup/PassiveRateGroup.cpp
+++ b/Svc/PassiveRateGroup/PassiveRateGroup.cpp
@@ -20,7 +20,12 @@
 
 namespace Svc {
 PassiveRateGroup::PassiveRateGroup(const char* compName)
-    : PassiveRateGroupComponentBase(compName), m_cycles(0), m_maxTime(0), m_numContexts(0) {}
+    : PassiveRateGroupComponentBase(compName), m_cycles(0), m_maxTime(0), m_numContexts(0) {
+    // Initialize port duration high water marks to zero
+    for (FwIndexType port = 0; port < NUM_RATEGROUPMEMBEROUT_OUTPUT_PORTS; port++) {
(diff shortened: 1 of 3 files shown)


LeStarch:
@timcanham @Kronos3, this changes from the per-run telemetry to a HWM.  This change expected?

timcanham:
I think it's more useful for determining margin, but it's also nice to have a running number to see what the steady state values are. I would add the running value as a second channel.

bdshenker:
@timcanham By "running number" do you mean latest sample or running average. @celskeggs What are your thoughts on adding this additional statistics?

timcanham:
I meant the latest sample. Essentially what was being pushed before. So ideally, both would be pushed, but if we had to pick one for performance reasons, it would be the high watermarks.

Svc/PassiveRateGroup/PassiveRateGroup.cpp
```diff
@@ -20,7 +20,12 @@
 
 namespace Svc {
 PassiveRateGroup::PassiveRateGroup(const char* compName)
-    : PassiveRateGroupComponentBase(compName), m_cycles(0), m_maxTime(0), m_numContexts(0) {}
+    : PassiveRateGroupComponentBase(compName), m_cycles(0), m_maxTime(0), m_numContexts(0) {
+    // Initialize port duration high water marks to zero
+    for (FwIndexType port = 0; port < NUM_RATEGROUPMEMBEROUT_OUTPUT_PORTS; port++) {
+        this->m_portDurationHighWaterMarks[port] = 0;
+    }
```
celskeggs: Is it insufficient to just initialize this using `, m_portDurationHighWaterMarks {}`? That might optimize better.

Svc/PassiveRateGroup/PassiveRateGroup.cpp
```diff
@@ -87,12 +95,28 @@ void PassiveRateGroup::CycleIn_handler(FwIndexType portNum, Os::RawTime& cycleSt
     }
 
     if (Svc::PassiveRateGroupCfg::PortCycleTime) {
-        this->tlmWrite_PortCycleTime(portTimes);
+        // Create array from HWM data for telemetry
+        PassiveRateGroup_CycleTime hwmArray;
```
celskeggs: Why create a second array? Why not make m_portDurationHighWaterMarks a `PassiveRateGroup_CycleTime` instance and directly populate it?

Svc/PassiveRateGroup/PassiveRateGroup.cpp
```diff
@@ -87,12 +95,28 @@ void PassiveRateGroup::CycleIn_handler(FwIndexType portNum, Os::RawTime& cycleSt
...
     this->tlmWrite_CycleCount(++this->m_cycles);
 }
 
+void PassiveRateGroup::CLEAR_PORT_HWM_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
+    // Clear all port duration high water marks
+    for (FwIndexType port = 0; port < NUM_RATEGROUPMEMBEROUT_OUTPUT_PORTS; port++) {
+        this->m_portDurationHighWaterMarks[port] = 0;
+    }
+    this->m_maxTime = 0;
+    this->m_cycles = 0;
```
celskeggs, line 114: This behavior is inconsistent with the name and description of the command. The command only indicates that the HWMs should be cleared, but this will clear everything.

celskeggs:
@bdshenker I think it's unnecessary for our use cases, but I can't immediately say that it will cause any problems for us to leave it in on Tim's behalf. I think we can put it in, and if we run into performance issues, we can add switch to having two different configuration options rather than just one.

Svc/PassiveRateGroup/PassiveRateGroup.fpp
```diff
@@ -19,7 +19,10 @@ module Svc {
     telemetry CycleCount: U32
 
     array CycleTime = [PassiveRateGroupOutputPorts] U32 default 0
-    telemetry PortCycleTime: CycleTime
+    telemetry PortCycleTime: CycleTime update on change
```
celskeggs: Needs to be clear in the name about whether this array contains HWMs

celskeggs (changes requested):
In addition to restoring the "latest measurement" channel, please consider the following changes.

Svc/PassiveRateGroup/PassiveRateGroup.cpp
bdshenker: @celskeggs I think we should clear all HWMs to keep them consistent. How about if I rename command into CLEAR_PASSIVE_RATE_GROUP_HWM?

celskeggs: Well, but not all of these are HWMs. Maybe rename the command to `CLEAR_STATISTICS`?

Svc/PassiveRateGroup/PassiveRateGroup.cpp
bdshenker: Restored "latest measurement" telemetry channels.
Updated initialization of m_portDurationHighWaterMarks.

Svc/PassiveRateGroup/PassiveRateGroup.cpp
bdshenker: Done

Svc/PassiveRateGroup/PassiveRateGroup.cpp
bdshenker: Changed command name to CLEAR_STATISTICS.

Svc/PassiveRateGroup/PassiveRateGroup.fpp
bdshenker: Changed array names to PortCycleTimeLast and PortCycleTimeHWM.

celskeggs (approved):
Looks good to me. Pending review from the F Prime team.

Kronos3 (approved):
This fits my use-case.

LeStarch (dismissed):

Svc/PassiveRateGroup/PassiveRateGroup.cpp
```diff
@@ -87,12 +96,24 @@ void PassiveRateGroup::CycleIn_handler(FwIndexType portNum, Os::RawTime& cycleSt
...
 
     this->tlmWrite_MaxCycleTime(this->m_maxTime);
     this->tlmWrite_CycleTime(cycleTime);
     this->tlmWrite_CycleCount(++this->m_cycles);
 }
 
+void PassiveRateGroup::CLEAR_STATISTICS_cmdHandler(FwOpcodeType opCode, U32 cmdSeq) {
+    // Clear all port duration high water marks
+    for (FwIndexType port = 0; port < NUM_RATEGROUPMEMBEROUT_OUTPUT_PORTS; port++) {
+        this->m_portDurationHighWaterMarks[static_cast<FwSizeType>(port)] = 0;
```
LeStarch, line 111: This is unguarded write to a member also accessed by the above code.  Use mutex, atomic, or `guarded` port

Svc/PassiveRateGroup/PassiveRateGroup.cpp
```diff
@@ -87,12 +96,24 @@ void PassiveRateGroup::CycleIn_handler(FwIndexType portNum, Os::RawTime& cycleSt
...
+    }
+    this->m_maxTime = 0;
+    this->m_cycles = 0;
```
LeStarch, line 114: I am ok with clearing the maxTime, but I question clearing thw m_cycle variable.

LeStarch (changes requested):
Also note, the SDD is now out of sync

Svc/PassiveRateGroup/PassiveRateGroup.cpp
bdshenker: The philosophy behind clearing cycles was to reset all statistical counters at once for easy correlation, but we’re not committed to this channel, so I will implement your reommendation.
Notice: Due to schedule constraints and to avoid redundant effort, this pull request (PR) will be subsumed by "Update in support of metrics collection optimization". Your recommendation will be implemented in that PR. Once that PR is approved and merged, we can close and delete this one. Let me know if you have any concerns with this approach.

Svc/PassiveRateGroup/PassiveRateGroup.cpp
bdshenker: Good point. Unfortunately I cannot use guarded port recommendation since PassiveRategroup is a passive component. I will use mutex.
@LeStarch SDD is indeed out of sync: command and 2 new requirements. Need to add new U/Ts as well.

bdshenker: @LeStarch @celskeggs There is a tradeoff here:
1) Using mutex provides clean solution but adds ~30-60 ticks per cycle execution (RTI) due disabling and reenabling interrupts and a few checks. So cost is small.
2) Leaving it as is and accept/document race condition as benign: collision would result in stale telemetry for at most 2 execution cycles. 
I am planning to implement Mike's advice: using mutex unless I hear concerns.

celskeggs: I think using `guarded` or a mutex is a fine idea since it only has to be acquired once per cycle.

Issue #4676, opened 2026-01-31
Unnecessarily large number of files rebuilt when config file changes


celskeggs:
| | |
|:---|:---|
|**_F´ Version_**| v4.0.1a1-103-g89a9e3247 |
|**_Affected Component_**| build system |
---
## Problem Description

My project uses the fprime-vorago Profiler component, which means we need to override https://github.com/fprime-community/fprime-vorago/blob/main/config/ProfilerCfg.hpp. This configuration file is ONLY included in Profiler.cpp, not anywhere else. When I change it, however, I find that the CMake build scripts get regenerated and hundreds of files in my project need to be rebuilt.

## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-4.18.0-553.89.1.el8_10.x86_64-x86_64-with-glibc2.28
Python version: 3.9.25
CMake version: 3.26.0
Pip version: 25.3
Pip packages:
    fprime-tools==4.0.2a2
    fprime-gds==4.0.2a11.dev3+gc55a33b86
    fprime-fpp==3.1.0a12
Project submodules:
    https://github.com/nasa/fpp.git @ v3.1.0a11-175-ge23d8412b
    https://github.com/nasa/fprime.git @ v4.0.1a1-103-g89a9e3247
    https://github.com/fprime-community/fprime-baremetal @ 90370fb
    https://github.com/fprime-community/fprime-vorago @ 97e3789
```

## How to Reproduce

This can likely be reproduced with any configuration file. If not, I can provide more specific replication instructions.

## Expected Behavior

I expect that changing a header file only used in one CPP file only requires that CPP file to be rebuilt, and any dependent binaries re-linked. I don't expect other unrelated files to need to be rebuilt.

kadircanyildirm-crypto:
Reproduced on `devel` @ `264bb83`, though the trigger is broader than editing a config file: **any real CMake re-configure** recompiles the config module's downstream objects — but only in a project using `CONFIGURATION_OVERRIDES`.

I touched only `CMakeLists.txt` in two in-tree projects. No config file was edited, and each had a clean no-op rebuild immediately before:

| Project | `CONFIGURATION_OVERRIDES` | Objects recompiled |
| --- | --- | --- |
| `TestDeploymentsProject/Ref` | no | **0** |
| `cmake/test/data/TestConfigDeployment` | yes | **14** |

The re-configure re-runs the FPP autocoder over the whole `default_config` module — 43 files in one step:

```
[...]

Issue #3929, opened 2025-07-24
Additional CI: Modification of Included Files Causes a Redetect


LeStarch:
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Often users must purge the build cache to pick up changes to support files (e.g. .fppi).  We could test for this.

PR #5576, opened 2026-08-04
Stop rewriting overridden configuration files on every configure

kadircanyildirm-crypto:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4676 |

---
## Change Description

An overridden configuration destination is written twice per CMake run: once by the module that supplies the original file, once by the module that supplies the override. Each of them writes while the file holds the other one's content, so `ONLY_IF_DIFFERENT` can never skip either copy. The destination's timestamp moves on every configure, and everything including it is rebuilt.

Both paths now record which source should end up at the destination instead of copying straight away. A single pass, deferred to the end of the top-level directory, writes each destination once, at a point where every override is known. A destination that already holds the winning content is left alone.

## Rationale

The write cannot simply be skipped by whoever loses. When the original is processed its overrides have not been registered yet, and they cannot be: an override is rejected outright if the file it overrides has not been seen, so the module supplying the original is always processed first. Any decision made at that moment relies on the previous configure, which cannot tell "the override is still there" apart from "the override was removed". Each variant I worked through along those lines would leave a removed override stale in the build cache, which is the same class of problem as the one being fixed. Deferring the write until all modules are registered is what actually removes the ambiguity.

Worth noting because I got it wrong at first: this is two writers per **CMake run**, not per build. The `info-cache` sub-build is a separate CMake process that also materialises configuration into the main build cache, and it runs before the main configure does its own pass. So an overridden destination is written four times per top-level configure, not two. Both processes run this same code, so the single fix covers both.

On `devel` @ `a24a4c8`, `TestConfigDeployment`, which declares four overrides:

| | before | after |
| --- | --- | --- |
| Configuration destinations whose timestamp moves on a no-op re-configure | 4 | **0** |
| The same, after running the `info-cache` sub-build by hand | 4 | **0** |
| Objects recompiled by a no-op re-configure | **15 of 16** | **0** |

`TestDeployment`, which declares no overrides, moves nothing either way — the four that move map one-to-one onto the four declared overrides.

`test_override_survives_reconfigure` is the regression test. It fails on `devel` with only the test applied, which is the run I would check first — #3933 fixed a nearby build-system bug without one, and #3929 had to be opened to add it afterwards.

| | result |
| --- | --- |
| `test_config.py` with this change | 8 passed (was 7) |
| `test_config.py` with the fix reverted, test kept | the new test fails |
| `pytest -s` from `cmake/test`, the command the CI job runs | 43 passed (17m04s) |

Removing an override is worth reviewing specifically, since that is where the simpler fixes break. Dropping `DpCfg.hpp` from the project's `CONFIGURATION_OVERRIDES` and re-configuring restores F Prime's original file byte for byte, because the recorded map is rebuilt from scratch on every configure.

Formatting checked with `black==21.6b0`, the version `python-format.yml` pins.

## Future Work

`cmake_minimum_required` in the top-level `CMakeLists.txt` says 3.18, but `config_assembler.cmake` already used `file(COPY_FILE)` (3.21) and `cmake_path()` (3.20) before this change, so the real floor was already higher than the declared one. `cmake_language(DEFER)` used here needs 3.19 and is therefore covered. The stale declaration seems worth its own issue rather than a change smuggled in here.

I used Claude Code. It helped me trace where the duplicate write comes from, wrote this change and the test, and drafted this description. The work is limited to `cmake/config_assembler.cmake` and the configuration test module.

The output was heavily modified rather than used as-is. The first designs had the losing writer skip its copy; working through them showed each would leave a removed override stale in the build cache, so they were dropped in favour of deferring the write. Every number above comes from a run I did myself.


LeStarch (dismissed):

LeStarch:
@kadircanyildirm-crypto would you address the merge conflict?

cmake/test/src/test_config.py
```diff
@@ -87,6 +87,27 @@ def test_library_new_config(CONFIG_BUILD):
...
+    """Test that a re-configure does not disturb an overridden configuration file
+
+    An overridden destination has two candidate sources, the original and the
+    override. Writing both leaves the file rewritten on every configure, which
+    moves its timestamp and rebuilds everything including it.
+    """
+    overridden = CONFIG_BUILD["build"] / "F-Prime" / "default" / "config" / "DpCfg.hpp"
+    assert overridden.exists(), "Override destination was never produced"
+    before = overridden.stat().st_mtime
+    contents = overridden.read_bytes()
```
lestarch-autobot, line 111: [Test Quality] **suggestion** Baseline snapshot does not verify the override actually won the destination.

`contents` is whatever the fixture left at the destination. If a regression let the original source win from the first configure, this test would assert stability of the wrong content and still pass. Comparing against the override source makes the test self-contained; today only the compile check in `TestHeaderOverride` carries that burden.

```suggestion
    contents = overridden.read_bytes()
    override_source = (
        CONFIG_BUILD["source"] / "override" / "project" / "DpCfg.hpp"
    )
    assert (
        contents == override_source.read_bytes()
    ), "Destination does not hold the override's content"
```

cmake/test/src/test_config.py
```diff
@@ -87,6 +87,27 @@ def test_library_new_config(CONFIG_BUILD):
     cmake.assert_process_success(CONFIG_BUILD, targets=["TestLibraryNewConfig"])
 
 
+def test_override_survives_reconfigure(CONFIG_BUILD):
```
lestarch-autobot, line 101: [Test Quality] **must fix** Override-removal path has no automated regression test.

The PR description identifies removing an override as exactly the case where the simpler fixes break (a removed override left stale in the build cache), yet it was only verified manually. A future regression there would silently ship stale configuration content. Add a test that drops `DpCfg.hpp` from the project's `CONFIGURATION_OVERRIDES`, re-configures, and asserts the destination reverts to F Prime's original file byte for byte.

cmake/config_assembler.cmake
```diff
@@ -89,22 +89,94 @@ function(fprime__internal_process_configuration_source_set MODULE_NAME SOURCE_SE
...
+    list(LENGTH RECORDED_DESTINATIONS DESTINATION_COUNT)
+    if (DESTINATION_COUNT EQUAL 0)
+        return()
+    endif()
+    math(EXPR LAST_INDEX "${DESTINATION_COUNT} - 1")
+    foreach(DESTINATION_INDEX RANGE ${LAST_INDEX})
+        list(GET RECORDED_DESTINATIONS ${DESTINATION_INDEX} DESTINATION)
+        list(GET RECORDED_SOURCES ${DESTINATION_INDEX} SOURCE)
+        file(COPY_FILE "${SOURCE}" "${DESTINATION}" ONLY_IF_DIFFERENT)
+    endforeach()
```
lestarch-autobot: [Maintainability] **suggestion** `maint-clever-code`: manual index bookkeeping to walk two parallel lists.

The LENGTH/guard/math/RANGE/GET dance exists only to keep the two lists index-aligned by hand; `foreach(... IN ZIP_LISTS ...)` (CMake 3.17, below the 3.21 floor `file(COPY_FILE)` already requires) expresses the pairing directly and handles the empty case without the early return.

```suggestion
    foreach(DESTINATION SOURCE IN ZIP_LISTS RECORDED_DESTINATIONS RECORDED_SOURCES)
        file(COPY_FILE "${SOURCE}" "${DESTINATION}" ONLY_IF_DIFFERENT)
    endforeach()
```

cmake/config_assembler.cmake
```diff
@@ -89,22 +89,94 @@ function(fprime__internal_process_configuration_source_set MODULE_NAME SOURCE_SE
...
+    list(FIND RECORDED_DESTINATIONS "${DESTINATION}" DESTINATION_INDEX)
+    if (DESTINATION_INDEX EQUAL -1)
+        list(APPEND RECORDED_DESTINATIONS "${DESTINATION}")
+        list(APPEND RECORDED_SOURCES "${SOURCE}")
+    else()
+        list(REMOVE_AT RECORDED_SOURCES ${DESTINATION_INDEX})
+        list(INSERT RECORDED_SOURCES ${DESTINATION_INDEX} "${SOURCE}")
+    endif()
+    set_property(GLOBAL PROPERTY FPRIME_CONFIG_DESTINATIONS "${RECORDED_DESTINATIONS}")
+    set_property(GLOBAL PROPERTY FPRIME_CONFIG_SOURCES "${RECORDED_SOURCES}")
```
lestarch-autobot, line 189: [Maintainability] **could fix** `maint-structural-hazard`: two parallel global lists whose index alignment is the invariant.

`FPRIME_CONFIG_DESTINATIONS`/`FPRIME_CONFIG_SOURCES` only work while every edit touches both at the same index; a future change that appends/removes one without the other silently maps destinations to the wrong sources. A keyed property per destination (e.g. `set_property(GLOBAL PROPERTY "FPRIME_CONFIG_SOURCE_${DESTINATION}" "${SOURCE}")` plus a single destinations list) removes the FIND/REMOVE_AT/INSERT dance and the alignment invariant entirely.

cmake/config_assembler.cmake
```diff
@@ -89,22 +89,94 @@ function(fprime__internal_process_configuration_source_set MODULE_NAME SOURCE_SE
...
+#
+# Records the source that should end up at a configuration destination. The last source recorded for
+# a destination wins, which is the overriding one: a module supplying a CONFIGURATION_OVERRIDE is
+# always processed after the module supplying the file it overrides, since the override is rejected
+# outright when that file has not been registered yet.
+#
+# Copying here rather than at the flush would write an overridden destination twice per CMake run,
+# once with the original and once with the override. Neither write can be skipped by
+# `ONLY_IF_DIFFERENT` because the file holds the other one's content, so the destination's timestamp
+# moves on every configure and everything including it is rebuilt.
```
lestarch-autobot, line 167: [Maintainability] **could fix** `maint-misleading-comment`: multi-paragraph counterfactual rationale in the function header.

This paragraph describes what the code would do under a rejected design, not what it does; it will silently go stale if the flush strategy changes, and the same rationale already lives in the PR/commit message. One line ("copies are deferred so an overridden destination is written once, by the winning source") carries the needed context.

cmake/config_assembler.cmake
```diff
@@ -89,22 +89,94 @@ function(fprime__internal_process_configuration_source_set MODULE_NAME SOURCE_SE
...
+    if (DESTINATION_COUNT EQUAL 0)
+        return()
+    endif()
+    math(EXPR LAST_INDEX "${DESTINATION_COUNT} - 1")
+    foreach(DESTINATION_INDEX RANGE ${LAST_INDEX})
+        list(GET RECORDED_DESTINATIONS ${DESTINATION_INDEX} DESTINATION)
+        list(GET RECORDED_SOURCES ${DESTINATION_INDEX} SOURCE)
+        file(COPY_FILE "${SOURCE}" "${DESTINATION}" ONLY_IF_DIFFERENT)
+    endforeach()
+endfunction()
```
lestarch-autobot, line 219: (Anchored above the offending lines; the diff does not include lines 180-192.)

[Maintainability] **future work** `maint-misleading-comment`: header of `fprime_internal_get_configuration_destination` documents a `CONFIG_NAME` argument the function does not take (actual: `MODULE_NAME`, `NEW_CONFIG_NAME`) and omits the `DESTINATION_MODULE`/`DESTINATION_OVERRIDE`/`DESTINATION_DIRECTORY` values it sets in the caller. Preexisting, but this PR's new code now depends on those undocumented outputs, so the next editor will read a wrong contract.

cmake/test/src/test_config.py
lestarch-autobot, line 101: (Anchored on the new test; the diff does not include line 2.)

[Maintainability] **future work** `maint-misleading-comment`: the module header still reads `test_basic.py: Basic CMake tests` but the file is `test_config.py` holding configuration tests. Preexisting; a reader grepping for the basic tests lands here.

cmake/config_assembler.cmake
```diff
@@ -89,22 +89,94 @@ function(fprime__internal_process_configuration_source_set MODULE_NAME SOURCE_SE
...
+    if (NOT EXISTS "${DESTINATION}")
+        file(COPY_FILE "${SOURCE}" "${DESTINATION}")
+    endif()
+
+    # Scheduled once. Deferring to the top-level directory means every configuration module has been
+    # processed, and every override therefore known, by the time destinations are written.
+    get_property(FLUSH_SCHEDULED GLOBAL PROPERTY FPRIME_CONFIG_FLUSH_SCHEDULED)
+    if (NOT FLUSH_SCHEDULED)
+        set_property(GLOBAL PROPERTY FPRIME_CONFIG_FLUSH_SCHEDULED TRUE)
+        cmake_language(DEFER DIRECTORY "${CMAKE_SOURCE_DIR}"
```
lestarch-autobot, line 200: [Design] **could fix** Deferred flush leaves the destination holding the losing/old content for the rest of the configure pass.

On a fresh configure (or when a source changes), the destination keeps the original/previous content until the end-of-top-level flush. This is safe only while nothing reads configuration *content* at configure time; I traced the info-cache sub-build (its fpp targets run at build phase, after its own flush) and the include-path check (paths only), and found no configure-time content reader — but I could not verify this exhaustively. Consider stating this invariant in the function comment so a future configure-time reader does not silently see the losing content.

lestarch-autobot, line 200: [Design] **future work** Declared CMake minimum (3.18) does not cover this file's real floor.

`cmake_language(DEFER)` needs 3.19, and this file already used `file(COPY_FILE)` (3.21) and `cmake_path()` (3.20) before this PR, so the effective floor was already above the declared `cmake_minimum_required(VERSION 3.18)`. Preexisting, not widened here; the PR description already proposes a separate issue — recording so it is tracked.

cmake/config_assembler.cmake
```diff
@@ -89,22 +89,94 @@ function(fprime__internal_process_configuration_source_set MODULE_NAME SOURCE_SE
...
+
+    list(LENGTH RECORDED_DESTINATIONS DESTINATION_COUNT)
+    if (DESTINATION_COUNT EQUAL 0)
+        return()
+    endif()
+    math(EXPR LAST_INDEX "${DESTINATION_COUNT} - 1")
+    foreach(DESTINATION_INDEX RANGE ${LAST_INDEX})
+        list(GET RECORDED_DESTINATIONS ${DESTINATION_INDEX} DESTINATION)
+        list(GET RECORDED_SOURCES ${DESTINATION_INDEX} SOURCE)
+        file(COPY_FILE "${SOURCE}" "${DESTINATION}" ONLY_IF_DIFFERENT)
```
lestarch-autobot, line 217: [Documentation] **could fix** Build-system docs do not describe the new deferred single-write behavior for overridden configuration files.

`docs/user-manual/framework/configuring-fprime.md` and `docs/user-manual/build-system/` describe `CONFIGURATION_OVERRIDES` but never say when overridden destinations are materialized; this PR changes that to a single deferred write at the end of the top-level configure, and unchanged destinations are no longer rewritten. A short note in `configuring-fprime.md` would explain why no-op re-configures no longer rebuild dependents. The generated CMake API reference picks up the new `####` doc blocks automatically, so no action is needed there.

cmake/config_assembler.cmake
lestarch-autobot, line 200: [Documentation] **future work** Build-system SDD (`cmake/docs/sdd.md`) does not cover the configuration-assembly/override mechanism this PR redesigns.

The SDD predates `config_assembler.cmake` and never described override handling, so the gap is preexisting, not introduced here. The new deferred-flush design (global-property source map plus `cmake_language(DEFER)` at the top-level directory) is now load-bearing build-system behavior worth capturing when the SDD is next updated.

cmake/config_assembler.cmake
```diff
@@ -89,22 +89,94 @@ function(fprime__internal_process_configuration_source_set MODULE_NAME SOURCE_SE
...
+        list(APPEND RECORDED_SOURCES "${SOURCE}")
+    else()
+        list(REMOVE_AT RECORDED_SOURCES ${DESTINATION_INDEX})
+        list(INSERT RECORDED_SOURCES ${DESTINATION_INDEX} "${SOURCE}")
+    endif()
+    set_property(GLOBAL PROPERTY FPRIME_CONFIG_DESTINATIONS "${RECORDED_DESTINATIONS}")
+    set_property(GLOBAL PROPERTY FPRIME_CONFIG_SOURCES "${RECORDED_SOURCES}")
+
+    if (NOT EXISTS "${DESTINATION}")
+        file(COPY_FILE "${SOURCE}" "${DESTINATION}")
```
lestarch-autobot, line 192: [Correctness] **could fix** `correctness-other`: overridden destination holds the original (or previous configure's) content until the deferred flush runs.

Trigger: any configure-time logic that reads an overridden destination's bytes between module registration and the end of top-level directory processing (e.g. a project `CMakeLists.txt` doing `file(READ)`/`configure_file` on a config header). Before this PR the override content landed at registration time; now it lands only at the flush. No in-repo configure-time reader of these files exists today, so the consequence is only reachable from project code outside this repository.

cmake/config_assembler.cmake
lestarch-autobot, line 200: [Correctness] **future work** `correctness-other`: `cmake_language(DEFER)` requires CMake 3.19, but the top-level `cmake_minimum_required` still declares 3.18.

Trigger: configuring with CMake 3.18 (the declared minimum). Consequence: configure aborts with an unknown-subcommand error instead of a clear version message. The effective floor was already 3.21 before this PR (`file(COPY_FILE)`, `cmake_path()` in this same file), so this PR does not widen it — flagged as preexisting per the PR's own Future Work note; the declared minimum should be raised in a follow-up.

lestarch-autobot (changes requested):

### Per-agent results

| Documentation Currency | 0 | 0 | 1 | 1 | 2 | Go |
| Design | 0 | 0 | 1 | 1 | 2 | Go |
| Test Quality | 1 | 1 | 0 | 0 | 2 | No-Go |
| Correctness | 0 | 0 | 1 | 1 | 2 | Go |
| Maintainability | 0 | 1 | 2 | 2 | 5 | Go |
| **Totals** | 1 | 2 | 5 | 5 | 13 | **No-Go** |

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>

**Test Quality**
- Override-removal path has no automated regression test — https://github.com/nasa/fprime/pull/5576#discussion_r3752010955

</details>

### Merge readiness
**Merge readiness: No-Go** — Test Quality has 1 outstanding must-fix item (missing override-removal regression test).

- None — all registered reviewers ran.

---

All stations report in; one item on the checklist before this build is go for launch.

LeStarch (dismissed):

kadircanyildirm-crypto:
Rebased onto `devel`. The conflict was additive — #5589's new function and mine landed at the same spot; both are kept and theirs is unchanged. `cmake/test/src/test_config.py` passes 9/9.

LeStarch:
@kadircanyildirm-crypto would you format python and consider the missing test?

LeStarch (approved):

kadircanyildirm-crypto:
All four red checks are external-repo jobs, and none of them traces back to this change.

- **fprime-examples** (`run / Build`, `run / Unit Tests`) — `FlightExamples/CustomFraming/DecafDeframer/DecafDeframer.cpp:101` calls `data.advance(..)`. `Fw::Buffer::advance` landed on `devel` in ceee420e on Aug 11; this branch's base is Aug 10, so the method does not exist here yet.
- **YAMCS Reference** (`Test`) — `test_cdh_core.py:77`, `assert 81 == (79 + 1)`. The same assertion fails on `devel` itself in [run 31506686144](https://github.com/nasa/fprime/actions/runs/31506686144) (26d4611c), which does not carry this change.
- **Zephyr** — fails inside the `get_pr_branch` setup action with `curl: (60) SSL certificate problem: self-signed certificate`, before anything is configured or built.

The branch is 24 commits behind `devel` — happy to rebase if you would like a fresh run.

*AI use per `AI_POLICY.md`. Tool: Claude Code. Type of assistance: reading the four failing job logs and drafting this comment. Level of modification: the run IDs, file paths and line numbers above are quoted from those logs and checked against the repositories.*

LeStarch (approved):

Issue #5509, opened 2026-07-27
Refactor LinearBufferBase


bevinduckett:
| | |
|:---|:---|
|**_F´ Version_**| devel |
|**_Affected Component_**|  LinearBufferBase |
---
## Feature Description

1. Change the getBuffAddr() & getCapacity() methods of the LinearBufferBase class from virtual to concrete
2. Add a SERIALIZABLE_INLINE macro to  default/config/FpConfig.h to allow projects to select whether (and which) options should be used to request inlining of the serializeFrom / deserializeTo function

This will also require an FPP update (to change the PortBuffers created by fpp-to-cpp)

## Rationale

Serialization is done everything in F' - currently a serializeFrom(U32) takes 137 cycles longer than a memcpy to store 4 bytes. Repeating that throughout the system has significant performance impacts. 

Implementing changes 1 & 2 (with the macro set to `__attribute__((always_inline))`) in a deployment built using[ llvm-vorago-arm-toolchain](https://github.com/fprime-community/llvm-vorago-arm-toolchain) with `-Os` enabled resulted bringing that difference down to 3 to 15 cycles (depending on the test approach used) with a 10% increase in image size (without `__attribute__((always_inline))`, the performance improvement was ~ 45% and the image size decreased)

bevinduckett:
Alternative mechanism for improving serializeFrom() performance were considered, with results shown below 

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns="http://www.w3.org/TR/REC-html40"
xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=PowerPoint.Slide>
<meta name=Generator content="Microsoft PowerPoint 15">
<style>

</style>
</head>

<body>



  |   | Redundant Calls Refactor | Devirtualization Refactor | Inline serializeFrom | Inline getBuffAddr & getCapacity | % Duration Decreased | % Binary Increased
-- | -- | -- | -- | -- | -- | -- | --
1 | vorago-serialize-fixes | x |   |   |   | 23% | 0%
2 | vorago-serialize-always-inline |   |   | x |   | 62% | 25%
3 | vorago-serialize-fixes-inline | x |   | x |   | 63% | 20%
4 | vorago-serialize-fixes-inline-v2 | x |   | x | x | 63% | 20%
[...]

PR #5584, opened 2026-08-04
Refactor LinearBufferBase  to remove virtual functions

bevinduckett:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/5509  |
|**_Documentation Included (y/n)_**| n/a  |
|**_Generative AI was used in this contribution (y/n)_**| yes  |

---
## Feature Description

1. Change the getBuffAddr() & getCapacity() methods of the LinearBufferBase class from virtual to concrete
2. Add a SERIALIZABLE_INLINE macro to  default/config/FpConfig.h to allow projects to select whether (and which) options should be used to request inlining of the serializeFrom / deserializeTo function
3. Modified the constructor for the LinearBufferBase to require buffAddr & capacity be provided (to make failure faster if a developer forgets  

This also requires an FPP update.

## Rationale

Serialization is done everything in F' - currently a serializeFrom(U32) takes 137 cycles longer than a memcpy to store 4 bytes. Repeating that throughout the system has significant performance impacts. 

Implementing changes 1 & 2 (with the macro set to `__attribute__((always_inline))`) in a deployment built using[ llvm-vorago-arm-toolchain](https://github.com/fprime-community/llvm-vorago-arm-toolchain) with `-Os` enabled resulted bringing that difference down to 7 to 15 cycles (depending on the test approach used) with a 9% increase in image size (without `__attribute__((always_inline))`, the performance improvement was ~ 45% and the image size decreased)

__NOTE__: This PR cannot be tested or merged without https://github.com/nasa/fpp/pull/1056

AI was used to assist with the changes (particularly to the classes that inherited from LinearBufferBase) and to review & propose improvements


bocchino:
@bevinduckett It looks like the file `fprime/FppTestProject/FppTest/component/tests/PortBufferTests.cpp` is missing from the branch. When I try to build against the updated FPP version, I get a build failure on this file.

bocchino:
When I remove that file from the build, I can generate the build, but then (on Mac OS/clang) I see compiler errors due to narrowing integer casts in assertions. Can you make sure that this branch builds against the updated FPP version? You can use the procedure described here: https://github.com/nasa/fpp/wiki/Adding-New-Features-to-FPP#building-f-prime-with-an-unreleased-version-of-fpp.

bevinduckett:
@bocchino - after the updates, all steps in  https://github.com/nasa/fpp/wiki/Adding-New-Features-to-FPP#building-f-prime-with-an-unreleased-version-of-fpp work for me after checking out and building [fprime-issue-5509-serial-buffer](https://github.com/nasa/fpp/tree/fprime-issue-5509-serial-buffer)

Fw/Types/Serializable.cpp
```diff
@@ -298,54 +327,58 @@ SerializeStatus LinearBufferBase::serializeSize(const FwSizeType size, Endiannes
...
 
-SerializeStatus LinearBufferBase::deserializeTo(U8& val, Endianness mode) {
+FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::deserializeTo(U8& val, Endianness mode) {
     // check for room
-    if (this->getSize() == this->m_deserLoc) {
+    const Serializable::SizeType size = this->m_serLoc;
+    if (size == this->m_deserLoc) {
         return FW_DESERIALIZE_BUFFER_EMPTY;
-    } else if (this->getSize() - this->m_deserLoc < static_cast<Serializable::SizeType>(sizeof(val))) {
+    } else if (size - this->m_deserLoc < static_cast<Serializable::SizeType>(sizeof(val))) {
```
bocchino, line 323: We should assert that `size >= this->m_deserLoc`. Otherwise there could be integer overflow.

FppTestProject/FppTest/struct/PrimitiveTest.cpp
```diff
@@ -35,20 +35,20 @@ class PrimitiveTest : public ::testing::Test {
         ASSERT_EQ(s.get_mI16(), testI16);
         ASSERT_EQ(s.get_mF64(), testF64);
     }
-
     void assertUnsuccessfulSerialization(T& s, U32 bufSize) {
-        // Avoid creating an array of size zero
+        // Array padded by 1 to bypass the LinearBufferBase check that
```
bocchino: I'm not sure what's going on here. Is this a new check added to LinearBufferBase? What is the motivation? Also, it looks like the logic on line 42 has been updated to avoid this issue in a different way. So is the padding necessary?

bevinduckett: It's a new check, the motivation is that now LinearBufferBase is relying on children classes to set/update m_buffAddr & m_capacity to valid values at the correct times - so LinearBufferBase's constructor checks that capacity is 0 **and** buffAddr is a nullptr (i.e. not a valid buffer, used for ports with no arguments)  **or** that both capacity and buffAddr indicate a valid buffer 

Which required this update

bocchino: Thanks. Now I see that the `bufSize + 1` was already there. I think that was done to avoid a size-zero array, which C++ does not like. So why was the comment updated from `Avoid creating an array of size zero` (which still seems true) to `Array padded by 1 to bypass the LinearBufferBase check ...` which does not seem correct? It seems like we satisfy the LinearBufferBase check in line 42, by passing in a null pointer if and only if bufSize is 0. This seems unrelated to the allocated size of the array, which is still one bigger than it needs to be to avoid ever having a size zero array.

bevinduckett: you're correct, I put the comment in the wrong place, I'll fix that

Fw/Types/Serializable.cpp
bevinduckett: added size check for all de-serialize functions

FppTestProject/FppTest/struct/PrimitiveTest.cpp
bevinduckett: fixed

bocchino: Thank you!

Fw/Types/Serializable.cpp
bevinduckett: the first fix introduced a UT failure, that's now been fixed

Fw/Types/Serializable.cpp
```diff
@@ -328,49 +328,52 @@ FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::serializeSize(co
 // deserialization routines
 
 FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::deserializeTo(U8& val, Endianness mode) {
-    // check for room
+    U8* buffAddr = this->m_buffAddr;
     const Serializable::SizeType size = this->m_serLoc;
+    FW_ASSERT(buffAddr != nullptr && size >= this->m_deserLoc, static_cast<FwAssertArgType>(size),
```
bocchino: Let's split this compound assertion into two assertions.

Fw/Types/Serializable.cpp
```diff
@@ -328,49 +328,52 @@ FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::serializeSize(co
...
     val = buffAddr[this->m_deserLoc + 0];
     this->m_deserLoc += static_cast<Serializable::SizeType>(sizeof(val));
     return FW_SERIALIZE_OK;
 }
 
 FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::deserializeTo(I8& val, Endianness mode) {
-    // check for room
+    U8* buffAddr = this->m_buffAddr;
     const Serializable::SizeType size = this->m_serLoc;
+    FW_ASSERT(buffAddr != nullptr && size >= this->m_deserLoc, static_cast<FwAssertArgType>(size),
```
bocchino: Split into two assertions.

Fw/Types/Serializable.cpp
```diff
@@ -445,15 +449,16 @@ FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::deserializeTo(I3
 #if FW_HAS_64_BIT == 1
 
 FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::deserializeTo(U64& val, Endianness mode) {
+    U8* buffAddr = this->m_buffAddr;
+    FW_ASSERT(buffAddr != nullptr && this->m_serLoc >= this->m_deserLoc, static_cast<FwAssertArgType>(this->m_serLoc),
```
bocchino: Split into two assertions.

Fw/Types/Serializable.cpp
```diff
@@ -509,16 +514,17 @@ FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::deserializeTo(F6
 }
 
 FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::deserializeTo(bool& val, Endianness mode) {
-    // check for room
+    U8* buffAddr = this->m_buffAddr;
     const Serializable::SizeType size = this->m_serLoc;
+    FW_ASSERT(buffAddr != nullptr && size >= this->m_deserLoc, static_cast<FwAssertArgType>(size),
```
bocchino: Split into two assertions.

Fw/Types/Serializable.cpp
bocchino: Split into two assertions.

Fw/Types/Serializable.cpp
```diff
@@ -328,49 +328,52 @@ FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::serializeSize(co
...
     this->m_deserLoc += static_cast<Serializable::SizeType>(sizeof(val));
     return FW_SERIALIZE_OK;
 }
 
 #if FW_HAS_16_BIT == 1
 FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::deserializeTo(U16& val, Endianness mode) {
-    // check for room
+    U8* buffAddr = this->m_buffAddr;
     const Serializable::SizeType size = this->m_serLoc;
+    FW_ASSERT(buffAddr != nullptr && size >= this->m_deserLoc, static_cast<FwAssertArgType>(size),
```
bocchino: Split into two assertions.

Fw/Types/Serializable.cpp
```diff
@@ -399,16 +402,17 @@ FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::deserializeTo(I1
 #endif
 #if FW_HAS_32_BIT == 1
 FW_SERIALIZE_FORCE_INLINE_LLB SerializeStatus LinearBufferBase::deserializeTo(U32& val, Endianness mode) {
-    // check for room
+    U8* buffAddr = this->m_buffAddr;
     const Serializable::SizeType size = this->m_serLoc;
+    FW_ASSERT(buffAddr != nullptr && size >= this->m_deserLoc, static_cast<FwAssertArgType>(size),
```
bocchino: Split into two assertions.

Fw/Types/Serializable.cpp
bevinduckett: see commit 130068d039d8b8a5c3c852552dc442996122ddc9

Fw/Types/Serializable.cpp
bevinduckett: see commit 130068d039d8b8a5c3c852552dc442996122ddc9

Fw/Types/Serializable.cpp
bevinduckett: see commit 130068d039d8b8a5c3c852552dc442996122ddc9

Fw/Types/Serializable.cpp
bevinduckett: see commit 130068d039d8b8a5c3c852552dc442996122ddc9

Fw/Types/Serializable.cpp
bevinduckett: see commit 130068d039d8b8a5c3c852552dc442996122ddc9

Fw/Types/Serializable.cpp
bevinduckett: see commit 130068d039d8b8a5c3c852552dc442996122ddc9

Fw/Types/Serializable.cpp
bevinduckett: see commit 130068d039d8b8a5c3c852552dc442996122ddc9

Fw/Types/Serializable.cpp
bocchino: Thanks, I was looking at the wrong version.

default/config/FpConfig.h
```diff
@@ -149,6 +149,16 @@ extern "C" {
 #define POSIX_THREADS_ENABLE_NAMES (1)  //!< Enable/Disable assigning names to threads
 #endif
 
+// Hint to the compiler to always inline LinearBufferBase serialization &
+// deserialization methods
+#define FW_SERIALIZE_FORCE_INLINE_LLB
```
bocchino: Should this macro be spelled LBB (for LinearBufferBase) instead of LLB?

bevinduckett: yes - will fix

bevinduckett: fixed

bocchino:
@bevinduckett There are some code analysis warnings. Can you take a look? They may all be instances of the same coding issue in a template.

bevinduckett:
> @bevinduckett There are some code analysis warnings. Can you take a look? They may all be instances of the same coding issue in a template.

@bocchino  - Those warnings are because the copy constructor of LinearBufferTemplate sets the member variables via setBuff rather than directly copying them. I've added a comment & a suppression

bevinduckett:
@bocchino  - The Zephyr CI checks are failing because the FpConfig.h file needs to be updated, see https://github.com/fprime-community/fprime-zephyr-reference/pull/39

bevinduckett:
@LeStarch / @bocchino 
All checks save two are passing / skipped: 
The [External Repo: Zephyr Reference / Zephyr Build (all boards) (pull_request)](https://github.com/nasa/fprime/actions/runs/31550950683/job/93982062764?pr=5584)  check failed because https://github.com/fprime-community/fprime-zephyr-reference/pull/41 hasn't been merged 
The [External Repo: YAMCS Reference / Build (pull_request)](https://github.com/nasa/fprime/actions/runs/31550950432/job/93973352117?pr=5584) check failed due to the CI issue tracked in https://github.com/nasa/fprime/issues/5587

bocchino:
> The [External Repo: Zephyr Reference / Zephyr Build (all boards) (pull_request)](https://github.com/nasa/fprime/actions/runs/31550950683/job/93982062764?pr=5584) check failed because https://github.com/fprime-community/fprime-zephyr-reference/pull/41 hasn't been merged

It's merged now, and I started a new CI run.

bocchino (approved):
Looks great! Thanks for all the hard work on this PR.

thomas-bc:
## Release notes

see https://github.com/fprime-community/fprime-zephyr-reference/pull/40/changes

LeStarch (approved):

Issue #5591, opened 2026-08-05
Drv::UdpSocket: configured send timeout is discarded when send port is 0 (respond-to-sender mode)


nurdymuny:
| | |
|:---|:---|
|**_F´ Version_**| devel @ cb17b6c |
|**_Affected Component_**| Drv::UdpSocket |
---
## Problem Description

`UdpSocket::openProtocol` applies the configured send timeout inside `if (port != 0)` (`Drv/Ip/UdpSocket.cpp:142`, `:166`). In the documented respond-to-sender configuration the send port *is* 0, so `setupTimeouts` never runs and `SO_SNDTIMEO` is never set — on a socket that will go on to send. The timeout the caller passed to `configureSend` is silently discarded.

`Drv/Ip/docs/sdd.md:179-184` documents passing 0 to `configureSend` as the supported way to respond to the source port of received datagrams, and `configureSend` accepts `timeout_seconds` / `timeout_microseconds`. `IpSocket::configure` stores them (`Drv/Ip/IpSocket.cpp:61-62`); `openProtocol` never applies them on this path.

The in-tree test exercises exactly this today: `Drv/Ip/test/ut/TestUdp.cpp:150` configures the receiver with `configureSend("127.0.0.1", 0, 0, 100)`, and that socket sends at line 174. The requested 100 µs send timeout is not on the file descriptor.

Measured with `getsockopt(SO_SNDTIMEO)` on the fd returned by `open()`:

| configuration | requested | applied |
|:---|:---|:---|
| `configureSend(addr, 60211, 1, 0)` + `configureRecv` | 1.000000 s | 1.000000 s |
| `configureSend(addr, 0, 1, 0)` + `configureRecv` | 1.000000 s | 0.000000 s |

`0` means block indefinitely, so a send on a wedged socket blocks the calling task rather than returning after the configured timeout.

## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
Python version: 3.12.3
CMake version: 3.26.0
Pip version: 26.2
Pip packages:
    fprime-tools==4.3.0a5
    fprime-gds==4.2.2a3
    fprime-fpp==3.3.0a12
Project submodules:
```

## How to Reproduce

1. Configure a `Drv::UdpSocket` with `configureRecv(addr, port)` and `configureSend(addr, 0, 1, 0)`.
2. Call `open()`.
[...]

PR #5593, opened 2026-08-05
fix(Drv/Ip): apply the UDP send timeout when no send port is configured

tekinertekin:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5591 |

---
## Change Description

`UdpSocket::openProtocol` applied the configured timeouts inside `if (port != 0)`,
so `SO_SNDTIMEO` was never set on a socket configured the respond-to-sender way.
That configuration is documented in `Drv/Ip/docs/sdd.md`: pass a send port of 0 to
`configureSend` and the socket replies to the source of the last datagram it
received. Such a socket does send, and the timeout the caller passed is stored by
`IpSocket::configure` and then dropped.

This moves the `setupTimeouts` call out of that block so it runs for every socket
the class opens, and adds a unit test that reads the value back with `getsockopt`.

`Drv/Ip/UdpSocket.cpp` +9/-6, `Drv/Ip/test/ut/TestUdp.cpp` +27.

## Rationale

`TcpClientSocket` and `TcpServerSocket` both call `setupTimeouts` unconditionally.
The UDP guard was the odd one out, and it conflated two different questions: "is a
send address configured" and "should this socket's options be set". The first has
nothing to do with the second.

Built and run locally on macOS (AppleClang) with `fprime-util build --ut` and
`fprime-util check` in `Drv/Ip`. I also built the unpatched file with the new test
in place, to confirm the test really catches the bug:

    without the fix:  [ FAILED ]  SendTimeout.TestSendTimeoutAppliedWithoutSendPort
                      getsockopt reported tv_sec = 0, configured value was 1
                      8 of 9 UDP tests passed

    with the fix:     [   OK   ]  SendTimeout.TestSendTimeoutAppliedWithoutSendPort
                      9 of 9 UDP tests passed, check exited 0

The `Drv/Ip` TCP tests pass in both cases.

Worth knowing while reviewing: `TestEphemeralPorts` already walks this path.
Line 150 configures the receiver with `configureSend("127.0.0.1", 0, 0, 100)` and
that socket goes on to send a reply, but nothing read `SO_SNDTIMEO` back, so the
suite passed over the bug.

The new test is guarded on `TGT_OS_TYPE_VXWORKS`, matching `setupTimeouts`, which
is a no-op there.

## Future Work

`setupSocketOptions` sits behind the same `if (port != 0)` guard, so `SO_REUSEADDR`
is also skipped when the send port is 0 — and that is the case where the socket is
the one that binds, which is what the option is for. I left it out of this PR
because changing it alters bind behaviour for existing receive-only sockets, and
that felt like your call rather than mine. Happy to fold it in here or open a
separate issue, whichever you prefer.

- **Type of assistance:** code generation, debugging assistance, test authoring
- **Scope:** `Drv/Ip/UdpSocket.cpp` (moving the `setupTimeouts` call out of the
  port guard) and `Drv/Ip/test/ut/TestUdp.cpp` (the new `SendTimeout` test)
- **Tool:** Claude (Anthropic)
- **Level of modification:** reviewed and verified rather than used as-is. I built
  and ran the suite locally, and separately built the unpatched source with the new
  test to confirm it fails without the change.

### DIFF
--- a/Drv/Ip/UdpSocket.cpp
+++ b/Drv/Ip/UdpSocket.cpp
@@ -138,6 +138,15 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
         return SOCK_FAILED_TO_GET_SOCKET;
     }
 
+    // Apply the configured timeouts. This socket may send even when no send port was
+    // configured, because a send port of 0 means "reply to the source of the last
+    // datagram received", so the timeout must not be tied to `port != 0` below.
+    status = this->setupTimeouts(socketFd);
+    if (status != SOCK_SUCCESS) {
+        (void)::close(socketFd);
+        return status;
+    }
+
     // May not be sending in all cases
     if (port != 0) {
         // Set up the address port and name
@@ -162,12 +171,6 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
             return SOCK_FAILED_TO_SET_SOCKET_OPTIONS;
         }
 
-        // Now apply timeouts
-        status = this->setupTimeouts(socketFd);
-        if (status != SOCK_SUCCESS) {
-            (void)::close(socketFd);
-            return status;
-        }
         static_assert(sizeof(m_addr_send) == sizeof(address), "Send address must match the local address structure");
         (void)memcpy(&this->m_addr_send, &address, sizeof(this->m_addr_send));
     }
--- a/Drv/Ip/test/ut/TestUdp.cpp
+++ b/Drv/Ip/test/ut/TestUdp.cpp
@@ -183,6 +183,33 @@ TEST(Ephemeral, TestEphemeralPorts) {
     receiver.close(recv_fd);
 }
 
+#ifndef TGT_OS_TYPE_VXWORKS
+// A send port of 0 means "reply to the source of the last datagram received", so such a
+// socket does send and the configured send timeout has to reach it. It used to be applied
+// only when a non-zero send port was configured.
+TEST(SendTimeout, TestSendTimeoutAppliedWithoutSendPort) {
+    const U32 timeout_seconds = 1;
+    const U32 timeout_microseconds = 0;
+
+    Drv::UdpSocket socket;
+    Drv::SocketDescriptor fd;
+    U16 recv_port = Drv::Test::get_free_port(true);
+    ASSERT_NE(0, recv_port);
+
+    socket.configureRecv("127.0.0.1", recv_port);
+    ASSERT_EQ(socket.configureSend("127.0.0.1", 0, timeout_seconds, timeout_microseconds), Drv::SOCK_SUCCESS);
+    ASSERT_EQ(socket.open(fd), Drv::SOCK_SUCCESS);
+
+    struct timeval timeout = {};
+    socklen_t timeout_size = sizeof(timeout);
+    ASSERT_EQ(::getsockopt(fd.fd, SOL_SOCKET, SO_SNDTIMEO, &timeout, &timeout_size), 0) << std::strerror(errno);
+    EXPECT_EQ(static_cast<U32>(timeout.tv_sec), timeout_seconds);
+    EXPECT_EQ(static_cast<U32>(timeout.tv_usec), timeout_microseconds);
+
+    socket.close(fd);
+}
+#endif
+
 int main(int argc, char** argv) {
     ::testing::InitGoogleTest(&argc, argv);
     return RUN_ALL_TESTS();


LeStarch (approved):

Drv/Ip/UdpSocket.cpp
```diff
@@ -138,6 +138,15 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
...
+    // configured, because a send port of 0 means "reply to the source of the last
+    // datagram received", so the timeout must not be tied to `port != 0` below.
+    status = this->setupTimeouts(socketFd);
+    if (status != SOCK_SUCCESS) {
+        (void)::close(socketFd);
+        return status;
+    }
+
     // May not be sending in all cases
     if (port != 0) {
```
lestarch-autobot, line 151: [Design] **future work** `setupSocketOptions` remains gated by `if (port != 0)`, diverging from the SDD.

(Anchored above the offending line; the diff does not include it.) `Drv/Ip/docs/sdd.md` states `IP_SOCKET_OPTIONS` are "applied to all sockets (TCP server, TCP client, and UDP)", but a receive-only / respond-to-sender UDP socket (`port == 0`) — the case that actually binds — skips them. Preexisting, not introduced by this PR, and the author flags it in the PR's Future Work section; recording so the SDD/code divergence is tracked (either scope the SDD claim to TCP or move the call, per the security tradeoff noted in issue #5591).

Drv/Ip/test/ut/TestUdp.cpp
```diff
@@ -183,6 +183,33 @@ TEST(Ephemeral, TestEphemeralPorts) {
     receiver.close(recv_fd);
 }
 
+#ifndef TGT_OS_TYPE_VXWORKS
+// A send port of 0 means "reply to the source of the last datagram received", so such a
+// socket does send and the configured send timeout has to reach it. It used to be applied
+// only when a non-zero send port was configured.
```
lestarch-autobot, line 189: [Maintainability] **suggestion** `maint-misleading-comment`: the last sentence narrates the bug history, not the code.

"It used to be applied only when a non-zero send port was configured" describes the pre-fix behavior; once merged, no such code exists and the sentence reads as a claim about code the reader cannot find. That history belongs in the commit/PR, not a comment the next engineer must reconcile with the file.

```suggestion
// A send port of 0 means "reply to the source of the last datagram received", so such a
// socket does send and the configured send timeout has to reach it.
```

Drv/Ip/test/ut/TestUdp.cpp
```diff
@@ -183,6 +183,33 @@ TEST(Ephemeral, TestEphemeralPorts) {
     receiver.close(recv_fd);
 }
 
+#ifndef TGT_OS_TYPE_VXWORKS
```
lestarch-autobot, line 186: [Maintainability] **could fix** `maint-inconsistent-local-convention`: unexplained platform exclusion — first OS guard in this file, with no rationale.

No other test in TestUdp.cpp is platform-guarded, and nothing states why VxWorks is excluded (getsockopt/SO_SNDTIMEO readback support?). The next engineer cannot tell whether the exclusion is load-bearing or can be removed when porting; a one-line reason on the guard fixes that.

Drv/Ip/test/ut/TestUdp.cpp
```diff
@@ -183,6 +183,33 @@ TEST(Ephemeral, TestEphemeralPorts) {
...
+TEST(SendTimeout, TestSendTimeoutAppliedWithoutSendPort) {
+    const U32 timeout_seconds = 1;
+    const U32 timeout_microseconds = 0;
+
+    Drv::UdpSocket socket;
+    Drv::SocketDescriptor fd;
+    U16 recv_port = Drv::Test::get_free_port(true);
+    ASSERT_NE(0, recv_port);
+
+    socket.configureRecv("127.0.0.1", recv_port);
```
lestarch-autobot, line 199: [C++ Design] **suggestion** CPP-32 (check return values of fallible operations): the `SocketIpStatus` returned by `configureRecv()` is ignored.

The neighboring `configureSend()` and `open()` calls assert their status; doing the same here reports a setup failure at the failing line instead of downstream.

```suggestion
    ASSERT_EQ(socket.configureRecv("127.0.0.1", recv_port), Drv::SOCK_SUCCESS);
```

Drv/Ip/UdpSocket.cpp
lestarch-autobot, line 151: [Correctness] **future work** `setupSocketOptions` is still inside this `if (port != 0)` guard, so the `IP_SOCKET_OPTIONS` from `IpCfg.hpp` (e.g. `SO_REUSEADDR`) are never applied when the send port is 0 — exactly the receive-only / reply-to configuration where the socket is the one that binds.

(Anchored above the offending line; the diff does not include line 169.)

finding-class: correctness-predicate. Trigger: `configureRecv(...)` + `configureSend(addr, 0, ...)` (or receive-only), then close and re-`open` while the address is still in use. Consequence: `bind` fails and `open` returns `SOCK_FAILED_TO_BIND` where a REUSEADDR-bearing socket would succeed. Preexisting — the guard is unchanged by this PR (the PR description already flags it as future work).

Drv/Ip/UdpSocket.cpp
```diff
@@ -138,6 +138,15 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
         return SOCK_FAILED_TO_GET_SOCKET;
     }
 
+    // Apply the configured timeouts. This socket may send even when no send port was
+    // configured, because a send port of 0 means "reply to the source of the last
+    // datagram received", so the timeout must not be tied to `port != 0` below.
```
lestarch-autobot, line 143: [Documentation] **suggestion** New comment says the reply-to address comes from the *last* datagram received, but the documented behavior is the *first*.

`Drv/Ip/docs/sdd.md` ("Support for Ephemeral Ports") and the `configureSend` Doxygen both document that the send address latches on the first received datagram, and `recvProtocol` only sets it while `m_addr_send.sin_port == 0`. The new comment (and the matching one in `TestUdp.cpp`) should say "first" to stay consistent with the SDD.

```suggestion
    // Apply the configured timeouts. This socket may send even when no send port was
    // configured, because a send port of 0 means "reply to the source of the first
    // datagram received", so the timeout must not be tied to `port != 0` below.
```

Drv/Ip/UdpSocket.cpp
```diff
@@ -138,6 +138,15 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
...
+    status = this->setupTimeouts(socketFd);
```
lestarch-autobot, line 144: [Documentation] **could fix** Behavior change not reflected in `Drv/Ip/docs/sdd.md` (PR template marks Documentation Included: n).

The send timeout is now applied to every opened UDP socket, including the send-port-0 "reply to sender" mode described in the SDD's "Support for Ephemeral Ports" section. That section could state that the `configureSend` timeout arguments now take effect in this mode, so users know the timeout is honored.

Drv/Ip/UdpSocket.cpp
```diff
@@ -138,6 +138,15 @@ SocketIpStatus UdpSocket::openProtocol(SocketDescriptor& socketDescriptor) {
...
+    if (status != SOCK_SUCCESS) {
```
lestarch-autobot, line 145: [Documentation] **could fix** `IpSocket::setupTimeouts` Doxygen (`Drv/Ip/IpSocket.hpp`) is now stale for UDP.

Its `\brief` reads "setup the socket timeout properties of the opened outgoing socket", but this PR widens the call to every opened UDP socket, including receive-only sockets with no send configuration. Update the `\brief` to say the timeouts are applied to every opened socket. (Anchored here; the diff does not include `IpSocket.hpp`.)

lestarch-autobot (approved):

### Per-agent results

| F Prime C/C++ Design | 0 | 1 | 0 | 0 | 1 | Go |
| Documentation Currency | 0 | 1 | 2 | 0 | 3 | Go |
| Design | 0 | 0 | 0 | 1 | 1 | Go |
| Correctness | 0 | 0 | 0 | 1 | 1 | Go |
| Maintainability | 0 | 1 | 1 | 0 | 2 | Go |
| **Totals** | 0 | 3 | 3 | 2 | 8 | **Go** |

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

### Merge readiness
**Merge readiness: Go** — all nine reviewers completed with zero outstanding must-fix findings.

- None — every registered reviewer ran.

---

Clean pass through all nine stations — this fix is go for docking.

Issue #5309, opened 2026-06-18
Component queue overflow in FpySequencer


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 4.2 |
|**_Affected Component_**|  FpySequencer |
---
## Problem Description

* A busy loop in a sequence in the FpySequencer will prevent all queue messages other than the sequence's own directives from being handled (except for the manual CANCEL command or port).
* If this continues for a while, the checkTimers or tlmWrite or health ports will fill up the queue, causing an overflow.
* As a rule of thumb, if the sequence is not interacting with the outside FPrime system through commands or telemetry or parameters, it should be strictly guaranteed that nothing it can do will cause the FSW to crash. This contract is violated by this issue.

## Expected behavior
* Should be impossible to crash FSW like this

youio:
Hi @zimri-leisher, are you still working on this problem? No worries if you're still interested or if you've already made some progress, but if its on the back burner I would be interested in helping out. Just let me know, thank you!

zimri-leisher:
This issue will take a decent bit of thought and careful design to solve correctly. If you're open to iterating back and forth on the design, I'm open to helping you work through it. However, it might turn out to be a bunch more difficult than expected to do this properly.
I think the most promising idea right now is having some sort of behavior where if the component queue would overflow, you could terminate the currently running sequence.

youio:
Okay I see, I'll look into it and see what I can do. I'll let you know if anything seems promising. Thank you!

youio:
Hi @zimri-leisher, apologies for the delay. I just drafted a [pull request](#5613) with a possible solution for this issue along the lines of the suggestion you gave. The implementation is currently untested and I detailed a couple of possible concerns, but I wanted to check in to see if you think it would be viable. Let me know if it needs some tweaks or if it should go in another direction. Thank you!

PR #5613, opened 2026-08-09
Fix FpySequencer Queue Overflow

youio:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5309  |
|**_Documentation Included (y/n)_**|n  |

---
## Change Description

This PR attempts to fix the potential queue overflow issue in `FpySequencer` if executing a sequence with a busy loop. The approach is to check queue capacity whenever a new external signal is submitted. If the message load is within a margin of the queue's total depth it sends the cancel signal to terminate the currently running sequence. This PR overrides the pre-message hooks defined in `FpySequenceComponentBaseAC.cpp` to send a cancel signal if queue capacity is within a given margin.

## Rationale

As detailed in the corresponding issue, a busy loop in an fpy sequence will prevent other actions from being handled which will eventually cause an overflow. 

## Implementation Caveats/Concerns

The main concern with this implementation is that it uses the pre-message hooks which seem to be intended for the user. Future users will still be able to add their own functionality to the hooks but it doesn't feel very encapsulated. Unfortunately I was unable to find another way to cleanly insert functionality into the generated code.

AI was used for code understanding.


Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -523,4 +525,49 @@ bool FpySequencer::isRunningState(State state) {
            this->sequencer_getState() == State::RUNNING_PAUSED || this->sequencer_getState() == State::RUNNING_SLEEPING;
 }
 
+//----------------------------------------------------
+// Wrapper functions for monitoring queue capacity
+//----------------------------------------------------
+
+void FpySequencer::checkTimers_handlerBase(FwIndexType portNum, U32 context) {
+    if (this->m_queue.getMessagesAvailable() > this->m_queue.getDepth() - LOAD_MARGIN) {
+        this->sequencer_sendSignal_cmd_CANCEL();
```
zimri-leisher: need to check: which thread is handlerbase runnign on? is it okay that sendsignal is running on that thread?

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -7,6 +7,8 @@
 #include <Svc/FpySequencer/FpySequencer.hpp>
 #include <new>
 
+#define LOAD_MARGIN 4
```
zimri-leisher: In CPP, it's idiomatic to use static constexpr instead of #define. I would also put this in FpySequencerCfg.fpp to make it configurable by projects.

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -831,6 +831,19 @@ class FpySequencer : public FpySequencerComponentBase {
...
+    
+    void cmdResponseIn_handlerBase(FwIndexType portNum,
+                                   FwOpcodeType opCode,
+                                   U32 cmdSeq,
+                                   const Fw::CmdResponse& response);
+    void pingIn_handlerBase(FwIndexType portNum, U32 key);
+    void seqRunIn_handlerBase(FwIndexType portNum,
+                              const Fw::StringBase& filename,
+                              const Svc::SeqArgs& args);
+    void tlmWrite_handlerBase(FwIndexType portNum, U32 context);
```
zimri-leisher: do we need to add these checks to other things like commands that can also go on the queue? my instinct is no, but i'd encourage you to ponder it for a bit

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -831,6 +831,19 @@ class FpySequencer : public FpySequencerComponentBase {
     // dispatches a command, returns whether successful or not
     Fw::Success sendCmd(FwOpcodeType opcode, const U8* argBuf, FwSizeType argBufSize);
 
+    // HandlerBase wrappers for queue capacity monitoring
+    void FpySequencer::checkTimers_handlerBase(FwIndexType portNum, U32 context);
```
zimri-leisher: remove FpySequencer::

zimri-leisher (changes requested):
Nice start, and good work. Some changes suggested. The most important thing is to add tests. At the very least, prove that a running sequence gets cancelled if the queue comes close to overflowing in a UT.

Svc/FpySequencer/FpySequencer.cpp
youio: The sendsignal command only adds the signal to the component's queue which is roughly the same thing that the handlerbase does itself so I believe this should be okay.

Svc/FpySequencer/FpySequencer.hpp
youio: The directive dispatch calls for a sequence have a priority of 6 in the queue and these are the only calls that have a priority lower than that, so I believe they should be the only ones that could clog up the queue. All of the other message types have priority 7 or higher unless I'm missing something, so I believe it should be impossible for them to build up due to a busy sequence and cause an overflow.

youio:
Hi @zimri-leisher, I've incorporated your suggestions. I ran into a snag with the previous implementation where the handlerbase functions weren't getting overridden when called on a `FpySequencerComponentBase' pointer, so I had to shift to using the pre-message hook functions. It feels a little messy but I was able to show that it successfully prevents queue overflow on a busy sequence. Lmk if you think this implementation is acceptable or if we need to find another way. Thank you!

zimri-leisher:
This looks pretty good, thanks @youio 
@Brian-Campuzano can you take a look?

default/config/FpySequencerCfg.fpp
```diff
@@ -8,6 +8,8 @@ module Svc {
         dictionary constant MAX_STACK_SIZE = 65535
         @ the maximum number of bytes in a directive
         dictionary constant MAX_DIRECTIVE_SIZE = 2048
+        @ how close the queue can get to capacity before cancelling the current sequence
+        dictionary constant LOAD_MARGIN = 4
```
zimri-leisher, line 12: I would rename this to something more clear. QUEUE_MARGIN perhaps?

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -431,6 +431,31 @@ void FpySequencer::tlmWrite_handler(FwIndexType portNum,  //!< The port number
...
+    if (this->m_queue.getMessagesAvailable() > this->m_queue.getDepth() - Fpy::LOAD_MARGIN) {
+        this->sequencer_sendSignal_cmd_CANCEL();
+    }
+}
+
+void FpySequencer::checkTimers_preMsgHook(FwIndexType portNum, U32 context) {
+    checkOverflow();
+}
+
+void FpySequencer::cmdResponseIn_preMsgHook(FwIndexType portNum,
```
zimri-leisher, line 448: can we just do this for checkTimers and tlmWrite for now? cmdResponseIn should be safe, it won't come in on a rate group

zimri-leisher (changes requested):

sylvesterkaczmarek (commented):
`getDepth() - Fpy::LOAD_MARGIN` can underflow when a deployment configures a queue depth smaller than the margin, which disables the overflow protection on exactly those small queues. Could this handle `depth <= LOAD_MARGIN` explicitly (or use an equivalent non-underflowing comparison) and add a small-depth regression?

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -431,6 +431,31 @@ void FpySequencer::tlmWrite_handler(FwIndexType portNum,  //!< The port number
     this->tlmWrite_Debug_StackSize(this->m_debug.stackSize);
 }
 
+// ----------------------------------------------------------------------
+// Pre-message hook overrides for typed async input ports
+// ----------------------------------------------------------------------
+
+void FpySequencer::checkOverflow() {
+    if (this->m_queue.getMessagesAvailable() > this->m_queue.getDepth() - Fpy::LOAD_MARGIN) {
```
lestarch-autobot, line 439: [Documentation] **must fix** `Svc/FpySequencer/docs/sdd.md` does not describe the new automatic sequence cancellation on queue pressure.

(Anchored on the code that creates the staleness; the diff does not touch the SDD.) The SDD is the per-component contract and currently says CANCEL happens only by command (`## Commands`) — after this PR a running sequence is also cancelled autonomously whenever `tlmWrite`, `checkTimers`, or `cmdResponseIn` arrives with fewer than `Svc::Fpy::LOAD_MARGIN` free queue slots. Operators need to know this from the SDD: add a short section (e.g. `## Queue Overflow Protection`) stating the trigger condition, that the sequence ends via the CANCEL path, and that `LOAD_MARGIN` in `FpySequencerCfg.fpp` tunes the threshold. The PR template also marks `Documentation Included: n`.

default/config/FpySequencerCfg.fpp
```diff
@@ -8,6 +8,8 @@ module Svc {
         dictionary constant MAX_STACK_SIZE = 65535
         @ the maximum number of bytes in a directive
         dictionary constant MAX_DIRECTIVE_SIZE = 2048
+        @ how close the queue can get to capacity before cancelling the current sequence
```
lestarch-autobot, line 11: [Documentation] **could fix** Doc comment for `LOAD_MARGIN` does not state its unit or exact trigger.

This comment becomes the dictionary description operators see. The implementation cancels when free queue slots drop below `LOAD_MARGIN` (`getMessagesAvailable() > getDepth() - LOAD_MARGIN`), so spell out the unit (queue slots) and the affected ports.

```suggestion
        @ number of free message-queue slots below which the running sequence is cancelled
        @ (checked when a tlmWrite, checkTimers, or cmdResponseIn message is enqueued)
```

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -537,6 +537,23 @@ class FpySequencer : public FpySequencerComponentBase {
                           U32 context           //!< The call order
                           ) override;
 
+    // ----------------------------------------------------------------------
+    // Pre-message hook overrides for typed async input ports
+    // ----------------------------------------------------------------------
+
+    void checkOverflow();
```
lestarch-autobot, line 544: [Documentation] **could fix** New `checkOverflow()` has no Doxygen comment, unlike every other declaration in this header.

```suggestion
    //! Cancel the running sequence when fewer than Fpy::LOAD_MARGIN queue slots remain
    void checkOverflow();
```

Svc/FpySequencer/FpySequencer.hpp
```diff
@@ -537,6 +537,23 @@ class FpySequencer : public FpySequencerComponentBase {
...
+    //! Pre-message hook for async input port checkTimers
+    void checkTimers_preMsgHook(FwIndexType portNum, U32 context) override;
+
+    //! Pre-message hook for async input port cmdResponseIn
+    void cmdResponseIn_preMsgHook(FwIndexType portNum,
+                                  FwOpcodeType opCode,
+                                  U32 cmdSeq,
+                                  const Fw::CmdResponse& response) override;
+
+    // ! Pre-message hook for async input port tlmWrite
```
lestarch-autobot, line 554: [Documentation] **could fix** `// !` is not a Doxygen marker; this comment is dropped from generated docs.

```suggestion
    //! Pre-message hook for async input port tlmWrite
```

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -431,6 +431,31 @@ void FpySequencer::tlmWrite_handler(FwIndexType portNum,  //!< The port number
     this->tlmWrite_Debug_StackSize(this->m_debug.stackSize);
 }
 
+// ----------------------------------------------------------------------
+// Pre-message hook overrides for typed async input ports
+// ----------------------------------------------------------------------
+
+void FpySequencer::checkOverflow() {
```
lestarch-autobot, line 438: [Maintainability] **could fix** `maint-unclear-naming`: `checkOverflow()` reads as a query but its effect is to cancel the running sequence.

A reader scanning the three hooks sees `checkOverflow()` and expects a predicate or a diagnostic; the actual behavior (sending `cmd_CANCEL`) is only visible in the body. A name that states the action, e.g. `cancelSequenceIfQueueNearFull()`, lets the next engineer understand each hook from its one-line body. Rename touches the `.hpp` declaration and the three call sites below.

Svc/FpySequencer/FpySequencer.cpp
lestarch-autobot, line 439: [Maintainability] **suggestion** `maint-clever-code`: unsigned subtraction `getDepth() - Fpy::LOAD_MARGIN` hides a wraparound case in the guard.

`FwSizeType` is unsigned, so the intended reading ("fewer than `LOAD_MARGIN` slots left") only holds when `depth > LOAD_MARGIN`; a reader must reason about the wrap for small configured depths to know what the branch does (also raised in @sylvesterkaczmarek's review). Moving the margin to the other side states the intent directly and needs no such reasoning.

```suggestion
    if (this->m_queue.getMessagesAvailable() + Fpy::LOAD_MARGIN > this->m_queue.getDepth()) {
```

Svc/FpySequencer/FpySequencer.hpp
lestarch-autobot, line 544: [Maintainability] **suggestion** `maint-inconsistent-local-convention`: undocumented helper filed under the "Pre-message hook overrides" section.

Every other declaration in this class carries a `//!` brief; `checkOverflow()` is the only bare one, and it sits under a banner that says the entries are hook overrides, so a reader will assume it is one. A brief comment naming what it does (and that it runs on the caller's thread, before enqueue) removes the mismatch.

```suggestion
    //! Cancel the running sequence when the message queue is within LOAD_MARGIN of capacity.
    //! Called from the pre-message hooks below, on the calling thread, before the message is enqueued.
    void checkOverflow();

```

Svc/FpySequencer/FpySequencer.hpp
lestarch-autobot, line 554: [Maintainability] **could fix** `maint-inconsistent-local-convention`: `// !` is not the `//!` Doxygen marker used on every neighboring declaration.

The stray space turns this into a plain comment, so `tlmWrite_preMsgHook` is the one hook in the section that drops out of generated docs.

```suggestion
    //! Pre-message hook for async input port tlmWrite
```

Svc/FpySequencer/FpySequencer.cpp
lestarch-autobot, line 439: [Architecture] **could fix** Sequencer fault policy (`checkOverflow` → `cmd_CANCEL`) runs in the caller's thread via `*_preMsgHook`, not on the component's own thread.

`FpySequencer` is an event-driven `active` component using the Cyclic Notification pattern (`checkTimers`/`tlmWrite` are `async` `Svc.Sched`). Pre-message hooks execute in the invoking context before the message is queued, so this check and the `sequencer_sendSignal_cmd_CANCEL()` enqueue now execute on the rate group's (or command responder's) thread and reach into `m_queue` from outside the component thread — the coupling the `async` Sched port exists to avoid (see "Cyclic Notification Pattern" and "Event-Driven Work" in `docs/user-manual/framework/component-and-port-selection.md`). `Os::Queue` is thread-safe, so this reads as unconventional primitive use rather than a data race. Consider keeping the policy on the component thread instead — e.g. check `m_queue.getMessagesAvailable()` in `dispatchStatement()` before the next directive self-enqueue — or use the FPP `hook` overflow behavior on these ports rather than a bespoke margin in caller context.

default/config/FpySequencerCfg.fpp
lestarch-autobot: [Maintainability] **Concur** — also in scope for maintainability (`maint-unclear-naming`): `LOAD_MARGIN` does not say what is loaded or what unit the margin is in; a name tying it to the message queue (e.g. `QUEUE_MARGIN`, or `QUEUE_CANCEL_MARGIN` since the effect is cancelling the sequence) reads correctly at both use sites; severity from my scope: **could fix**.

Svc/FpySequencer/FpySequencer.cpp
lestarch-autobot, line 439: [C++ Design] **must fix** CPP-27 (bounded arithmetic): `getDepth() - Fpy::LOAD_MARGIN` underflows on small queues.

`Os::Queue::getDepth()` returns unsigned `FwSizeType`, so a deployment whose queue depth is
`<= LOAD_MARGIN` wraps the right-hand side to a huge value and the guard never fires — the new
protection silently disappears exactly where the queue is tightest. An additive comparison
cannot wrap.

```suggestion
    if (this->m_queue.getMessagesAvailable() + static_cast<FwSizeType>(Fpy::LOAD_MARGIN) > this->m_queue.getDepth()) {
```

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -431,6 +431,31 @@ void FpySequencer::tlmWrite_handler(FwIndexType portNum,  //!< The port number
...
+    if (this->m_queue.getMessagesAvailable() > this->m_queue.getDepth() - Fpy::LOAD_MARGIN) {
+        this->sequencer_sendSignal_cmd_CANCEL();
+    }
+}
+
+void FpySequencer::checkTimers_preMsgHook(FwIndexType portNum, U32 context) {
+    checkOverflow();
```
lestarch-autobot, line 445: [C++ Design] **could fix** CPP-26 (style guide): member call not qualified with `this->`.

The rest of this file qualifies member access consistently; same at lines 452 and 456.

```suggestion
    this->checkOverflow();
```

Svc/FpySequencer/FpySequencer.hpp
lestarch-autobot, line 544: [C++ Design] **could fix** CPP-26 (style guide): `checkOverflow` has no Doxygen brief.

Every other declaration in this section carries a `//!` brief, and this one is a private
helper rather than a `_preMsgHook` override, so a brief keeps the section heading honest.

```suggestion
    //! Cancel the running sequence if the message queue is near capacity
    void checkOverflow();
```

Svc/FpySequencer/FpySequencer.hpp
lestarch-autobot, line 554: [C++ Design] **could fix** CPP-26 (style guide): `// !` is not a Doxygen marker, so this brief is dropped.

```suggestion
    //! Pre-message hook for async input port tlmWrite
```

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
```diff
@@ -5556,4 +5556,106 @@ TEST_F(FpySequencerTester, popSerializable_differentPorts) {
...
+    FwSizeType depth = qptr->getDepth();
+
+    // Attempt to overload the queue
+    for (FwSizeType i = 0; i < depth + 1; i++) {
+        invoke_to_tlmWrite(0, 0);
+        for (U8 j = 0; j < (qptr->getMessagesAvailable() < 4 ? qptr->getMessagesAvailable() : 4); j++) {
+            this->dispatchOne(this->cmp);
+        }
+    }
+}
```
lestarch-autobot, line 5576: [Test Quality] **must fix** `QueueOverflow_TlmWrite` invokes `invoke_to_tlmWrite` in a loop but contains no `ASSERT_*`/`EXPECT_*`.

The test only fails if the queue asserts; it never verifies that `checkOverflow()` fired `sequencer_sendSignal_cmd_CANCEL()` and the busy sequence actually stopped. Because each iteration also drains up to 4 messages, the margin may never be reached and the test would still pass. Assert the observable outcome (state returns to `IDLE`, `SequenceCancelled` emitted) so a regression in the hook or threshold is caught. (best-effort fix; verify before applying)

```suggestion
    // the flood must have tripped the pre-message hook and cancelled the sequence
    dispatchUntilState(State::IDLE);
    ASSERT_EVENTS_SequenceCancelled_SIZE(1);
}
```

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
```diff
@@ -5556,4 +5556,106 @@ TEST_F(FpySequencerTester, popSerializable_differentPorts) {
...
+    FwSizeType depth = qptr->getDepth();
+
+    // Attempt to overload the queue
+    for (FwSizeType i = 0; i < depth + 1; i++) {
+        invoke_to_checkTimers(0, 0);
+        for (U8 j = 0; j < (qptr->getMessagesAvailable() < 4 ? qptr->getMessagesAvailable() : 4); j++) {
+            this->dispatchOne(this->cmp);
+        }
+    }
+}
```
lestarch-autobot, line 5595: [Test Quality] **must fix** `QueueOverflow_CheckTimers` invokes `invoke_to_checkTimers` in a loop but contains no `ASSERT_*`/`EXPECT_*`.

The test only fails if the queue asserts; it never verifies that `checkOverflow()` fired `sequencer_sendSignal_cmd_CANCEL()` and the busy sequence actually stopped. Because each iteration also drains up to 4 messages, the margin may never be reached and the test would still pass. Assert the observable outcome (state returns to `IDLE`, `SequenceCancelled` emitted) so a regression in the hook or threshold is caught. (best-effort fix; verify before applying)

```suggestion
    // the flood must have tripped the pre-message hook and cancelled the sequence
    dispatchUntilState(State::IDLE);
    ASSERT_EVENTS_SequenceCancelled_SIZE(1);
}
```

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
```diff
@@ -5556,4 +5556,106 @@ TEST_F(FpySequencerTester, popSerializable_differentPorts) {
...
+    FwSizeType depth = qptr->getDepth();
+
+    // Attempt to overload the queue
+    for (FwSizeType i = 0; i < depth + 1; i++) {
+        invoke_to_cmdResponseIn(0, 0, 0, Fw::CmdResponse::OK);
+        for (U8 j = 0; j < (qptr->getMessagesAvailable() < 4 ? qptr->getMessagesAvailable() : 4); j++) {
+            this->dispatchOne(this->cmp);
+        }
+    }
+}
```
lestarch-autobot, line 5614: [Test Quality] **must fix** `QueueOverflow_CmdResponseIn` invokes `invoke_to_cmdResponseIn` in a loop but contains no `ASSERT_*`/`EXPECT_*`.

The test only fails if the queue asserts; it never verifies that `checkOverflow()` fired `sequencer_sendSignal_cmd_CANCEL()` and the busy sequence actually stopped. Because each iteration also drains up to 4 messages, the margin may never be reached and the test would still pass. Assert the observable outcome (state returns to `IDLE`, `SequenceCancelled` emitted) so a regression in the hook or threshold is caught. (best-effort fix; verify before applying)

```suggestion
    // the flood must have tripped the pre-message hook and cancelled the sequence
    dispatchUntilState(State::IDLE);
    ASSERT_EVENTS_SequenceCancelled_SIZE(1);
}
```

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
```diff
@@ -5556,4 +5556,106 @@ TEST_F(FpySequencerTester, popSerializable_differentPorts) {
...
+    FwSizeType depth = qptr->getDepth();
+
+    // Attempt to overload the queue
+    for (FwSizeType i = 0; i < depth + 20; i++) {
+        invoke_to_tlmWrite(0, 0);
+        for (U8 j = 0; j < (qptr->getMessagesAvailable() < 4 ? qptr->getMessagesAvailable() : 4); j++) {
+            this->dispatchOne(this->cmp);
+        }
+    }
+}
```
lestarch-autobot, line 5636: [Test Quality] **must fix** `QueueOverflow_longBusySequence` invokes `invoke_to_tlmWrite` in a loop but contains no `ASSERT_*`/`EXPECT_*`.

The test only fails if the queue asserts; it never verifies that `checkOverflow()` fired `sequencer_sendSignal_cmd_CANCEL()` and the busy sequence actually stopped. Because each iteration also drains up to 4 messages, the margin may never be reached and the test would still pass. Assert the observable outcome (state returns to `IDLE`, `SequenceCancelled` emitted) so a regression in the hook or threshold is caught. (best-effort fix; verify before applying)

```suggestion
    // the flood must have tripped the pre-message hook and cancelled the sequence
    dispatchUntilState(State::IDLE);
    ASSERT_EVENTS_SequenceCancelled_SIZE(1);
}
```

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
```diff
@@ -5556,4 +5556,106 @@ TEST_F(FpySequencerTester, popSerializable_differentPorts) {
...
+    }
+
+    // Attempt to overload the queue
+    for (FwSizeType i = 0; i < Fpy::LOAD_MARGIN; i++) {
+        sendCmd_RUN(0, 0, Fw::String("test.bin"), BlockState::BLOCK);
+        for (U8 j = 0; j < (qptr->getMessagesAvailable() < 4 ? qptr->getMessagesAvailable() : 4); j++) {
+            this->dispatchOne(this->cmp);
+        }
+    }
+}
```
lestarch-autobot, line 5659: [Test Quality] **must fix** `QueueOverflow_CmdRun` pre-fills the queue and sends `RUN` commands but asserts nothing.

No `ASSERT_CMD_RESPONSE*`, state check, or event assertion follows the loop, so the test cannot distinguish "commands are safe near the margin" from "nothing happened". State what this test is meant to demonstrate (e.g. `ASSERT_CMD_RESPONSE_SIZE(Fpy::LOAD_MARGIN)` with `EXECUTION_ERROR` for `RUN` while running, and whether the sequence is or is not cancelled) and assert it after the loop.

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
```diff
@@ -5556,4 +5556,106 @@ TEST_F(FpySequencerTester, popSerializable_differentPorts) {
...
+
 }  // namespace Svc
```
lestarch-autobot, line 5661: [Test Quality] **suggestion** No test pins the `checkOverflow()` threshold (`> depth - LOAD_MARGIN`) on either side of the boundary.

The new tests flood the queue without checking when the cancel fires, so an off-by-one in the comparison or a change to `LOAD_MARGIN` would go unnoticed. A deterministic boundary test can fill without dispatching and observe the extra `CANCEL` signal via `getMessagesAvailable()`, then confirm the sequence is cancelled. (best-effort fix; verify before applying)

```suggestion
TEST_F(FpySequencerTester, QueueOverflow_thresholdBoundary) {
    // Busy sequence
    allocMem();
    add_NO_OP();
    add_GOTO(0);
    writeAndRun();

    Os::Queue* qptr = this->tester_get_m_componentQueue_ptr();
    FwSizeType threshold = qptr->getDepth() - Fpy::LOAD_MARGIN;

    // Fill (without dispatching) up to exactly the threshold: hook must not fire
    while (qptr->getMessagesAvailable() < threshold) {
        invoke_to_tlmWrite(0, 0);
    }
    ASSERT_EQ(qptr->getMessagesAvailable(), threshold);
    // available == threshold is not > threshold: still no CANCEL signal enqueued
    invoke_to_tlmWrite(0, 0);
    ASSERT_EQ(qptr->getMessagesAvailable(), threshold + 1);
    // now above threshold: CANCEL signal + tlmWrite message are both enqueued
    invoke_to_tlmWrite(0, 0);
    ASSERT_EQ(qptr->getMessagesAvailable(), threshold + 3);

    dispatchUntilState(State::IDLE);
    ASSERT_EVENTS_SequenceCancelled_SIZE(1);
}

}  // namespace Svc
```

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
```diff
@@ -5556,4 +5556,106 @@ TEST_F(FpySequencerTester, popSerializable_differentPorts) {
     ASSERT_EQ(m_serialOutHistory[0].portNum, 4);
 }
 
+TEST_F(FpySequencerTester, QueueOverflow_TlmWrite) {
```
lestarch-autobot, line 5559: [Test Quality] **could fix** `QueueOverflow_TlmWrite`, `_CheckTimers`, `_CmdResponseIn` and `_longBusySequence` repeat the same 15-line flood/drain body, differing only in the invoked port.

Factor the busy-sequence setup and flood/drain loop into a Tester helper taking the port-invoke callable (or a `TEST_P` over the three hooks) so the assertions added to fix the coverage gap live in one place.

Svc/FpySequencer/FpySequencer.cpp
lestarch-autobot, line 439: [Correctness] **must fix** `correctness-arithmetic`: `getDepth() - Fpy::LOAD_MARGIN` is unsigned (`FwSizeType`) and wraps when a deployment sets the queue depth below `LOAD_MARGIN` (default 4). Trigger: depth 3, margin 4 -> RHS is ~SIZE_MAX, so `getMessagesAvailable() > RHS` is never true and `cmd_CANCEL` is never sent; the new overflow protection is silently disabled on exactly the smallest queues. Rewriting as an addition keeps the same threshold without the wrap.

```suggestion
    if (this->m_queue.getMessagesAvailable() + Fpy::LOAD_MARGIN > this->m_queue.getDepth()) {
```

Svc/FpySequencer/FpySequencer.cpp
```diff
@@ -431,6 +431,31 @@ void FpySequencer::tlmWrite_handler(FwIndexType portNum,  //!< The port number
     this->tlmWrite_Debug_StackSize(this->m_debug.stackSize);
 }
 
+// ----------------------------------------------------------------------
+// Pre-message hook overrides for typed async input ports
+// ----------------------------------------------------------------------
+
+void FpySequencer::checkOverflow() {
+    if (this->m_queue.getMessagesAvailable() > this->m_queue.getDepth() - Fpy::LOAD_MARGIN) {
+        this->sequencer_sendSignal_cmd_CANCEL();
```
lestarch-autobot, line 440: [Correctness] **could fix** `correctness-other`: `sequencer_sendSignal_cmd_CANCEL()` enqueues onto the same `m_queue` being guarded, and this hook runs on the caller's thread before the port message itself is enqueued, so every over-threshold call adds two messages. Trigger: with the sequencer thread not yet dispatching, the first hook call at `depth-3` fills to `depth-1`, and the second call's CANCEL fills the queue so the port message `send` fails and asserts -- the 4-slot margin yields ~1 message of headroom, and repeated CANCELs are redundant (and ignored in IDLE). Consider sending CANCEL once per episode (e.g. a flag cleared when the sequence leaves RUNNING) or sizing the margin for two slots per caller.

Svc/FpySequencer/FpySequencer.cpp
lestarch-autobot, line 439: [Design] **must fix** — `design-fault-handling-gap`

The margin does not cover the fan-in of the mitigation itself. Each hook that trips enqueues **two** messages (the priority-9 CANCEL plus the original port message), and three ports are hooked, so up to 6 messages can land after the threshold before the sequencer thread drains — but `LOAD_MARGIN` is 4. Concretely: with a directive message in flight (odd occupancy) the queue sits at `depth-3` when `tlmWrite` and `checkTimers` fire on the same rate-group tick from a higher-priority thread; `tlmWrite` adds CANCEL+msg (`depth-1`), `checkTimers` adds CANCEL+msg → overflow → `assert`, i.e. the exact crash #5309 is meant to prevent. Nothing stops a second/third CANCEL being sent while the first is still queued.
Fix options: send CANCEL at most once per near-full episode (e.g. an atomic flag armed here and cleared when the state machine leaves RUNNING), and/or require and document `LOAD_MARGIN >= 2 * (hooked ports) + 1` and raise the default accordingly, with a UT that trips two hooks back-to-back at the threshold.

cc @LeStarch @thomas-bc — low-confidence on exact scheduling interleaving; please sanity-check the fan-in arithmetic.

Svc/FpySequencer/FpySequencer.cpp
lestarch-autobot, line 440: [Design] **suggestion** — `design-intent-mismatch`

The stated intent (#5309, PR body) is to abort a *busy-looping sequence*. This hook raises `cmd_CANCEL` on any queue pressure, in any state: `FpySequencerStateMachine.fppi` documents `cmd_CANCEL` as "raised in all states except IDLE" and `IDLE` has no `on cmd_CANCEL`, so an IDLE sequencer whose queue fills for an unrelated reason (thread starvation) now spends its last free slots on no-op signals instead of the incoming messages, hastening the overflow. Consider gating the signal on `isRunningState(sequencer_getState())` (best-effort cross-thread read is acceptable here since the fallback is the existing assert), and update the `cmd_CANCEL` doc comment in the `.fppi` to reflect that it is now raised from a pre-message hook rather than only from the CANCEL command.

Svc/FpySequencer/FpySequencer.cpp
lestarch-autobot, line 439: [Operational] **must fix** — ops-config-extreme

`getDepth()` and `LOAD_MARGIN` are unsigned (`FwSizeType`); any instance with `queue size` ≤ 4 (legal in FPP) wraps the RHS to ~2^64, so the guard never fires and the queue asserts exactly on the smallest-depth instances the fix is meant to protect. Reorder to add on the LHS (`messagesAvailable ≤ depth`, so no overflow):

```suggestion
    if (this->m_queue.getMessagesAvailable() + Fpy::LOAD_MARGIN > this->m_queue.getDepth()) {
```
(Concurs with the concern raised in the earlier human review; no agent thread exists yet at this site.)

Svc/FpySequencer/FpySequencer.cpp
lestarch-autobot, line 440: [Operational] **could fix** — ops-failure-path (judgment call; @LeStarch @thomas-bc)

The hook runs on the *caller's* thread and cannot tell a busy-looping sequence from a starved sequencer thread. On a strict-priority RTOS with a 10 Hz `checkTimers` + 1 Hz `tlmWrite` and `queue size 10`, any preemption of the sequencer for ≳0.6 s cancels whatever is in flight — including a `SLEEPING` wait-for-burn-time sequence, a `VALIDATING` run (`EXECUTION_ERROR` returned to a blocking RUN), or a legitimately queued `cmdResponseIn` burst — and emits the same `SequenceCancelled` event as a ground CANCEL, so the operator cannot distinguish the two.

Smallest remedy: raise a dedicated `WARNING_HI` event (e.g. `QueueNearFull(messagesAvailable, depth)`) from the hook, and state the auto-cancel + depth-sizing rule in `docs/sdd.md` (untouched by this PR).

default/config/FpySequencerCfg.fpp
lestarch-autobot, line 12: [Operational] **suggestion** — ops-timing-resource

Each hook invocation above threshold enqueues **two** messages (the port message plus a priority-9 `cmd_CANCEL`), so with `LOAD_MARGIN = 4` and `checkTimers`/`tlmWrite` on the same rate group the margin is spent within one tick: tick N fills to `depth-1`; if the sequencer thread does not dispatch before tick N+1, the next hook enqueues `CANCEL` (`depth`) then its own message → the original overflow assert. The protection window is therefore exactly one rate-group period, and `PriorityQueue::getMessagesAvailable()` reads the heap size unlocked so the count seen may already be stale by one message. Document the sizing rule so integrators size depth/margin deliberately:

```suggestion
        @ queue-fill threshold that cancels the running sequence; each trigger enqueues 2 messages,
        @ so keep >= 2 * (rate-group-driven ports) + 1 and queue size well above it
        dictionary constant LOAD_MARGIN = 4
```

Svc/FpySequencer/FpySequencer.cpp
lestarch-autobot: [Security] **Concur** — also in scope for ground-driven DoS: on a deployment with depth <= LOAD_MARGIN the wrapped guard never fires, so a ground-uploaded busy loop still overflows the `assert` queue; severity from my scope: **must fix**.

Svc/FpySequencer/FpySequencer.cpp
lestarch-autobot: [Security] **Concur** — also in scope for ground-reachable asserts: a ground-uploaded busy loop plus two rate-group ports firing in one cycle (`tlmWrite` then `checkTimers`, each pushing CANCEL + message) fills the last slots before the lower-priority sequencer thread pops CANCEL, so the `assert` overflow this PR targets remains reachable (and CANCEL is discarded in IDLE); severity from my scope: **must fix**. cc @LeStarch @thomas-bc @bitWarrior — please confirm the rate-group/sequencer priority assumption.

Svc/FpySequencer/FpySequencer.cpp
lestarch-autobot: [Summary] **Duplicate** — consolidated into https://github.com/nasa/fprime/pull/5613#discussion_r3973377796.

lestarch-autobot: [Summary] **Duplicate** — consolidated into https://github.com/nasa/fprime/pull/5613#discussion_r3973377796.

lestarch-autobot: [Summary] **Duplicate** — consolidated into https://github.com/nasa/fprime/pull/5613#discussion_r3973377796.

lestarch-autobot: [Summary] **Severity escalated to must fix** — duplicate threads consolidated here were tagged **must fix** by C++ Design (CPP-27, https://github.com/nasa/fprime/pull/5613#discussion_r3973379600), Correctness (https://github.com/nasa/fprime/pull/5613#discussion_r3973385727), Operational (https://github.com/nasa/fprime/pull/5613#discussion_r3973396377) and Security (concurrence: ground-driven DoS). This thread is the canonical home of the unsigned-underflow finding and is treated as **must fix**.

Svc/FpySequencer/FpySequencer.hpp
lestarch-autobot: [Summary] **Duplicate** — consolidated into https://github.com/nasa/fprime/pull/5613#discussion_r3973375135.

lestarch-autobot: [Summary] **Duplicate** — consolidated into https://github.com/nasa/fprime/pull/5613#discussion_r3973375135.

lestarch-autobot: [Summary] **Severity escalated to suggestion** — the consolidated Maintainability thread (https://github.com/nasa/fprime/pull/5613#discussion_r3973377812) tagged this **suggestion** and also notes the helper sits under the "Pre-message hook overrides" banner; this thread is treated as **suggestion**.

Svc/FpySequencer/FpySequencer.hpp
lestarch-autobot: [Summary] **Duplicate** — consolidated into https://github.com/nasa/fprime/pull/5613#discussion_r3973375141.

lestarch-autobot: [Summary] **Duplicate** — consolidated into https://github.com/nasa/fprime/pull/5613#discussion_r3973375141.

lestarch-autobot (changes requested):

### Per-agent results

| F Prime C/C++ Design | 1 | 0 | 3 | 0 | 4 | No-Go |
| Documentation Currency | 1 | 0 | 3 | 0 | 4 | No-Go |
| Design | 1 | 1 | 0 | 0 | 1 | No-Go |
| Architecture | 0 | 0 | 1 | 0 | 1 | Go |
| Test Quality | 5 | 1 | 1 | 0 | 7 | No-Go |
| Correctness | 1 | 0 | 1 | 0 | 2 | No-Go |
| Operational | 1 | 1 | 1 | 0 | 1 | No-Go |
| Maintainability | 0 | 2 | 3 | 0 | 5 | Go |
| **Totals** | 8 | 4 | 9 | 0 | 21 | **No-Go** |

Totals count each consolidated finding once (raw per-agent sums: 12 / 5 / 13 / 0 / 27). The 2 Security must-fix entries are concurrences on the C++ Design and Correctness threads and carry the canonical threads' severity.

Duplicates consolidated this run: 7 (threads closed by the §5h post-pass; `resolveReviewThread` returned FORBIDDEN for this token, so each duplicate carries a linking reply but remains unresolved — maintainers may resolve them)

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>
<summary>Outstanding must-fix items (8)</summary>

**Maintainability** (escalated to must fix by consolidation)
- `getDepth() - Fpy::LOAD_MARGIN` is unsigned and wraps when the queue depth is ≤ `LOAD_MARGIN`, silently disabling the overflow guard on the smallest queues; rewrite as `getMessagesAvailable() + LOAD_MARGIN > getDepth()` — also: F Prime C/C++ Design (CPP-27), Correctness, Operational, Security — https://github.com/nasa/fprime/pull/5613#discussion_r3973377796

**Documentation Currency**
- `Svc/FpySequencer/docs/sdd.md` does not describe the new automatic sequence cancellation on queue pressure — https://github.com/nasa/fprime/pull/5613#discussion_r3973375112

**Design**
- **Human design adjudication required.** `LOAD_MARGIN` (4) does not cover the fan-in of the mitigation itself: three hooked ports each enqueue CANCEL + original message, so two hooks tripping back-to-back can still overflow and assert — https://github.com/nasa/fprime/pull/5613#discussion_r3973389489

**Test Quality**
- `QueueOverflow_TlmWrite` loops `invoke_to_tlmWrite` but contains no `ASSERT_*`/`EXPECT_*` — https://github.com/nasa/fprime/pull/5613#discussion_r3973383036
- `QueueOverflow_CheckTimers` loops `invoke_to_checkTimers` but contains no `ASSERT_*`/`EXPECT_*` — https://github.com/nasa/fprime/pull/5613#discussion_r3973383051
- `QueueOverflow_CmdResponseIn` loops `invoke_to_cmdResponseIn` but contains no `ASSERT_*`/`EXPECT_*` — https://github.com/nasa/fprime/pull/5613#discussion_r3973383056
- `QueueOverflow_longBusySequence` loops `invoke_to_tlmWrite` but contains no `ASSERT_*`/`EXPECT_*` — https://github.com/nasa/fprime/pull/5613#discussion_r3973383072
- `QueueOverflow_CmdRun` pre-fills the queue and sends `RUN` commands but asserts nothing — https://github.com/nasa/fprime/pull/5613#discussion_r3973383078

</details>

### Merge readiness
**Merge readiness: No-Go** — 8 outstanding must-fix items across Maintainability (consolidated arithmetic defect), Documentation Currency, Design, and Test Quality.

---

Ten agents, one queue, zero margin for wraparound — fix the arithmetic and this sequencer is cleared for launch.

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: Agreed

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: Agreed

zimri-leisher: This is true and worth thinking more about. @youio does checking in dispatchStatement work, could you test?

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: Agreed

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
zimri-leisher: Good idea, worth trying

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
zimri-leisher: Also good idea, same with all these "check sequence was cancelled" comments

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
zimri-leisher: Agreed. Also if you're going to fix the underflow/overflow please add a test for that, and make sure that the test fails before you fix it, and succeeds after

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: Hmm. This is actually a problem with this design. It doesn't fundamentally stop the queue from being overfilled, it just tries to prevent it by cancelling the sequence. If a bunch of port calls come in at the same time, the sequencer could still be overwhelmed. However, I guess the question is what can we really do if we suddenly get flooded with port calls? Any component would die obviously--unless it drops or otherwise hooks on queue being full instead of asserting. So the only way to certainly prevent this is to do that on every async input port. But we probably don't want to do that: we do actually want to allow it to assert if the queue is overfull. The only thing we *don't* want is for the queue to fill up *while a sequence is running*. So maybe the guarantee we should make is if we do assert, a sequence will not be running. But I don't think we can make that guarantee, because in order for a sequence to stop running, we need to add a signal to the queue and it needs to get processed, which might not happen soon enough and other messages could get added to the queue.

Honestly, I'm looping back around to the design we have here. It's a simple best-effort approach and I don't think there's any possible way to do better in a guaranteed way except by dropping messages, which we don't want to do.

The issue the AI brings up is real though, for each port call that comes in when we're over the margin, we add that port call plus the cancel signal to the queue, filing it even faster. Hopefully people just set their margin conservatively.

zimri-leisher:
@youio lots of feedback from the AI, a lot of it is right. My main feedback: can we make this optional and off by default? Maybe a LOAD_MARGIN of 0 means that it never does this behavior.

Sorry this is a larger review load than I expected. This is kinda what I meant originally when I said we'd have to go back and forth on this. It's a complicated issue, there's no obvious solution

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: Agreed, and we should go with the documentation/default/UT route

Svc/FpySequencer/FpySequencer.cpp
zimri-leisher: Agreed, should be gated on isRunningState, and agreed on doc update

zimri-leisher: Agreed, we want a dedicated SequenceCancelledDueToQueueNearFull event

zimri-leisher:
@youio had a discussion with FPrime team and decided that the right way to do this is to have an atomic variable in the checkTimers/writeTlm handlers. They should be synchronous handlers, and they should check if the atomic var is not set before pushing to an internal checkTimersInternal/writeTlmInteral async port. If it is set, they should drop the port call. And then the actual handler for checkTimersInternal/writeTlmInternal should unset the atomic var (there should be two, one for each port). This way we guarantee only one message from each of these ports will be on the queue

sylvesterkaczmarek:
That direction makes sense to me. The per-port atomic gate should bound those periodic queue entries directly, without coupling the protection to sequence cancellation. Happy to re-review once the revision is pushed.

youio:
@zimri-leisher I see that makes sense, thanks for the feedback. I think I'll go ahead and close this pull request so I can work on the new implementation with a clean slate and I'll let you know when its ready for review. @sylvesterkaczmarek I'll ping you as well. Thanks again!

sylvesterkaczmarek:
Sounds good. Ping me when the replacement PR is ready and I’ll re-review.

Issue #5249, opened 2026-06-05
Make delegate pattern optional by moving OSAL to use configuration and inheritance


celskeggs:
| | |
|:---|:---|
|**_F´ Version_**| latest |
|**_Affected Component_**| Os |
---
## Background

The OSAL in F Prime uses a "delegate" pattern, where the implementation of each OSAL interface is selected at link time. Since the implementation is not known at _compile_ time for any compilation unit that uses the interface, indirect code with virtual function calls gets generated. This object code is less efficient than an implementation that knows the exact implementation and data layout.

The reason we select the OSAL interface implementations at link time is because it allows platforms that mix UT and deployment compilation to use different OSAL implementations for deployment binaries and UT binaries. However, this is not needed for some platforms, which would rather have the performance benefit of knowing the OSAL interface implementation at compile time.

## Feature Description

After discussion with @LeStarch and @thomas-bc earlier this week, I want to propose using inheritance and configuration instead of link-time delegation. We can continue supporting link-time delegation as an option for platforms that need it and for backwards compatibility, while allowing compile-time implementation selection for other platforms.

The current class hierarchy for RawTime is as follows (abbreviated for clarity):

```
RawTimeInterface
\- RawTime (with storage size FW_RAW_TIME_HANDLE_MAX_SIZE)
\- StubRawTime
\- PosixRawTime
\- TimerRawTime (from fprime-vorago)

configuration:
integer FW_RAW_TIME_HANDLE_MAX_SIZE defaults to 56
```

The RawTime class currently provides the delegate implementation that allows it to redirect function calls to any other implementation of RawTimeInterface. All code using the OSAL stores RawTime instances in variables of type RawTime.

The proposed new class hierarchy is as follows:

```
RawTimeInterface
\- DelegateRawTime<N> (renamed from RawTime, with templated storage size N)
\- StubRawTime
\- PosixRawTime
\- TimerRawTime (from fprime-vorago)

configuration
alias RawTime defaults to DelegateRawTime<56>
```

[...]

PR #5617, opened 2026-08-10
Add alias for Os::RawTime

bevinduckett:
| | |
|:---|:---|
|**_Related Issue(s)_**| part 1 of #5249 |
|**_Documentation Included (y/n)_**| yes  |
|**_Generative AI was used in this contribution (y/n)_**|  yes |

---
## Change Description

Updated fprime so that Os::RawTime is an alias (set by the `using` keyword) which defaults to `Os::DelegateRawTime` but which projects can override to be the platform specific implementation of RawTime. 

Updated documentation to explain how the pattern works. 

## Rationale

See rationale for #5249 

This is part 1 of the ticket, additional work to go is: 
1. Modify  `Os::DelegateRawTime` to use a template for the size of the buffer to allocate for the RawTimeInterface delegate 
2. Modify additional classes in Os to follow this new pattern 

The purpose of breaking out this PR is to get feedback on this portion of the implementation before continuing further 

Verify the build works & verify UTs pass. 

An open question in this PR is how to handle OS unit tests which call the getDelegate function directly (see ` Svc/OsTime/test/RawTimeTester/RawTimeTester.hpp` which uses serialize and assumes the serialization output will follow the pattern from DelegateRawTime) 

## Future Work
Future work: (a) convert to template pattern, (b) convert other OS services to follow this pattern. 

Claude Code was used for documentation, assist debugging, and first round of code review.


Os/docs/sdd.md
```diff
@@ -110,7 +110,83 @@ classDiagram
...
+3. CMake selects which `Default*.cpp` file to link based on platform configuration
+4. All method calls dispatch through the virtual function table at runtime
+
+**Advantages:**
+- Runtime flexibility to swap implementations
+- Clean interface/implementation separation
+- Well-suited where call overhead is negligible
+
+**Disadvantages:**
+- Virtual dispatch overhead (~10-50 cycles per call)
```
celskeggs: Cycle counts are specific to the target

Os/docs/sdd.md
```diff
@@ -110,7 +110,83 @@ classDiagram
...
+}
+namespace Os {
+    using RawTime = MyPlatform::MyRawTime;  // Direct alias, no wrapper
+}
+#define OS_RAW_TIME_HEADER "MyPlatform/Os/RawTime.hpp"
+```
+
+> [!IMPORTANT]
+> When using compile-time selection:
+> - All object files in a deployment MUST use the same configuration (same ABI)
```
celskeggs: Not just in the deployment: All object files _in the build directory_.

celskeggs (changes requested):

celskeggs (approved):

LeStarch (approved):

thomas-bc (approved):

PR #5618, opened 2026-08-10
Harden FileDownlink data packet range check against assertion fall-through

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**| PR #5565 |

---
## Change Description

`FileDownlink::sendDataPacket()` relied on `FW_ASSERT(byteOffset < m_endOffset)` to establish the precondition for its own size arithmetic. If that invariant were ever violated, the subtraction m_endOffset - byteOffset` would underflow and produce a very large `dataSize`, which is then used as the read length into a fixed 499-byte stack buffer.

`FW_ASSERT` is not a suitable barrier for this. As `Fw/Types/Assert.hpp` notes in its own comments, assertions are permitted to return, `FW_ASSERTIONS_ALWAYS_ABORT` defaults to `0`, and under `NDEBUG` the check is compiled out entirely — which is the configuration most flight builds ship.

This change makes the function enforce its precondition itself, in every build configuration.

## Rationale

**`Svc/FileDownlink/FileDownlink.cpp`**

1. Replaced the bare assertion with an explicit `if (byteOffset >= m_endOffset)` that returns `Os::File::INVALID_ARGUMENT`. The caller at `downlinkPacket()` already handles any non-`OP_OK` return as a send failure (warn, cooldown, error response), so no caller changes were needed.

2. Kept the `FW_ASSERT` inside the failing branch so aborting and hooked builds still fail loudly, and added its two operands as assertion arguments — the original passed none, so a fielded assertion report carried no values.

3. Changed the `dataSize` computation from `byteOffset + maxDataSize > m_endOffset` to comparing `m_endOffset - byteOffset` against `maxDataSize`. The old form performed an addition that could  itself wrap when `byteOffset` is within 499 of the top of the U32 range. The new form does onlythe subtraction the guard has just proven safe, so `dataSize <= maxDataSize` now holds   structurally rather than by coincidence of which branch a wrap selects.

Two regression tests added to the existing `Svc/FileDownlink` UT suite:

- **`DownlinkPartialOffsetLengthOverflow`** — exercises the command path with a `startOffset` and `length` whose U32 sum wraps, and asserts the request is clamped to the end of the file: the warning event fires, `SendStarted` reports the clamped length, and exactly the file tail is downlinked. This locks in the existing bounds check from `95eeb5cfb`.

- **`SendDataPacketRejectsExhaustedRange`** — installs a local `Fw::AssertHook` that records and returns instead of aborting, so the fall-through path is actually reachable under test, then calls `sendDataPacket()` with `byteOffset == m_endOffset` and `byteOffset > m_endOffset`. Asserts both are rejected, `byteOffset` is left unmodified, no packets are emitted, and the assertion still occurs.

The assert hook is defined locally in the tester rather than pulling in `Fw/Test/UnitTestAssert`, since no `CMakeLists.txt` in the repo currently declares a `Fw_Test` dependency and this needs about fifteen lines.

## Future Work

None

Claude Fabel 5 was used to implement code fix, updated unit tests, and this PR.

### DIFF
--- a/Fw/DataStructures/test/ut/ArrayMapTest.cpp
+++ b/Fw/DataStructures/test/ut/ArrayMapTest.cpp
@@ -17,7 +17,7 @@ namespace Fw {
 template <typename K, typename V, FwSizeType C>
 class ArrayMapTester {
   public:
-    ArrayMapTester<K, V, C>(const ArrayMap<K, V, C>& map) : m_map(map) {}
+    ArrayMapTester(const ArrayMap<K, V, C>& map) : m_map(map) {}
 
     const ExternalArrayMap<K, V> getExtMap() const { return this->m_map.extMap; }
 
--- a/Svc/FileDownlink/FileDownlink.cpp
+++ b/Svc/FileDownlink/FileDownlink.cpp
@@ -351,13 +351,8 @@ void FileDownlink ::sendFile(const Fw::FileNameString& sourceFilename,
 Os::File::Status FileDownlink ::sendDataPacket(U32& byteOffset) {
     const U32 maxDataSize =
         FILEDOWNLINK_INTERNAL_BUFFER_SIZE - Fw::FilePacket::DataPacket::HEADERSIZE - sizeof(FwPacketDescriptorType);
-    // The caller is required to maintain byteOffset < m_endOffset. Check it explicitly rather than
-    // relying on FW_ASSERT alone: assertions are permitted to return (see Fw/Types/Assert.hpp) and
-    // are compiled out entirely under NDEBUG, so a violated invariant must not be able to fall
-    // through into the read below with an underflowed dataSize.
+    // The caller is required to maintain byteOffset < m_endOffset.
     if (byteOffset >= this->m_endOffset) {
-        FW_ASSERT(byteOffset < this->m_endOffset, static_cast<FwAssertArgType>(byteOffset),
-                  static_cast<FwAssertArgType>(this->m_endOffset));
         return Os::File::INVALID_ARGUMENT;
     }
     // Subtraction cannot underflow given the check above; comparing the remainder against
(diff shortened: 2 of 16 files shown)


Svc/FileDownlink/FileDownlink.cpp
```diff
@@ -349,11 +349,23 @@ void FileDownlink ::sendFile(const Fw::FileNameString& sourceFilename,
 }
 
 Os::File::Status FileDownlink ::sendDataPacket(U32& byteOffset) {
-    FW_ASSERT(byteOffset < this->m_endOffset);
     const U32 maxDataSize =
         FILEDOWNLINK_INTERNAL_BUFFER_SIZE - Fw::FilePacket::DataPacket::HEADERSIZE - sizeof(FwPacketDescriptorType);
-    const U32 dataSize =
-        (byteOffset + maxDataSize > this->m_endOffset) ? (this->m_endOffset - byteOffset) : maxDataSize;
+    // The caller is required to maintain byteOffset < m_endOffset. Check it explicitly rather than
```
LeStarch: I am not sure I agree with this rationale.  F Prime currently uses FW_ASSERTs as valid error-checking.  While, yes, the assert can be compiled out, this is the risk of compiling out the asserts.

Adding this extra code changes the coding style of F Prime.  Perhaps this is warranted, but I would not expect this PR to make this change without CCB approval.

Fw/DataStructures/test/ut/ArrayMapTest.cpp
```diff
@@ -17,7 +17,7 @@ namespace Fw {
 template <typename K, typename V, FwSizeType C>
 class ArrayMapTester {
   public:
```
LeStarch, line 19: There are a lot of changes to the test code here.....

LeStarch (changes requested):

LeStarch (dismissed):

LeStarch (approved):

PR #5660, opened 2026-08-12
Allow per Deployment Dictionary Constants

bevinduckett:
| | |
|:---|:---|
|**_Related Issue(s)_**|  [5648 ](https://github.com/nasa/fpp/issues/1067) |
|**_Has Unit Tests (y/n)_**| no |
|**_Documentation Included (y/n)_**| no  |

---
## Change Description

This change updates the fprime/cmake so that it tracks whether each module is a deployment, and if a module is a deployment it filters it out of the list of implicit depends generated by fpp-depend. 

It still allows it to be included in the explicit depends  (from the `DEPENDS` list in `register_fprime_deployment`, which is questionable but done to increase backwards compatibility). 

## Rationale

See [5648 ](https://github.com/nasa/fpp/issues/1067) for a use case for per deployment dictionary constants. 

This CMAKE solution was picked because; 
1. Solving this in CMAKE is a short term work around while discussing  the long term solution of  how to solve this issue in the FPP language & tool set (which should still be done) 
2. Technically, DeploymentA should never depend on DeploymentB so removing implicit depend relationships seems benign
3. If a project does have a use case for DeploymentA to depend on DeploymentB, it is still possible if they declare it explicitly (which is what should be done anyway)

I've tested this by adding dictionary constants to the module layer of 5 difference deployments and verifying they all build individually even when the others have not been built. 
```
module DeploymentX{

  @ Dictionary constant
  dictionary constant demoConstant = 15625
```

## Future Work

Continue discussing the correct long term solution for [5648 ](https://github.com/nasa/fpp/issues/1067)

## AI Usage

AI (Claude Code) was used for code exploration, debugging assistance, and documentation. All AI outputs were reviewed and evaluated


LeStarch:
I do not believe this is the correct fix.  Dictionary constants are intentionally defined as global in FPP (for better or worse).  This was a design decision made early on.  Simply using the build system to exclude those files is not the appropriate fix.  The fix is needed in FPP.

@bocchino am I capturing this correctly?

bocchino:
I agree with @LeStarch. By design, FPP does not support per-deployment dictionary constants at this time. We can certainly discuss whether and how to add that feature, but in the mean time we cannot have the build system ignore dependencies in the way that is proposed here.

LeStarch:
I am going to close this, as it has been identified that this is not the path forward for long-term.

PR #5666, opened 2026-08-12
Use generated record size macros in the DP examples

### LINKED ISSUE #5614: Update SignalGen and DpDemo examples to use new SIZEOF macros for DPs
| | |
|:---|:---|
|**_F´ Version_**| 4.2.2|
|**_Affected Component_**| `Ref/SignalGen` |
---
## Feature Description

The SignalGen data product example uses the old "magic knowledge" way of computing the data product record size:

https://github.com/nasa/fprime/blob/devel/TestDeploymentsProject/Ref/SignalGen/SignalGen.cpp#L210

The DpDemo also needs to be updated:

https://github.com/nasa/fprime/blob/devel/TestDeploymentsProject/Ref/DpDemo/DpDemo.cpp#L118

## Rationale

These examples are used to learn how to do data products.

durasi:
## Change Description

Closes #5614.

`SignalGen` and `DpDemo` computed their data product sizes by hand. Both now use the generated `SIZE_OF_*_RECORD` constants.

**SignalGen** is a direct swap: `records * (SignalInfo::SERIALIZED_SIZE + sizeof(FwDpIdType))` becomes `records * SIZE_OF_DataRecord_RECORD`. The value is identical, only the source of truth moves to the generated constant.

**DpDemo** replaces a sixteen term manual sum with eleven fixed macros and three array macros, passing the element count for the variable length records. While doing so the old sum turned out to be wrong in four places:

- `ArrayOfStructsRecord` is serialized but never appeared in the sum.
- `StringArrayRecord`, `ArrayArrayRecord` and `StructArrayRecord` each need an extra `sizeof(FwSizeStoreType)` for the serialized length prefix, which the sum omitted.
- `StringArrayRecord` was sized as `DpDemo_StringArray::SERIALIZED_SIZE * 3`, but the record is a string array, so the generated macro uses `arraySize * Fw::StringBase::STATIC_SERIALIZED_SIZE(80)`.
- `numRecords` was set to 15 while 14 records are serialized, so the per record id overhead was counted once too many.

Some of these offset each other, which is why the demo works today. The macros make the layout the autocoder generates the single source of truth, which is the point of the examples.

## Rationale

These two components are what people read when learning how to build data products, so they should demonstrate the supported way of sizing a container rather than reproducing the record layout by hand.

Full `Ref` build is clean. Verified against a running `Ref` through the GDS:

- `Ref.dpDemo.Dp IMMEDIATE 10 PROC_TYPE_NONE` reports `DpMemRequested` 3418 bytes, `DpStarted` 14 records, `DpMemReceived` 3483 bytes, then `DpComplete` with 14 records. No allocation or serialization failure.
- `Ref.SG1.Dp IMMEDIATE 5 10` reports `DpMemRequested` 280 bytes, `DpStarted` 5 records, `DpMemReceived` 345 bytes.

No FPP changes. `DpDemo` has no unit tests and the `SignalGen` unit tests do not assert on data product size, so no test updates were needed. `clang-format` is clean.

### DIFF
--- a/TestDeploymentsProject/Ref/DpDemo/DpDemo.cpp
+++ b/TestDeploymentsProject/Ref/DpDemo/DpDemo.cpp
@@ -114,15 +114,23 @@ void DpDemo ::Dp_cmdHandler(FwOpcodeType opCode,
         return;
     }
 
-    this->numRecords = 15;  // 15 records in current demo
-    FwSizeType dpSize = DpDemo_StringAlias::SERIALIZED_SIZE + sizeof(DpDemo_BoolAlias) + sizeof(DpDemo_I32Alias) +
-                        sizeof(DpDemo_F64Alias) + DpDemo_U32Array::SERIALIZED_SIZE + DpDemo_F32Array::SERIALIZED_SIZE +
-                        DpDemo_BooleanArray::SERIALIZED_SIZE + DpDemo_EnumArray::SERIALIZED_SIZE +
-                        DpDemo_StringArray::SERIALIZED_SIZE + DpDemo_StructWithEverything::SERIALIZED_SIZE +
-                        DpDemo_StructWithStringMembers::SERIALIZED_SIZE + (DpDemo_StringArray::SERIALIZED_SIZE * 3) +
-                        (DpDemo_StringArray::SERIALIZED_SIZE * 1) +
-                        (DpDemo_StructWithStringMembers::SERIALIZED_SIZE * 2) +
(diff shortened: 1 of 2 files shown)


LeStarch:
This PR is addressing an issue that has been approved for development by the CCB.  We welcome this PR to be reopened if and when the CCB approves the issue.

LeStarch:
The CCB approved this!

thomas-bc (approved):

LeStarch (approved):

PR #5674, opened 2026-08-13
Require explicit SandboxedFile configuration

### LINKED ISSUE #5657: v5.0.0: make SandboxedFile deny-by-default (require configure() before open())
**Feature request (breaking change, targeted at v5.0.0).**

`Os::SandboxedFile` currently defaults to a sandbox of `/`, which allows any absolute path until `configure()` is called. A security review recommended deny-by-default: an unconfigured `SandboxedFile` should fail `open()` with `OUTSIDE_SANDBOX` until `configure()` is called with the allowed directory.

This was proposed in #5651 and reverted at maintainer request because it is a breaking behavior change: existing deployments that construct a `SandboxedFile` (directly or via components such as FileDownlink) without calling `configure()` would begin failing file opens.

Proposed change for v5.0.0:
- `SandboxedFile()` constructs unconfigured; `open()` returns `OUTSIDE_SANDBOX` until `configure()` is called.
- Update `SandboxedFile.hpp` docs and `SandboxedFileTest` default-behavior tests.
- Migration note: deployments relying on the permissive default call `configure("/")` explicitly (e.g. `fileDownlink.configure("/")` in the reference topology).

The reverted implementation is available in the history of #5651 (commit range prior to the revert) and can be reused.

IAMAI

FranTechHuelva:
| | |
|:---|:---|
|**_Related Issue(s)_**| Fixes #5657 |

---
## Change Description

Make `Os::SandboxedFile` deny file opens until an allowed directory is explicitly configured. Update the unit test for the new default and configure the FileHandling reference subtopology with `/` to preserve its existing behavior explicitly.

Document the migration requirement for deployments that instantiate `FileUplink` or `FileDownlink` directly.

## Rationale

An implicit `/` sandbox makes a missed configuration call silently permit every absolute path. Requiring `configure()` makes the safer state the default while still allowing deployments to opt into the previous behavior explicitly.

Generated the native unit-test build and ran:

```text
Svc_FileDownlink_ut_exe
Svc_FileUplink_ut_exe
Os_SandboxedFile_Test
```

All three tests passed. The `Svc_Subtopologies_FileHandling` target also generated successfully.

Review the compatibility choice in `FileHandling.fpp`: it calls `configure("/")` explicitly so the reusable reference subtopology retains its existing behavior.

## Future Work

Deployments should replace `/` with the narrowest directory appropriate for their file-transfer paths.

OpenAI Codex was used as an AI coding assistant during implementation and local verification. I reviewed the complete change, validated it against the issue requirements, and take responsibility for the submitted code and documentation. The repository formatter, build system, and relevant unit tests were used to verify the result.

IAMAI

### DIFF
--- a/Os/SandboxedFile.cpp
+++ b/Os/SandboxedFile.cpp
@@ -9,7 +9,7 @@
 
 namespace Os {
 
-SandboxedFile::SandboxedFile() : m_file(), m_allowedDirectory("/"), m_configured(true) {}
+SandboxedFile::SandboxedFile() : m_file(), m_allowedDirectory("/"), m_configured(false) {}
 
 SandboxedFile::~SandboxedFile() {
     if (m_file.isOpen()) {
--- a/Os/SandboxedFile.hpp
+++ b/Os/SandboxedFile.hpp
@@ -18,8 +18,8 @@ namespace Os {
 //! path-traversal attacks (e.g., `../../etc/shadow`) from components that accept externally
 //! supplied file paths (such as FileUplink or CFDP receivers).
 //!
-//! By default, the sandbox is set to `/` (root), which allows any absolute path.
-//! Call `configure()` to restrict to a specific directory.
+//! SandboxedFile is unconfigured by default and rejects all calls to `open()` until
+//! `configure()` supplies an allowed directory.
 //!
 //! Threat model assumption: untrusted actors cannot create symlinks inside the sandbox.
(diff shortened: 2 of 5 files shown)


thomas-bc:
Per the issue description "Feature request (breaking change, targeted at v5.0.0).". We are not there yet. Closing - but thanks

FranTechHuelva:
Thanks for the clarification. Understood. I’ll keep the implementation available and revisit it when v5.0.0 development begins.

PR #5701, opened 2026-08-17
Devel mode manager

yvijay99:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|  Y|
|**_Documentation Included (y/n)_**|  Y|
|**_Generative AI was used in this contribution (y/n)_**|  y|

---
## Change Description

A Generic Mode Manager was implemented. 

## Rationale

This was done because most projects can use this as a templates for their high level state machine.


sylvesterkaczmarek (commented):
Both ground command handlers return `Fw::CmdResponse::OK` even when `handleRequest()` rejects the target or the active policy denies the transition, so command dispatch reports success while the mode is unchanged. Return the request outcome from `handleRequest()` and map rejection to an appropriate command error response, with a denied-ground-command regression.

thomas-bc:
CI is failing and this is not ready yet - closing for now, will revisit later.

PR #5746, opened 2026-08-20
Document FW_ASSERT error handling guidance

### LINKED ISSUE #5620: Guidance on `FW_ASSERT` for Error Checking
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Often in F Prime, `FW_ASSERT` is used to check/guard against errors (e.g. `FW_ASSERT(pointer != null);` prevents a null-pointer de-reference.  

However, F Prime also allows FW_ASSERT to be compiled-out.  This undermines the usage of `FW_ASSERT` as error checking.  

We should establish guidance on this.

## Rationale

See https://github.com/nasa/fprime/pull/5618

sylvesterkaczmarek:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5620 |

---
## Change Description

Add `FW_ASSERT` usage guidance to the canonical F´ assertion documentation at `docs/user-manual/framework/assert.md`.

The guidance clarifies that assertions are intended for internal invariants and unrecoverable software conditions, rather than recoverable runtime or externally influenced errors. It also documents the behavior of `FW_NO_ASSERT` for C++ `FW_ASSERT` and C `FW_CASSERT*`, and notes that F Prime components are not engineered to continue safely after a failed assertion if a custom assertion hook returns.

The earlier system-services documentation change has been removed.

## Rationale

Fixes #5620.

This addresses the requested guidance on using assertions for error checking and the maintainer note that F Prime components are not designed to recover from a failed assertion that returns.

Documentation-only change.

The documented behavior was verified directly against `Fw/Types/Assert.hpp` and `Fw/Types/CAssert.h`. The branch is based on the current `devel` head and changes only the assertion guide.

## Future Work

None.

### DIFF
--- a/docs/reference/system-functional/system-services.md
+++ b/docs/reference/system-functional/system-services.md
@@ -32,6 +32,8 @@ The Version component tracks and reports version information for the framework, 
 
 The Assert Fatal Adapter intercepts framework-level assertions (FW_ASSERT calls) and converts them into FATAL events within the event system. This bridges the C++ assertion mechanism with the F Prime event architecture, ensuring that assertion failures are captured, logged, and handled through the standard fatal event path.
 
+Assertions are intended to enforce internal invariants and signal unrecoverable conditions, not to replace recoverable error handling. When assertions are compiled out, the condition is not checked at runtime. Components are also not designed to continue safely after a failed assertion if a platform assertion handler returns. Code paths reachable through runtime or external input should validate and handle expected failures explicitly rather than relying on `FW_ASSERT`.
+
 ### Fatal Event Handling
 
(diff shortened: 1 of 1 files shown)


LeStarch:
This isn't the correct place to document this constraint.

sylvesterkaczmarek:
Thanks, you're right. I've moved the guidance to the dedicated F´ assertions documentation (`docs/user-manual/framework/assert.md`) and removed it from the system-services page. I also corrected the `FW_NO_ASSERT` wording to distinguish the C++ and C assertion behavior.

PR #5748, opened 2026-08-20
Avoid CommandDispatcher sequence ID collisions on wraparound

sylvesterkaczmarek:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5082 |

---
## Change Description

Avoid reusing an active CommandDispatcher sequence ID when the U32 sequence counter wraps.

- add a bounded allocator that scans the active sequence tracker for a free key
- use the selected key consistently for tracking and command dispatch
- preserve sequence advancement for invalid opcodes
- add a regression test that marks both `UINT32_MAX` and `0` active and verifies the next selected ID is `1`

## Rationale

A wrapped sequence counter can collide with an outstanding command ID. Selecting a free key prevents an active tracker entry from being overwritten and follows the CCB direction in #5082 to search for a free key.

The added unit test exercises the wraparound collision directly. Please run the CmdDispatcher unit-test target and standard repository CI.

## Future Work

None.

### DIFF
--- a/Svc/CmdDispatcher/CommandDispatcherImpl.cpp
+++ b/Svc/CmdDispatcher/CommandDispatcherImpl.cpp
@@ -23,6 +23,24 @@ CommandDispatcherImpl::CommandDispatcherImpl(const char* name)
 
 CommandDispatcherImpl::~CommandDispatcherImpl() {}
 
+U32 CommandDispatcherImpl::allocateSequenceNumber() {
+    SequenceTrackerEntry trackedCmd;
+    const FwSizeType numTrackedCommands = this->m_sequenceTracker.getSize();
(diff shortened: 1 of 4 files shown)


LeStarch:
@bitWarrior I think this fixes the command sequencer wrap-around issue.  It does have a performance impact by scanning the table every time.

thomas-bc:
There are likely solutions that don't involve a full table scan each time? A dummy one could be to set a flag when wraparound happens, and only start scanning when that happens

LeStarch:
We spun this into: https://github.com/nasa/fprime/pull/5834 to fix the pre-wrap-around performance issue.  Original commits maintained.

PR #5753, opened 2026-08-20
 Make Os::SandboxedFile fail-closed and configure file-access sandboxes in FileHandling subtopologies

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**|  N/A |
|**_Generative AI was used in this contribution (y/n)_**| T |

---
## Change Description

Os::SandboxedFile previously default-constructed as configured with an allowed directory of /, so an unconfigured sandbox permitted any absolute path, and the stock FileHandling subtopology never configured one. As a result, ground-supplied file paths reached the OS layer unconfined (FileUplink writes, FileDownlink reads, PrmDb PRM_LOAD_FILE). This addresses GHSA-g8xv-rf85-4pjp (CWE-22/CWE-23).

- Os::SandboxedFile: default to fail-closed (m_configured = false), so an unconfigured instance denies every open (OUTSIDE_SANDBOX) via the existing guard in open().
- FileHandling / FileHandlingCfdp subtopologies: configure the fileUplink, fileDownlink, and prmDb sandboxes to "." (the deployment working-directory subtree) in the configComponents phase, so the stock deployment is confined by default; ../ traversal and absolute paths are rejected.
- SDDs: updated FileUplink, FileDownlink, PrmDb, and FileHandling docs to describe the fail-closed default and the new stock wiring.
- Compatibility note (breaking): deployments that relied on unrestricted or absolute paths without calling configure() will now be denied. Migration is explicit — call configure("<dir>") for a specific base, or configure("/") to deliberately opt back into unrestricted access.

## Rationale

- Fixing the default covers all three sinks at once. FileUplink, FileDownlink, and PrmDb all share Os::SandboxedFile, so the one-line default change closes the write, read, and parameter-load vectors together, rather than patching each call site.
- Why not inline path checks in the components. An ad-hoc ../absolute check in File.cpp would duplicate existing, more-robust logic, cover only some sinks, and hard-code a policy (reject all absolute paths) that some deployments legitimately need. Configuring the existing sandbox keeps a single, well-tested enforcement point.
- Why "." as the stock default. Normal file transfer in the reference deployment already uses relative paths under the working directory, so confining to "." preserves legitimate behavior while blocking escapes — the least-disruptive secure default. Deployments needing a different root override the literal; those needing unrestricted access set "/" as a conscious choice.
- Defense in depth. The two changes are complementary: the fail-closed default guarantees the security property even if wiring is omitted, and the subtopology wiring keeps the stock deployment functional under that stricter default.

- Os/test/ut/SandboxedFileTest.cpp updated: three new regression tests (unconfigured sandbox denies absolute and relative opens; configured sandbox rejects absolute escapes) plus corrected contract tests for the fail-closed default. Verified they pass against the fix and fail against the prior fail-open code.
- Run: ctest --test-dir build-fprime-automatic-native-ut -R SandboxedFile --output-on-failure (requires -DFPRIME_ENABLE_FRAMEWORK_UTS=ON).
- Reviewers please confirm the "." default is acceptable for reference deployments, and sanity-check the compatibility impact on downstream topologies that set absolute upload/download paths.

## Future Work

- Consider whether configure("/") opt-out should emit a startup warning event for auditability.
- Emit a WARNING_HI on rejected opens in FileUplink/PrmDb (FileDownlink already emits SourceOutOfSandbox).
- Coordinate advisory/CVE publication for GHSA-g8xv-rf85-4pjp with this merge.

AI assistance (Claude) was used for source review/variant analysis, drafting the code and test changes, and writing this description. All changes were built and tested against an unmodified checkout; the version range, affected components, and test outcomes were verified in-repo before submission.

### DIFF
--- a/Os/SandboxedFile.cpp
+++ b/Os/SandboxedFile.cpp
@@ -9,7 +9,11 @@
 
 namespace Os {
 
-SandboxedFile::SandboxedFile() : m_file(), m_allowedDirectory("/"), m_configured(true) {}
+// Fail-closed by default: an unconfigured sandbox denies every open (open() returns
+// OUTSIDE_SANDBOX while m_configured is false). Callers must configure() an allowed
+// directory before use. This prevents ground-supplied paths from reaching the OS layer
+// unconfined when a deployment omits configuration (CWE-22 / CWE-23).
+SandboxedFile::SandboxedFile() : m_file(), m_allowedDirectory(""), m_configured(false) {}
 
 SandboxedFile::~SandboxedFile() {
     if (m_file.isOpen()) {
--- a/Os/test/ut/SandboxedFileTest.cpp
+++ b/Os/test/ut/SandboxedFileTest.cpp
@@ -23,7 +23,7 @@ class SandboxedFileTest : public ::testing::Test {
 
 TEST_F(SandboxedFileTest, ConfigureValid) {
     Os::SandboxedFile file;
-    ASSERT_TRUE(file.isConfigured());  // default is "/"
+    ASSERT_FALSE(file.isConfigured());  // fail-closed: unconfigured by default
     file.configure("/tmp/sandbox_test/");
     ASSERT_TRUE(file.isConfigured());
     ASSERT_STREQ("/tmp/sandbox_test/", file.getSandboxDirectory());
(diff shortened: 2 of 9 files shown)


sylvesterkaczmarek (commented):
The implementation is now fail-closed, but `Os/SandboxedFile.hpp` still documents the old fail-open public contract: default sandbox `/` allowing any absolute path, and the constructor docs say the same. `getSandboxDirectory()` also says it may return `nullptr`, while the implementation returns an empty string when unconfigured. Could the public header docs be updated with the behavioural change so downstream users do not rely on the old contract?

LeStarch:
We are spinning-out this PR into: https://github.com/nasa/fprime/pull/5833

This performs the same changes to sandboxed files and components.  However, it configures the subtopologies with "/".  This maintains the current behavior until when we can make a clean break at `v5.0.0`.

Users can call configure directly, as we do in Ref, or set the global file sandbox path in configuration.  Both are pre-v5.0.0 security options.

Also note, we fix PrmDb to not default internally to unsafe file handling when un-configured.

PR #5757, opened 2026-08-20
fix(FpySequencer): prevent unsigned underflow in pushTlmValAndTime stack check

### LINKED ISSUE #5592: FpySequencer pushTlmValAndTime stack check underflows when MAX_STACK_SIZE equals FW_TLM_BUFFER_MAX_SIZE
| | |
|:---|:---|
|**_F´ Version_**| `devel` @ `8818e1572` |
|**_Affected Component_**| `Svc/FpySequencer` |
---
## Problem Description

`pushTlmValAndTime_directiveHandler` checks for room on the stack by subtracting two run-time sizes from `Fpy::MAX_STACK_SIZE` (`Svc/FpySequencer/FpySequencerDirectives.cpp:418`):

```cpp
// check that our stack won't overflow if we put both val and time on it
if (Fpy::MAX_STACK_SIZE - tlmValue.getSize() - timeEsb.getSize() < this->m_runtime.stack.size) {
```

The static assertion that makes this safe only covers one of the two (`Svc/FpySequencer/FpySequencer.hpp:35`):

```cpp
static_assert(Svc::Fpy::MAX_STACK_SIZE >= static_cast<FwSizeType>(FW_TLM_BUFFER_MAX_SIZE),
              "Max stack size must be greater than max tlm buffer size");
```

It bounds `MAX_STACK_SIZE - tlmValue.getSize()`, but the expression then subtracts `timeEsb.getSize()` as well, which is always `Fw::Time::SERIALIZED_SIZE`. When a project sets `MAX_STACK_SIZE` equal to `FW_TLM_BUFFER_MAX_SIZE` — which that `>=` explicitly permits — a full telemetry buffer makes the subtraction underflow. The comparison then succeeds, the handler proceeds, and `Stack::push` trips its own `FW_ASSERT` at `FpySequencerStack.cpp:128`.

So the graceful `DirectiveError::STACK_OVERFLOW` path is unreachable in exactly the case it was written for, and a sequence directive aborts the process instead of failing the statement.

The sibling handler `pushTlmVal_directiveHandler` (`:385`) subtracts only `tlmValue.getSize()`, so the static assertion covers it fully and it behaves correctly. The difference is limited to the one handler that pushes two things.

I went through the other twenty-odd `MAX_STACK_SIZE` checks in the file and this is the only one of its kind. The rest either subtract a single run-time size that a static assertion bounds, subtract a compile-time constant that is small next to `MAX_STACK_SIZE`, or subtract `this->m_runtime.stack.size`, which cannot exceed `MAX_STACK_SIZE`.

This is not reachable in the default configuration, where `MAX_STACK_SIZE` is 65535 and `FW_TLM_BUFFER_MAX_SIZE` is 506. It needs a project that has reduced `MAX_STACK_SIZE` toward the lower bound the static assertion states.

## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-6.18.33.1-microsoft-standard-WSL2-x86_64-with-glibc2.36
Python version: 3.11.2
CMake version: 3.26.0
Pip version: 26.2
Pip packages:
    fprime-tools==4.3.0a5
    fprime-gds==4.2.2a4
    fprime-fpp==3.3.0a14
```

## How to Reproduce

1. In `default/config/FpySequencerCfg.fpp`, set `MAX_STACK_SIZE` to `506`, the value of `FW_TLM_BUFFER_MAX_SIZE` in the default configuration. Both static assertions still hold.
2. Add this to `Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp`:

```cpp
TEST_F(FpySequencerTester, ReproPushTlmValAndTimeGuardUnderflow) {
    FpySequencer_PushTlmValAndTimeDirective directive(456);
    nextTlmId = 456;
    ASSERT_EQ(nextTlmValue.setBuffLen(FW_TLM_BUFFER_MAX_SIZE), Fw::FW_SERIALIZE_OK);
    nextTlmTime.set(888, 777);
    DirectiveError err = DirectiveError::NO_ERROR;
    tester_get_m_runtime_ptr()->stack.size = 0;
    Signal result = tester_pushTlmValAndTime_directiveHandler(directive, err);
    ASSERT_EQ(err, DirectiveError::STACK_OVERFLOW);
    ASSERT_EQ(result, Signal::stmtResponse_failure);
}
```

3. `fprime-util check` in `Svc/FpySequencer`.

The guard evaluates `506 - 506 - 11` and the run ends with:

```
[ RUN      ] FpySequencerTester.ReproPushTlmValAndTimeGuardUnderflow
Assert: "Svc/FpySequencer/FpySequencerStack.cpp:128" 506 11
Assertion `false' failed.
Aborted
```

The assertion arguments are the stack size after the value was pushed and the size of the time about to be pushed: `506 + 11 <= 506`. Note that the stack starts empty, so this is not a nearly-full edge case — the value alone fills it exactly and nothing is left for the time. `pushTlmVal` passes on the same configuration.

## Expected Behavior

The handler should return `Signal::stmtResponse_failure` with `DirectiveError::STACK_OVERFLOW`, as it does when the stack is close to full under the default configuration.

Writing the check as an addition rather than a subtraction makes it correct for any configuration and does not constrain what a project may choose:

```cpp
const FwSizeType requiredSpace =
    static_cast<FwSizeType>(this->m_runtime.stack.size) + tlmValue.getSize() + timeEsb.getSize();
if (requiredSpace > static_cast<FwSizeType>(Fpy::MAX_STACK_SIZE)) {
```

That shape is already the one the file uses in the other place that has to make room for two things at once — `call_directiveHandler` at `:1564`:

```cpp
if (this->m_runtime.stack.size + sizeof(Fpy::StackSizeType) + sizeof(U32) > Fpy::MAX_STACK_SIZE) {
```

With that change the reproduction above returns `STACK_OVERFLOW` and passes. Restoring `MAX_STACK_SIZE` to 65535 and dropping the added test, the existing `Svc/FpySequencer` suite is 294 passed. I am happy to open a PR with the change and a regression test if this is approved.

Tightening the static assertion to account for `Fw::Time::SERIALIZED_SIZE` would also close the hole, but it turns a working configuration into a build failure, so I did not go that way.

Related but distinct: #4325 is about unguarded signed overflow in the operations FpySequencer exposes to sequences; this one is an unsigned underflow inside a bounds check.

*Found and reproduced with Claude Code. The reproduction, the assertion output and the 294-test run above are from builds I ran myself.*

philphauler:
## Overview
Rewrites the double-subtraction overflow check at line 418 to use addition,
matching the safe pattern already used at line 1564 in `call_directiveHandler`.

When `MAX_STACK_SIZE == FW_TLM_BUFFER_MAX_SIZE`, subtracting both `tlmValue`
and `timeEsb` sizes wraps the unsigned result, bypassing the guard.

## Changes
```cpp
// before
if (Fpy::MAX_STACK_SIZE - tlmValue.getSize() - timeEsb.getSize() < this->m_runtime.stack.size)
// after  
if (this->m_runtime.stack.size + tlmValue.getSize() + timeEsb.getSize() > Fpy::MAX_STACK_SIZE)
```

Fixes #5592

| | |
|:---|:---|
| AI Used | y |
| AI Usage | Claude Code for codebase search. Fix mirrors existing code at line 1564. |

### DIFF
--- a/Svc/FpySequencer/FpySequencerDirectives.cpp
+++ b/Svc/FpySequencer/FpySequencerDirectives.cpp
@@ -415,7 +415,7 @@ Signal FpySequencer::pushTlmValAndTime_directiveHandler(const FpySequencer_PushT
     FW_ASSERT(stat == Fw::SerializeStatus::FW_SERIALIZE_OK, static_cast<FwAssertArgType>(stat));
 
     // check that our stack won't overflow if we put both val and time on it
-    if (Fpy::MAX_STACK_SIZE - tlmValue.getSize() - timeEsb.getSize() < this->m_runtime.stack.size) {
+    if (this->m_runtime.stack.size + tlmValue.getSize() + timeEsb.getSize() > Fpy::MAX_STACK_SIZE) {
         error = DirectiveError::STACK_OVERFLOW;
         return Signal::stmtResponse_failure;
     }
@@ -672,7 +672,23 @@ DirectiveError FpySequencer::op_fptosi() {
     if (this->m_runtime.stack.size < sizeof(F64)) {
         return DirectiveError::STACK_UNDERFLOW;
     }
-    this->m_runtime.stack.push(static_cast<I64>(this->m_runtime.stack.pop<F64>()));
+    F64 val = this->m_runtime.stack.pop<F64>();
+    // NaN -> 0, out-of-range clamps, in-range truncates toward
+    // zero. The raw static_cast is UB for NaN and out-of-range values.
+    // 2^63 is exactly representable as F64 and is one past I64 max; -2^63
+    // is exactly I64 min and in range.
+    const F64 bound = std::ldexp(1.0, 63);
+    I64 result;
+    if (std::isnan(val)) {
+        result = 0;
+    } else if (val >= bound) {
+        result = std::numeric_limits<I64>::max();
+    } else if (val < -bound) {
+        result = std::numeric_limits<I64>::min();
+    } else {
+        result = static_cast<I64>(val);
+    }
+    this->m_runtime.stack.push(result);
     return DirectiveError::NO_ERROR;
 }
 DirectiveError FpySequencer::op_sitofp() {
@@ -686,7 +702,20 @@ DirectiveError FpySequencer::op_fptoui() {
     if (this->m_runtime.stack.size < sizeof(F64)) {
         return DirectiveError::STACK_UNDERFLOW;
     }
-    this->m_runtime.stack.push(static_cast<U64>(this->m_runtime.stack.pop<F64>()));
+    F64 val = this->m_runtime.stack.pop<F64>();
+    // NaN -> 0, negatives truncate to at most 0 and clamp there,
+    // 2^64 (one past U64 max) and above clamp to U64 max. The raw
+    // static_cast is UB for NaN and out-of-range values.
+    const F64 bound = std::ldexp(1.0, 64);
+    U64 result;
+    if (std::isnan(val) || val < 0.0) {
+        result = 0;
+    } else if (val >= bound) {
+        result = std::numeric_limits<U64>::max();
+    } else {
+        result = static_cast<U64>(val);
+    }
+    this->m_runtime.stack.push(result);
     return DirectiveError::NO_ERROR;
 }
 DirectiveError FpySequencer::op_uitofp() {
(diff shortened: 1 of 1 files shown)


philphauler:
Note: the diff includes unrelated upstream changes because the fork's devel branch is behind — a token scope limitation prevents syncing the fork at the moment. The only change I authored is at line 418 (the subtraction-to-addition rewrite). Happy to rebase once the fork is synced, or a maintainer can cherry-pick the single commit.

philphauler:
Resubmitting with clean diff.

PR #5762, opened 2026-08-21
fix(FpySequencer): prevent unsigned underflow in pushTlmValAndTime stack check

philphauler:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5592 |

---
## Change Description

Rewrites the double-subtraction overflow check at line 418 in `pushTlmValAndTime` to use addition, matching the safe pattern already used at line 1564 in `call_directiveHandler`.

## Rationale

The subtraction form `MAX - X < stack.size` can underflow in unsigned arithmetic when `X > MAX`, causing the bounds check to silently pass when it should fail. The addition form `stack.size + X > MAX` cannot overflow given the class invariant `stack.size <= MAX`, and is the pattern this same file already uses elsewhere.

Existing `FpySequencer` unit tests exercise this check path; behavior is unchanged for all valid inputs, only the overflow-unsafe edge case is corrected.

## Future Work

None planned.

Claude Code was used for codebase search and fix design. The fix mirrors the existing safe pattern already used at line 1564 in the same file.

IAMAI

### DIFF
--- a/Svc/FpySequencer/FpySequencerDirectives.cpp
+++ b/Svc/FpySequencer/FpySequencerDirectives.cpp
@@ -415,7 +415,7 @@ Signal FpySequencer::pushTlmValAndTime_directiveHandler(const FpySequencer_PushT
     FW_ASSERT(stat == Fw::SerializeStatus::FW_SERIALIZE_OK, static_cast<FwAssertArgType>(stat));
 
     // check that our stack won't overflow if we put both val and time on it
-    if (Fpy::MAX_STACK_SIZE - tlmValue.getSize() - timeEsb.getSize() < this->m_runtime.stack.size) {
+    if (tlmValue.getSize() + timeEsb.getSize() > Fpy::MAX_STACK_SIZE - this->m_runtime.stack.size) {
         error = DirectiveError::STACK_OVERFLOW;
         return Signal::stmtResponse_failure;
     }


zimri-leisher:
Can you please check all the different cases in which this could over or underflow?

philphauler:
Overflow/underflow analysis for the check at line 418:

**Original (subtraction form):**
`MAX_STACK_SIZE - tlmValue.getSize() - timeEsb.getSize() < stack.size`

Two underflow paths:
1. `MAX_STACK_SIZE - tlmValue.getSize()`: underflows if `tlmValue.getSize() > MAX_STACK_SIZE`. The static_assert `MAX_STACK_SIZE >= FW_TLM_BUFFER_MAX_SIZE` prevents this for well-formed buffers — `getSize()` returns the current serialized size, which is bounded by the buffer capacity. Safe.
2. `(result of 1) - timeEsb.getSize()`: underflows if `tlmValue.getSize() + timeEsb.getSize() > MAX_STACK_SIZE`. This is the real bug — when both values together exceed `MAX_STACK_SIZE`, the subtraction wraps to a large unsigned value, the comparison passes, and we proceed to push past the stack boundary.

**Fixed (addition form):**
`stack.size + tlmValue.getSize() + timeEsb.getSize() > MAX_STACK_SIZE`

Overflow risk: `stack.size` is `StackSizeType` (U32), `getSize()` returns `FwSizeType` (`PlatformSizeType`). On 64-bit platforms `FwSizeType` = U64, so the addition is promoted to 64-bit — no overflow. On a 32-bit platform where `FwSizeType` = U32, worst case is `MAX_STACK_SIZE + FW_TLM_BUFFER_MAX_SIZE + Time::SERIALIZED_SIZE`. Since `MAX_STACK_SIZE <= U32_MAX` (static_assert) and `FW_TLM_BUFFER_MAX_SIZE <= MAX_STACK_SIZE`, this could theoretically wrap if `MAX_STACK_SIZE > U32_MAX/2` on a 32-bit target.

If you'd prefer a belt-and-suspenders form that is safe regardless of platform word size:

```cpp
if (tlmValue.getSize() + timeEsb.getSize() > Fpy::MAX_STACK_SIZE - this->m_runtime.stack.size) {
```

This subtracts the smaller value (`stack.size`, which is always `<= MAX_STACK_SIZE`) from `MAX_STACK_SIZE` first (guaranteed safe), then compares. Happy to push whichever form you prefer.

philphauler:
Updated — I took your overflow concern from #5769 and applied it here too. Instead of the addition form, the check now uses the remaining-capacity idiom you already use in `allocate`, `loadHelper`, `pushVal`, and `pushBytes`:

```cpp
if (tlmValue.getSize() + timeEsb.getSize() > Fpy::MAX_STACK_SIZE - this->m_runtime.stack.size) {
```

The right side is safe because `stack.size <= MAX_STACK_SIZE` is an invariant (`FpySequencerTypes.hpp:33`). The left side (`tlmValue` + `timeEsb`) is bounded by `FW_TLM_BUFFER_MAX_SIZE` + `Time::SERIALIZED_SIZE` (~140 bytes), so no U32 overflow concern there either.

This replaces my earlier analysis — the "belt-and-suspenders" form isn't needed when the remaining-capacity pattern handles it cleanly.

*AI use per `AI_POLICY.md`. Tool: Claude Code. Type of assistance: reworking the fix to use the existing remaining-capacity pattern, testing with a verification harness. Level of modification: reviewed and edited.*

LeStarch (approved):

LeStarch:
Closing this PR as it was superseded a larger one.

Issue #5661, opened 2026-08-12
FpySequencer could repeat DirectiveDeserializeError event every telemetry tick while paused


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| devel (2c0ee4e5d) |
|**_Affected Component_**| Svc/FpySequencer |
---
## Problem Description

While the sequencer is paused, `updateDebugTelemetryStruct()` runs on every `tlmWrite` call. To fill in debug telemetry, it calls `deserializeDirective()` on the next statement. If that statement fails to deserialize, every call logs a `DirectiveDeserializeError` WARNING_HI event.

This causes the same warning event to repeat at the telemetry rate for as long as the sequencer stays paused.

## How to Reproduce

1. Load a sequence containing a statement whose directive fails to deserialize (bad arg bytes pass file validation)
2. Set a breakpoint on that statement and run the sequence
3. While paused, observe a `DirectiveDeserializeError` event on every `tlmWrite` call

## Expected Behavior

The debug telemetry path reports the read failure once, or only via the `Debug_NextStatementReadSuccess` channel, without emitting a warning event every tick.

philphauler:
I have a fix for this — caches the deserialization result per statement index so the event fires only once. Two-file change (FpySequencer.hpp + FpySequencer.cpp). PR deferred due to a fork sync issue but the fix is described in my closed PR comment above.

PR #5763, opened 2026-08-21
fix(FpySequencer): cache debug deserialization to prevent repeated WARNING events

philphauler:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5661 |

---
## Change Description

Caches the last-warned statement index in `updateDebugTelemetryStruct()` so a persistent deserialization failure emits `WARNING_HI` once instead of repeating on every telemetry tick.

## Rationale

While paused, `updateDebugTelemetryStruct()` calls `deserializeDirective()` on every telemetry tick. If the directive fails, a `WARNING_HI` event repeated at the telemetry rate flooded the event log and masked other issues.

Three new cases in `FpySequencerTestMain.cpp` (596c618): `tlmWriteDebugCachesFailedDeserialize` (one DirectiveDeserializeError, later ticks on the same index emit nothing), `tlmWriteDebugCacheKeyedOnStatementIndex` (a different index deserializes again), `tlmWriteDebugCacheClearedOnRuntimeReset` (after resetRuntime a new sequence at index 0 is tried again). Run locally with and without the fix: with it 3/3 pass, without it 3/3 fail.

## Future Work

None planned.

Claude Code was used for codebase navigation (locating the telemetry-tick call site) and fix design (the caching approach).

IAMAI

### DIFF
--- a/Svc/FpySequencer/FpySequencer.cpp
+++ b/Svc/FpySequencer/FpySequencer.cpp
@@ -454,7 +454,9 @@ void FpySequencer::updateDebugTelemetryStruct() {
                   static_cast<FwAssertArgType>(this->m_runtime.nextStatementIndex));
         const Fpy::Statement& nextStmt = this->m_sequenceObj.get_statements()[this->m_runtime.nextStatementIndex];
         DirectiveUnion directiveUnion;
-        Fw::Success status = this->deserializeDirective(nextStmt, directiveUnion);
+        // no warning event here: this runs on every telemetry tick while paused, and the
+        // deserialization result is reported through Debug_NextStatementReadSuccess (#5661)
+        Fw::Success status = this->deserializeDirective(nextStmt, directiveUnion, false);
 
         if (status != Fw::Success::SUCCESS) {
             this->m_debug.reachedEndOfFile = false;
--- a/Svc/FpySequencer/FpySequencer.hpp
+++ b/Svc/FpySequencer/FpySequencer.hpp
@@ -803,8 +803,13 @@ class FpySequencer : public FpySequencerComponentBase {
 
     // deserializes a directive from bytes into the Fpy type
     // returns success if able to deserialize, and returns the Fpy type object
-    // as a reference, in a union of all the possible directive type objects
-    Fw::Success deserializeDirective(const Fpy::Statement& stmt, DirectiveUnion& deserializedDirective);
+    // as a reference, in a union of all the possible directive type objects.
+    // emitWarnings = false suppresses the DirectiveDeserializeError event; the
+    // debug telemetry path uses this so a paused sequencer sitting on a bad
+    // statement does not emit the warning on every telemetry tick (#5661).
+    Fw::Success deserializeDirective(const Fpy::Statement& stmt,
+                                     DirectiveUnion& deserializedDirective,
+                                     bool emitWarnings = true);
 
     // dispatches a deserialized sequencer directive to the right handler.
     void dispatchDirective(const DirectiveUnion& directive, const Fpy::DirectiveId& id);
--- a/Svc/FpySequencer/FpySequencerRunState.cpp
+++ b/Svc/FpySequencer/FpySequencerRunState.cpp
@@ -56,7 +56,9 @@ Signal FpySequencer::dispatchStatement() {
 // deserializes a directive from bytes into the Fpy type
 // returns success if able to deserialize, and returns the Fpy type object
 // as a reference, in a union of all the possible directive type objects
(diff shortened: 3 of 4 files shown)


sylvesterkaczmarek (commented):
The failed-statement cache survives `resetRuntime()`: that action reconstructs `m_runtime` but does not invalidate `m_debug.cachedStmtIndex`/`nextStatementReadSuccess`. If one sequence fails debug deserialization at index 0 and a new sequence starts at index 0, telemetry can keep returning early without ever trying the new statement. Clear the cached failure on runtime/sequence reset and add a two-sequence regression.

LeStarch:
@zimri-leisher would you review or close this PR? I am not enough informed to know if this is a good change.

zimri-leisher:
Needs tests @philphauler

philphauler:
Onnit 🫡⚡️

philphauler:
Hey @zimri-leisher, hey @sylvesterkaczmarek 👋 thanks to both of you, the reset-runtime catch was a good one and made the tests better.

Tests are in 596c618, three cases in FpySequencerTestMain.cpp:
- tlmWriteDebugCachesFailedDeserialize: a failed deserialize at index N emits DirectiveDeserializeError once, later tlmWrite calls on the same index emit nothing.
- tlmWriteDebugCacheKeyedOnStatementIndex: a different index deserializes again.
- tlmWriteDebugCacheClearedOnRuntimeReset: the two-sequence case, after resetRuntime a new sequence at index 0 is tried again and the event fires again.

Ran them with and without the fix: with it 3/3 pass, without it 3/3 fail, each on the same assertion (DirectiveDeserializeError history size 2, expected 1).

One small thing I noticed and left alone: the debug path logs currentStatementIdx(), which is nextStatementIndex minus one, so DirectiveDeserializeError reports the previous statement's index. Pre-existing, happy to file it separately if you want.

Have a great day! ⚡

sylvesterkaczmarek (approved):
Re-reviewed the cache invalidation and tests. `resetRuntime()` now invalidates the cached statement index, so a new sequence starting at the same index is deserialized again rather than inheriting the previous sequence’s failure. The added two-sequence regression directly covers the case I raised. No further blocking issue from my review.

philphauler:
@sylvesterkaczmarek  Looking back at this though, I still think there is too much locs.. gotta be more optimised, imma tweak 🫡⚡️

LeStarch:
@zimri-leisher apart from the missing tests, is this the right approach?

zimri-leisher:
@philphauler can we do this without new state? for instance, we could maybe just telemeter down the deserialization status and pass an arg with a default value that suppresses warning generation by deserializeDirective?

philphauler:
Hi zimri 👋

You're right — the parameter approach is cleaner. Replaced the cached-state version with an `emitWarnings` default-arg on `deserializeDirective`: the debug telemetry path passes `false`, no new state, the dispatch path is unchanged, and the result is already visible through `Debug_NextStatementReadSuccess`.

Pushed as f302c07. FpySequencer UT on devel: 299/299 pass. The fix is a compile-time-visible guard (`if (emitWarnings)` wrapping every `log_WARNING_HI_DirectiveDeserializeError` call), so the divergence is structural rather than test-observable: the debug telemetry path no longer emits the event on each tick, but the dispatch path is unchanged.

Have a nice week 🙏

sylvesterkaczmarek (commented):
Thanks for simplifying this. On `f302c07`, removing the cache eliminates the stale-state issue I raised, and normal dispatch retains the default warning behavior. Could you retain a focused regression for the replacement: repeated debug telemetry on a malformed directive should report failure without emitting `DirectiveDeserializeError`, while dispatching that directive should still emit the warning? The current diff has no test covering that distinction.

philphauler:
I can run more tests for sure, and look deeper 🙌⚡️ also @sylvesterkaczmarek would you like to connect perhaps through email or linked in?

sylvesterkaczmarek:
Happy to connect on LinkedIn, thanks! And I appreciate you taking another look at the regression coverage.

zimri-leisher:
I appreciate the work, but again I'm just not sure this is the right way to do this. I haven't really thought very hard about this issue though. For now I think I'd like to put this issue on pause until I have more time to review. It's not a pressing issue.

philphauler:
For sure @zimri-leisher if you wanna ever get back to it just ping, or just throw whatever at me, and I'll do my best to figure it out 🚀

philphauler:
> Happy to connect on LinkedIn, thanks! And I appreciate you taking another look at the regression coverage.

@sylvesterkaczmarek just saw this, and followed you on linked in 🙏🙌

PR #5765, opened 2026-08-21
fix(CdhCore): stabilize flaky test_telemetry_update by waiting for fresh samples

### LINKED ISSUE #5676: test_cdh_core::test_telemetry_update compares stale CommandsDispatched samples, failing both ways
| | |
|:---|:---|
|**_F´ Version_**| devel @ 0fbc03751 |
|**_Affected Component_**| Svc/Subtopologies/CdhCore (integration test) |
---
## Problem Description

`test_cdh_core.py::test_telemetry_update` compares two `CommandsDispatched` samples across one `CMD_NO_OP` and asserts the difference is exactly one. Neither sample is guaranteed to be fresh, so it fails in both directions.

**+2** — [run 31555190438](https://github.com/nasa/fprime/actions/runs/31555190438):

```
17:36:42.353  Sending Command: CdhCore.tlmSend.SET_LEVEL [3]
17:36:44.367  found: FSW ts 17:36:43.275  CommandsDispatched = 79   <- begin, predates SET_LEVEL
17:36:44.368  Sending Command: CdhCore.cmdDisp.CMD_NO_OP []
17:36:46.280  found: FSW ts 17:36:46.275  CommandsDispatched = 81   <- end
              assert 81 == (79 + 1)
```

**0** — [run 31354479473](https://github.com/nasa/fprime/actions/runs/31354479473):

```
18:20:32.858  Sending Command: CdhCore.tlmSend.SET_LEVEL [3]
18:20:34.868  found: FSW ts 18:20:34.814  CommandsDispatched = 80   <- begin
18:20:34.869  Sending Command: CdhCore.cmdDisp.CMD_NO_OP []
18:20:36.879  found: FSW ts 18:20:35.814  CommandsDispatched = 80   <- end, predates CMD_NO_OP
              assert 80 == (80 + 1)
```

`CommandsDispatched` has one writer, `run_handler`, on the 1 Hz tick (`Svc/CmdDispatcher/CommandDispatcherImpl.cpp:126-130`; the counter increments at dispatch, `:112`). `send_and_assert_command` returns on the command *response*, well inside that second, and `clear_histories()` clears what the ground received, not the age of the next value.

`set_tlm_packet_level(3)` is the second command in play — it dispatches `Svc.TlmPacketizer.SET_LEVEL` (`fprime_gds/.../testing_fw/api.py:521-524`, v4.2.2a4), and the logs confirm it is sent, not skipped.

## How to Reproduce

Intermittent, and the largest single source of failures in this job. Of 17 recent failing `External Repo: YAMCS Reference` runs, 11 had a failed `Test` job and 7 of those were this test:

[31506686144](https://github.com/nasa/fprime/actions/runs/31506686144) (a `push` on `devel` @ 26d4611c — no PR content involved), [31555190438](https://github.com/nasa/fprime/actions/runs/31555190438), [31535964024](https://github.com/nasa/fprime/actions/runs/31535964024), [31532491247](https://github.com/nasa/fprime/actions/runs/31532491247), [31419754184](https://github.com/nasa/fprime/actions/runs/31419754184), [31415992880](https://github.com/nasa/fprime/actions/runs/31415992880), [31354479473](https://github.com/nasa/fprime/actions/runs/31354479473).

## Expected Behavior

One `CMD_NO_OP` moves the channel by exactly one. That holds if both endpoints are sampled after the command they bracket has been counted — e.g. read until two consecutive samples agree before taking `begin`, and read until the value changes for `end`, instead of taking the first update that arrives.

Happy to open that PR if the approach looks right. I have not, because I cannot run the integration suite locally; the above is from the run logs and the sources.

---

*AI use per `AI_POLICY.md`. Tool: Claude Code. Type of assistance: reading the failing job logs, reconstructing the two timelines, tracing the channel to its single writer, drafting this issue. Scope: diagnosis only, no code change. Level of modification: every timestamp, path, line number and run ID was read out of the logs and checked against the repositories.*

philphauler:
## Overview
`test_telemetry_update` fails intermittently because `await_telemetry` can return a stale sample. The test then sees +2 or +0 instead of +1.

7 of 11 recent `External Repo: YAMCS Reference` failures are this test (#5676).

## Changes
- **Baseline:** reads until two consecutive samples agree, flushing staleness
- **End:** waits until the counter has actually advanced past the baseline, not just any sample

Both loops are bounded by `await_telemetry`'s timeout.

Fixes #5676

| | |
|:---|:---|
| AI Used | y |
| AI Usage | Claude Code for log analysis and fix design. The consecutive-agreement pattern was suggested in the issue. |

### DIFF
--- a/Svc/Subtopologies/CdhCore/test/int/test_cdh_core.py
+++ b/Svc/Subtopologies/CdhCore/test/int/test_cdh_core.py
@@ -58,20 +58,14 @@ def test_command_and_event_with_many_args(fprime_test_api: IntegrationTestAPI):
 
 def test_telemetry_update(fprime_test_api: IntegrationTestAPI):
     """Test that we can receive telemetry updates with expected values"""
-
-    cmd_dispatched_channel = fprime_test_api.get_telemetry_pred("CommandsDispatched")
     fprime_test_api.set_tlm_packet_level(3)
-    # Clear stale telemetry, then wait for fresh telemetry after command completes
     fprime_test_api.clear_histories()
-
-    begin_result = fprime_test_api.await_telemetry(cmd_dispatched_channel, timeout=3)
-    begin_tlm_val = begin_result.val_obj.val
-
-    # Send command and wait for completion with assert
+    # Read two samples so the second is guaranteed fresh (not stale from prior test)
+    samples = fprime_test_api.await_telemetry_count(2, "CommandsDispatched", timeout=10)
+    assert len(samples) >= 2
+    before = samples[-1].val_obj.val
     fprime_test_api.send_and_assert_command(
         f"{fprime_test_api.get_mnemonic('Svc.CommandDispatcher')}.CMD_NO_OP"
     )
-    # Clear stale telemetry, then wait for fresh telemetry after command completes
     fprime_test_api.clear_histories()
-    end_result = fprime_test_api.await_telemetry(cmd_dispatched_channel, timeout=3)
-    assert end_result.val_obj.val == begin_tlm_val + 1
+    fprime_test_api.assert_telemetry("CommandsDispatched", value=before + 1, timeout=10)


kadircanyildirm-crypto:
The implementation matches what I proposed in the issue — baseline by consecutive agreement, end by waiting past the baseline — thanks for picking it up.

Two notes from reading the diff against fprime-gds 4.3.0:

- "Both loops are bounded by `await_telemetry`'s timeout" is true per iteration only. Every pass restarts a fresh 3-second window, so the loop as a whole has no wall-clock bound if samples keep arriving. With `CommandsDispatched`'s single 1 Hz writer that should not happen here in practice, but a deadline carried across iterations would make the bound real rather than assumed.

- On timeout `await_telemetry` returns `None` (`testing_fw/api.py:788` — "On timeout, the search will return None"), so the next `curr.val_obj.val` raises `AttributeError` instead of a readable assert. The pre-change code had the same exposure on its two awaits; the loops multiply the opportunities. An `assert ... is not None` after each await would keep a stall diagnosable in CI logs.

Neither changes the approach; with the second one addressed I'd expect the 7/11 failure signature from the issue to disappear.

*AI use per `AI_POLICY.md`. Tool: Claude Code. Type of assistance: reading the PR diff and the `fprime_gds` source cited above, drafting this comment. Scope: comment only, no code change. Level of modification: reviewed and edited; the `api.py` behavior was checked against the installed package.*

philphauler:
Thanks for the thorough review — both points were spot on. I've reworked the fix to address them directly:

**Unbounded loops → removed entirely.** Replaced with `await_telemetry_count(2, "CommandsDispatched", timeout=10)` which collects two samples under a single deadline. Reading two guarantees the last one is fresh (not stale from a prior test cycle), which was the root cause of the flake.

**None vulnerability → eliminated.** The post-command check now uses `assert_telemetry("CommandsDispatched", value=before + 1, timeout=10)` which is the framework's own value-match primitive with a built-in timeout — no raw `await_telemetry` calls remain.

Net result: **-6 lines**, no loops, no None exposure, one function changed. The diff now only touches `test_telemetry_update`.

*AI use per `AI_POLICY.md`. Tool: Claude Code. Type of assistance: reworking the fix based on review feedback, verifying `await_telemetry_count` and `assert_telemetry` signatures against fprime-gds 4.3.0 source. Level of modification: reviewed and edited.*

kadircanyildirm-crypto:
Verified against fprime-gds 4.3.0 — both `await_telemetry_count` and `assert_telemetry` are there, and since the counter is cumulative the post-command `clear_histories()` can't lose the +1 (later samples still carry it). This reads cleaner than the original test. LGTM.

*AI use per `AI_POLICY.md`. Tool: Claude Code. Type of assistance: verifying the two API calls against the installed package, drafting this comment. Level of modification: reviewed and edited.*

philphauler:
Let's go! 🚀

Svc/Subtopologies/CdhCore/test/int/test_cdh_core.py
```diff
@@ -58,20 +58,14 @@ def test_command_and_event_with_many_args(fprime_test_api: IntegrationTestAPI):
 
 def test_telemetry_update(fprime_test_api: IntegrationTestAPI):
     """Test that we can receive telemetry updates with expected values"""
-
-    cmd_dispatched_channel = fprime_test_api.get_telemetry_pred("CommandsDispatched")
     fprime_test_api.set_tlm_packet_level(3)
```
thomas-bc, line 61: Did you trace the issue down w.r.t what was happening ? I wonder if this `set_tlm_packet_level` (which sends a command in the YAMCS deployment case) could be causing the race condition on the CommandsDispatched value

thomas-bc (approved):

Svc/Subtopologies/CdhCore/test/int/test_cdh_core.py
philphauler: Well, lemme check on that and get back.

Is there also any way to get in touch with you guys ? I've been hitting various repos on your end with tiny tweaks here and there and I have some cool ideas under my hood 🥹😅

PR #5766, opened 2026-08-21
feat(Os): SandboxedFile deny-by-default (v5.0.0)

### LINKED ISSUE #5657: v5.0.0: make SandboxedFile deny-by-default (require configure() before open())

philphauler:
## Overview
`Os::SandboxedFile` currently defaults to a sandbox of `/`, allowing any absolute path until `configure()` is called. A security review recommended deny-by-default (#5657).

This is the change proposed in #5651 (later reverted), reapplied with the migration path noted.

## Changes
- **SandboxedFile.cpp**: constructor initializes `m_configured(false)` and `m_allowedDirectory("")` instead of `true`/`"/"`
- **SandboxedFile.hpp**: updated doc comment to reflect deny-by-default
- **SandboxedFileTest.cpp**: `ConfigureValid` asserts `!isConfigured()` initially; `DefaultConfigAllowsAnyAbsolutePath` → `UnconfiguredRejectsOpen` (verifies OUTSIDE_SANDBOX)

## Migration (v4 → v5)
Deployments that construct `SandboxedFile` without calling `configure()` must add:
```cpp
sandboxedFile.configure("/");  // restores v4.x permit-all behavior
```
Components using `SandboxedFile` (FileDownlink, FileUplink, CfdpManager) should configure to their intended directory at startup.

## Testing
Updated unit tests verify the new default. Existing tests for configured behavior are unchanged.

Fixes #5657

| | |
|:---|:---|
| AI Used | y |
| AI Usage | Claude Code for tracing SandboxedFile usage and the reverted #5651 history. |

### DIFF
--- a/Os/SandboxedFile.cpp
+++ b/Os/SandboxedFile.cpp
@@ -9,7 +9,7 @@
 
 namespace Os {
 
-SandboxedFile::SandboxedFile() : m_file(), m_allowedDirectory("/"), m_configured(true) {}
+SandboxedFile::SandboxedFile() : m_file(), m_allowedDirectory(""), m_configured(false) {}
 
 SandboxedFile::~SandboxedFile() {
     if (m_file.isOpen()) {
--- a/Os/SandboxedFile.hpp
+++ b/Os/SandboxedFile.hpp
@@ -43,9 +43,10 @@ class SandboxedFile {
     SandboxedFile(const SandboxedFile&) = delete;             //!< Non-copyable (owns file handle)
     SandboxedFile& operator=(const SandboxedFile&) = delete;  //!< Non-copy-assignable
 
-    //! \brief Construct a SandboxedFile with default sandbox of `/`
+    //! \brief Construct an unconfigured SandboxedFile (deny-by-default)
     //!
-    //! The default allows any absolute path. Call `configure()` to restrict.
+    //! `open()` returns `OUTSIDE_SANDBOX` until `configure()` is called.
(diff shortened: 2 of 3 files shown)


thomas-bc (commented):
As stated in the ticket, this is a breaking change targeted for v5, and we are not there yet. Closing, but thanks.

philphauler:
All g thankss! Will take this as feedback for future stuff 🙏🙌

Issue #5767, opened 2026-08-21
FpySequencer: 11 remaining unsigned-underflow stack checks (same class as #5592)


philphauler:
## Summary

PR #5762 (pending, fixing #5592) rewrites the double-subtraction overflow check at line 418 to use safe addition. However, 11 other stack-overflow checks in the same file use the same subtraction-before-comparison pattern and are theoretically vulnerable to the same unsigned underflow.

## Affected Lines

**Group A** — `MAX_STACK_SIZE - dynamicSize < stack.size` (underflows if `dynamicSize > MAX_STACK_SIZE`):

| Line | Context | Risk |
|------|---------|------|
| 385 | `pushTlmVal` — `tlmValue.getSize()` | Medium — size from TLM channel |
| 445 | `pushPrm` — `prmValue.getSize()` | Medium — size from PRM |
| 457 | `pushCmdResponse` — `sizeof(CmdResponse)` | Low — compile-time constant |
| 1392 | `pushTime/cmdResponse` — `sizeof(CmdResponse)` | Low — compile-time constant |
| 1410 | `pushTime` — `Fw::Time::SERIALIZED_SIZE` | Low — compile-time constant |
| 1442 | `pushRand` — `sizeof(U32)` | Low — compile-time constant |

**Group B** — `val > MAX_STACK_SIZE - stack.size` (underflows if `stack.size > MAX_STACK_SIZE`):

| Line | Context | Risk |
|------|---------|------|
| 1249 | `allocate` — `directive.get_size()` | Low — stack.size bounded |
| 1279 | `storeRelConstOffset` — `size` | Low |
| 1315 | `pushVal` — `directive.get__valSize()` | Low |
| 1514 | `discard` — `byteCount` | Low |
| 1651 | `return` — `returnValSize` | Low |

## Assessment

The compile-time-constant cases (457, 1392, 1410, 1442) cannot underflow in practice — `sizeof(CmdResponse)` and `SERIALIZED_SIZE` are always much smaller than `MAX_STACK_SIZE`. The dynamic cases (385, 445) carry the same theoretical risk as the fixed line 418: a TLM or PRM value whose serialized size exceeds `MAX_STACK_SIZE` would wrap the subtraction.

Group B is safe by construction (`stack.size` cannot exceed `MAX_STACK_SIZE` after every push is guarded), but the subtraction form is still formally undefined on overflow for unsigned types.

## Proposed Fix

[...]

PR #5769, opened 2026-08-21
fix(FpySequencer): rewrite 11 remaining subtraction-form stack checks to addition

philphauler:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5767 |

---
## Change Description

Rewrites the seven remaining `MAX_STACK_SIZE - X < stack.size` stack checks in `FpySequencerDirectives.cpp` to the remaining-capacity form `X > MAX_STACK_SIZE - stack.size`, the idiom the file already uses in `allocate`, `loadHelper`, `pushVal` and `pushBytes`.

## Rationale

`MAX - X` underflows in unsigned arithmetic when `X > MAX`. `MAX - stack.size` cannot underflow because `stack.size <= MAX` is an invariant of the stack, and the addition form `stack.size + X` can itself wrap (zimri-leisher's review). The request that can exceed MAX is pushTlmValAndTime with a full telemetry buffer: it needs `FW_TLM_BUFFER_MAX_SIZE + Fw::Time::SERIALIZED_SIZE` bytes, and FpySequencer.hpp only requires `MAX_STACK_SIZE >= FW_TLM_BUFFER_MAX_SIZE`, so a configuration with the two equal is legal and there the old form silently accepts the push and the stack asserts.

Boundary tests (exact fit succeeds, one byte over reports STACK_OVERFLOW) for pushTlmVal, pushTlmValAndTime, pushPrm and the stackCmd response-slot check, plus `pushTlmValAndTimeFullBuffer`: a full telemetry buffer plus time on an empty stack, expecting STACK_OVERFLOW whenever the request cannot fit and the exact resulting size otherwise. Run locally three ways: default config (MAX_STACK_SIZE 65535) with the fix, 299 tests pass; MAX_STACK_SIZE 506 (equal to FW_TLM_BUFFER_MAX_SIZE in the UT build) with the fix, the test passes with STACK_OVERFLOW; the same config without the fix, the old check passes and the stack asserts (`FpySequencerStack.cpp:128`, 506 11) and the UT binary aborts. Other tests hard-code offsets for the default size, so the 506 run needs a gtest filter on this test.

## Future Work

None planned.

Claude Code was used for exhaustive grep of the subtraction pattern across the file. The transformation is mechanical (same fix applied uniformly).

IAMAI

### DIFF
--- a/Svc/FpySequencer/FpySequencerDirectives.cpp
+++ b/Svc/FpySequencer/FpySequencerDirectives.cpp
@@ -382,7 +382,7 @@ Signal FpySequencer::pushTlmVal_directiveHandler(const FpySequencer_PushTlmValDi
         return Signal::stmtResponse_failure;
     }
 
-    if (Fpy::MAX_STACK_SIZE - tlmValue.getSize() < this->m_runtime.stack.size) {
+    if (tlmValue.getSize() > Fpy::MAX_STACK_SIZE - this->m_runtime.stack.size) {
         error = DirectiveError::STACK_OVERFLOW;
         return Signal::stmtResponse_failure;
     }
@@ -415,7 +415,7 @@ Signal FpySequencer::pushTlmValAndTime_directiveHandler(const FpySequencer_PushT
     FW_ASSERT(stat == Fw::SerializeStatus::FW_SERIALIZE_OK, static_cast<FwAssertArgType>(stat));
 
     // check that our stack won't overflow if we put both val and time on it
-    if (Fpy::MAX_STACK_SIZE - tlmValue.getSize() - timeEsb.getSize() < this->m_runtime.stack.size) {
+    if (tlmValue.getSize() + timeEsb.getSize() > Fpy::MAX_STACK_SIZE - this->m_runtime.stack.size) {
         error = DirectiveError::STACK_OVERFLOW;
         return Signal::stmtResponse_failure;
     }
@@ -442,7 +442,7 @@ Signal FpySequencer::pushPrm_directiveHandler(const FpySequencer_PushPrmDirectiv
         return Signal::stmtResponse_failure;
     }
 
-    if (Fpy::MAX_STACK_SIZE - prmValue.getSize() < this->m_runtime.stack.size) {
+    if (prmValue.getSize() > Fpy::MAX_STACK_SIZE - this->m_runtime.stack.size) {
         error = DirectiveError::STACK_OVERFLOW;
         return Signal::stmtResponse_failure;
     }
@@ -454,7 +454,7 @@ Signal FpySequencer::pushPrm_directiveHandler(const FpySequencer_PushPrmDirectiv
 Signal FpySequencer::constCmd_directiveHandler(const FpySequencer_ConstCmdDirective& directive, DirectiveError& error) {
     // the cmd response code will be pushed to the stack when it comes back, so make sure
     // there is room for it now, before the cmd is dispatched
-    if (Fpy::MAX_STACK_SIZE - sizeof(Fw::CmdResponse::SerialType) < this->m_runtime.stack.size) {
+    if (sizeof(Fw::CmdResponse::SerialType) > Fpy::MAX_STACK_SIZE - this->m_runtime.stack.size) {
         error = DirectiveError::STACK_OVERFLOW;
         return Signal::stmtResponse_failure;
     }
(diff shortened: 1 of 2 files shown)


zimri-leisher:
I recall originally doing this to prevent overflow. Could this overflow the stack size type now? Is there a way to totally avoid over/underflow?

philphauler:
Good point — the addition form can overflow `StackSizeType` (U32).

**Original subtraction form** (`MAX_STACK_SIZE - size < stack.size`): zimri's concern is correct — this was written to prevent overflow of the subtraction result. It's safe when `size <= MAX_STACK_SIZE`, but underflows when `size > MAX_STACK_SIZE` (a malformed directive could provide this).

**Addition form** (`stack.size + size > MAX_STACK_SIZE`): both operands are `StackSizeType` (U32). If `stack.size + size > U32::MAX`, the sum wraps and the comparison passes incorrectly. This happens when `MAX_STACK_SIZE > U32::MAX/2` — unlikely in practice but not impossible on a configuration that sets a large stack.

**Safe form** — widen one operand so the addition can't wrap:

```cpp
if (static_cast<U64>(this->m_runtime.stack.size) + size > Fpy::MAX_STACK_SIZE) {
```

This avoids both the original underflow (subtraction from a potentially smaller value) and the new overflow (U32 wrap), with zero runtime cost — the cast is a register promotion. I'll update all 11 checks to use this form and push the update.

philphauler:
Good catch — you're absolutely right that the addition form can overflow U32. After looking more carefully at the file, I noticed you already use the remaining-capacity idiom in `allocate`, `loadHelper`, `pushVal`, and `pushBytes`:

```cpp
if (size > Fpy::MAX_STACK_SIZE - this->m_runtime.stack.size) {
```

This form is safe because `stack.size <= MAX_STACK_SIZE` is an invariant (enforced by the static_assert at `FpySequencerTypes.hpp:33`), so the subtraction never underflows. It avoids both the original underflow (`MAX - size` wraps when `size > MAX`) and the addition overflow (`stack + size` wraps on U32).

I've updated the PR to use this same pattern for all 7 remaining checks that had the reversed subtraction form. The diff is now +7/-7, and every overflow check in the file uses the same idiom. No type widening needed.

For the double-operand case at `pushTlmValAndTime`, the left side is `tlmValue.getSize() + timeEsb.getSize()` — bounded by `FW_TLM_BUFFER_MAX_SIZE` + `Time::SERIALIZED_SIZE` (~140 bytes), so U32 overflow isn't a concern there.

*AI use per `AI_POLICY.md`. Tool: Claude Code. Type of assistance: analyzing the existing overflow-check patterns in the file, writing a Python test harness to verify the fix catches all 4 underflow cases the original misses. Level of modification: reviewed and edited.*

LeStarch (dismissed):

zimri-leisher:
@philphauler  are there tests proving that no overflow can occur? you should make sure to write the tests such that the overflow occurs without the fix, and then with the fix it doesn't.

zimri-leisher (changes requested):
Needs tests

philphauler:
@zimri-leisher Got it! 🫡
Imma take this into my flow from now on and perform every time. 

Anything else or other feedback super much appreciated, I'm hella active and wanna open my eyes and ears and my contributions to be meaningful 🙏

philphauler:
Hey @zimri-leisher 👋 fair ask, and it took some digging to find where the two forms actually disagree, which was a fun one.

Tests are in aec964b and 3feceb5 (the branch was rebased once to clean up a commit trailer, code unchanged). The boundary ones (exact fit accepted, one byte over rejected) pass with both forms in the default config, because with MAX_STACK_SIZE 65535 nothing a directive can push comes close to the stack size, so the subtraction never underflows there.

The case where they diverge is pushTlmValAndTime with a full telemetry buffer: it needs FW_TLM_BUFFER_MAX_SIZE + Fw::Time::SERIALIZED_SIZE bytes, and FpySequencer.hpp:35 only requires MAX_STACK_SIZE >= FW_TLM_BUFFER_MAX_SIZE, so a config with the two equal is legal. pushTlmValAndTimeFullBuffer pushes a full buffer plus time onto an empty stack and expects STACK_OVERFLOW whenever the request cannot fit. With MAX_STACK_SIZE set to 506 (FW_TLM_BUFFER_MAX_SIZE in the UT build):
- without the fix: the subtraction underflows, the check passes, and the stack asserts: `Assert: "Svc/FpySequencer/FpySequencerStack.cpp:128" 506 11`, UT binary aborted
- with the fix: STACK_OVERFLOW, test passes

In the default config the same test checks the exact resulting size instead, so it runs everywhere. (A few other tests hard-code offsets that assume the default size, so the 506 run only makes sense with a gtest filter on this test.)

Thanks for pushing on this, have a great day! ⚡

LeStarch (approved):

LeStarch:
@zimri-leisher I am going to merge this. I see tests, as you asked, and see the corrected code that I agree fixes potential bugs.

philphauler:
Eyyyy let's go 🚀🚀🚀

Issue #5690, opened 2026-08-14
Unguarded FW_ASSERT in CfdpManager


bitWarrior:
| | |
|:---|:---|
|**_F´ Version_**| v.42.2 devel |
|**_Affected Component_**| `Svc::Ccsds::Cfdp::CfdpManager` (CFDP receive path) |
---
## Problem Description

A zero-length CFDP **Class 2 FileData PDU** can crash the flight software via a reachable assertion (denial of service, CWE-617). Ref: [GHSA-mh5x-2m6h-8267](https://github.com/nasa/fprime/security/advisories/GHSA-mh5x-2m6h-8267).

A FileData PDU that carries a complete 4-byte file offset but **zero file-data octets** (its PDU payload length equals the encoded offset length) deserializes successfully with `dataSize == 0`. `FileDataPdu::fromSerialBuffer()` subtracts the offset size from the payload length and, when they are equal, yields `m_dataSize == 0` (this empty-payload case is itself covered by the existing `FileDataEmptyPayload` unit test, so it is accepted as syntactically valid).

When such a PDU arrives before any Metadata, `Engine::recvInit()` starts a Class 2 (R2) receive transaction with no declared file size. The bounds check in `Transaction::rProcessFd()` uses `bound = std::numeric_limits<FileSize>::max()` in that state, so a zero size passes:

```cpp
const FileSize bound = this->m_flags.rx.md_recv ? this->m_fsize
                                                : std::numeric_limits<FileSize>::max();
if ((offset > bound) || ((bound - offset) < dataSize)) { /* reject */ }   // false for offset=0, dataSize=0
```
The Class 2 handler then records the interval via `CfdpChunkList::add(offset, dataSize)` = `add(0, 0)`. `add()` only guards against integer rollover (`(offset + size) >= offset`, which is `0 >= 0` → true), so the zero-size chunk flows into `insert()` → `combineNext()`:

```cpp
FileSize chunk_end = chunk->offset + chunk->size;      // 0 + 0 = 0
/* Assert no rollover, only possible as a bug */
FW_ASSERT(chunk_end > chunk->offset, ...);             // 0 > 0 is false -> assert fires (args 0 0)
```

The strict `>` fails for a zero-length interval, the assertion fires (`Chunk.cpp:212`, args `0 0`), and the deployment's FATAL handling terminates the process.

Full processing chain:

```
Drv::TcpServer -> ComStub -> FrameAccumulator / TcDeframer -> SpacePacketDeframer
[...]

PR #5771, opened 2026-08-21
fix(SpacePacketFramer): guard zero-size data at runtime, not only via assert

philphauler:
## Summary

- Convert the `FW_ASSERT(data.getSize() > 0)` in `SpacePacketFramer::dataIn_handler` to a
  runtime guard that logs a throttled `EmptyDataReceived` warning and returns the buffer to
  the caller
- The CCSDS Packet Data Length field encodes "octets minus 1", so a zero-octet payload
  cannot be represented; the `static_cast<U16>(data.getSize() - 1)` underflows to 65535
  when the assert is compiled out under `FW_NO_ASSERT`
- Add the `EmptyDataReceived` event to the FPP model with `throttle 5`
- Add unit test `EmptyDataDropped` covering the zero-size path

## Motivation

This is the same defense-in-depth pattern applied in #5585, #5633, #5690, and #5766:
inputs that can be influenced by upstream components or external sources should be
guarded at runtime, not only by assertions that may be compiled out.

## Test plan

- [ ] New `EmptyDataDropped` test: sends a zero-size buffer, asserts no packet emitted,
  buffer returned, warning event fired, no allocation attempted
- [ ] Existing `testNominalFraming` and `testOversizedAllocatorBufferIsTrimmed` pass
  unchanged

### DIFF
--- a/Svc/Ccsds/SpacePacketFramer/SpacePacketFramer.cpp
+++ b/Svc/Ccsds/SpacePacketFramer/SpacePacketFramer.cpp
@@ -25,14 +25,21 @@ SpacePacketFramer ::~SpacePacketFramer() {}
 // ----------------------------------------------------------------------
 
 void SpacePacketFramer ::dataIn_handler(FwIndexType portNum, Fw::Buffer& data, const ComCfg::FrameContext& context) {
+    // CCSDS Space Packet requires at least 1 data octet (the Packet Data Length
+    // field encodes "number of octets minus 1", so size 0 cannot be represented).
+    // Guard at runtime so the subtraction at packetDataLength never underflows,
+    // even in FW_NO_ASSERT builds.
+    if (data.getSize() == 0) {
+        this->log_WARNING_HI_EmptyDataReceived();
+        this->dataReturnOut_out(0, data, context);
+        return;
+    }
+
     SpacePacketHeader header;
     Fw::SerializeStatus status;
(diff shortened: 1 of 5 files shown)


thomas-bc (commented):
Requests going to the Framer are coming from framework/project code (i.e. not untrusted input from the radio) - therefore I would contend that requesting framing of 0-length data is a coding error, since the SpacePacketProtocol doesn't allow it. It absolutely should be caught during development and guarded against in project code. However I do understand the desire not to crash out of the sky for something that can just be skipped and logged....

@bitWarrior thoughts here?

philphauler:
Thanks @thomas-bc — you're right that 0-length framing is a coding error (Space Packet can't represent it) and should be caught in project code.

The guard here is defense-in-depth for the case it slips through: in a FW_NO_ASSERT build the original FW_ASSERT(size > 0) compiles to no-op and size - 1 underflows to 0xFFFF, walking past the buffer. This now logs EmptyDataReceived and returns instead of asserting or walking OOB.

Happy to adjust the event severity or keep it as assert-only if you prefer the hard-fail during dev — just wanted to avoid the silent OOB in release.

cc @bitWarrior

LeStarch:
This PR is incorrect. Since framer is taking input from flight software, a fail-fast assert is the correct way to establish the expected contract with the rest of the system.

PR #5775, opened 2026-08-22
Serialize the CRC in Utils::CRCChecker so .CRC32 files are portable

### LINKED ISSUE #5774: Svc/FileWorker writes the .CRC32 sidecar big-endian but validates it host-endian, so its own files fail CRC on little-endian targets
| | |
|:---|:---|
|**_F´ Version_**| `devel` @ 78e2b18 (2026-08-22) |
|**_Affected Component_**| `Svc/FileWorker`, `Utils/CRCChecker`, `Os/ValidateFile` |
---
## Problem Description

`Svc::FileWorker` writes the `<file>.CRC32` sidecar in **big-endian** byte order but validates it in **host-endian** byte order, so on any little-endian target a file FileWorker just wrote can never be read back or verified. The integrity check reports corruption for data that is perfectly intact.

Write path — `FileWorker.cpp:498` writes the raw `HashBuffer` bytes produced by `Utils::Hash::finalize(HashBuffer&)`, which serialises through `Fw::LinearBufferBase::serializeFrom(U32, Endianness::BIG)`.

Read path — `FileWorker.cpp:66` and `:106` call `Utils::verify_checksum` → `Utils::read_crc32_from_file`, which does:

```cpp
// Utils/CRCChecker.cpp:119
f.read(reinterpret_cast<U8*>(&checksum_from_file), 4);   // raw host-endian U32
```

On x86_64 / aarch64-LE — every CI platform and the Ref deployment — those two are byte-swapped. `readIn_handler` emits `log_WARNING_HI_CrcFailed` and returns `readDoneOut(FW_STATUS_FAILED_CRC, 0)`, refusing to deliver intact data; `verifyIn_handler` returns `FW_STATUS_FAILED_CRC`.

**The repo's own fixtures show which convention the read path expects:**

```
$ python3 -c "import zlib;print(hex(zlib.crc32(open('Svc/FileWorker/nominalread.bin','rb').read())))"
0xbc0905f4
$ xxd Svc/FileWorker/nominalread.bin.CRC32
00000000: f405 09bc                                ....
```

The checked-in sidecar is little-endian (`f4 05 09 bc`). `writeBufferHashToFile` would have written `bc 09 05 f4`. So the fixtures and the read path agree with each other, and the write path disagrees with both.

This is not caught today because `Svc/FileWorker/test/ut/FileWorkerTester.cpp` never does a write-then-read round trip — `testWriting` only re-reads the file body and never touches the `.CRC32` file, while `testReadFile` / `testVerify` consume the pre-made little-endian fixtures.

## Wider inconsistency (why I'm asking rather than sending a patch)

The two conventions are both used across the framework, so fixing this properly is a convention decision rather than a one-line change:

- **Big-endian `HashBuffer` bytes** — `Utils::Hash::finalize` / `setHashValue`; `Os/ValidateFileCommon.cpp:104` `writeHash` and `:75-86` `readHash`, used by `Svc::ComLogger` and `Svc::BufferLogger` via `Os::ValidatedFile`.
- **Host-endian raw `U32`** — `Utils/CRCChecker.cpp:91` (write) and `:119` (read); internally consistent, but incompatible with `Os::ValidateFile`, and the resulting `.CRC32` files are not portable across endianness.
- `Svc::FileWorker` mixes both: `:498` and `:373` big-endian, `:66` and `:106` host-endian.

So `Utils::CRCChecker`-produced sidecars cannot be validated by `Os::ValidateFile::validate()` and vice versa, independently of the FileWorker bug.

I have a fix ready but there are two defensible directions and the choice affects an on-disk format, so I'd rather ask first:

1. **Make FileWorker write host-endian** — smallest change, matches `Utils::CRCChecker` and the existing checked-in fixtures, no fixture churn. Leaves `CRCChecker` ↔ `Os::ValidateFile` incompatible and leaves `.CRC32` files endian-dependent.
2. **Standardise on the big-endian `HashBuffer` form** — matches `Os::ValidateFile`, makes sidecars portable across targets, and is the framework-wide serialisation convention. Requires converting `Utils::CRCChecker` and regenerating the checked-in `.CRC32` fixtures, and changes an on-disk format for existing deployments.

My preference is (2) for correctness with (1) as the low-risk option, but this is a project decision. Either way I'd add the missing write→read round-trip test, which is what would have caught this.

Related but distinct: #5715 lists `Svc/FileWorker/FileWorker.cpp:66` as `DESIGN` ("relaxing the CRC32 sidecar requirement is an interface contract change"). That's about whether the sidecar is required at all; this issue is about the byte order it is written in.

## Context / Environment

Reproduced by byte-order inspection of the checked-in fixtures and by a standalone harness replicating both code paths verbatim — no F´ build required, so `fprime-util version-check` isn't meaningful here. Source read at `devel` HEAD on 2026-08-22.

## How to Reproduce

1. `python3 -c "import zlib;print(hex(zlib.crc32(open('Svc/FileWorker/nominalread.bin','rb').read())))"` → `0xbc0905f4`
2. `xxd Svc/FileWorker/nominalread.bin.CRC32` → `f4 05 09 bc` (little-endian; the write path emits `bc 09 05 f4`)
3. In-tree: in `FileWorkerTester.cpp::testWriting`, after the existing `invoke_to_writeIn(...); doDispatch();`, add `invoke_to_readIn(0, Fw::String("testwrite.txt"), buf); doDispatch();` — `readDoneOut` reports `FW_STATUS_FAILED_CRC` and `ASSERT_EVENTS_CrcFailed_SIZE(1)` fires on a file the component itself just wrote.

## Expected Behavior

A file written by `Svc::FileWorker` should validate against its own `.CRC32` sidecar on every supported target, and `.CRC32` files should mean the same thing to `Utils::CRCChecker` and `Os::ValidateFile`.

---

*AI usage: I used an AI assistant to help search the tree for sibling call sites and to draft this report. The endianness divergence, the fixture byte-order check and the proposed directions are my own analysis and I have verified each claim against the source at `devel` HEAD; I can explain and defend every line above.*

winklemad:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5774 |

---
## Change Description

`Utils::create_checksum_file` wrote the CRC by casting the `U32` to a `U8*`, and `read_crc32_from_file` read it back the same way, so the contents of a `.CRC32` file depended on the endianness of the processor. This change serializes the value through `Fw::SerialBuffer` on write and deserializes it on read, so the file holds the same bytes on every target.

`Svc/FileWorker/nominalread.bin.CRC32` is regenerated in the new byte order (`f4 05 09 bc` → `bc 09 05 f4`; verified against `zlib.crc32` of `nominalread.bin`, which is `0xbc0905f4`).

I also dropped the tracked `Svc/FileWorker/testfile.txt.CRC32`. It is listed in that directory's own `.gitignore`, its data file `testfile.txt` is generated at test time and is not in the repository, and neither FileWorker unit-test suite produces or reads it — I deleted it locally and both suites still pass from clean. Happy to restore it if it is there on purpose.

`Svc/FileWorker/FileWorker.cpp` is deliberately untouched: its write path is already correct.

## Rationale

Fixes #5774.

A `.CRC32` sidecar is not private to `CRCChecker`. `Utils::Hash::finalize(HashBuffer&)`, `Os::ValidateFile`, `Os::ValidatedFile` and `Svc::ComLogger` all produce or consume the same file through F Prime serialization, which is big-endian. `CRCChecker` was the one component using the host's raw `U32` bytes.

`Svc::FileWorker` sits on both sides of that split. It writes the sidecar via `Utils::Hash::finalize(HashBuffer&)` (`FileWorker.cpp:498`, big-endian) but validates it via `Utils::verify_checksum` (`FileWorker.cpp:66` and `:106`, host-endian). On a little-endian target the two disagree, so `readIn_handler` logs `CrcFailed` and returns `FW_STATUS_FAILED_CRC` for a file the component itself just wrote.

This is the same defect class as #4931 ("the stored value is determined by the endianness of the processor, not a generic. Fix is to use a serial buffer in F Prime to serialize the U32 then write it out"), fixed for `Svc::PrmDb` in #4956. This change follows that pattern, including its `FW_ASSERT` on the serialize status.

I originally proposed the opposite fix in #5774 and corrected myself in a comment there: moving `FileWorker` to host-endian breaks the three existing `Os::ValidateFile::validate(...) == VALIDATION_OK` assertions in `FileWorkerTester.cpp` (`:238-239`, `:297-298`, `:362-363`), which confirms the big-endian side is the one that agrees with the framework.

Two new tests, both of which fail on `devel` and pass with this change on a little-endian host:

- `Svc/FileWorker` `Nominal.testWriteReadRoundTrip` — writes a file through `writeIn`, then reads the same file back through `readIn`, asserting no `CrcFailed` event and `FW_STATUS_DONE_READ`, then re-checks it through `verifyIn`. Before the fix it reports `FW_STATUS_FAILED_CRC` (status 7 instead of 14) with one `CrcFailed` event. This round trip is the gap that let the bug survive: nothing exercised write-then-read on the same file.
- `Utils` `CRCCheckerTest.ChecksumFileMatchesFrameworkFormat` — asserts the bytes on disk match what `Utils::Hash` produces for the same data, that `Os::ValidateFile::validate` accepts a file written by `create_checksum_file`, and that `read_crc32_from_file` correctly reads a file written by `Os::ValidateFile::createValidation`.

The existing `CRCChecker` tests added in #5480 needed no changes and all still pass. Worth noting for review: none of them hard-code a byte order, because each is a self-consistent `CRCChecker` write-then-read round trip (`GoldenValue` asserts the `U32` value `0xCBF43926`, not its layout on disk). That is why the suite stayed green while the format was wrong, and it is what the new test above is for.

Suites run locally: `Utils` (16/16), `Svc/FileWorker` nominal and error (11/11 and 9/9, including the three `Os::ValidateFile` assertions above), `Os` (3/3 executables). Also re-ran `Svc/FileWorker` after deleting all generated test files, to confirm nothing depends on a leftover sidecar.

**Compatibility impact, stated plainly:** this changes the on-disk format of `.CRC32` files written by `Utils::CRCChecker`. A sidecar produced by an older build on a little-endian target will no longer validate after this change, and vice versa — the same break #4956 accepted for `PrmDb`'s parameter file. Files written by `Utils::Hash` / `Os::ValidateFile` / `Svc::ComLogger` are unaffected, since those were already in this format. If a compatibility path is wanted (for example accepting the byte-swapped value on read for one release, behind a config option), say the word and I will add it.

## Future Work

None required for this fix. If maintainers want the migration path described above for existing flight-side `.CRC32` files, I am happy to submit it as a separate issue-backed change.

I used an AI assistant (Claude, via Claude Code) while working on this change, and I reviewed and verified everything before submitting.

- **Navigation and analysis:** I used it to trace the producers and consumers of the `.CRC32` sidecar across `Utils`, `Os` and `Svc`, and to pull up the #4931 / #4956 precedent. It is also what led me down the wrong path initially — I posted an incorrect fix direction on #5774 and corrected it myself after building the tree and watching the three `Os::ValidateFile` assertions fail.
- **Code:** the change to `Utils/CRCChecker.cpp` is small and I wrote it to mirror #4956's `serializeFrom` / `deserializeTo` pattern; the assistant helped me confirm the `Fw::SerialBuffer` API. The two new unit tests were drafted with its help and then edited by me — I set what they assert, added the clean-slate file removal in the FileWorker test after hitting a stale-file problem, and confirmed each one fails on unmodified `devel` for the right reason before the fix.
- **Verification:** all of it is mine. I built the unit tests locally, captured the failing output before the change and the passing output after, ran the `Utils`, `Svc/FileWorker` and `Os` suites, and regenerated the fixture from `zlib.crc32` of the data file rather than hand-editing bytes.

I understand this change and can answer questions on any part of it.


sylvesterkaczmarek (commented):
This changes the on-disk `.CRC32` format from host-endian bytes to F Prime serialization. Existing checksum files produced on little-endian systems will therefore fail verification after an upgrade. Could the reader accept the legacy representation as a fallback, or should this include an explicit migration/versioning story for persisted CRC sidecars?

winklemad:
Thanks for looking at this — that's the right question to ask about a persisted format, so let me lay out what I found.

The serialized form isn't a new format being introduced here; it's the one the rest of the framework already writes to `.CRC32`. `Utils::Hash::finalize` builds the buffer with `bufferOut.serializeFrom(~(this->hash_handle))` (`Utils/Hash/Crc32/HashImpl.cpp:40-46`), and `Os::validateFile` persists that buffer verbatim through `writeHash` (`Os/ValidateFileCommon.cpp:93-111`). `Utils::CRCChecker` was the one place writing `reinterpret_cast<U8*>(&checksum)` instead, so it was the outlier.

The practical consequence is that on a little-endian host the two halves already disagreed. For CRC `0xB70B4C26`, `Utils::Hash` writes `B7 0B 4C 26` and the old `CRCChecker` wrote `26 4C 0B B7`. A sidecar produced by one could not be verified by the other, which is the bug the round-trip test in `Svc/FileWorker` pins: the component writes its checksum via `Utils::Hash` and reads it back via `Utils::verify_checksum`, and that failed before this change.

So the migration surface is narrower than a format change across the board. It's files written by `CRCChecker` specifically, on a little-endian host, and those files were already unreadable by `Os::validateFile`. Adding a legacy fallback to the reader would mean keeping a byte order that only one component ever produced and that no other part of F Prime accepts, and on a big-endian host the two encodings are identical so a fallback would be indistinguishable from the correct one.

That said, whether existing deployments need a transition is a call for the maintainers, not me. If you'd rather have it, I'm happy to add a reader fallback that accepts the byte-swapped value and re-writes the sidecar in serialized form, or to gate it behind a config option. Just say which shape you want.

Also rebased onto devel. The earlier `UTs [ubuntu-22.04]` red was `PosixFileTest / Functionality.RandomizedInterfaceTesting`, the known random failure from #2835 — `Utils_ut_exe` passed in that same run — and devel now carries #5831, which addresses that test.

sylvesterkaczmarek:
Thanks for tracing the existing write path. You're right: `Hash::finalize` already serializes into `HashBuffer`, and `writeHash` persists that representation, so `CRCChecker` was the outlier. My earlier migration concern doesn't hold given that existing canonical format.

winklemad:
Appreciate you tracing the write path with me. Agreed — the serialized form is the existing canonical representation that `Hash::finalize`/`writeHash` already persist, so having `CRCChecker` read it back the same way just closes the round-trip. Glad the migration concern doesn't bite here.

sylvesterkaczmarek:
Thanks, and likewise. Tracing the existing Hash::finalize and writeHash path made the compatibility story much clearer. With that canonical representation confirmed, the CRCChecker change reads as a consistency fix rather than a format migration.

Utils/CRCChecker.cpp
```diff
@@ -89,9 +90,15 @@ crc_stat_t create_checksum_file(const char* const fname) {
...
+    // Write checksum file. Serialize the value rather than writing the raw U32 bytes so that the
+    // file contents do not depend on the endianness of the processor.
+    U8 checksum_data[sizeof(checksum)];
+    Fw::SerialBuffer checksum_buffer(checksum_data, sizeof(checksum_data));
+    Fw::SerializeStatus ser_stat = checksum_buffer.serializeFrom(checksum);
+    FW_ASSERT(Fw::FW_SERIALIZE_OK == ser_stat, static_cast<FwAssertArgType>(ser_stat));
+
+    bytes_to_write = checksum_buffer.getSize();
+    stat = f.write(checksum_buffer.getBuffAddr(), bytes_to_write);
     if (stat != Os::File::OP_OK || sizeof(checksum) != bytes_to_write) {
```
lestarch-autobot, line 102: [Maintainability] **suggestion** `maint-structural-hazard`: write length and post-write check use two different size sources.

`bytes_to_write` is now taken from `checksum_buffer.getSize()`, but the short-write check two lines later still compares against `sizeof(checksum)`. They agree only because a serialized `U32` is 4 bytes; the first change that serializes anything more into the buffer (a format version, a length) would make every write report `FAILED_FILE_CRC_WRITE` with no compiler help. Checking against the same buffer the write came from removes the hidden coupling.

```suggestion
    if (stat != Os::File::OP_OK || checksum_buffer.getSize() != bytes_to_write) {
```

Svc/FileWorker/test/ut/FileWorkerTester.cpp
```diff
@@ -377,6 +377,59 @@ void FileWorkerTester ::testAppending() {
     ASSERT_EVENTS_WriteTimeout_SIZE(0);
 }
 
+void FileWorkerTester ::testWriteReadRoundTrip() {
+    // Regression: the component must be able to read back a file it wrote itself. The write path
+    // stores the checksum through Utils::Hash, while readIn_handler checks it through
+    // Utils::verify_checksum, so both must agree on the on-disk layout of the .CRC32 file.
+    const char* fnameChar = "testroundtrip.txt";
+    const char* hashFileChar = "testroundtrip.txt.CRC32";
```
lestarch-autobot, line 385: [Maintainability] **suggestion** `maint-inconsistent-local-convention`: sidecar name hard-codes the `.CRC32` extension.

Every other test in this file derives the hash file name from `Utils::Hash::getFileExtensionString()` (`testWriting`, `testWritingOffset`, `testAppending`). If `HASH_EXTENSION_STRING` is reconfigured, this test's clean-slate `remove()` silently misses the real sidecar and the "stale file" scenario the comment below warns about comes back. Using the same derivation keeps one source of truth.

```suggestion
    std::string hashFileNameStr = std::string(fnameChar) + std::string(Utils::Hash::getFileExtensionString());
    const char* hashFileChar = hashFileNameStr.c_str();
```

Svc/FileWorker/test/ut/FileWorkerTester.cpp
```diff
@@ -377,6 +377,59 @@ void FileWorkerTester ::testAppending() {
     ASSERT_EVENTS_WriteTimeout_SIZE(0);
 }
 
+void FileWorkerTester ::testWriteReadRoundTrip() {
+    // Regression: the component must be able to read back a file it wrote itself. The write path
+    // stores the checksum through Utils::Hash, while readIn_handler checks it through
+    // Utils::verify_checksum, so both must agree on the on-disk layout of the .CRC32 file.
```
lestarch-autobot, line 383: [Maintainability] **could fix** `maint-misleading-comment`: three-line lead-in comment where one or two lines carry the point.

Inline comments should be succinct (1-2 lines). The mechanism detail (which helper each path uses) is already visible in the asserts below and in `FileWorker.cpp`; it will drift the first time either path changes.

```suggestion
    // Regression: readIn must accept a .CRC32 sidecar written by writeIn (Utils::Hash writes it,
    // Utils::verify_checksum reads it, so both must agree on the on-disk layout).
```

Utils/test/ut/CRCCheckerTester.cpp
```diff
@@ -113,6 +114,42 @@ TEST_F(CRCCheckerTest, GoldenValue) {
     ASSERT_STREQ(extended.toChar(), hashFileName(TEST_FILE).toChar());
 }
 
+TEST_F(CRCCheckerTest, ChecksumFileMatchesFrameworkFormat) {
+    // A .CRC32 file is not private to CRCChecker: Utils::Hash and Os::ValidateFile write and read
+    // the same sidecar for the same data files. Its contents must therefore be the serialized
+    // (big-endian) form of the CRC rather than a raw copy of the host's U32 bytes.
```
lestarch-autobot, line 120: [Maintainability] **could fix** `maint-misleading-comment`: three-line lead-in comment where one or two lines carry the point.

Inline comments should be succinct (1-2 lines); the rest of this file uses one-liners (`GoldenValue`). The test name and the per-step comments below already say what is checked.

```suggestion
    // The .CRC32 sidecar is shared with Utils::Hash and Os::ValidateFile, so it must hold the
    // serialized (big-endian) CRC, not the host's raw U32 bytes.
```

Utils/test/ut/CRCCheckerTester.cpp
```diff
@@ -113,6 +114,42 @@ TEST_F(CRCCheckerTest, GoldenValue) {
...
+    Utils::HashBuffer expectedBuffer;
+    Utils::Hash::hash(data, sizeof(data), expectedBuffer);
+    U8 onDisk[sizeof(U32)] = {};
+    Os::File crcFile;
+    ASSERT_EQ(crcFile.open(hashFileName(TEST_FILE).toChar(), Os::File::OPEN_READ), Os::File::OP_OK);
+    FwSizeType readSize = sizeof(onDisk);
+    ASSERT_EQ(crcFile.read(onDisk, readSize), Os::File::OP_OK);
+    ASSERT_EQ(readSize, static_cast<FwSizeType>(sizeof(onDisk)));
+    crcFile.close();
+    ASSERT_EQ(memcmp(onDisk, expectedBuffer.getBuffAddr(), sizeof(onDisk)), 0);
```
lestarch-autobot, line 135: [Test Quality] **suggestion** On-disk byte layout is asserted only relative to `Utils::Hash`, not against a fixed golden byte sequence.

The test's stated property is that the sidecar holds the big-endian serialized CRC. Comparing against `Utils::Hash::finalize(HashBuffer&)` proves the two producers agree, but both go through `Fw::SerialBuffer::serializeFrom`, so a change in framework serialization would keep this passing while the on-disk format silently changed. Pinning the literal bytes of `0xCBF43926` makes the format itself the oracle (same spirit as `GoldenValue`).

```suggestion
    const U8 expectedBytes[] = {0xCB, 0xF4, 0x39, 0x26};  // 0xCBF43926 serialized big-endian
    ASSERT_EQ(memcmp(onDisk, expectedBytes, sizeof(onDisk)), 0);
    ASSERT_EQ(memcmp(onDisk, expectedBuffer.getBuffAddr(), sizeof(onDisk)), 0);
```

Utils/CRCChecker.cpp
```diff
@@ -89,9 +90,15 @@ crc_stat_t create_checksum_file(const char* const fname) {
         return FAILED_FILE_CRC_OPEN;
     }
 
-    // Write  checksum  file
-    bytes_to_write = sizeof(checksum);
-    stat = f.write(reinterpret_cast<U8*>(&checksum), bytes_to_write);
+    // Write checksum file. Serialize the value rather than writing the raw U32 bytes so that the
+    // file contents do not depend on the endianness of the processor.
```
lestarch-autobot, line 94: [Documentation] **suggestion** The on-disk `.CRC32` format is now a public contract, but no doc comment states it.

`create_checksum_file` / `read_crc32_from_file` in `Utils/CRCChecker.hpp` carry no `\brief`/format doc, and this comment says only "serialized". Name the format (F Prime serialization, big-endian) and the interoperating producers/consumers (`Utils::Hash`, `Os::ValidateFile`, `Svc::ComLogger`) so the next reader cannot re-introduce the host-endian write. Please also add matching doc comments on the two declarations in `CRCChecker.hpp` (not in this diff, so no suggestion block for it).

```suggestion
    // Write checksum file as an F Prime-serialized (big-endian) U32 so the layout is host-independent
    // and matches the .CRC32 files produced/consumed by Utils::Hash, Os::ValidateFile and Svc::ComLogger.
```

Utils/CRCChecker.cpp
```diff
@@ -119,14 +126,21 @@ crc_stat_t read_crc32_from_file(const char* const fname, U32& checksum_from_file
...
+    U8 checksum_data[sizeof(checksum_from_file)];
     FwSizeType checksum_from_file_size = static_cast<FwSizeType>(sizeof(checksum_from_file));
-    stat = f.read(reinterpret_cast<U8*>(&checksum_from_file), checksum_from_file_size);
+    stat = f.read(checksum_data, checksum_from_file_size);
     if (stat != Os::File::OP_OK || checksum_from_file_size != sizeof(checksum_from_file)) {
         f.close();
         return FAILED_FILE_CRC_READ;
     }
 
+    // Deserialize the value to match the serialized form written by create_checksum_file
```
lestarch-autobot, line 138: [Documentation] **must fix** `Svc/FileWorker/docs/sdd.md` does not reflect the changed `.CRC32` contract this read path now enforces.

`Svc::FileWorker` is the only consumer of `read_crc32_from_file` (via `verify_checksum` in `readIn_handler` / `verifyIn_handler`). Its SDD ("Verify checksum", "write a validation CRC file") never states the sidecar layout, and after this change `readIn`/`verifyIn` reject sidecars written by earlier little-endian builds of this same path (the PR's own stated compatibility impact). Add a short note under `writeIn_handler` step 5 / `readIn_handler` step 4 stating the `.CRC32` file holds the F Prime-serialized (big-endian) CRC shared with `Utils::Hash`/`Os::ValidateFile`, and that pre-change host-endian sidecars must be regenerated. Anchored here because the SDD is not in the diff.

Utils/CRCChecker.cpp
```diff
@@ -89,9 +90,15 @@ crc_stat_t create_checksum_file(const char* const fname) {
...
     }
 
-    // Write  checksum  file
-    bytes_to_write = sizeof(checksum);
-    stat = f.write(reinterpret_cast<U8*>(&checksum), bytes_to_write);
+    // Write checksum file. Serialize the value rather than writing the raw U32 bytes so that the
+    // file contents do not depend on the endianness of the processor.
+    U8 checksum_data[sizeof(checksum)];
+    Fw::SerialBuffer checksum_buffer(checksum_data, sizeof(checksum_data));
+    Fw::SerializeStatus ser_stat = checksum_buffer.serializeFrom(checksum);
```
lestarch-autobot, line 97: [Design] **suggestion** Missed reuse: the sidecar format this PR aligns to is *defined* by `Utils::HashBuffer` (what `Utils::Hash::finalize(HashBuffer&)` emits and what `Os::ValidateFile` writes/reads). Re-deriving it here through a parallel `Fw::SerialBuffer` of a `U32` matches today only by construction; if the hash implementation or `HASH_DIGEST_LENGTH` ever changes, `CRCChecker` silently drifts from `ValidateFile` again — the exact failure mode #5774 fixes. Writing the `HashBuffer` directly expresses the shared contract and removes the assert-on-serialize path.

```suggestion
    // Write checksum file in the HashBuffer form shared with Utils::Hash / Os::ValidateFile so
    // that the file contents do not depend on the endianness of the processor.
    Utils::HashBuffer checksum_buffer;
    hash.finalize(checksum_buffer);

    bytes_to_write = checksum_buffer.getSize();
    stat = f.write(checksum_buffer.getBuffAddr(), bytes_to_write);
```

Utils/CRCChecker.cpp
```diff
@@ -119,14 +126,21 @@ crc_stat_t read_crc32_from_file(const char* const fname, U32& checksum_from_file
...
+    stat = f.read(checksum_data, checksum_from_file_size);
     if (stat != Os::File::OP_OK || checksum_from_file_size != sizeof(checksum_from_file)) {
         f.close();
         return FAILED_FILE_CRC_READ;
     }
 
+    // Deserialize the value to match the serialized form written by create_checksum_file
+    Fw::SerialBuffer checksum_buffer(checksum_data, sizeof(checksum_data));
+    checksum_buffer.fill();
+    Fw::SerializeStatus ser_stat = checksum_buffer.deserializeTo(checksum_from_file);
```
lestarch-autobot, line 141: [Design] **suggestion** Same reuse point on the read side: `Utils::HashBuffer` already provides `asBigEndianU32()` for exactly this conversion (used by `Svc::FileWorker`'s own hash handling), so the sidecar bytes can be interpreted through the type that owns the format instead of a second `Fw::SerialBuffer` + `fill()` + `deserializeTo` sequence.

```suggestion
    // Interpret the bytes as the HashBuffer form written by create_checksum_file / Os::ValidateFile
    Utils::HashBuffer checksum_buffer(checksum_data, static_cast<FwSizeType>(sizeof(checksum_data)));
    checksum_from_file = checksum_buffer.asBigEndianU32();
```

Utils/CRCChecker.cpp
```diff
@@ -12,6 +12,7 @@
 #include <Fw/FPrimeBasicTypes.hpp>
 #include <Fw/Types/Assert.hpp>
 #include <Fw/Types/FileNameString.hpp>
+#include <Fw/Types/SerialBuffer.hpp>
```
lestarch-autobot, line 15: [Design] **future work** (Anchored above; refers to `Utils::CRCChecker` as a whole.) `create_checksum_file` / `read_crc32_from_file` / `verify_checksum` are a second, independent implementation of what `Os::ValidateFile::createValidation` / `validate` already do (hash a file, write/read the `.CRC32` sidecar). #5774 is the cost of that duplication drifting. Consider a follow-up that makes `CRCChecker` a thin wrapper over `Os::ValidateFile` (or deprecates it in favor of `ValidateFile` for `Svc::FileWorker`).

Utils/CRCChecker.cpp
```diff
@@ -89,9 +90,15 @@ crc_stat_t create_checksum_file(const char* const fname) {
         return FAILED_FILE_CRC_OPEN;
     }
 
-    // Write  checksum  file
-    bytes_to_write = sizeof(checksum);
-    stat = f.write(reinterpret_cast<U8*>(&checksum), bytes_to_write);
+    // Write checksum file. Serialize the value rather than writing the raw U32 bytes so that the
+    // file contents do not depend on the endianness of the processor.
+    U8 checksum_data[sizeof(checksum)];
```
lestarch-autobot: [C++ Design] **suggestion** CPP-19 (explicit initialization): uninitialized local array.

`checksum_data` is declared without an initializer. It is fully written by `serializeFrom` before it is read, so no defect today, but CPP-19 asks every local to be explicitly initialized. Alternatively, `Utils::HashBuffer` + `hash.finalize(HashBuffer&)` already yields exactly these serialized bytes (it is what `Os::writeHash` persists), which would avoid the raw array altogether.

```suggestion
    U8 checksum_data[sizeof(checksum)] = {};
```

Utils/CRCChecker.cpp
```diff
@@ -119,14 +126,21 @@ crc_stat_t read_crc32_from_file(const char* const fname, U32& checksum_from_file
         return FAILED_FILE_CRC_OPEN;
     }
 
-    // Read  checksum  file
+    // Read checksum file
+    U8 checksum_data[sizeof(checksum_from_file)];
```
lestarch-autobot: [C++ Design] **suggestion** CPP-19 (explicit initialization): uninitialized local array.

`checksum_data` is declared without an initializer. The short-read check on the next lines guards every consumer, so no defect today, but CPP-19 asks every local to be explicitly initialized.

```suggestion
    U8 checksum_data[sizeof(checksum_from_file)] = {};
```

Svc/FileWorker/test/ut/FileWorkerTester.cpp
```diff
@@ -377,6 +377,59 @@ void FileWorkerTester ::testAppending() {
...
+    const char* fnameChar = "testroundtrip.txt";
+    const char* hashFileChar = "testroundtrip.txt.CRC32";
+    const FwSizeType dataSize = 1024;
+    U8 data[dataSize];
```
lestarch-autobot: [C++ Design] **suggestion** CPP-19 (explicit initialization): uninitialized local array.

Test-only local array declared without an initializer (filled by the loop below). CPP-19 applies to UT sources built with the flight toolchain config; a `= {}` keeps it explicit.

```suggestion
    U8 data[dataSize] = {};
```

Svc/FileWorker/test/ut/FileWorkerTester.cpp
```diff
@@ -377,6 +377,59 @@ void FileWorkerTester ::testAppending() {
...
+    this->invoke_to_writeIn(0, fname, writeBuffer, 0, false);
+    this->component.doDispatch();
+    ASSERT_from_writeDoneOut_SIZE(1);
+    ASSERT_from_writeDoneOut(0, FileWorkerStatus::FW_STATUS_DONE_WRITE, dataSize);
+    ASSERT_EVENTS_WriteFileError_SIZE(0);
+    ASSERT_EVENTS_WriteValidationError_SIZE(0);
+
+    // Read the same file back through the component
+    this->clearHistory();
+    U8 readData[dataSize];
```
lestarch-autobot: [C++ Design] **suggestion** CPP-19 (explicit initialization): uninitialized local array.

Test-only local array declared without an initializer (populated by `readIn`, then compared with `memcmp`). CPP-19 asks for explicit initialization; `= {}` also makes the `memcmp` well-defined if the read ever comes back short.

```suggestion
    U8 readData[dataSize] = {};
```

Utils/CRCChecker.cpp
```diff
@@ -89,9 +90,15 @@ crc_stat_t create_checksum_file(const char* const fname) {
...
-    stat = f.write(reinterpret_cast<U8*>(&checksum), bytes_to_write);
+    // Write checksum file. Serialize the value rather than writing the raw U32 bytes so that the
+    // file contents do not depend on the endianness of the processor.
+    U8 checksum_data[sizeof(checksum)];
+    Fw::SerialBuffer checksum_buffer(checksum_data, sizeof(checksum_data));
+    Fw::SerializeStatus ser_stat = checksum_buffer.serializeFrom(checksum);
+    FW_ASSERT(Fw::FW_SERIALIZE_OK == ser_stat, static_cast<FwAssertArgType>(ser_stat));
+
+    bytes_to_write = checksum_buffer.getSize();
+    stat = f.write(checksum_buffer.getBuffAddr(), bytes_to_write);
```
lestarch-autobot, line 101: [Operational] **could fix** `ops-failure-path`: after this upgrade, legacy `.CRC32` sidecars are unreadable on board and nothing in-repo regenerates them.

Scenario: a deployment whose stored sidecars were written by the pre-change `create_checksum_file` on a little-endian target upgrades to this commit. `verify_checksum` then returns `FAILED_FILE_CRC_CHECK` for every such file, and its only production callers (`Svc::FileWorker::readIn_handler` / `verifyIn_handler`) emit `CrcFailed(9)` and return `FW_STATUS_FAILED_CRC`, so the file cannot be read or verified.
No production code calls `create_checksum_file`, so repairing a 4-byte sidecar costs a full re-uplink of the data file through `writeIn` (1 MiB file to fix 4 bytes) unless the project adds its own regeneration step.
Operational judgment call. Smallest remedy: upgrade guidance stating legacy sidecars must be regenerated before the new build reads those files (ground-side, or on board via `Os::ValidateFile::createValidation`). FileWorker-written sidecars are unaffected: that path already stored the big-endian `Utils::Hash`/`HashBuffer` layout.

Utils/CRCChecker.cpp
lestarch-autobot: [Operational] **Concur** — also in scope for operational consequences: scope the note to sidecars written by the pre-change `create_checksum_file`; `Svc::FileWorker`'s own write path already stored the big-endian `Utils::Hash`/`HashBuffer` layout, so its sidecars were unreadable *before* this fix, not after it.

lestarch-autobot (changes requested):

### Per-agent results

| F Prime C/C++ Design | 0 | 4 | 0 | 0 | 4 | Go |
| Documentation Currency | 1 | 1 | 0 | 0 | 2 | No-Go |
| Design | 0 | 2 | 0 | 1 | 0 | Go |
| Operational | 0 | 0 | 2 | 0 | 2 | Go |
| Maintainability | 0 | 2 | 2 | 0 | 4 | Go |
| **Totals** | 1 | 10 | 4 | 1 | 13 | **No-Go** |

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>

**Documentation Currency**
- `Svc/FileWorker/docs/sdd.md` does not state the `.CRC32` sidecar layout (F Prime-serialized, big-endian) that `readIn`/`verifyIn` now enforce, nor that pre-change host-endian sidecars must be regenerated — also: Operational — https://github.com/nasa/fprime/pull/5775#discussion_r3973288815

</details>

### Merge readiness
**Merge readiness: No-Go** — Documentation Currency has 1 outstanding must-fix item (SDD not updated for the changed `.CRC32` contract).

---

Byte order sorted on every target; one SDD note and this checksum is cleared for flight.

winklemad:
Thanks for the detailed pass. Pushed a21b669 addressing the review:

- **Documentation Currency (must-fix):** added a "Checksum (`.CRC32`) file format" section to `Svc/FileWorker/docs/sdd.md` documenting the serialized, big-endian sidecar layout that `readIn`/`verifyIn` enforce, and — covering the Operational note too — that pre-change host-endian sidecars must be regenerated before the current build reads their data files (ground-side, or on board via `Os::ValidateFile::createValidation`), while `Svc::FileWorker`'s own `Utils::Hash` sidecars are unaffected. Flipped the PR's "Documentation Included" flag to y.
- **CPP-19 (explicit initialization):** zero-initialized the four local buffers — `checksum_data` in `create_checksum_file`/`read_crc32_from_file`, and `data`/`readData` in the round-trip test. Each is fully written before use, so no behavior change.

On the Design "future work" — folding `CRCChecker` into a thin wrapper over `Os::ValidateFile` to remove the duplicate hash/sidecar implementation: agreed that's the right long-term direction, but it's a larger refactor with its own compatibility surface, so I'd keep it to a separate issue-backed change rather than widen this fix.

Utils/CRCChecker.cpp
lestarch-autobot: [Operational] Fixed in a21b66966538f1ff91204d7506059d0ff6108f15.

Utils/CRCChecker.cpp
lestarch-autobot: [C++ Design] Fixed in a21b66966538f1ff91204d7506059d0ff6108f15.

Utils/CRCChecker.cpp
lestarch-autobot: [C++ Design] Fixed in a21b66966538f1ff91204d7506059d0ff6108f15.

Utils/CRCChecker.cpp
lestarch-autobot: [Operational] Fixed in a21b66966538f1ff91204d7506059d0ff6108f15 (the new SDD section covers the scoping requested in my concurrence; leaving resolution to the owning reviewer).

Svc/FileWorker/test/ut/FileWorkerTester.cpp
lestarch-autobot: [C++ Design] Fixed in a21b66966538f1ff91204d7506059d0ff6108f15.

Svc/FileWorker/test/ut/FileWorkerTester.cpp
lestarch-autobot: [C++ Design] Fixed in a21b66966538f1ff91204d7506059d0ff6108f15.

Svc/FileWorker/docs/sdd.md
```diff
@@ -97,3 +97,10 @@ Calling component passes in a file path *path* to read and a buffer for client t
 ### cancelIn_handler
 
 1. Set abort flag to true
+
+### Checksum (`.CRC32`) file format
+
+`writeIn_handler` writes a `.CRC32` sidecar alongside each file, and `readIn_handler` / `verifyIn_handler` validate the file against it. The sidecar stores the file's CRC-32 (a `U32`) in F Prime's serialized, big-endian representation — the same layout produced and consumed by `Utils::Hash` and `Os::ValidateFile` — so its contents do not depend on the endianness of the target that wrote it. `Svc::FileWorker` reaches this format through `Utils::CRCChecker`, whose `create_checksum_file` / `read_crc32_from_file` serialize and deserialize the value through `Fw::SerialBuffer`.
```
lestarch-autobot: [Documentation] **suggestion** New SDD section misattributes `writeIn_handler`'s sidecar write to `Utils::CRCChecker`.

`writeIn_handler` writes the `.CRC32` file via `writeBufferHashToFile` / `getHash` (`Utils::Hash` + `HashBuffer`, `FileWorker.cpp:203`, `:373`); only the read/verify side reaches `Utils::CRCChecker` (`Utils::verify_checksum`, `:73`, `:120`). As written the section also contradicts its own compatibility note below, which correctly says FileWorker's own sidecars come from `Utils::Hash`.

```suggestion
`writeIn_handler` writes a `.CRC32` sidecar alongside each file through `Utils::Hash` / `Utils::HashBuffer`, and `readIn_handler` / `verifyIn_handler` validate the file against it through `Utils::verify_checksum` (`Utils::CRCChecker`). The sidecar stores the file's CRC-32 (a `U32`) in F Prime's serialized, big-endian representation — the same layout produced and consumed by `Utils::Hash` and `Os::ValidateFile` — so its contents do not depend on the endianness of the target that wrote it. `Utils::CRCChecker` reaches that same layout by serializing and deserializing the value through `Fw::SerialBuffer` in `create_checksum_file` / `read_crc32_from_file`.
```

Utils/CRCChecker.cpp
lestarch-autobot: [Documentation] Fixed in a21b66966538f1ff91204d7506059d0ff6108f15 — the new "Checksum (`.CRC32`) file format" section in `Svc/FileWorker/docs/sdd.md` documents the serialized big-endian layout and the regeneration requirement for pre-change sidecars.

lestarch-autobot (approved):

### Per-agent results

| F Prime C/C++ Design | 0 | 4 | 0 | 0 | 0 | Go |
| Documentation Currency | 1 | 2 | 0 | 0 | 2 | Go |
| Design | 0 | 2 | 0 | 1 | 3 | Go |
| Operational | 0 | 0 | 2 | 0 | 0 | Go |
| Maintainability | 0 | 2 | 2 | 0 | 4 | Go |
| **Totals** | 1 | 11 | 4 | 1 | 10 | **Go** |

<details>

| F Prime C/C++ Design | 4 | 0 | 0 | 0 | 0 | 0 |
| Documentation Currency | 1 | 1 | 1 | 0 | 0 | 0 |
| Design | 0 | 3 | 0 | 0 | 0 | 0 |
| Test Quality | 0 | 1 | 0 | 0 | 0 | 0 |
| Operational | 2 | 0 | 0 | 0 | 0 | 0 |
| Maintainability | 0 | 4 | 0 | 0 | 0 | 0 |

</details>

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

### Merge readiness
**Merge readiness: Go** — all ten reviewers completed with zero outstanding must-fix items; the run-1 SDD must-fix is fixed in a21b669. Remaining open threads are non-blocking suggestions / could-fix items for maintainer discretion.

---

Sidecar bytes now agree on every target — CRC verified, this one is go for docking.

winklemad:
Good catch on the attribution — corrected in 8a9049f. The SDD section now says `Svc::FileWorker` writes the sidecar through `Utils::Hash` (`writeBufferHashToFile`) and validates it through `Utils::verify_checksum` (which reads it back via `Utils::CRCChecker::read_crc32_from_file`), rather than implying the write goes through `CRCChecker`. That also lines it up with the compatibility note below, which already credits FileWorker's own sidecars to `Utils::Hash`.

Svc/FileWorker/docs/sdd.md
lestarch-autobot: Fixed in 8a9049f0cfafcd70a39c3eee8d2cc48796208fa9.

Svc/FileWorker/docs/sdd.md
```diff
@@ -97,3 +97,10 @@ Calling component passes in a file path *path* to read and a buffer for client t
 ### cancelIn_handler
 
 1. Set abort flag to true
+
+### Checksum (`.CRC32`) file format
+
+`writeIn_handler` writes a `.CRC32` sidecar alongside each file, and `readIn_handler` / `verifyIn_handler` validate the file against it. The sidecar stores the file's CRC-32 (a `U32`) in F Prime's serialized, big-endian representation — the same layout produced and consumed by `Utils::Hash` and `Os::ValidateFile` — so its contents do not depend on the endianness of the target that wrote it. `Svc::FileWorker` writes the sidecar through `Utils::Hash` (`writeBufferHashToFile`) and validates it through `Utils::verify_checksum`, which reads the value back with `Utils::CRCChecker::read_crc32_from_file`; both sides serialize and deserialize the `U32` through `Fw::SerialBuffer`.
```
lestarch-autobot, line 103: **could fix** — `Utils::CRCChecker::read_crc32_from_file` is not a real symbol: `read_crc32_from_file` is a free function in namespace `Utils` (`Utils/CRCChecker.hpp`), and `Utils::Hash` serializes through `HashBuffer`, not `Fw::SerialBuffer`. Suggest naming the real API so readers can find it.

```suggestion
`writeIn_handler` writes a `.CRC32` sidecar alongside each file, and `readIn_handler` / `verifyIn_handler` validate the file against it. The sidecar stores the file's CRC-32 (a `U32`) in F Prime's serialized, big-endian representation — the same layout produced and consumed by `Utils::Hash` and `Os::ValidateFile` — so its contents do not depend on the endianness of the target that wrote it. `Svc::FileWorker` writes the sidecar through `Utils::Hash` (`writeBufferHashToFile`) and validates it through `Utils::verify_checksum`, which reads the value back with `Utils::read_crc32_from_file` (`Utils/CRCChecker.hpp`); both sides use F Prime serialization for the `U32`.
```

lestarch-autobot (approved):

### Per-agent results

| F Prime C/C++ Design | 0 | 4 | 0 | 0 | 0 | Go |
| Documentation Currency | 1 | 2 | 1 | 0 | 2 | Go |
| Design | 0 | 2 | 0 | 1 | 3 | Go |
| Operational | 0 | 0 | 2 | 0 | 0 | Go |
| Maintainability | 0 | 2 | 2 | 0 | 4 | Go |
| **Totals** | 1 | 11 | 5 | 1 | 10 | **Go** |

<details>

| F Prime C/C++ Design | 0 | 0 | 0 | 0 | 0 | 0 |
| Documentation Currency | 1 | 1 | 1 | 0 | 0 | 0 |
| Design | 0 | 3 | 0 | 0 | 0 | 0 |
| Test Quality | 0 | 1 | 0 | 0 | 0 | 0 |
| Operational | 0 | 0 | 0 | 0 | 0 | 0 |
| Maintainability | 0 | 4 | 0 | 0 | 0 | 0 |

</details>

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

### Merge readiness
**Merge readiness: Go** — all ten reviewers completed on 8a9049f with zero outstanding must-fix items; the run-2 SDD attribution suggestion is fixed in 8a9049f. Remaining open threads are non-blocking suggestions / could-fix items for maintainer discretion. Note: threads marked `Fixed in` remain visually open because `resolveReviewThread` is denied for the review token.

---

Three passes, one byte order, all systems nominal — cleared for docking, Commander.

sylvesterkaczmarek (approved):
Rechecked the current revision. The CRC sidecar now uses the same serialized representation as `Utils::Hash`, the regression covers both write/read directions through the framework validation path, and the SDD now states the legacy-sidecar migration boundary accurately. My earlier compatibility concern is resolved.

LeStarch (approved):

winklemad:
Thanks @LeStarch, and thanks to everyone who reviewed — tracing the `Hash::finalize`/`writeHash` path together made the compatibility story clear. Glad to get `CRCChecker` reading back the same canonical big-endian form the rest of the framework already writes.

Issue #5322, opened 2026-06-23
Config CI: Fix build failures when FPRIME_ENABLE_TEXT_LOGGERS=OFF


lestarch-autobot:
## Background

PR #5321 adds a CI matrix workflow that tests non-default `FpConfig.h` configuration toggles. Two profiles are currently **disabled** because they cause build failures:

| Profile | Flags | Error |
|---------|-------|-------|
| `minimal` | `-DFW_OBJECT_NAMES=0 -DFW_OBJECT_REGISTRATION=0 -DFW_QUEUE_REGISTRATION=0 -DFW_PORT_TRACING=0 -DFW_ENABLE_TEXT_LOGGING=0 -DFW_SERIALIZABLE_TO_STRING=0 -DFPRIME_ENABLE_TEXT_LOGGERS=OFF` | `error: symbol CdhCore is not defined` |
| `no-text-logging` | `-DFW_ENABLE_TEXT_LOGGING=0 -DFW_SERIALIZABLE_TO_STRING=0 -DFPRIME_ENABLE_TEXT_LOGGERS=OFF` | `error: symbol CdhCore is not defined` |

## Root Cause

Both profiles set `FPRIME_ENABLE_TEXT_LOGGERS=OFF`, which removes text logger components from the build. The `TestDeploymentsProject` topology references `CdhCore` which depends on text loggers being present. When they are removed, the symbol becomes undefined.

## What Needs to Be Done

1. Make `TestDeploymentsProject` (and any other test topologies) conditionally include text-logger-dependent components based on the `FPRIME_ENABLE_TEXT_LOGGERS` setting.
2. Once fixed, re-enable the profiles by removing the `"disabled": true` field from `ci/config-profiles.json`.

## Relevant Files

- `ci/config-profiles.json` — profile definitions (disabled entries have `"disabled": true` and a `"reason"` field)
- `.github/workflows/config-test.yml` — matrix workflow that filters disabled profiles
- `default/config/FpConfig.h` — configuration header with `#ifndef` guards
- PR #5321 — the config CI workflow that surfaced these failures

## Acceptance Criteria

- Both `minimal` and `no-text-logging` profiles build and pass unit tests against framework and Ref
- `"disabled": true` removed from both entries in `ci/config-profiles.json`

arham0254A:
Hi, I would like to take a look at this issue. Could I be assigned?

thomas-bc:
[...]

Issue #5677, opened 2026-08-13
Config Test has passed every PR since 2026-07-08 without building anything


kadircanyildirm-crypto:
| | |
|:---|:---|
|**_F´ Version_**| devel @ 0fbc03751 |
|**_Affected Component_**| CI |
---
## Problem Description

The `Config Test` check has been green on every pull request since 2026-07-08 without building anything.

`config-test.yml` builds its matrix like this:

```bash
MATRIX=$(cat ci/config-profiles.json | jq -c '[.[] | select(.disabled != true) | select(.name == "minimal")]')
echo "empty=$(echo "$MATRIX" | jq -r 'length == 0')" >> "$GITHUB_OUTPUT"
```

`minimal` carries `"disabled": true` — since #5321 added it on 2026-06-27, with the reason in the file and tracked in #5322. The two `select` filters cancel out, `empty` is `true`, and both real jobs are gated off by `if: needs.generate-matrix.outputs.empty != 'true'`. Skipped jobs do not fail a workflow, so the check reports success.

#5381 pinned the pull-request leg to `minimal` on 2026-07-08; `minimal` was already disabled at that commit.

## How to Reproduce

Every run since, on every trigger:

```
generate-matrix               success
Build [${{ matrix.name }}]    skipped
UTs [${{ matrix.name }}]      skipped
```

530 runs have concluded `success` this way — 395 `pull_request`, 134 `push`, 1 `schedule`. I opened 23 spread across the five weeks; all identical. The workflow's own 21:00 cron is empty for the same reason.

Config coverage itself is fine — `config-test-nightly.yml` runs all 8 enabled profiles and passes ([31587000443](https://github.com/nasa/fprime/actions/runs/31587000443)). What is missing is the pull-request leg, and a green check saying otherwise.

## Expected Behavior

`Config Test` on a pull request should either exercise a profile or not claim to. Two ways, and the pick looks like yours given #5641:

1. Point the pull-request leg at a profile that is enabled — costs 2 jobs per PR, which is what the workflow was written to cost.
2. Drop the `pull_request` and `schedule` triggers until #5322 lands, leaving the nightly to cover the profiles — frees queue time and removes a check that cannot fail.

Either way an empty selection is worth failing on rather than skipping, so disabling a profile cannot quietly remove a PR check again.

Happy to open the PR once you have picked one.

---

[...]

PR #5778, opened 2026-08-24
Build CdhCore without textLogger when FPRIME_ENABLE_TEXT_LOGGERS is off

kadircanyildirm-crypto:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5322 |

---
## Change Description

`Svc/Subtopologies/CdhCore/CMakeLists.txt` returned before registering anything when
`FPRIME_ENABLE_TEXT_LOGGERS` was off, so every topology referencing the subtopology
failed FPP with `error: symbol CdhCore is not defined` — the failure both disabled
profiles in `ci/config-profiles.json` share.

FPP has no conditional compilation, so the optional instance is expressed as two
thin variants around a shared include:

- `CdhCore.fpp` keeps `textLogger`; new `CdhCoreNoTextLogger.fpp` omits it. The
  bodies moved verbatim into `CdhCoreInstances.fppi` (module members) and
  `SubtopologyMembers.fppi` (topology members), so each variant is ~10 lines.
  CMake selects the variant from `FPRIME_ENABLE_TEXT_LOGGERS`.
- Same treatment for the one in-repo consumer of `CdhCore.textLogger`:
  `TestDeploymentsProject/Ref/Top/topology.fpp` keeps the
  `text event connections` specifier, `topology-no-text-logger.fpp` omits it, and
  the body moved verbatim into `RefTopologyMembers.fppi`. The `Ports_RateGroups`
  enum moved to `instances.fpp` (module level, shared by both variants).
- Both profiles re-enabled in `ci/config-profiles.json`; the conditional presence
  is noted in the subtopology `sdd.md`.

Base IDs are untouched, and with the option on the topology is unchanged: all
eight generated `Ref/Top` `*Ac.*` files are byte-identical before and after this
change; the two dictionary JSONs differ only in their `git describe` version
stamp (`v4.3.0` vs `v4.3.0-dirty`, from the then-uncommitted working tree).
`Svc.PassiveTextLogger` is the only component the option removes that any in-repo
`.fpp` references — there are no `Svc.ActiveTextLogger` instantiations.

## Rationale

Restores the behavior `cmake/options.cmake` documents for the option (exclude the
two text logger components) instead of the current effect (any deployment using
CdhCore loses command dispatch, events, telemetry and health), and completes both
items in #5322: the test topologies build with the option off, and the two
profiles are re-enabled — so `config-test.yml` exercises the `minimal` profile on
this PR itself.

All on `devel` @ `7d8f579`, fprime-tools 4.3.0 / fprime-fpp 3.3.0 (the versions
`requirements.txt` pins), `fprime-util generate <args>` + `fprime-util build --all`
in `TestDeploymentsProject`:

| configuration | before | after |
|---|---|---|
| default | builds | builds — generated `Ref/Top` outputs byte-identical (see above) |
| `-DFPRIME_ENABLE_TEXT_LOGGERS=OFF` alone | `symbol CdhCore is not defined` | builds |
| `minimal` profile, verbatim | same error | builds |
| `no-text-logging` profile, verbatim | same error | builds |
| framework root, `minimal` args | — | builds (806/806) |
| framework `fprime-util check --all`, `minimal` args | — | 100% — 129/129 ctest targets |
| Ref `fprime-util check --all`, `minimal` args | — | 100% — 2/2 |

Editing one of the new `.fppi` files retriggers topology generation and the
downstream compiles (fpp dep-info tracks includes; checked on the
`no-text-logging` build cache). With the option off, `locs.fpp` locates every
CdhCore instance in `CdhCoreNoTextLogger.fpp` and the generated dictionary
contains no `textLogger` entry.

Review focus: the three `.fppi` extractions are pure moves — `git diff
--color-moved` renders them as such, and the byte-identical generated outputs
confirm it semantically.

## Future Work

The deployment cookiecutter templates in fprime-tools also hardcode
`text event connections instance CdhCore.textLogger`
(`cookiecutter-fprime-deployment*/Top/topology.fpp:43`), so freshly generated
projects still will not build with the option off; that needs the same variant
treatment in fprime-tools.

Type of assistance: diagnosis of the CMake/FPP failure chain, drafting the file
restructure, and running the build/test matrix above. Tool: Claude Code. Scope:
the files in this PR. Level of modification: the moved blocks were kept verbatim
(verified by include-expansion diff and by the byte-identical generated outputs);
the author reviewed and approved the change and this description.


philphauler:
this also unblocks #5677 right? the minimal and no-text-logging profiles get re-enabled since CdhCore can now build without the text logger. clean split with the fppi includes.

checked against my #5765 in the same dir — no conflict, the integration tests dont reference textLogger directly so both topology variants should work fine.

*AI use per `AI_POLICY.md`. Tool: Claude Code. Type of assistance: reading the PR diff to check for conflicts with #5765. Level of modification: reviewed and edited.*

kadircanyildirm-crypto:
Partly — it removes the condition that made #5677 visible, rather than the reporting flaw itself. With `minimal` re-enabled, the PR-triggered `Config Test` builds framework and Ref again, so the "green with zero builds" state no longer occurs on PRs. Two pieces of #5677 remain: `generate-matrix` still concludes success if the matrix ever comes back empty, and most of the remaining profiles pass `FW_*` values that nothing reads (details in the issue thread), so what they exercise is the default build under another name.

Thanks for checking against #5765 — agreed, the integration tests go through the dictionary rather than the instance list, so they hold for either variant.

*AI use per `AI_POLICY.md`. Tool: Claude Code. Type of assistance: drafting this reply. Level of modification: reviewed and edited.*

philphauler:
Eyyy let's go ⚡️🚀

thomas-bc (commented):
This is a lot to navigate, and as you point out not well supported by FPP (hence the CMake logic around it). I'd be in favor of saying CdhCore and Ref both include text loggers - in you need CdhCore without text logger, you need to unfurl the subtopology manually

kadircanyildirm-crypto:
Agreed — with FPP having no conditionals, the CMake selection isn't worth the framework carrying. Happy to adopt the policy that CdhCore and Ref always include the text logger.

That does change #5322's scope, though: its acceptance criteria are to re-enable the `minimal` and `no-text-logging` profiles, but both pass `FPRIME_ENABLE_TEXT_LOGGERS=OFF` to a topology that instantiates CdhCore, so under this policy they can't be re-enabled — only removed. I'll close this PR and open a small one that drops the two profiles and documents the constraint in CdhCore's sdd.md, if that works for you and the issue owner.

(Separately: the fprime-tools deployment cookiecutters wire `CdhCore.textLogger`, so they'd inherit the same requirement — worth an issue there if the policy is settled.)

*AI use per `AI_POLICY.md`. Tool: Claude Code. Assistance: drafting; re-checked `ci/config-profiles.json` and the open PRs on `devel` before posting. Level of modification: reviewed and edited.*

Issue #5804, opened 2026-08-29
SDLS: default ClearText decryptor remains routable alongside a project decryptor; null cipher use is silent; downlink SA follows frame context


bitWarrior:
| | |
|:---|:---|
|**_F´ Version_**| devel (cd3845fc2) |
|**_Affected Component_**| `Svc/Subtopologies/ComCcsdsSdls` (docs), `Svc/Ccsds/SdlsSaRouter`, `Svc/Ccsds/ClearTextDecryptor`, `Svc/Ccsds/ClearTextEncryptor`, `Svc/Ccsds/CcsdsSdlsFramer` |
---
## Problem Description

The SDLS layer selects which decryptor handles an uplink frame solely from the 16-bit SA index carried in the frame itself (`CcsdsSdlsDeframer.cpp:32-41` → `SdlsSaRouter.cpp:36-44`). That is by design — the SA map (`SdlsCfg.SaMap`) is the policy — but three things around it make the default configuration easy to carry into a flight build unintentionally:

1. **Adoption guidance says "add", not "replace".** `ComCcsdsSdls/docs/sdd.md` §2.4 tells projects that "to route additional SAs to additional decryptors" they should extend `SdlsCfg.SaMap` / `SaRouterPortCount` and connect the new ports. §2.5 documents the default map as `{ SA 0 -> PLAINTEXT, SA 1 -> UNCONNECTED }`. A project that follows §2.4 literally — supplies a real decryptor on SA 1 and leaves the default in place — ends up with `ClearTextDecryptor` still mapped on SA 0. Because `ClearTextDecryptor::decryptIn_handler` returns `SdlsStatus::SUCCESS` unconditionally (`ClearTextDecryptor.cpp:29`), any frame tagged SA 0 is passed downstream as if it had been authenticated. The `[!WARNING]` at the top of the SDD says the default provides no security and must be overridden, but nothing says that leaving it *alongside* a real decryptor turns it into a bypass of that decryptor rather than a fallback.

2. **A null cipher handling traffic is silent.** Routing to `ClearTextDecryptor` / `ClearTextEncryptor` is a success path: no event, no `errorNotify`. `SdlsSaRouter` only emits `TrackingTableFull` and `UntrackedBufferReturned`. A misconfigured flight build gives no telemetry hint that frames are being accepted (or emitted) in the clear.

[...]

PR #5807, opened 2026-08-30
SDLS hardening: NullCipherInUse events, SA_INDEX authoritative on downlink, replace-not-add guidance

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5804 |

---
## Change Description

Three hardening changes to the SDLS layer, per #5804:

1. **Null cipher is loud.** `Svc.Ccsds.ClearTextDecryptor` and `ClearTextEncryptor` now raise a throttled `NullCipherInUse` WARNING_HI event (carrying the SA index) for every frame they handle. Both components gain the `Fw.Event` interface and a `timeCaller` port. Previously a frame passing through the null cipher was a silent success path with no telemetry trace.

2. **Downlink SA selection is configuration-only.** `Svc.Ccsds.CcsdsSdlsFramer` now encrypts every frame with its `SA_INDEX` parameter. A frame context `saIndex` that is set and differs from the parameter is reported with a throttled `ContextSaIndexIgnored` WARNING_LO event and ignored; an unset context (`SaIndexUnset`) raises nothing. Previously the context value took precedence, so anything upstream that populated `saIndex` (the field is written by the uplink deframer) could select the encryptor. No in-tree component carries an uplink context through to the framer today, so this closes a latent coupling rather than changing observed behavior in any shipped topology. Requirement `SVC-CCSDS-SDLS-FRAMER-002` reworded; `-009` added. `ComCfg.fpp` comment on `saIndex` updated to match.

3. **Docs say "replace", not "add".** `ComCcsdsSdls/docs/sdd.md` §2.4 is retitled *Replacing the Default Decryptor and Encryptor* and now carries a `[!WARNING]` that any SA left mapped to a `ClearText*` component is an unauthenticated path for the whole link, with instructions to remap it to `UNCONNECTED` or remove it when adding SAs. §2.5 notes that the SA map is the allow-list (unmapped SAs are already rejected with `UNKNOWN_SA`) and that the downlink SA comes from `SA_INDEX`. `SVC-COMCCSDSSDLS-005` carries the same caveat. `ClearText*` SDDs gain an Events section and requirement `-005`.

## Rationale

The SA router dispatches on the SA index read from the incoming frame, which is correct by design — the SA map is the policy. But the surrounding defaults made a dangerous configuration easy to reach: the subtopology guidance told integrators to route *additional* SAs to *additional* decryptors without saying to remove the default null cipher, nothing in telemetry revealed when the null cipher was in use, and the downlink framer let a context-supplied SA override its configured one. This change makes a misconfiguration visible, removes the latent downlink override, and corrects the guidance.

All three component suites pass; full native framework build (`fprime-util build`) passes with the new ports.

- `Svc_Ccsds_ClearTextDecryptor_ut_exe` / `Svc_Ccsds_ClearTextEncryptor_ut_exe` — 2/2 each. `*PassThrough` tests now also assert exactly one `NullCipherInUse` with the request's SA index.
- `Svc_Ccsds_CcsdsSdlsFramer_ut_exe` — 7/7. `Frame__ContextSa` rule now picks a context SA that differs from `TEST_PARAM_SA_INDEX` (27) and asserts the parameter reaches the encryption helper and is recorded in the context; the ignored request is asserted at most once per application (the event is throttled, so the 10,000-step randomized scenario must tolerate zero late in the run) and exactly once on a fresh tester via `assertContextSaIgnoredReported()`. `Frame__ParameterSa` is unchanged and still asserts no events.

Review focus:
- **Behavior change in `CcsdsSdlsFramer`:** the context `saIndex` no longer selects the encryptor. If any out-of-tree project relies on an upstream component setting `saIndex` to choose a downlink SA per packet, it will now get `ContextSaIndexIgnored` and encrypt on `SA_INDEX`. I believe this was never a supported use (the field is documented as written by deframers), but it is worth a maintainer call. A per-VC or allow-listed override could be added later if a real need appears.
- Throttle values (5) on the two new events.
- Adding `Fw.Event`/`timeCaller` to the `ClearText*` components means deployments importing `ComCcsdsSdls` must have their event/time pattern connections cover the `decryptor`/`encryptor` instances — standard for any component with events, but new for these two.

## Future Work

- None planned; the per-VC SA allow-list suggested in #5804 is intentionally not implemented, since the SA map already serves as the allow-list.

Claude Code was used to draft the implementation, unit tests, and documentation changes, and to run the unit test suites and framework build. All changes were reviewed by the author.


Svc/Ccsds/CcsdsSdlsFramer/CcsdsSdlsFramer.cpp
```diff
@@ -38,13 +38,16 @@ void CcsdsSdlsFramer ::comStatusIn_handler(FwIndexType portNum, Fw::Success& con
...
 
 void CcsdsSdlsFramer ::dataIn_handler(FwIndexType portNum, Fw::Buffer& data, const ComCfg::FrameContext& context) {
-    // Determine the security association index: use the context's value when set, otherwise the SA_INDEX parameter
-    const U16 unsetSaIndex = static_cast<U16>(ComCfg::SaIndexUnset);
-    U16 saIndex = context.get_saIndex();
-    if (saIndex == unsetSaIndex) {
-        Fw::ParamValid valid = Fw::ParamValid::INVALID;
-        saIndex = this->paramGet_SA_INDEX(valid);
-        FW_ASSERT(FW_PARAM_OK(valid), static_cast<FwAssertArgType>(valid.e));
+    // The configured SA_INDEX parameter is authoritative for the encryption path. The frame context's
```
thomas-bc, line 41: I don't think this is desired. Context value **should** be authoritative. Otherwise there always needs to be an operator in the loop to update the SA, which is not practical.

context is controlled by the application, if it is set it should be respected imo

thomas-bc (changes requested):

Svc/Ccsds/CcsdsSdlsFramer/CcsdsSdlsFramer.cpp
thomas-bc: if the docs say otherwise then the docs should be updated. Using frame context for this is the correct design imo

Svc/Ccsds/CcsdsSdlsFramer/CcsdsSdlsFramer.fpp
```diff
@@ -2,16 +2,22 @@ module Svc {
...
     @ 4. Allocate a frame buffer and prepend the security association index (16 bits) to the encrypted data
     @ 5. Pass the SDLS frame to the next component in the pipeline
     passive component CcsdsSdlsFramer {
 
-        @ Security association index used when the incoming frame context does not specify one
+        @ Security association index used to encrypt every downlink frame. This parameter is
+        @ authoritative: an SA index carried in the incoming frame context does not select the encryptor.
         param SA_INDEX: U16 default 0
 
+        @ Event raised when the incoming frame context requests an SA index other than the configured SA_INDEX
```
LeStarch, line 16: If we don't want to support application-level SA selection, we must remove it from the context....not merely add an event saying we ignored it.

LeStarch (approved):

Svc/Ccsds/CcsdsSdlsFramer/CcsdsSdlsFramer.cpp
LeStarch: CCSDS leaves this to the "Application Layer" to determine. However, it does make it clear that the selection of SA is a coordinated activity.  In our case, it would be coordinated through setting/saving the parameter.

Thus, I believe this PR is the correct approach.  I could concede that reading sa from context could be configure-opt-in. However, this may be more than we need for now.

LeStarch (approved):

Svc/Ccsds/CcsdsSdlsFramer/CcsdsSdlsFramer.cpp
thomas-bc: I really disagree here - this will hold for as long as we do only one type of encryption (i.e. AES). As soon as we try to add key exchanges, or even just more than 1 key for AES, it will break here. The application layer needs to be able to choose what scheme is used. It doesn't seem practical or even doable for the ground to attempt to command that (again, in the case of more than 1 key or algorithm.

thomas-bc: > the selection of SA is a coordinated activity

Maybe I'm mistaken here, but I don't read coordinated as "commanded". Let me read up a little more...

thomas-bc: My understanding is what's coordinated is what set of SA's is allowed. What the application chooses to use, that's up to the application. If we only allow one SA at a time, and that needs to be commanded, the spacecraft can't use 2 different keys at the same time, or re-key while also transmitting, or change any behavior without ground intervention.

thomas-bc: Or even AES for most things and cleartext for a subset of things (e.g. some events / telemetry or whatnot) –which is a use case we discussed– that would be made impossible by this change. We should sit down to understand where our difference of interpretation is coming from.

LeStarch:
We spun the changes that are approved into: https://github.com/nasa/fprime/pull/5832

The question of SA supplied by only parameter or app layer should be a separate PR and run through CCB.

PR #5808, opened 2026-08-30
Drop Svc.Ping input ports on overflow so the Health system gets to react

Willmac16:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**| - |

---
## Change Description

Set Svc.Ping input ports to `drop` rather than `assert`

## Rationale

This ensures the Health system gets to react to the queue overflow at the configured timeout threshold, rather than the assert hook firing & taking down the FSW/making cleanup much more complicated.

I think in general it'd be good to have some tests into how the system responds when compute is under contention/severely under provisioned to ensure the system fails gracefully where possible, rather than abruptly asserting.

## Future Work

I will likely have more suggestions for changes to overflow behavior in upstream components in subsequent PRs. 

The actual changes to the FPrime repo were made in vim w/ the help of ripgrep. However, [the overall topology analysis](https://github.com/Willmac16/fprime-topo-analysis) I'm working on, which reminded me of the "pings should drop" point, is tooling written by AI.

### DIFF
--- a/Svc/Health/Health.fpp
+++ b/Svc/Health/Health.fpp
@@ -11,7 +11,7 @@ module Svc {
     output port PingSend: [HealthPingPorts] Svc.Ping
 
     @ Ping return port
-    async input port PingReturn: [HealthPingPorts] Svc.Ping
+    async input port PingReturn: [HealthPingPorts] Svc.Ping drop
 
     @ Run port
     sync input port Run: Svc.Sched
--- a/Svc/Health/Health.fpp
+++ b/Svc/Health/Health.fpp
@@ -11,7 +11,7 @@ module Svc {
     output port PingSend: [HealthPingPorts] Svc.Ping
 
     @ Ping return port
-    async input port PingReturn: [HealthPingPorts] Svc.Ping drop
+    async input port PingReturn: [HealthPingPorts] Svc.Ping
 
     @ Run port
     sync input port Run: Svc.Sched
--- a/.github/skills/fprime-component-design-fpp/SKILL.md
+++ b/.github/skills/fprime-component-design-fpp/SKILL.md
@@ -92,7 +92,7 @@ for port definition syntax.
 | Need | Port | Import |
 |---|---|---|
 | Rate-group scheduling | `sync input port run: Svc.Sched` | `import Svc.Sched` |
-| Health ping | `async input port pingIn: Svc.Ping` | `import Svc.Ping` |
+| Health ping | `async input port pingIn: Svc.Ping drop` | `import Svc.Ping` |
 | Time access | `time get port timeCaller` | (special port) |
 | Command handling | `command recv port cmdIn` | (special port) |
 | Event logging | `event port eventOut` | (special port) |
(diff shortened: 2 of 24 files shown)


Willmac16:
@zimri-leisher my answer would be: set the ping priority just below the threshold you can't live w/o. 
```
        @ Ping in port
        # TODO should ping have highest prio? or lowest?
        async input port pingIn: Svc.Ping priority 10 drop
```

In my mind all three of these are below ping (everything else above):
* `tlmWrite` should be drop, lowest priority
* `seqRunIn` ideally should operate w/ a hook that returns the seq req. back to the dispatcher
* `checkTimers` should probably be drop (assuming the logic queries time rather than relying upon getting every single tick)

zimri-leisher:
> @zimri-leisher my answer would be: set the ping priority just below the threshold you can't live w/o.
> 
> ```
>         @ Ping in port
>         # TODO should ping have highest prio? or lowest?
>         async input port pingIn: Svc.Ping priority 10 drop
> ```
> 
> In my mind all three of these are below ping (everything else above):
> 
> * `tlmWrite` should be drop, lowest priority
> * `seqRunIn` ideally should operate w/ a hook that returns the seq req. back to the dispatcher
> * `checkTimers` should probably be drop (assuming the logic queries time rather than relying upon getting every single tick)

I think one of the difficulties with the `ping` system is that you have to set it so it detects the thing that you care about. for instance, for the sequencer, with the ping being the highest prio it will only detect if the actual event loop is stuck (stack overflow on the thread, some data race somehow). And putting it on the lowest prio will detect if the queue is not being emptied completely (this is not an option, because the sequence directives go on the queue, so it would cause health to fail if the sequence ran in a blocking manner for a while). I'm not sure there's a good rule of thumb for every component. Perhaps the standard should be an FPP configurable priority for the health port?

sylvesterkaczmarek (commented):
Should `Health.PingReturn` be excluded from this blanket `drop` change? This is Health’s own receive queue, not the monitored component’s queue. If it is full, a valid reply can be discarded, so the ping `cycleCount` is never reset and Health can emit `HLTH_PING_WARN` / `HLTH_PING_LATE` for a component that actually responded. That turns Health’s own queue pressure into a false fault. I think this port needs separate handling or a regression covering Health-queue overflow.

Svc/Health/Health.fpp
```diff
@@ -11,7 +11,7 @@ module Svc {
     output port PingSend: [HealthPingPorts] Svc.Ping
 
     @ Ping return port
-    async input port PingReturn: [HealthPingPorts] Svc.Ping
+    async input port PingReturn: [HealthPingPorts] Svc.Ping drop
```
LeStarch: I would prefer if this particular change were pushed to a new PR.

In all other cases, a drop is effectively the same as an unprocessed-ping, and health is designed to handle that correctly.  Thus having them assert just fast-forwards an otherwise clearer error.

However, this change could cause spurious changes in an otherwise working system.  Having this assert here means we make it easier to see the sizing errors in this queue by failing-fast....so this change is worth a discussion at the CCB.

LeStarch (changes requested):
I would ask we revert the one change to health, and then merge the rest.  We can break out that change into a new issue.

LeStarch (approved):

LeStarch:
Made the reversion. Please spin to a new PR if you really want health to drop responses.

Issue #5816, opened 2026-09-03
Integer overflow in LinearBufferBase::serializeFrom() can cause buffer overflow


Arslan8:
LinearBufferBase::serializeFrom() performs an overflow-prone bounds check before copying a caller-provided buffer:

if (this->m_serLoc + length > this->getCapacity()) {
    return FW_SERIALIZE_NO_ROOM_LEFT;
}

Because m_serLoc and length are unsigned size types, a sufficiently large length can cause m_serLoc + length to wrap to a small value. The capacity check may then incorrectly succeed, after which the copy operation is called with the original large length and can write beyond the destination buffer.

### How to Reproduce
Create a LinearBufferBase-derived serialization buffer with a small capacity.
Advance m_serLoc by serializing some data.
Call serializeFrom() with a length large enough that m_serLoc + length wraps around.
The capacity check can pass even though length exceeds the remaining destination capacity.
Expected Behavior

serializeFrom() should reject any length larger than the remaining capacity using an overflow-safe bounds check, for example:

if (this->m_serLoc > this->getCapacity() ||
    length > this->getCapacity() - this->m_serLoc) {
    return FW_SERIALIZE_NO_ROOM_LEFT;
}

PR #5819, opened 2026-09-03
fix(Fw/Types): use overflow-safe bounds checks in LinearBufferBase serialization

philphauler:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5816 |

---
## Change Description

Fixes two unsigned integer overflow vulnerabilities in `Fw/Types/Serializable.cpp` where `serLoc + length > capacity` can wrap to zero and silently pass the bounds check. Uses the subtraction form (`length > capacity - serLoc`) which cannot underflow because `serLoc <= capacity` is a class invariant.

Two sites fixed (+5/-2 lines):
1. `serializeFrom(const U8*, length)` — single subtraction-form check
2. `serializeFrom(const LinearBufferBase&)` — two-step check for `size + sizeof(FwSizeStoreType)` via an `available` intermediate

## Rationale

In unsigned arithmetic, `a + b > max` wraps to a small value when `a + b` overflows, making the guard silently ineffective. The subtraction form `b > max - a` is safe when `a <= max` (the invariant `serLoc <= capacity` holds), and is already the established idiom in this codebase.

The existing `LinearBufferBase` unit tests cover the affected serialization paths. Behavior is identical for all valid inputs; only the overflow edge case is corrected.

## Future Work

None planned.

Claude Code was used for codebase search (finding all `serLoc + length > capacity` instances) and fix design.

IAMAI

### DIFF
--- a/Fw/Types/Serializable.cpp
+++ b/Fw/Types/Serializable.cpp
@@ -255,7 +255,8 @@ LinearBufferBase::serializeFrom(const U8* buff,
     }
 
     // make sure we have enough space
-    if (this->m_serLoc + length > this->m_capacity) {
+    // Note: m_serLoc <= m_capacity is a class invariant, so this subtraction cannot underflow.
+    if (length > this->m_capacity - this->m_serLoc) {
         return FW_SERIALIZE_NO_ROOM_LEFT;
     }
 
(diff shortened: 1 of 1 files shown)


LeStarch (approved):

LeStarch (approved):

philphauler:
Hey @LeStarch 👋 just saw this was merged yesterday, thank you! I pushed three tests to the branch after the merge (766a947: length near max, buffer size near max, exact fit; without the fix the first one crashes the UT binary on the memmove past the buffer). If you would like them in devel I can open a small test-only PR, otherwise nothing more to do here. Have a great day!

Issue #5849, opened 2026-09-06
Svc::FileManager lacks the optional sandbox confinement its sibling file components have; FileHandling sandboxDir does not cover it


bitWarrior:
| | |
|:---|:---|
|**_F´ Version_**| devel (14e385d99), v4.3.0 |
|**_Affected Component_**| `Svc/FileManager`, `Svc/Subtopologies/FileHandling` |
---
## Problem Description

`Svc::FileManager` is the only file-handling component without an optional path-confinement (sandbox) configuration. All of its command handlers pass the ground-supplied path argument directly to `Os::FileSystem` / `Os::File` / `Os::Directory` calls:

| Command | Site |
|---|---|
| CreateDirectory | `FileManager.cpp:62` |
| RemoveFile | `FileManager.cpp:78` |
| MoveFile | `FileManager.cpp:101` |
| RemoveDirectory | `FileManager.cpp:116` |
| AppendFile | `FileManager.cpp:135` |
| FileSize | `FileManager.cpp:151` |
| ListDirectory | `FileManager.cpp:175` (plus `getFileSize` at `:457`) |
| CalculateCrc | `FileManager.cpp:202` |
| GenerateDp | `FileManager.cpp:247` |

Operating on operator-specified paths is FileManager's documented purpose, and the commands require command authority — so this is expected behavior, not a defect. However:

1. `Svc::FileUplink`, `Svc::FileDownlink`, and `Svc::PrmDb` all offer directory confinement via `Os::SandboxedFile` / `configureSandbox`, so FileManager is now the inconsistent one.
2. The `FileHandling` subtopology documents `FileHandlingConfig::Paths::sandboxDir` as the knob that restricts file access, and instantiates `fileManager` (`FileHandling.fpp:35`) — but the knob cannot cover it, because FileManager has no configure hook. A deployment that sets `sandboxDir` may reasonably believe its file access is confined when the file-management command set is not.

## Context / Environment

```
devel @ 14e385d99; also present in v4.3.0
```

## How to Reproduce

Send any FileManager command with an absolute path outside the deployment's data area (e.g. `FileSize /etc/hostname`, or `CreateDirectory /tmp/anywhere`); it succeeds. Unit tests demonstrating this on all handler paths were provided by the reporter and can be adapted as regression tests for the fix.

## Expected Behavior

Give FileManager the same optional, fail-open confinement as its sibling components:

[...]

PR #5850, opened 2026-09-06
Add fail-closed sandbox confinement to Svc::FileManager command paths

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5849 |

---
## Change Description

Gives `Svc::FileManager` the same directory-confinement configuration its sibling file components (`FileUplink`, `FileDownlink`, `PrmDb`) already have, per #5849.

- **New `FileManager::configure(const char* sandboxDir)`**, mirroring the `FileUplink`/`Os::SandboxedFile` pattern: **fail-closed** until called; configuring `"/"` permits any path (historical behavior).
- **All ten command path sites confined.** Every path argument (`CreateDirectory`, `RemoveFile`, both args of `MoveFile` and `AppendFile`, `RemoveDirectory`, `FileSize`, `ListDirectory`, `CalculateCrc`, `GenerateDp`) is canonicalized with `Os::FilePathUtils::resolveFromCwd` and checked with `checkContainment` before any OS call, via a single `resolveInSandbox()` helper. The canonical path is what reaches `Os::FileSystem`/`Os::File`/`Os::Directory`; events continue to echo the path as commanded.
- **Rejection is loud and side-effect free**: new `PathOutsideSandbox` (WARNING_HI, carries the rejected path and the configured directory — appended to `Events.fppi` so existing event IDs are unchanged), `Errors` telemetry incremented, `VALIDATION_ERROR` command response, no filesystem operation.
- **Subtopologies configure it open by default** for backwards compatibility: `FileHandling` calls `fileManager.configure(FileHandlingConfig::Paths::sandboxDir)` (default `"/"`; the constant's comment now lists `fileManager`), and `FileHandlingCfdp` calls `fileManager.configure("/")`, matching how each already handles `prmDb`.
- **Docs**: new Security Considerations section in the FileManager SDD; the `FileHandling` and `FileHandlingCfdp` SDD warning blocks now cover `fileManager` and show the restricting call.

## Rationale

FileManager was the only file-handling component with no optional sandbox: a deployment that set `FileHandlingConfig::Paths::sandboxDir` to a restricted directory confined uplink, downlink, and the parameter database — but the file-management command set could still operate on any path the process can reach. Operating on operator-specified paths is FileManager's documented purpose, so the default remains unrestricted through the subtopologies; this change makes confinement *possible* and consistent, and makes an unconfigured (out-of-subtopology) instantiation fail loudly rather than silently unrestricted, matching the fail-closed convention `Os::SandboxedFile` adopted.

`Svc_FileManager_ut_exe`: 32/32 pass (28 existing + 4 new). Full native framework build passes, which compiles both subtopologies' generated `configComponents` phases.

- `Sandbox.rejectsOutsidePaths` — with the sandbox set to a test directory: absolute-outside and `..`-escaping paths are rejected on every command (including per-argument checks on `MoveFile`/`AppendFile`), with the event/response/telemetry contract asserted and the filesystem verified untouched; a contained command still succeeds.
- `Sandbox.failClosed` — an unconfigured component rejects commands.
- `Sandbox.resolvesPaths` — `..` segments that stay inside the sandbox resolve and the command succeeds on the canonical path.
- `Sandbox.openRootIsUnrestricted` — with `configure("/")` (the subtopology default), the exact operations from the motivating report (create/size/move/remove at arbitrary absolute paths under `/tmp`) still succeed, pinning the backwards-compatibility guarantee.
- Existing tests run with the sandbox configured to `"/"` in the tester constructor (as the subtopologies do) and pass unchanged, demonstrating behavioral compatibility.

Review focus:
- **Fail-closed default**: out-of-tree projects that instantiate `Svc.FileManager` directly (both in-tree instantiations are in the subtopologies, which are wired here) must add a `configure()` call or every command fails with `PathOutsideSandbox`. This is deliberate — it matches `FileUplink` — but it is the migration-visible part of the change.
- Containment is textual (no `realpath`), the same limitation as `Os::SandboxedFile`; symlinks inside the sandbox are out of scope.
- `PathOutsideSandbox` placement at the end of `Events.fppi` (event IDs stable; dictionaries gain one event).

Regression tests derive from the reporter's PoC in the tracking issue.

## Future Work

- Consider a follow-on aligning `Svc::CmdSequencer`/`Svc::DpCatalog` and other path-taking components with the same pattern if desired.

Claude Code was used to draft the implementation, unit tests, and documentation changes, and to run the unit test suite and framework build. All changes were reviewed by the author.

### DIFF
--- a/Svc/FileManager/FileManager.cpp
+++ b/Svc/FileManager/FileManager.cpp
@@ -65,8 +65,9 @@ void FileManager ::configure(const char* sandboxDir) {
     // Ensure trailing '/'
     const FwSizeType resolvedLen = Fw::StringUtils::string_length(resolved, Os::FilePathUtils::MAX_PATH_LENGTH);
     FW_ASSERT(resolvedLen > 0);
-    FW_ASSERT(resolvedLen + 2 <= Os::FilePathUtils::MAX_PATH_LENGTH);
+    FW_ASSERT(resolvedLen < Os::FilePathUtils::MAX_PATH_LENGTH);
     if (resolved[resolvedLen - 1] != '/') {
+        FW_ASSERT(resolvedLen + 2 <= Os::FilePathUtils::MAX_PATH_LENGTH);
         resolved[resolvedLen] = '/';
         resolved[resolvedLen + 1] = '\0';
     }
--- a/Svc/FileManager/FileManager.cpp
+++ b/Svc/FileManager/FileManager.cpp
@@ -16,7 +16,9 @@
 #include <Fw/FPrimeBasicTypes.hpp>
 #include "Fw/Types/Assert.hpp"
 #include "Fw/Types/ExternalString.hpp"
+#include "Fw/Types/StringUtils.hpp"
 #include "Os/Directory.hpp"
+#include "Os/FilePathUtils.hpp"
 #include "Svc/FileManager/FileManager.hpp"
 #include "config/FileManagerConfig.hpp"
 
@@ -31,6 +33,8 @@ FileManager ::FileManager(const char* const compName  //!< The component name
     : FileManagerComponentBase(compName),
       commandCount(0),
       errorCount(0),
+      m_sandboxDir(""),
+      m_sandboxConfigured(false),
       m_listState(IDLE),
       m_totalEntries(0),
       m_currentOpCode(0),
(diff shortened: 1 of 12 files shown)


Svc/FileManager/FileManager.cpp
```diff
@@ -49,17 +53,69 @@ FileManager ::FileManager(const char* const compName  //!< The component name
...
+    // Resolve the sandbox directory (relative paths resolve against CWD)
+    char resolved[Os::FilePathUtils::MAX_PATH_LENGTH];
+    const Os::FilePathUtils::Status resolveStatus =
+        Os::FilePathUtils::resolveFromCwd(sandboxDir, resolved, sizeof(resolved));
+    FW_ASSERT(resolveStatus == Os::FilePathUtils::VALID, static_cast<FwAssertArgType>(resolveStatus));
+
+    // Ensure trailing '/'
+    const FwSizeType resolvedLen = Fw::StringUtils::string_length(resolved, Os::FilePathUtils::MAX_PATH_LENGTH);
+    FW_ASSERT(resolvedLen > 0);
+    FW_ASSERT(resolvedLen + 2 <= Os::FilePathUtils::MAX_PATH_LENGTH);
```
LeStarch: This is only required if path[resolvedLen - 1]  !=  "/".  Otherwise it is not required as we are not adding to the string in the case that the path is well-formed (ends in /).

LeStarch (changes requested):
One minor nit.

LeStarch (approved):

PR #5852, opened 2026-09-07
Add sandbox confinement to Svc::FileManager

### LINKED ISSUE #5849: Svc::FileManager lacks the optional sandbox confinement its sibling file components have; FileHandling sandboxDir does not cover it

Moferanoluwa:
| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime#5849 |

---
## Change Description

`Svc::FileManager` was the only file-handling component without optional path confinement. All nine of its command handlers (`CreateDirectory`, `RemoveFile`, `MoveFile`, `RemoveDirectory`, `AppendFile`, `FileSize`, `ListDirectory`, `CalculateCrc`, `GenerateDp`) passed ground-supplied paths straight to `Os::FileSystem`/`Os::File`/`Os::Directory`, and the `FileHandling` subtopology's `sandboxDir` knob did not cover it.

- Added `FileManager::configure(sandboxDir)`, fail-closed like `FileUplink`/`FileDownlink`/`PrmDb`. Since `FileManager` doesn't go through `Os::File` uniformly (most handlers call `Os::FileSystem` statics or `Os::Directory` directly), it validates paths itself via a new private `checkSandbox()` helper built on `Os::FilePathUtils::resolveFromCwd` + `checkContainment`, rather than wrapping `Os::SandboxedFile` (which only covers `Os::File::open`).
- Every path argument is checked, including both legs of `MoveFile`/`AppendFile` (both are checked independently, not short-circuited, so a violation on either side is reported) and the per-entry path `ListDirectory` constructs while walking a directory (defense in depth against a platform that yields a `.`/`..` entry — POSIX's `Os::Directory` already filters these, but the check doesn't assume that). Rejections emit a new `PathOutsideSandbox` warning event. `RemoveFile`'s `ignoreErrors` flag only waives missing-file errors, not the sandbox check.
- Extracted `Os::FilePathUtils::resolveDirectory()` (resolve + ensure trailing `/`) out of `Os::SandboxedFile::configure()` so `FileManager::configure()` doesn't reimplement that logic.
- `FileHandling` now configures `fileManager`'s sandbox to `FileHandlingConfig::Paths::sandboxDir` (`"/"` by default) alongside its siblings, and `Ref` reconfigures it to `"."` like `fileUplink`/`fileDownlink`/`prmDb`.

## Rationale

Fixes #5849. A deployment that sets `FileHandlingConfig::Paths::sandboxDir` (or reconfigures it) to restrict file access could reasonably believe the whole `FileHandling` subtopology is confined, when `FileManager`'s command set — which can create/remove/move/append/CRC any path, and package arbitrary files into data products — was not.

Added to `Svc/FileManager/test/ut`: fail-closed-by-default rejection, an escaping absolute path rejected on a single-path command, a path inside the configured sandbox still succeeding, both legs of `MoveFile` reported independently when both escape, `ignoreErrors=true` not bypassing the check on `RemoveFile`, and the three structurally distinct code paths (`ListDirectory`'s async open, `GenerateDp`'s always-OK-response convention, `CalculateCrc`'s direct `Os::File` use). Also added `test_file_manager_sandbox` to the `Ref` integration suite, mirroring the existing `test_prm_db_sandbox`.

This dev environment can't natively build F' (no Windows platform support in the CMake build), so I set up WSL Ubuntu to verify:
- `Os_FilePathUtils_Test` (24/24) and `Os_SandboxedFile_Test` (11/11) still pass after extracting `resolveDirectory()`.
- `Svc_FileManager_ut_exe` (36/36, all pre-existing tests plus the 8 new sandbox tests) passes.
- The `Ref` deployment builds and links cleanly end-to-end, which exercises both the `FileHandling.fpp` `configComponents` phase change and the `RefTopology.cpp` wiring (these aren't touched by a component-level `fprime-util check`, so this was the only way to compile-check them).

I did **not** run the new `test_file_manager_sandbox` Ref integration test itself — that needs `fprime-gds`, which isn't set up in this environment. It follows the same pattern as `test_prm_db_sandbox` immediately above it in the same file, which does pass in CI.

## Future Work

None.

- **Type**: Traced the confinement gap through the sibling components (`FileUplink`, `FileDownlink`, `PrmDb`) and the recent `Os::SandboxedFile` fail-closed rework (#5833), designed and implemented the fix, wrote the unit/integration tests, and verified the build in WSL since this machine can't natively build F'.
- **Scope**: `FileManager`, `Os::FilePathUtils`/`Os::SandboxedFile` (extracted a shared helper, no behavior change there), `FileHandling` subtopology wiring, `Ref` topology/integration test.
- **Level of modification**: I reviewed and take responsibility for every line; verification was via the actual test suite and a full deployment build, not just code reading.

IAMAI

### DIFF
--- a/Os/FilePathUtils.cpp
+++ b/Os/FilePathUtils.cpp
@@ -172,6 +172,27 @@ Status resolveFromCwd(const char* path, char* resolvedOut, FwSizeType resolvedSi
     return resolvePath(path, "/", resolvedOut, resolvedSize);
 }
 
+Status resolveDirectory(const char* directory, char* resolvedOut, FwSizeType resolvedSize) {
+    const Status resolveStatus = resolveFromCwd(directory, resolvedOut, resolvedSize);
+    if (resolveStatus != VALID) {
+        return resolveStatus;
+    }
+
+    const FwSizeType resolvedLen = Fw::StringUtils::string_length(resolvedOut, resolvedSize);
+    if (!isValidCString(resolvedLen, resolvedSize)) {
+        return INVALID_PATH;
+    }
+    FW_ASSERT(resolvedLen > 0);  // resolveFromCwd always yields at least "/"
+    if (resolvedOut[resolvedLen - 1] != '/') {
+        if (resolvedLen + 2 > resolvedSize) {
+            return TOO_LONG;
+        }
+        resolvedOut[resolvedLen] = '/';
+        resolvedOut[resolvedLen + 1] = '\0';
+    }
+    return VALID;
+}
+
 // Internal containment check on already-resolved paths.
 // Verifies that resolvedPath starts with allowedDirectory as a prefix,
 // and that the match occurs at a `/` boundary.
--- a/Os/FilePathUtils.hpp
+++ b/Os/FilePathUtils.hpp
@@ -68,6 +68,19 @@ Status resolvePath(const Fw::ConstStringBase& path, const Fw::ConstStringBase& b
 //!
 Status resolveFromCwd(const char* path, char* resolvedOut, FwSizeType resolvedSize);
 
+//! \brief Resolve a directory path and ensure it ends with a trailing separator
+//!
+//! Convenience wrapper for configuring a sandbox: resolves the directory via
+//! resolveFromCwd() and appends a trailing `/` if not already present, so the
+//! result can be passed directly as the allowedDirectory argument to checkContainment().
+//!
+//! \param directory: input directory path (may be relative or absolute)
+//! \param resolvedOut: output buffer for the resolved directory (with trailing `/`)
+//! \param resolvedSize: size of the output buffer
+//! \return VALID on success, INVALID_PATH or TOO_LONG on failure
+//!
+Status resolveDirectory(const char* directory, char* resolvedOut, FwSizeType resolvedSize);
+
 //! \brief Check whether an already-resolved path is within an allowed directory
 //!
 //! Both arguments must already be canonical absolute paths (output of resolvePath
--- a/Os/SandboxedFile.cpp
+++ b/Os/SandboxedFile.cpp
@@ -3,9 +3,7 @@
 // \brief Implementation of directory-sandboxed file wrapper
 // ======================================================================
 #include <Fw/Types/Assert.hpp>
-#include <Fw/Types/StringUtils.hpp>
 #include <Os/SandboxedFile.hpp>
-#include <cstring>
 
 namespace Os {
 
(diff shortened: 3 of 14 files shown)


thomas-bc:
Duplicate of #5850 - closing

Issue #5851, opened 2026-09-07
FpySequencer parameter reads check the wrong output port


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| `devel`; also present at `cd3845fc2ba957aa65751c4851046ee1c7d3ef3b` |
|**_Affected Component_**| `Svc::FpySequencer` |

## Problem Description

`pushPrm_directiveHandler()` checks `isConnected_prmGet_OutputPort(0)` but calls `getParam_out(0, ...)`, which is a different port. This can incorrectly reject a connected parameter-read port or pass the guard with the actual port disconnected, potentially hitting the generated connection assertion.

## Context / Environment

Source inspection only; no runtime reproduction or `fprime-util version-check` performed.

## How to Reproduce

1. Inspect [`pushPrm_directiveHandler()`](https://github.com/nasa/fprime/blob/cd3845fc2ba957aa65751c4851046ee1c7d3ef3b/Svc/FpySequencer/FpySequencerDirectives.cpp#L430).
2. Compare the guarded port (`prmGet`) with the invoked port (`getParam`).

## Expected Behavior

Check `isConnected_getParam_OutputPort(0)` before invoking `getParam_out`.

Hashim1999164:
hey, opened a pr that checks getParam instead of prmGet: https://github.com/nasa/fprime/pull/5865

PR #5865, opened 2026-09-08
Check getParam connection in PUSH_PRM

Hashim1999164:
| | |
|:---|:---|
|**_Related Issue(s)_**| #5851 |

---
## Change Description

PUSH_PRM checked isConnected_prmGet_OutputPort before calling getParam_out. prmGet is the component param port. Sequence param reads go through getParam.

## Rationale

Fixes the wrong port check in #5851 so a disconnected getParam is caught and a disconnected prmGet does not block PUSH_PRM.

Extended the pushPrm unit test. One case leaves prmGet unconnected and still succeeds. One case leaves getParam unconnected and returns PRM_GET_NOT_CONNECTED.

## Future Work

None.

Used an assistant to locate the handler and add the unit tests. I checked the port names against FpySequencer.fpp.

IAMAI

### DIFF
--- a/Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
+++ b/Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
@@ -288,6 +288,15 @@ TEST_F(FpySequencerTester, pushPrm) {
     ASSERT_EQ(result, Signal::stmtResponse_success);
     ASSERT_EQ(err, DirectiveError::NO_ERROR);
     ASSERT_EQ(tester_get_m_runtime_ptr()->stack.size, Fpy::MAX_STACK_SIZE);
+
+    tester_get_m_runtime_ptr()->stack.size = 1;
+    clearHistory();
+
+    this->component.set_getParam_OutputPort(0, nullptr);
+    result = tester_pushPrm_directiveHandler(directive, err);
+    ASSERT_EQ(result, Signal::stmtResponse_failure);
+    ASSERT_EQ(err, DirectiveError::PRM_GET_NOT_CONNECTED);
+    this->component.set_getParam_OutputPort(0, this->get_from_getParam(0));
 }
 
 TEST_F(FpySequencerTester, cmd) {
--- a/Svc/FpySequencer/FpySequencerDirectives.cpp
+++ b/Svc/FpySequencer/FpySequencerDirectives.cpp
@@ -428,7 +428,6 @@ Signal FpySequencer::pushTlmValAndTime_directiveHandler(const FpySequencer_PushT
 }
 
 Signal FpySequencer::pushPrm_directiveHandler(const FpySequencer_PushPrmDirective& directive, DirectiveError& error) {
-    // PUSH_PRM reads sequence params through getParam, not the component prmGet port
     if (!this->isConnected_getParam_OutputPort(0)) {
         error = DirectiveError::PRM_GET_NOT_CONNECTED;
         return Signal::stmtResponse_failure;
--- a/Svc/FpySequencer/FpySequencerDirectives.cpp
+++ b/Svc/FpySequencer/FpySequencerDirectives.cpp
@@ -428,7 +428,8 @@ Signal FpySequencer::pushTlmValAndTime_directiveHandler(const FpySequencer_PushT
 }
 
 Signal FpySequencer::pushPrm_directiveHandler(const FpySequencer_PushPrmDirective& directive, DirectiveError& error) {
-    if (!this->isConnected_prmGet_OutputPort(0)) {
+    // PUSH_PRM reads sequence params through getParam, not the component prmGet port
+    if (!this->isConnected_getParam_OutputPort(0)) {
         error = DirectiveError::PRM_GET_NOT_CONNECTED;
         return Signal::stmtResponse_failure;
     }


LeStarch:
@Hashim1999164 thanks for the contribution.  Next time, please consult the [contribution guidelines](https://github.com/nasa/fprime/blob/devel/CONTRIBUTING.md) as the issue this addresses has not passed CCB and thus has not been approved for implementation yet.

Since this is an obvious bug fix, I will recommend we proceed with this PR, however; in the future please check the status of an issue before submitting a PR.

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
```diff
@@ -235,6 +235,22 @@ TEST_F(FpySequencerTester, pushPrm) {
     ASSERT_EQ(result, Signal::stmtResponse_failure);
     ASSERT_EQ(err, DirectiveError::STACK_OVERFLOW);
     err = DirectiveError::NO_ERROR;
+
+    tester_get_m_runtime_ptr()->stack.size = 1;
+    clearHistory();
+
+    // prmGet is the component param port. PUSH_PRM should still succeed without it.
+    this->component.set_prmGet_OutputPort(0, nullptr);
```
LeStarch: I don't think we want this here.  This is the valid case.

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
```diff
@@ -235,6 +235,22 @@ TEST_F(FpySequencerTester, pushPrm) {
...
+    result = tester_pushPrm_directiveHandler(directive, err);
+    ASSERT_EQ(err, DirectiveError::NO_ERROR);
+    ASSERT_EQ(result, Signal::stmtResponse_success);
+    this->component.set_prmGet_OutputPort(0, this->get_from_prmGet(0));
+    err = DirectiveError::NO_ERROR;
+
+    this->component.set_getParam_OutputPort(0, nullptr);
+    result = tester_pushPrm_directiveHandler(directive, err);
+    ASSERT_EQ(result, Signal::stmtResponse_failure);
+    ASSERT_EQ(err, DirectiveError::PRM_GET_NOT_CONNECTED);
```
LeStarch, line 298: I think we need to restore the previous port set here after this test case.

LeStarch (changes requested):
I think there are two issues with the tests.  Will confirm via CI.

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
Hashim1999164: yeah dropped that. the succeed path is already covered above.

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
Hashim1999164: yep restoring getParam after this now.

Hashim1999164:
yeah my bad on the CCB thing. I just saw the wrong port and jumped. wont skip that next time.

tests are updated: dropped the extra succeed case and getParam gets reconnected after the not connected check.

Hashim1999164:
rebased this onto devel. kept the exact fit stack case that landed there plus the getParam disconnect check, and reconnect the port after it.

LeStarch (approved):

LeStarch:
I removed the comment because there was no need to refer to the original (broken) code in the new fixed code.

Svc/FpySequencer/test/ut/FpySequencerTestMain.cpp
```diff
@@ -288,6 +288,15 @@ TEST_F(FpySequencerTester, pushPrm) {
     ASSERT_EQ(result, Signal::stmtResponse_success);
     ASSERT_EQ(err, DirectiveError::NO_ERROR);
     ASSERT_EQ(tester_get_m_runtime_ptr()->stack.size, Fpy::MAX_STACK_SIZE);
+
+    tester_get_m_runtime_ptr()->stack.size = 1;
+    clearHistory();
+
+    this->component.set_getParam_OutputPort(0, nullptr);
```
lestarch-autobot, line 295: This line is what aborts the UT. `set_getParam_OutputPort(0, nullptr)` calls the autocoded `OutputPrmGetPort::addCallPort`, whose first statement is `FW_ASSERT(callPort != nullptr)` (`PrmGetPortAc.cpp:109`), so the process dies before `pushPrm_directiveHandler` ever runs and the new `isConnected_getParam_OutputPort` check is never exercised.

The framework has no port disconnect, so the port must be reconstructed in place — the same idiom `WasmSequencerTester::disconnectGetParam` uses. Friendship is not inherited by the `TEST_F` subclass, so this has to be a `FpySequencerTester` member (like `tester_get_m_runtime_ptr`):

```suggestion
    tester_disconnect_getParam();
```

`FpySequencerTester.hpp` (next to `tester_get_m_runtime_ptr()`):
```cpp
    //! Reset getParam to unconnected (framework has no disconnect; reconstruct port in place)
    void tester_disconnect_getParam();
```

`FpySequencerTester.cpp`:
```cpp
void FpySequencerTester::tester_disconnect_getParam() {
    Fw::OutputPrmGetPort& port = this->cmp.m_getParam_OutputPort[0];
    port.~OutputPrmGetPort();
    new (&port) Fw::OutputPrmGetPort();
    port.init();
}
```

Verified locally on this branch: all 299 `Svc_FpySequencer_ut_exe` tests pass with these three changes. The existing reconnect line below is fine as-is.

PR #5936, opened 2026-09-10
Add alias for Os::Mutex

jrussino:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/5945, https://github.com/nasa/fprime/issues/5249 |
|**_Has Unit Tests (y/n)_**| n |
|**_Documentation Included (y/n)_**| n |
|**_Generative AI was used in this contribution (y/n)_**| y |

---
## Change Description

This PR applies the same pattern used for `Os::RawTime` in https://github.com/nasa/fprime/pull/5617 to `Os::Mutex` :

Update fprime so that `Os::Mutex` is an alias (set by the `using` keyword) which defaults to `Os::DelegateMutex` but which projects can override to be a platform specific implementation of Mutex.

## Rationale

See rationale for https://github.com/nasa/fprime/issues/5249

This PR is being proposed now as it is expected to provide a performance improvement needed for a specific project using F Prime. 

## Testing/Review Recommendations

Verify the build works & verify UTs pass.

## Future Work

Future work: convert other OS services to follow this pattern as described in https://github.com/nasa/fprime/issues/5249#issuecomment-5287632693.

## AI Usage (see [policy](https://github.com/nasa/fprime/blob/devel/AI_POLICY.md))

Claude Code (Opus 4.8) was used to generate an initial draft of the proposed change to Os::Mutex based on the pattern used for Os::RawTime in https://github.com/nasa/fprime/pull/5617, and also for help gathering internal performance metrics on the impact of this change.

PR #5955, opened 2026-09-12
Update additional classes in Os to offer the alias option

bevinduckett:
| | |
|:---|:---|
|**_Related Issue(s)_**|  5249 |

---
## Change Description

Other PRs related to adding OS aliases (https://github.com/nasa/fprime/issues/5249) 
https://github.com/nasa/fprime/pull/5617 added the Os::RawTime alias
https://github.com/nasa/fprime/pull/5936 is for the Os::Mutex alias 

This PR covers the remainder of the "simple" cases notes in https://github.com/nasa/fprime/issues/5249, specifically it adds aliases for: 

Os::RawTime
Os::File
Os::FileSystem
Os::Directory
Os::ConditionVariable
Os::Console
Os::CountingSemaphore
Os::Cpu
Os::Memory

## Rationale

See discussion in https://github.com/nasa/fprime/issues/5249

Verify the build works & verify UTs pass.

## Future Work

Future work is to decide whether to convert the more complex classes, [see ](https://github.com/nasa/fprime/issues/5249#issuecomment-5287632693)

AI was used to assist in refactoring / boilerplate comment & code generation  and to assist with debugging  build system / unit test errors. All UT changes have been reviewed


Os/test/ut/filesystem/FileSystemRules.cpp
```diff
@@ -370,14 +370,13 @@ void Os::Test::FileSystem::Tester::GetSetWorkingDirectory::action(Os::Test::File
...
+    // Change working directory
     std::string other_dir = state.get_random_directory().path;
-    std::string expected_cwd = original_cwd + "/" + other_dir;
     status = Os::FileSystem::getSingleton().changeWorkingDirectory(other_dir.c_str());
     ASSERT_EQ(status, Os::FileSystem::Status::OP_OK) << "Failed to change working directory";
     status = Os::FileSystem::getSingleton().getWorkingDirectory(cwdBuffer, cwdSize);
-    ASSERT_EQ(status, Os::FileSystem::Status::OP_OK) << "Failed to get new working directory";
-    ASSERT_EQ(expected_cwd.compare(cwdBuffer), 0) << "getWorkingDirectory did not return the expected directory";
+    ASSERT_EQ(status, Os::FileSystem::Status::OP_OK) << "Failed to change working directory back";
+    ASSERT_TRUE(other_dir.compare(cwdBuffer)) << "getWorkingDirectory did not return the expected directory";
```
lestarch-autobot: [Test Quality] **must fix** `ASSERT_TRUE(other_dir.compare(cwdBuffer))` passes by construction and re-introduces the tautology fixed in https://github.com/nasa/fprime/pull/5706.

`other_dir` is relative (`filesystem_test_directory/...`, CommonTests.cpp:38) while `getWorkingDirectory` returns an absolute path, so `compare()` is always non-zero and the assert never fails. If the paths ever matched, `compare()==0` would fail the test, so the check is also inverted. This PR's diff drops `expected_cwd` and the exact comparison; restore them (the line-378 failure message is also wrong again).

```suggestion
    // Change working directory. Tracked paths are relative to the original working directory.
    std::string other_dir = state.get_random_directory().path;
    std::string expected_cwd = original_cwd + "/" + other_dir;
    status = Os::FileSystem::getSingleton().changeWorkingDirectory(other_dir.c_str());
    ASSERT_EQ(status, Os::FileSystem::Status::OP_OK) << "Failed to change working directory";
    status = Os::FileSystem::getSingleton().getWorkingDirectory(cwdBuffer, cwdSize);
    ASSERT_EQ(status, Os::FileSystem::Status::OP_OK) << "Failed to get new working directory";
    ASSERT_EQ(expected_cwd.compare(cwdBuffer), 0) << "getWorkingDirectory did not return the expected directory";
```

Os/Posix/test/ut/PosixFileTests.cpp
```diff
@@ -93,14 +93,14 @@ void setUp(bool requires_io) {
...
-    else if (requires_io && not check_permissions(BASE_PATH, R_OK | W_OK)) {
+    else if (requires_io && not check_permissions(BASE_PATH, R_OK & W_OK)) {
         GTEST_SKIP() << "Cannot read/write in directory: " << BASE_PATH;
     }
     OWNER_PID = ::getpid();
     int signals[] = {SIGQUIT, SIGABRT, SIGTERM, SIGINT, SIGHUP};
     for (unsigned long i = 0; i < FW_NUM_ARRAY_ELEMENTS(signals); i++) {
         // Could not register signal handler
-        if (signal(signals[i], cleanup) == SIG_ERR) {
+        if (signal(SIGQUIT, cleanup) == SIG_ERR) {
```
lestarch-autobot: [Test Quality] **must fix** The loop registers `cleanup` for `SIGQUIT` five times; `SIGABRT`/`SIGTERM`/`SIGINT`/`SIGHUP` get no handler (reverts the https://github.com/nasa/fprime/pull/5706 fix).

An `FW_ASSERT` abort (or Ctrl-C) in the owning process now leaves `/tmp/fprime/fprime-os-file-test` behind, and every later run then hits the `Test file exists` `GTEST_SKIP()` at line 93, so the Posix File suite silently skips instead of running.

```suggestion
        if (signal(signals[i], cleanup) == SIG_ERR) {
```

Os/Posix/test/ut/PosixFileTests.cpp
```diff
@@ -93,14 +93,14 @@ void setUp(bool requires_io) {
         GTEST_SKIP() << "Test file exists: " << non_random_filename->c_str();
     }
     // IO required and cannot read/write to BASE_PATH then skip
-    else if (requires_io && not check_permissions(BASE_PATH, R_OK | W_OK)) {
+    else if (requires_io && not check_permissions(BASE_PATH, R_OK & W_OK)) {
```
lestarch-autobot: [Test Quality] **suggestion** `R_OK & W_OK` is `4 & 2 == 0` (`F_OK`), so this skip guard only checks that `BASE_PATH` exists, not that it is readable/writable (reverts the https://github.com/nasa/fprime/pull/5706 fix).

On an unwritable `/tmp/fprime` the IO tests fail with open errors instead of skipping with the intended message.

```suggestion
    else if (requires_io && not check_permissions(BASE_PATH, R_OK | W_OK)) {
```

Os/Generic/docs/sdd.md
```diff
@@ -28,7 +28,7 @@ Requirement | Description | Verification Method
 ----------- | ----------- | -------------------
 PQ-001 | The PriorityQueue shall implement the Os::QueueInterface for compatibility with F´ components | Inspection, Unit Test
 PQ-002 | The PriorityQueue shall support message prioritization with higher priority messages dequeued before lower priority messages | Unit Test
-PQ-003 | The PriorityQueue shall allocate memory during create and deallocate it during destruction using the registered `Fw::MemAllocator` | Inspection
+PQ-003 | The PriorityQueue shall allocate memory dynamically using new/delete during create and destruction | Inspection
```
lestarch-autobot: [Documentation] **must fix** PQ-003 reverted to a statement the code contradicts.

This line undoes the #5713 correction: `PriorityMemQueue.cpp` still allocates via `allocator.checkedAllocate(...)` / `allocator.deallocate(...)` on the registered `Fw::MemAllocator` (see `allocateArrays`, `teardownInternal`), not `new`/`delete`. Looks like a merge artifact; restore the devel text.

```suggestion
PQ-003 | The PriorityQueue shall allocate memory during create and deallocate it during destruction using the registered `Fw::MemAllocator` | Inspection
```

default/config/OsDelegateFile.hpp
```diff
@@ -0,0 +1,48 @@
...
+// ======================================================================
+#ifndef CONFIG_OS_DELEGATEFILE_HPP
+#define CONFIG_OS_DELEGATEFILE_HPP
+
+//!< Forward declaration of the link-time delegate
+//!< Default: select the File implementation at link time
+namespace Os {
+
+class DelegateFile;
+using File = DelegateFile;
```
lestarch-autobot: [Documentation] **must fix** `Os/docs/sdd.md` §5.2.2 still lists RawTime as the only compile-time-selectable service.

This PR adds `OsDelegate{File,FileSystem,Directory,ConditionVariable,Console,CountingSemaphore,Cpu,Memory}.hpp`, but the SDD's **Currently Supported Services** bullet list (and §5.2.1 "Most OSAL services use link-time selection") still describe RawTime as the sole alias. Please extend the list to the new services and note the naming convention (`config/OsDelegate<Service>.hpp` + `OS_<SERVICE>_HEADER`) so the SDD matches the shipped configuration set. (Anchored here; `Os/docs/sdd.md` is not in the diff.)

Os/FileInterface.hpp
```diff
@@ -0,0 +1,519 @@
+// ======================================================================
+// \title Os/FileInterface.hpp
```
lestarch-autobot: [Documentation] **could fix** `Os/docs/sdd.md` §5.1 table links point at the old `File.hpp` layout.

The Interface row links `Os::FileInterface` to `../File.hpp#L27` and the Wrapper row links `Os::File` to `../File.hpp#L225`; after this PR the interface lives in `Os/FileInterface.hpp` and the wrapper is `Os::DelegateFile` in `Os/DelegateFile.hpp` (`Os::File` is now the configured alias). Update the two links/names (and the §5.1 diagram's `File` box, which is now `DelegateFile`).

Os/File.hpp
```diff
@@ -1,602 +1,41 @@
...
+// and CFG must never include either of them (that would form a cycle).
 // ======================================================================
 #ifndef Os_File_hpp_
 #define Os_File_hpp_
 
-#include <Fw/FPrimeBasicTypes.hpp>
-#include <Fw/Types/ConstStringBase.hpp>
-#include <Os/Os.hpp>
-#include <Utils/Hash/Hash.hpp>
+#include "Os/FileInterface.hpp"
```
lestarch-autobot: [Documentation] **suggestion** `docs/how-to/integrate/implement-osal.md` still tells implementers to read the interface from `Os/File.hpp`.

The §2.1 TIP ("Look at each interface header in `Os/` (e.g., `Os/Mutex.hpp`, `Os/File.hpp`, `Os/Task.hpp`)") and the closing line of the interface table ("Refer to each interface header in `Os/`") predate this split: `Os/File.hpp` is now a thin alias header and the pure-virtual contract is in `Os/FileInterface.hpp` (same for `*Interface.hpp` of the other services touched here). Point both sentences at the `Os/<Service>Interface.hpp` headers.

Os/FileSystemInterface.cpp
```diff
@@ -0,0 +1,272 @@
...
+}  // end copyFileData
+
+}  // namespace
+
+// ------------------------------------------------------------
+// Convenience static wrappers built directly on the virtual instance API
+// ------------------------------------------------------------
+
+FileSystemInterface::Status FileSystemInterface::removeDirectory(const char* path) {
+    return Os::FileSystem::getSingleton()._removeDirectory(path);
```
lestarch-autobot: [Documentation] **must fix** Compile-time selection now has undocumented requirements beyond "inherit from the interface".

The static convenience wrappers resolve through `Os::FileSystem::getSingleton()` (same for `Cpu`/`Memory`; `Console` additionally needs static `write()`), and `Os/Os.cpp` calls `Os::<Service>::init()`. A project that aliases `Os::FileSystem` to its own class per `implement-osal.md` Step 3 / `Os/docs/sdd.md` §5.2.2 IMPORTANT ("MUST inherit from the corresponding interface class") will fail to link. Document in both places that singleton services must also provide `static void init()` and `static <Impl>& getSingleton()` (see `Os/Stub/FileSystem.hpp` for the minimal shape).

Os/ValidateFileCommon.cpp
```diff
@@ -135,9 +135,7 @@ ValidateFile::Status translateStatus(File::Status status, StatusFileType type) {
                 case File::OTHER_ERROR:
                     return ValidateFile::OTHER_ERROR;
                 default:
-                    // Unlisted statuses (e.g. NOT_SUPPORTED, INVALID_ARGUMENT) can
-                    // legitimately come from the OS layer; report rather than assert
-                    return ValidateFile::OTHER_ERROR;
+                    FW_ASSERT(false, status);
```
lestarch-autobot: [Design] **must fix** `design-behavioral-regression`: this hunk is not part of the OSAL alias work. It is a leftover of the stale `Revert "Merge branch 'devel' into multi-os-alias"` (af3538a1c1) and reinstates `FW_ASSERT(false, status)` that #5710 replaced with `return ValidateFile::OTHER_ERROR` (here and at line 158), so a `NOT_SUPPORTED`/`INVALID_ARGUMENT` from an OS layer once again halts flight software. The same revert also undoes #5705/#5713 in `Os/Posix/error.cpp`, `Os/Stub/ConditionVariable.cpp`, `Os/Generic/test/ut/QueueRulesDefinitions.hpp`, `Os/Generic/test/ut/PriorityMemQueueTests.cpp`, `Os/Generic/docs/sdd.md` (PQ-003) and the test regressions Test Quality flagged. Please restore these files to `devel` so the diff contains only the alias/delegate changes.

Os/CpuInterface.cpp
```diff
@@ -0,0 +1,17 @@
+// ======================================================================
+// \title Os/CpuInterface.cpp
+// \brief common convenience-wrapper implementation for Os::Cpu
+// ======================================================================
+#include <Os/Cpu.hpp>
+
+namespace Os {
+
+CpuInterface::Status CpuInterface::getCount(FwSizeType& cpu_count) {
+    return Os::Cpu::getSingleton()._getCount(cpu_count);
```
lestarch-autobot: [Design] **must fix** `design-code-mismatch`: the static convenience API now depends on `Os::Cpu::getSingleton()`/`init()`, but that static surface exists only on `DelegateCpu` and `StubCpu`; it is not part of `CpuInterface`, and `LinuxCpu`/`DarwinCpu` do not provide it. The compile-time selection documented in `config/OsDelegateCpu.hpp` therefore fails to compile for any in-tree concrete implementation. Same pattern in `MemoryInterface.cpp`, `FileSystemInterface.cpp`, and `DelegateConsole` (where `getSingleton()` also carries `Fw::Logger::registerLogger`, so a compile-time `Os::Console` would silently lose logger output). Suggest declaring `static Os::Cpu& getSingleton(); static void init();` on the interface and defining them in `CpuInterface.cpp` against the configured alias (holding `static Os::Cpu s_singleton;`), so both selection modes share one contract. @LeStarch @thomas-bc

Os/ConsoleInterface.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include <Fw/Logger/Logger.hpp>
+#include <Os/Os.hpp>
+#include "config/OsDelegateConsole.hpp"
+
+namespace Os {
+//! \brief Base class for storing implementation specific handle information
+struct ConsoleHandle {};
+
+// \brief Interface defining the properties of the console
+class ConsoleInterface : public Fw::Logger {
```
lestarch-autobot: [Design] **suggestion** `design-scope-creep`: the survey in #5249 recommended `Os::Console` land in its own PR because of the `Fw::Logger` coupling. Moving the `Fw::Logger` base onto `ConsoleInterface` makes every platform console (Posix, out-of-tree) an `Fw::Logger` and changes their object layout, which is a broader contract change than the other seven classes. If bundling is intentional, please state so (and the layout impact) in the PR description; otherwise split Console out. @LeStarch @thomas-bc

default/config/PlatformCfg.fpp
```diff
@@ -18,7 +18,7 @@ constant FW_MUTEX_HANDLE_MAX_SIZE = 72
 constant FW_QUEUE_HANDLE_MAX_SIZE = 368
 
 @ Maximum size of a handle for Os::Directory
-constant FW_DIRECTORY_HANDLE_MAX_SIZE = 16
+constant FW_DIRECTORY_HANDLE_MAX_SIZE = 32
```
lestarch-autobot: [Design] **could fix** `design-margin-erosion`: `FW_DIRECTORY_HANDLE_MAX_SIZE` doubles (16 -> 32) for every platform and every `Os::Directory` instance, but the PR description does not say which handle grew or why. Please record the driving implementation and size in the description or the constant's doc comment so projects can size their overrides.

Os/ValidateFileCommon.cpp
lestarch-autobot: [Security] **must fix** Valid file I/O errors now abort validation instead of returning an error.

`NOT_SUPPORTED` and `INVALID_ARGUMENT` are valid `File::Status` values produced by the POSIX errno mapper, but neither switch handles them. A validation of an affected file or filesystem can now trigger `FW_ASSERT(false, status)` and terminate the flight process; the same regression exists in the `HashFileType` default below. Return `OTHER_ERROR` for unlisted statuses as before. Finding class: `general-vulnerability/unexpected-file-status-assert`.

```suggestion
                    return ValidateFile::OTHER_ERROR;
```

Os/DelegateConditionVariable.cpp
```diff
@@ -0,0 +1,50 @@
...
+    if (this->m_lock != nullptr && this->m_lock != &mutex) {
+        return Status::ERROR_DIFFERENT_MUTEX;
+    };
+    this->m_lock = &mutex;
+    return this->m_delegate.pend(mutex);
+}
+
+void DelegateConditionVariable::wait(Os::Mutex& mutex) {
+    FW_ASSERT(&this->m_delegate == reinterpret_cast<ConditionVariableInterface*>(&this->m_handle_storage[0]));
+    this->m_delegate.wait(mutex);
```
lestarch-autobot: [Security] **must fix** Keep the mutex-consistency check on the public wait path.

Calling `m_delegate.wait(mutex)` dispatches to the concrete backend and bypasses this wrapper's `pend()`, which records the first mutex and rejects later different ones. On POSIX, concurrent `pthread_cond_wait` calls with different mutexes on the same condition are undefined behavior, so the previously guarded misuse can now hang or corrupt synchronization state. Invoke the base helper on *this* object to dispatch through the checked `pend()`. Finding class: `general-vulnerability/undefined-synchronization`.

```suggestion
    ConditionVariableInterface::wait(mutex);
```

Os/ValidateFileCommon.cpp
lestarch-autobot: [C++ Design] **must fix** CPP-4 (`FW_ASSERT` is for programming errors only): asserting on an unlisted `File::Status` returned by the OS layer.

`status` comes from `Os::File` (e.g. `NOT_SUPPORTED`, `INVALID_ARGUMENT`, `OUTSIDE_SANDBOX` on a compile-time-selected implementation), so this is an environmental error, not a programming error. This re-introduces the assert removed in https://github.com/nasa/fprime/pull/5710 — commit af3538a1 (`Revert "Merge branch 'devel'..."`) undid that fix and d739adc5 did not restore this file. Same at line 158.

```suggestion
                    return ValidateFile::OTHER_ERROR;
```

Os/FileInterface.hpp
```diff
@@ -0,0 +1,519 @@
...
+    //! \brief copy constructor that copies the mode state tracked by this base class
+    FileInterface(const FileInterface& other);
+
+    //! \brief assignment operator that copies the mode state tracked by this base class
+    FileInterface& operator=(const FileInterface& other);
+
+    //! \brief determine if the file is open
+    //! \return true if file is open, false otherwise
+    //!
+    bool isOpen() const;
```
lestarch-autobot: [C++ Design] **must fix** CPP-26 (F Prime style): non-virtual base state (`m_mode`) that only `DelegateFile` maintains; concrete implementations hide `isOpen()` instead.

`PosixFile::open`/`close` never call `setMode()`, and `StubFile` declares its own `isOpen() const` (Os/Stub/File.hpp:61) hiding this one. When a platform aliases `Os::File` directly to an implementation (the compile-time path `config/OsDelegateFile.hpp` advertises), `FileInterface::isOpen()`, `readline()` (`NOT_OPENED`) and the mode checks in the `open(...)` convenience overloads silently stop working, and `calculateCrc` returns `NOT_SUPPORTED`. Either make `isOpen()`/mode bookkeeping virtual with the implementations responsible for `setMode()`, or document that the compile-time alias path is delegate-only.

Os/DirectoryInterface.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+    //! \brief copy constructor is forbidden
+    DirectoryInterface(const DirectoryInterface& other) = delete;
+
+    //! \brief assignment operator is forbidden
+    DirectoryInterface& operator=(const DirectoryInterface& other) = delete;
+
+    //! \brief Check if Directory is open or not
+    //! \return true if Directory is open, false otherwise
+    bool isOpen() const;
```
lestarch-autobot: [C++ Design] **must fix** CPP-26 (F Prime style): `m_is_open` is only set by `DelegateDirectory`; `PosixDirectory`/`StubDirectory` declare a non-const `bool isOpen();` that hides this member.

`PosixDirectory::open` never calls `setOpen(true)` (and `PosixDirectory::isOpen()` is declared in Os/Posix/Directory.hpp:54 with no definition), so with `Os::Directory` aliased to an implementation `read(Fw::StringBase&)`, `readDirectory()` and `getFileCount()` always return `NOT_OPENED`. Same class of issue as `FileInterface::isOpen()`.

Os/DirectoryInterface.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+void DirectoryInterface::setOpen(bool is_open) {
+    this->m_is_open = is_open;
+}
+
+DirectoryInterface::Status DirectoryInterface::read(Fw::StringBase& filename) {
+    if (not this->isOpen()) {
+        return Status::NOT_OPENED;
+    }
+    return this->read(const_cast<char*>(filename.toChar()), filename.getCapacity());
```
lestarch-autobot: [C++ Design] **future work** CPP-10 (avoid `const_cast`): writes through `const_cast<char*>(filename.toChar())`.

Moved verbatim from `Os/Directory.cpp`; not introduced here. A mutable-buffer accessor on `Fw::StringBase` (or reading into a local `char[FW_FIXED_LENGTH_STRING_SIZE]` and assigning) would remove the cast.

Os/FileInterface.cpp
```diff
@@ -0,0 +1,182 @@
...
+                                          FwSizeType length,
+                                          FileInterface::Mode requested_mode,
+                                          FileInterface::OverwriteType overwrite) {
+    FW_ASSERT(nullptr != filepath);
+    const FwSizeType string_len = static_cast<FwSizeType>(Fw::StringUtils::string_length(filepath, length));
+    FW_ASSERT(string_len < length, static_cast<FwAssertArgType>(string_len), static_cast<FwAssertArgType>(length));
+    FW_ASSERT(FileInterface::Mode::OPEN_NO_MODE < requested_mode &&
+              FileInterface::Mode::MAX_OPEN_MODE > requested_mode);
+    FW_ASSERT((0 <= overwrite) && (overwrite < OverwriteType::MAX_OVERWRITE_TYPE));
+    return this->open(filepath, requested_mode, overwrite);
```
lestarch-autobot: [Correctness] **suggestion** The 3-argument `open(path, mode, overwrite)` no longer bounds-checks the path string.

Before this PR `File::open(path, mode, overwrite)` forwarded here with `length = FileNameStringSize + 1`, so `string_length(filepath, length) < length` was asserted on every open. Now that signature *is* the pure-virtual delegate hook and `DelegateFile::open` forwards to the OS implementation with only a null check.
Callers using that form (`Svc/ActiveTextLogger/LogFile.cpp:111`, `Svc/FileWorker/FileWorker.cpp:444`, `Svc/Ccsds/CfdpManager/TransactionRx.cpp:368`, `Svc/PrmDb/PrmDbImpl.cpp:141`, `Svc/BufferLogger/BufferLoggerFile.cpp:122`) silently lose the over-long/unterminated path guard.
Consider keeping a checked non-virtual 3-arg `open` on `FileInterface` and giving the implementation hook a distinct name (as `Directory`/`FileSystem` do with `_`-prefixed hooks). Anchored on the checked overload; the offending declaration is the pure-virtual `open` in `FileInterface.hpp`.

Os/Stub/File.cpp
```diff
@@ -13,6 +13,10 @@ StubFile::Status StubFile::open(const char* filepath, StubFile::Mode open_mode,
     return status;
 }
 
+bool StubFile::isOpen() const {
+    return false;
+}
```
lestarch-autobot: [Correctness] **could fix** `StubFile::isOpen()` hides the non-virtual `FileInterface::isOpen()` and disagrees with the mode state the base class now tracks.

`FileInterface::isOpen()` is non-virtual and derives from `m_mode`, which this PR moved into the interface. This shadow always returns `false`, so a `StubFile` reached through a `FileInterface&` (e.g. `FileInterface::readline`, or the `DelegateFile` copy/assign path) reports the base state while direct `StubFile` callers get the shadow — two answers for one object. Since `StubFile::open` never returns `OP_OK`, `m_mode` stays `OPEN_NO_MODE` and the base `isOpen()` already yields `false`; the override is redundant and can be removed.

```suggestion
```

Os/ValidateFileCommon.cpp
lestarch-autobot: [Correctness] **Concur** — also in scope for functional correctness: `default:` at lines 138 and 158 now asserts on `File::NOT_SUPPORTED`/`INVALID_ARGUMENT`, which the Os layer legitimately returns (e.g. Stub `size`/`read`), halting flight software instead of reporting `OTHER_ERROR`.

Os/FileInterface.hpp
```diff
@@ -0,0 +1,519 @@
...
+    //! ```
+    //!
+    //! \note The base-class implementation returns `NOT_SUPPORTED`; the CRC
+    //! scratch state and algorithm live on `Os::DelegateFile` (see the design
+    //! note above), which overrides this method.
+    //!
+    //! \param crc: U32 bit value to fill with CRC
+    //! \return OP_OK on success otherwise error status
+    //!
+    virtual Status calculateCrc(U32& crc);
```
lestarch-autobot: [Operational] **suggestion** `ops-config-extreme`: under the documented compile-time alias (`OsDelegateFile.hpp` option 2), `Os::File` is the concrete class, so unless the platform overrides all three CRC methods these fall through to the base and return `NOT_SUPPORTED` — a legal configuration with no compile-time signal.

Framework consequence: `PrmDb::readParamFile` (`PrmDbImpl.cpp:446`) logs `PrmFileReadError(CRC_BUFFER)` and returns `ERROR` on every boot (no parameters load), and `FileManager::CALCULATE_CRC` always returns `EXECUTION_ERROR`. Smallest remedy: state in `OsDelegateFile.hpp` option 2 that a compile-time `Os::File` must override `calculateCrc`/`incrementalCrc`/`finalizeCrc` (or PrmDb/FileManager are non-functional). Judgment call — no in-repo platform selects this yet; @LeStarch @thomas-bc for confirmation.

Os/test/ut/filesystem/FileSystemRules.cpp
lestarch-autobot: [Correctness] **Concur** — also in scope for functional correctness: `compare()` is non-zero whenever the strings differ, and the relative `other_dir` always differs from the absolute cwd, so the assertion cannot fail and `getWorkingDirectory` is unverified.

Os/Posix/test/ut/PosixFileTests.cpp
lestarch-autobot: [Correctness] **Concur** — also in scope for functional correctness: loop variable `signals[i]` is never used, so `cleanup` is bound only to `SIGQUIT` and test files leak on `SIGABRT`/`SIGTERM`/`SIGINT`/`SIGHUP`.

Os/Posix/test/ut/PosixFileTests.cpp
lestarch-autobot: [Correctness] **Concur** — also in scope for functional correctness: `R_OK & W_OK` evaluates to `0` (`F_OK`), so the guard tests existence only and IO tests run against an unwritable directory.

Os/ValidateFileCommon.cpp
lestarch-autobot: [Operational] **Concur** — also in scope for operational consequences: reachable in flight via `ComLogger::closeFile` -> `createValidation` and `BufferLogger` hash files; a `NOT_SUPPORTED`/`INVALID_ARGUMENT` from `Posix/error.cpp` (ENOSYS/EINVAL) on one log rotation asserts the whole deployment instead of one `FileValidationError` event; severity from my scope: **must fix**.

Os/Posix/error.cpp
```diff
@@ -54,7 +54,7 @@ FileSystem::Status errno_to_filesystem_status(int errno_input) {
         case 0:
             status = FileSystem::Status::OP_OK;
             break;
-        // Adjacent cases below intentionally share a status via fallthrough
+        // All fall through are intended to fallback on OTHER_ERROR
```
lestarch-autobot: [Maintainability] **must fix** — Comment says the fallthroughs land on `OTHER_ERROR`, but the cases directly below map to `NO_PERMISSION`, `ALREADY_EXISTS`, `DOESNT_EXIST`, etc. The next engineer adding an errno will trust the comment and misplace the case. This line reverts a devel fix (#5705) — it came in via af3538a "Revert 'Merge branch devel'" and was not restored by d739adc.

```suggestion
        // Adjacent cases below intentionally share a status via fallthrough
```

Os/Posix/File.cpp
```diff
@@ -134,7 +134,7 @@ PosixFile::Status PosixFile::open(const char* filepath,
 }
 
 void PosixFile::close() {
-    // Only close file handles that are open
+    // Only close file handles that are not open
```
lestarch-autobot: [Maintainability] **must fix** — Comment states the opposite of the guard (`!= INVALID_FILE_DESCRIPTOR` closes handles that *are* open). Inverted comments on resource-release paths are how double-close/leak edits slip in. This line reverts a devel fix (#5705) — it came in via af3538a "Revert 'Merge branch devel'" and was not restored by d739adc.

```suggestion
    // Only close file handles that are open
```

Os/Stub/ConditionVariable.cpp
```diff
@@ -15,11 +15,11 @@ StubConditionVariable::Status StubConditionVariable::pend(Os::Mutex& mutex) {
     return StubConditionVariable::Status::ERROR_NOT_IMPLEMENTED;
 }
 void StubConditionVariable::notify() {
-    // Nobody can be waiting, because pend() always fails with ERROR_NOT_IMPLEMENTED.
+    // Nobody is waiting, because we assert if anyone tries to wait.
```
lestarch-autobot: [Maintainability] **must fix** — `pend()` five lines above returns `ERROR_NOT_IMPLEMENTED`; it does not assert. Same stale text at line 22 in `notifyAll()`. A reader relying on the asserted-never-waits invariant would wrongly conclude a caller cannot proceed past `pend()`. This line reverts a devel fix (#5705) — it came in via af3538a "Revert 'Merge branch devel'" and was not restored by d739adc.

```suggestion
    // Nobody is waiting, because pend() returns ERROR_NOT_IMPLEMENTED to anyone who tries.
```

Os/Generic/test/ut/PriorityMemQueueTests.cpp
```diff
@@ -2023,7 +2023,7 @@ TEST_F(PriorityMemQueueTestFixture, DisableDuringBlockingReceive) {
     // Disable priority 2 while receiver is blocked
     queue.m_handle.disablePriority(2);
 
-    // Send to disabled priority 2 - succeeds (disable affects receive only); message stays queued
+    // Send to disabled priority 2 - should be rejected or fall back to priority 0
```
lestarch-autobot: [Maintainability] **must fix** — `PriorityMemQueue::send()` has no enabled-priority check, so the send is neither rejected nor redirected; it succeeds and the message stays queued (which is why the result is discarded). The comment documents behavior the queue does not have, which will mislead whoever next touches disable semantics. This line reverts a devel fix (#5705) — it came in via af3538a "Revert 'Merge branch devel'" and was not restored by d739adc.

```suggestion
    // Send to disabled priority 2 - succeeds (disable affects receive only); message stays queued
```

Os/Generic/test/ut/PriorityMemQueueInputValidationTests.cpp
```diff
@@ -14,10 +14,13 @@
 #include "Fw/Types/String.hpp"
 #include "Os/Generic/PriorityMemQueue.hpp"
 
-// Test helper wrapping the public PriorityMemQueue::resetConfig() API
+// Test helper to access private members
```
lestarch-autobot: [Maintainability] **could fix** — The helper only calls the public `resetConfig()` (the body comment even says so); it accesses no private members. Same stale wording at `PriorityMemQueueTests.cpp:49`. Readers will hunt for a `friend` declaration that does not exist. This line reverts a devel fix (#5705) — it came in via af3538a "Revert 'Merge branch devel'" and was not restored by d739adc.

```suggestion
// Test helper wrapping the public PriorityMemQueue::resetConfig() API
```

Os/Generic/test/ut/QueueRulesDefinitions.hpp
```diff
@@ -2,13 +2,13 @@
 // \title Os/Generic/test/ut/QueueRulesDefinitions.hpp
 // \brief definitions for queue testing
 // ======================================================================
-#ifndef OS_GENERIC_TEST_UT_QUEUE_RULES_DEFINITIONS
-#define OS_GENERIC_TEST_UT_QUEUE_RULES_DEFINITIONS
+#ifndef OS_STUB_TEST_UT_QUEUE_RULES_DEFINITIONS
+#define OS_STUB_TEST_UT_QUEUE_RULES_DEFINITIONS
```
lestarch-autobot: [Maintainability] **suggestion** — Header guard says `OS_STUB_...` for a file in `Os/Generic/test/ut/`. A guard that names another directory invites a real collision if `Os/Stub/test/ut/` ever gains a `QueueRulesDefinitions.hpp` with the same guard (silent empty include). Also update the `#endif` comment on line 14. This line reverts a devel fix (#5705) — it came in via af3538a "Revert 'Merge branch devel'" and was not restored by d739adc.

```suggestion
#ifndef OS_GENERIC_TEST_UT_QUEUE_RULES_DEFINITIONS
#define OS_GENERIC_TEST_UT_QUEUE_RULES_DEFINITIONS
```

Os/DelegateConsole.cpp
```diff
@@ -0,0 +1,74 @@
...
+        Fw::Logger::registerLogger(&s_singleton);
+    }
+    return s_singleton;
+}
+
+// ------------------------------------------------------------------
+// Common virtual functions built on top of OS-specific functions
+// ------------------------------------------------------------------
+
+void ConsoleInterface::writeMessage(const Fw::ConstStringBase& message) {
```
lestarch-autobot: [Maintainability] **could fix** — The interface's convenience method is defined in the *delegate's* translation unit (same for `ConditionVariableInterface::wait` in `DelegateConditionVariable.cpp:45`), while `FileInterface`, `DirectoryInterface`, `FileSystemInterface`, `CpuInterface`, `MemoryInterface` each get their own `*Interface.cpp`. A project that selects a compile-time alias and drops `DelegateConsole.cpp` from the build will get an unexpected undefined-symbol link error for a method that lives on the interface. Suggest moving these into `ConsoleInterface.cpp` / `ConditionVariableInterface.cpp` (and listing them in `Os/CMakeLists.txt`) to match the other five.

Os/DelegateFile.cpp
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// ----------------------------------------------------------------------
+// CRC implementation
+//
+// Relocated from FileInterface so the scratch buffer is charged to this
+// wrapper's (unconstrained) storage rather than the fixed-size delegate
+// handle. The algorithm is unchanged; it drives I/O through the virtual
+// `this->read()`, which dispatches to the selected delegate.
+// ----------------------------------------------------------------------
```
lestarch-autobot: [Maintainability] **suggestion** — This is the third copy of the CRC-placement rationale (also `FileInterface.hpp:374-392` and `DelegateFile.hpp:169-181`), and this copy narrates the diff ("Relocated from", "algorithm is unchanged"), which becomes false the first time the algorithm changes. Three prose blocks drift independently; keep the one in `FileInterface.hpp` and point to it.

```suggestion
// ----------------------------------------------------------------------
// CRC implementation -- state lives here, not on FileInterface; see the
// design note in Os/FileInterface.hpp.
// ----------------------------------------------------------------------
```

Os/Generic/docs/sdd.md
lestarch-autobot: [Operational] **Concur** — also in scope for operational consequences: an operator sizing a deployment's `Fw::MemAllocator` pool from PQ-003 would omit the queue's allocation and hit a `checkedAllocate` failure at init.

Os/ValidateFileCommon.cpp
lestarch-autobot: [Summary] **Duplicate** — consolidated into https://github.com/nasa/fprime/pull/5955#discussion_r4085879026.

lestarch-autobot: [Summary] **Duplicate** — consolidated into https://github.com/nasa/fprime/pull/5955#discussion_r4085879026.

lestarch-autobot (changes requested):

### Per-agent results

| F Prime C/C++ Design | 3 | 0 | 0 | 1 | 4 | No-Go |
| Documentation Currency | 3 | 1 | 1 | 0 | 5 | No-Go |
| Design | 2 | 1 | 1 | 0 | 4 | No-Go |
| Test Quality | 2 | 1 | 0 | 0 | 3 | No-Go |
| Correctness | 3 | 2 | 1 | 0 | 6 | No-Go |
| Operational | 1 | 1 | 1 | 0 | 3 | No-Go |
| Maintainability | 4 | 2 | 2 | 0 | 8 | No-Go |
| **Totals** | 16 | 8 | 6 | 1 | 31 | **No-Go** |

Totals count each consolidated finding once (2 duplicate threads closed by the post-pass; concurrence replies remain counted in their agents' rows).

Duplicates consolidated this run: 2

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

<details>
<summary>Outstanding must-fix items (14)</summary>

**Design**
- **Human design adjudication required.** `Os/ValidateFileCommon.cpp` `default:` reinstates `FW_ASSERT(false, status)` (#5710 fix reverted by stale devel-merge revert af3538a); OS-layer `NOT_SUPPORTED`/`INVALID_ARGUMENT` halts flight software — also: Security, C++ Design, Correctness, Operational — https://github.com/nasa/fprime/pull/5955#discussion_r4085879026
- **Human design adjudication required.** Static convenience API depends on `getSingleton()`/`init()` that only `Delegate*`/`Stub*` provide; compile-time selection of `LinuxCpu`/`DarwinCpu` (and Memory/FileSystem/Console) does not compile — https://github.com/nasa/fprime/pull/5955#discussion_r4085879046

**Security Vulnerabilities**
- `DelegateConditionVariable::wait` drops the mutex-consistency check on the public wait path — https://github.com/nasa/fprime/pull/5955#discussion_r4085886286

**F Prime C/C++ Design**
- CPP-26: non-virtual `FileInterface::m_mode` maintained only by `DelegateFile`; concrete implementations hide `isOpen()` — https://github.com/nasa/fprime/pull/5955#discussion_r4085890410
- CPP-26: `DirectoryInterface::m_is_open` set only by `DelegateDirectory`; `PosixDirectory`/`StubDirectory` hide it with non-const `isOpen()` — https://github.com/nasa/fprime/pull/5955#discussion_r4085890421

**Documentation Currency**
- `Os/Generic/docs/sdd.md` PQ-003 reverted to a statement the code contradicts — also: Operational — https://github.com/nasa/fprime/pull/5955#discussion_r4085873236
- `Os/docs/sdd.md` §5.2.2 still lists RawTime as the only compile-time-selectable service — https://github.com/nasa/fprime/pull/5955#discussion_r4085873262
- Compile-time selection now requires `init()`/`getSingleton()` beyond "inherit from the interface"; undocumented — https://github.com/nasa/fprime/pull/5955#discussion_r4085873285

**Test Quality**
- `FileSystemRules.cpp` `ASSERT_TRUE(other_dir.compare(cwdBuffer))` is a tautology (re-introduces the bug fixed in #5706) — also: Correctness — https://github.com/nasa/fprime/pull/5955#discussion_r4085859337
- `PosixFileTests.cpp` signal loop registers `cleanup` for `SIGQUIT` five times; other signals get no handler (reverts #5706) — also: Correctness — https://github.com/nasa/fprime/pull/5955#discussion_r4085859386

**Maintainability**
- `Os/Posix/error.cpp` fallthrough comment says `OTHER_ERROR` but cases map elsewhere (reverts #5705) — https://github.com/nasa/fprime/pull/5955#discussion_r4085898618
- `Os/Posix/File.cpp` close-path comment inverted relative to the guard (reverts #5705) — https://github.com/nasa/fprime/pull/5955#discussion_r4085898654
- `Os/Stub/ConditionVariable.cpp` comment claims `pend()` asserts; it returns `ERROR_NOT_IMPLEMENTED` (reverts #5705) — https://github.com/nasa/fprime/pull/5955#discussion_r4085898674
- `PriorityMemQueueTests.cpp` comment documents an enabled-priority check `send()` does not have (reverts #5705) — https://github.com/nasa/fprime/pull/5955#discussion_r4085898698

</details>

### Merge readiness
**Merge readiness: No-Go** — 8 of 10 reviewers report outstanding must-fix findings (14 consolidated threads), most of them devel fixes (#5705/#5706/#5710/#5713) undone by the stale merge revert af3538a still present in the branch.

---

Eight services now share the alias/delegate helm; restore the reverted devel fixes and this ship is ready for its shakedown cruise.

thomas-bc (commented):
If that's possible, I'd vote for merging all the config headers that select these into a single header, so we don't overcrowd the default/config directory

PR #5992, opened 2026-09-17
Fw/DpContainer: bound the data size by what the header can store

### LINKED ISSUE #4957: DpContainer DataSize FwSizeType and FwSizeStoreType
| | |
|:---|:---|
|**_F´ Version_**| 4.2.0 |
|**_Affected Component_**| Fw/DpContainer  |
---
## Problem Description

In `Fw/DpContainer` the data section size is set and used internally as an `FwSizeType` but used as a `FwSizeStoreType` when serialized to disk. This causes an FW_ASSERT when serializing a header with a size larger than FwSizeStoreType can hold.

## Expected Behavior

The DataSize getter/setter fields on DpContainer should take `FwSizeStoreType` so that the user is unable to serialize a value that's too large.

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4957 |
|**_Documentation Included (y/n)_**| n (doxygen comments only; the Fw/Dp SDD already describes DataSize as serialized at `sizeof(FwSizeStoreType)`) |

---
## Change Description

`Fw::DpContainer` tracked the data size as an `FwSizeType`, but the header serializes it as an `FwSizeStoreType` (U16 by default). `setBuffer` gave the data buffer the full remaining capacity of the packet buffer with no regard to what the header can encode, so with a buffer larger than about 64 KB the generated record serializers could grow `m_dataSize` past the bound. `serializeHeader` then asserted on the `FW_SERIALIZE_FORMAT_ERROR` returned by `serializeSize`. The same assert is reachable from `DpWriter::performProcessing` and `DpCompressProc`.

This PR:

- Clamps the data buffer capacity in `setBuffer` to the new `DpContainer::MAX_DATA_SIZE` (the maximum of `FwSizeStoreType`). Record serialization now returns `FW_SERIALIZE_NO_ROOM_LEFT` at the bound, which callers already handle, instead of producing a header that cannot be serialized.
- Exposes `MAX_DATA_SIZE` as a public constant and asserts in `setDataSize` that the value fits, so hand-coded callers fail at the point of misuse rather than later in `serializeHeader`.
- Adds unit tests covering the capacity clamp, the no-room-left status one byte past the bound, a header round trip at exactly the bound, and the setter assert.

## Rationale

Fixes #4957. The fix proposed in the issue (change the accessors to `FwSizeStoreType`) was not adopted: the autocoded record serializers write `m_dataSize` directly, so narrowing the member would silently wrap in U16 arithmetic and produce a corrupt header with a valid hash. Bounding the capacity keeps the failure explicit and does not change the wire format or the dictionary.

- `Fw/Dp` unit tests: 5 pass, including the two new `DataSize.*` tests. The new tests skip themselves if a project configures `FwSizeStoreType` wide enough that the bound cannot be exercised.
- `Svc/DpWriter`, `Svc/DpCatalog`, `Svc/DpManager` and `Svc/DpCompressProc` unit tests pass unchanged, confirming the tighter setter does not break existing callers.
- Review note: the out-of-line `constexpr FwSizeType DpContainer::MAX_DATA_SIZE;` definition in the .cpp is required because the project builds as C++14 and gtest macros ODR-use the constant.

## Future Work

- The fpp generator emits `serializeSize(size)` under an `FW_ASSERT` in `serializeRecord_*` for array records, so an array record with more elements than `FwSizeStoreType` can represent asserts in the same way. That should return a status instead and belongs in an fpp issue.
- Per the CCB comment on #4957, the broader divergence between `FwSizeType` and `FwSizeStoreType` remains. Projects needing data products larger than 64 KB should set `FwSizeStoreType` to U32 in their configuration, which is a dictionary and file-format change.

Claude Code was used to analyze the issue, implement the fix, write the unit tests, and draft this description, under human direction and review. IAMAI


Fw/Dp/DpContainer.hpp
```diff
@@ -63,6 +66,10 @@ class DpContainer {
     //! Reserve space for the header, the header hash, and the data hash
     //! This is also the number of non-data bytes in the packet
     static constexpr FwSizeType MIN_PACKET_SIZE = Header::SIZE + 2 * HASH_DIGEST_LENGTH;
+    //! The maximum data size
+    //! The data size is stored in the header as an FwSizeStoreType, so the
+    //! data payload can never be larger than that type can represent
+    static constexpr FwSizeType MAX_DATA_SIZE = std::numeric_limits<FwSizeStoreType>::max();
```
thomas-bc-autowot, line 72: [Documentation] **could fix** `Fw/Dp/docs/sdd.md` does not document the new payload bound `MAX_DATA_SIZE` or the capacity clamp in `setBuffer`.

The SDD §5.1.1 header table only states that `DataSize` is serialized at `sizeof(FwSizeStoreType)`; it does not say the payload is now bounded by `DpContainer::MAX_DATA_SIZE` (65535 bytes with the default `FwSizeStoreType = U16`), that `setBuffer` clamps the data capacity to it (record serialization returns `FW_SERIALIZE_NO_ROOM_LEFT` at the bound), or that `setDataSize` asserts above it. The SDD is the component contract; a sentence after the header table (or under §2 Configuration, since `FwSizeStoreType` is the knob) would close the gap, e.g.:

```markdown
The data payload size is bounded by `DpContainer::MAX_DATA_SIZE`, the largest value
representable by `FwSizeStoreType` (65535 bytes for the default `U16`). `setBuffer`
clamps the data capacity to this bound, so serializing records past it returns
`FW_SERIALIZE_NO_ROOM_LEFT`; `setDataSize` asserts if given a larger size. Projects
that need larger data products must widen `FwSizeStoreType` in `FpConfig.fpp`.
```

Fw/Dp/DpContainer.cpp
```diff
@@ -129,7 +136,11 @@ void DpContainer::setBuffer(const Buffer& buffer) {
               static_cast<FwAssertArgType>(MIN_PACKET_SIZE));
     // Initialize the data buffer
     U8* const buffAddr = buffer.getData();
-    const FwSizeType dataCapacity = buffer.getSize() - MIN_PACKET_SIZE;
+    // Bound the data capacity by what the header can represent, so that
+    // serializing records reports FW_SERIALIZE_NO_ROOM_LEFT instead of
+    // growing the data size past what serializeHeader can encode
+    const FwSizeType availableCapacity = bufferSize - MIN_PACKET_SIZE;
+    const FwSizeType dataCapacity = (availableCapacity < MAX_DATA_SIZE) ? availableCapacity : MAX_DATA_SIZE;
```
thomas-bc-autowot, line 143: [Documentation] **could fix** `docs/user-manual/framework/data-products.md` §3.3 does not mention the data-size ceiling this clamp introduces.

§3.3 says `dpGet`/`dpRequest` return a buffer "large enough to store a data product packet with the requested data size", and §3.2.2 tells users to sum the `SIZE_OF_*_RECORD` constants to compute that size. With this change a request above `Fw::DpContainer::MAX_DATA_SIZE` (65535 bytes with the default `FwSizeStoreType = U16`) yields a container whose usable capacity is silently clamped, and record serialization then fails with `FW_SERIALIZE_NO_ROOM_LEFT`. A one-line note in §3.3 (after the `dpGet` / `dpRequest` descriptions) would set expectations, e.g.:

```markdown
The requested data size must not exceed `Fw::DpContainer::MAX_DATA_SIZE`, the largest
value representable by `FwSizeStoreType` (65535 bytes for the default `U16`); the
container's data capacity is clamped to this bound.
```

Fw/Dp/DpContainer.hpp
```diff
@@ -171,8 +178,12 @@ class DpContainer {
     }
 
     //! Set the data size
+    //! The data size must not exceed MAX_DATA_SIZE, because the header stores
+    //! it as an FwSizeStoreType
     void setDataSize(FwSizeType dataSize  //!< The data size
     ) {
+        FW_ASSERT(dataSize <= MAX_DATA_SIZE, static_cast<FwAssertArgType>(dataSize),
```
thomas-bc-autowot, line 185: [Documentation] **could fix** `docs/how-to/develop/data-products.md` "Compute Container Size" does not state the upper bound the computed size must respect.

That section walks users through `containerSize = RECORD_COUNT * (SIZE_OF_... )` and passing it to `dpGet_*`, and "Other Considerations" discusses container-size trade-offs, but neither mentions that the total is now bounded by `Fw::DpContainer::MAX_DATA_SIZE` (65535 bytes with the default `FwSizeStoreType = U16`) and that a larger `RECORD_COUNT` will hit `FW_SERIALIZE_NO_ROOM_LEFT` on the records that no longer fit. A sentence at the end of "Compute Container Size" would prevent that surprise, e.g.:

```markdown
The computed size must not exceed `Fw::DpContainer::MAX_DATA_SIZE` (65535 bytes with the
default `FwSizeStoreType = U16`); projects needing larger containers must widen
`FwSizeStoreType` in their `FpConfig.fpp`.
```

Fw/Dp/DpContainer.hpp
```diff
@@ -171,8 +178,12 @@ class DpContainer {
...
+                  static_cast<FwAssertArgType>(MAX_DATA_SIZE));
         this->m_dataSize = dataSize;
     }
```
thomas-bc-autowot, line 188: (Anchored above the offending line; the diff does not include line 190.)

[Documentation] **could fix** The `setBuffer` Doxygen comment (`//! Set the packet buffer`, line 190) does not reflect the new capacity clamp.

`setDataSize` and `MAX_DATA_SIZE` gained doc-comments describing the bound, but `setBuffer` — the public entry point whose behavior actually changed — still reads only "Set the packet buffer". Callers that hand it a buffer larger than `MIN_PACKET_SIZE + MAX_DATA_SIZE` now get a data buffer whose capacity is clamped, which only the `.cpp` implementation comment explains. Suggested header comment:

```cpp
    //! Set the packet buffer
    //! The data capacity is the buffer size minus MIN_PACKET_SIZE, clamped to
    //! MAX_DATA_SIZE; serializing past it returns FW_SERIALIZE_NO_ROOM_LEFT
    void setBuffer(const Buffer& buffer  //!< The packet buffer
    );
```

Fw/Dp/DpContainer.hpp
thomas-bc-autowot, line 72: [Operational] **suggestion** The clamp silently hands a `dpGet`/`dpRequest` producer a container smaller than the data size it asked for; no status, event, or doc surfaces it.

`dpGet(n)` allocates `getPacketSizeForDataSize(n)` bytes and returns `SUCCESS`, but the container it builds now has usable capacity `min(n, MAX_DATA_SIZE)` = 65535 with the default `FwSizeStoreType = U16`. A producer sized for a 128 KiB container gets 64 KiB, strands the other 50% of every allocation until dealloc, and records past the bound return `FW_SERIALIZE_NO_ROOM_LEFT` — a status the how-to example (`docs/how-to/develop/data-products.md`, `serializeRecord_SineRecord(...)` with the return dropped) never checks, so the PR-body claim that callers already handle it does not hold for the documented pattern. Pre-PR this configuration mismatch was a hard `FW_ASSERT` at the first `dpSend`; post-PR it is silent partial data products for the life of the mission with no operator-visible signal. Judgment call: the assert was the worse outcome, but the bound must be visible at the API surface so integrators widen `FwSizeStoreType` instead of discovering truncation in the archive; at minimum document the clamp here (the user-manual `dpGet`/`dpRequest` description should say the same).

```suggestion
    //! The maximum data size
    //! The data size is stored in the header as an FwSizeStoreType, so the
    //! data payload can never be larger than that type can represent.
    //! setBuffer clamps the data capacity to this value: a container obtained
    //! through dpGet/dpRequest with a data size above MAX_DATA_SIZE is not
    //! rejected; it accepts records only up to MAX_DATA_SIZE bytes and
    //! serializeRecord_* returns FW_SERIALIZE_NO_ROOM_LEFT past that point.
    //! Projects that need larger data products must widen FwSizeStoreType.
    static constexpr FwSizeType MAX_DATA_SIZE = std::numeric_limits<FwSizeStoreType>::max();
```

Fw/Dp/DpContainer.cpp
```diff
@@ -12,6 +12,13 @@
 
 namespace Fw {
 
+// ----------------------------------------------------------------------
+// Static constant definitions
+// ----------------------------------------------------------------------
+
+// Out-of-line definition required in C++14 when the constant is ODR-used
+constexpr FwSizeType DpContainer::MAX_DATA_SIZE;
```
thomas-bc-autowot, line 20: [Operational] **could fix** `MAX_DATA_SIZE` assumes `FwSizeStoreType` is no wider than `FwSizeType`; the configuration where it is wider is not rejected.

With `FwSizeStoreType = U64` on a platform whose `PlatformSizeType` is `U32` the initializer narrows `18446744073709551615` to `4294967295`; g++ `-std=c++14` emits `-Woverflow` on that line (reproduced), which is a build break under `fprime_add_validation_flags` (`-Werror`). Runtime effect in that configuration is benign (the bound saturates to the `FwSizeType` maximum, so no clamp), and `serializeSize` already tolerates the wider store type, so this is a config-extreme robustness item rather than a flight hazard. A `static_assert` makes the assumption explicit and fails at the right place instead of as a warning-turned-error.

```suggestion
constexpr FwSizeType DpContainer::MAX_DATA_SIZE;
static_assert(sizeof(FwSizeStoreType) <= sizeof(FwSizeType),
              "FwSizeStoreType must not be wider than FwSizeType");
```

thomas-bc-autowot (approved):

### Per-agent results

| Documentation Currency | 0 | 0 | 4 | 0 | 4 | Go |
| Operational | 0 | 1 | 1 | 0 | 2 | Go |
| Maintainability | 0 | 0 | 0 | 0 | 0 | Go |
| **Totals** | 0 | 1 | 5 | 0 | 6 | **Go** |

Documentation Currency's metadata reports `outstanding: 0`; its 4 `could fix` threads are all open, so the table shows 4.

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

### Merge readiness
**Merge readiness: Go** — all 10 reviewers completed with zero outstanding must-fix findings; 6 non-blocking threads (1 suggestion, 5 could fix) remain for maintainer discretion.

---

Header and payload finally agree on how much fits in the bottle — clear for the next stage burn, pending human eyes.

LeStarch (commented):
I don't understand this PR.  The size of the buffer is `FwSizeType`, and the size of the header field is also `FwSizeType`.  So why do we need all this extra code to check for something that is guaranteed by structure?

If we need a change, we should do a static_cast to assert the above assumption instead.

LeStarch:
I see the issue, the header is limited to `FwSizeStoreType`.  The current recommendation, however; is that we fix FwSizeType == FwSizeStoreType and as such we'd collapse back into the situation that we originally had.

bitWarrior:
@LeStarch Thanks, your second comment has it right. To make the reachable failure concrete:

- `setBuffer` gives the data buffer the full remaining capacity, and the autocoded `serializeRecord_*` functions grow `m_dataSize` (an `FwSizeType`) until that capacity is full.
- With any container buffer over ~64 KB and the default `FwSizeStoreType = U16`, `m_dataSize` passes 65535 and `serializeHeader` then asserts on the `FW_SERIALIZE_FORMAT_ERROR` from `serializeSize`. That is the assert in #4957, and it fires at `dpSend`, in `DpWriter::performProcessing`, and in `DpCompressProc`.

So the bug only exists while the two types differ. If the plan is to make `FwSizeType == FwSizeStoreType`, then this PR collapses to a no-op (the clamp never engages, the setter assert is always true, and the new tests skip themselves) and I'd rather close it in favor of that change. Is the unification scheduled for a release, and is there an issue I can point #4957 at?

If the types will stay separate for a while, I can narrow this to what you suggested: drop the capacity clamp and instead assert at the point of misuse (the setter, plus `setBuffer` rejecting a buffer whose data capacity cannot be represented). The operational-consequences bot flagged the same concern you did, that a silent clamp is worse than a visible failure for a misconfigured deployment, so I'm fine going that way. The remaining hole would be the autocoded array-record path, which asserts on `serializeSize(size)` instead of returning a status; I filed that separately as nasa/fpp#1113.

Happy to take either direction, just let me know which you'd prefer.

bitWarrior:
@LeStarch Gentle nudge on the question above. I don't see an issue or PR yet for making `FwSizeType == FwSizeStoreType`, so I'd like to know which way to take this:

1. If the unification is planned, I'll close this PR, open the tracking issue for it if that helps, and point #4957 at it.
2. If the types will stay separate for now, I'll narrow this PR to the assert-at-misuse form you suggested (no capacity clamp).

Either is fine with me; CI is green on the current head in the meantime.

LeStarch:
I am going to close this issue, as we have now fixed FwSizeType == FwSizeStoreType in configuration with a big warning for those who choose to diverge.

Also, we will make this explicit in the next breaking-change release.

PR #6001, opened 2026-09-20
WasmSequencer: Fix seqDoneOut reply

Kronos3:
| | |
|:---|:---|
|**_Related Issue(s)_**| Closes #5906  |
|**_Has Unit Tests (y/n)_**| y |
|**_Documentation Included (y/n)_**| y |
|**_Generative AI was used in this contribution (y/n)_**| y |

---
## Change Description

The `seqDoneOut` port was not being invoked in _all_ cases where a sequence was dispatched from a `seqRunIn` port. This PR fixes that. It also fixes some inconsistencies in the `fprime.h` header and adds some static assertions to validate that `fprime.h` is in sync with the framework.

## Rationale

Bug

## Future Work

N/A

## AI Usage (see [policy](https://github.com/nasa/fprime/blob/devel/AI_POLICY.md))

Claude Code was used to add tests and validate the fixes.

PR #6006, opened 2026-09-21
User config to dispatch command if sequence table is full

### LINKED ISSUE #5812: Modify command dispatchers to execute command anyway even if sequence table fills up
| | |
|:---|:---|
|**_F´ Version_**| v4.3.0 |
|**_Affected Component_**|  fprime-baremetal.Svc.PassiveCmdDispatcher and Svc.CmdDispatcher |
---
## Feature Description

Change behavior of the CmdDispatcher and PassiveCmdDispatcher to execute even if the sequence tracker is full. 

If the sequence tracker table is full, still send back an error code, independent of the status of the command as we will not have the ability to receive back the command status. The best current result to return is [EXECUTION_ERROR](https://github.com/nasa/fprime/blob/devel/Fw/Cmd/Cmd.fpp#L23), but it may be better to create a new command response enumerator to indicate "command executed but status unknown" 

Relevant change scope:
https://github.com/nasa/fprime/blob/devel/Svc/CmdDispatcher/CommandDispatcherImpl.cpp#L98-L103

https://github.com/fprime-community/fprime-baremetal/blob/main/fprime-baremetal/Svc/PassiveCmdDispatcher/PassiveCmdDispatcher.cpp#L154-L173

## Rationale

When a command is executing, it consumes a tracking table entry in the F Prime Command Dispatcher (or F Prime Baremetal PassiveCmdDispathcer). This entry only gets cleaned up when the command completes. If a number of commands are generated by an automated sequence on the main compute element in a short period of time that do not complete expeditiously, or if a coding defect results in commands that never terminate, it could result in exhaustion of the tracking tables, causing critical commands to be rejected at an inopportune time.

michaelkiper:
| | |
|:---|:---|
|**_Related Issue(s)_**|  https://github.com/nasa/fprime/issues/5812#issuecomment-5638784675 |

---
## Change Description

This PR resolves [this CCB comment](https://github.com/nasa/fprime/issues/5812#issuecomment-5638784675) with a caveat of the first bullet (see "Future Work" section):
```
CCB: recommendations:

1. Add a configuration to allow ports to bypass tracking (because the tracking is going to be discarded anyway e.g. uplink)
2. Add configuration (drop on full, dispatch-without-tracking) to allow projects to select this.
3. Add a new response type that says this command is untracked.
```

- Adds in `CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull = false`, as a user defined configuration option for whether to still dispatch a sequence command if the sequence tracker table is full.
     - If false (default), the CmdDispatcher will provide back a `Fw::CmdResponse::EXECUTION_ERROR` to the caller if the sequence tracker table is full and will not dispatch the command. This is the current behavior that already exists.
     - If true, the CmdDispatcher will still dispatch the command and return a `Fw::CmdResponse::DISPATCHED_UNTRACKED` response to the caller AFTER the command has already been dispatched. It occurs in this order as if the user sets `ExecuteCommandWhenSequenceTrackerTableIsFull = true`, that means they have critical commands that must execute immediately, so any caller handling occurs after the dispatching since that's a higher priority.
- Adds in `Fw::CmdResponse::DISPATCHED_UNTRACKED` response type to let the caller know that the command is dispatched but the status is going to be unchecked as the sequence tracker table is full.

## Rationale

resolves https://github.com/nasa/fprime/issues/5812

Rationale for behavior change request is in the above ticket.

```
fprime-util check -j5
```

## Future Work

Make the new behavior of "optional execute command even if sequence tracker table is full" configurable on a per-port basis instead of a global boolean. This is outside the scope of work for the project I work on (not sure if the name is approved for release), but the default configuration preserves the prior behavior of returning an error and not dispatching the command.

This would resolve point 1 of the CCB comment linked at the top of the PR.

Claude Code Opus 5 was used for searching the solution space for where the configuration file, `default/config/CommandDispatcherImplCfg.hpp` lives, identifying common patterns in the code base to replicate, e.g. asking it if there is a common syntax in the code base to change `if constexpr (! CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull) {` to a more obvious branching evaluation than using a logical NOT operator, and having it run a heuristical check to identify if I've made any logical errors in my implementation. 

Claude Code Opus 5 was also used in documentation and unit test generation - I reviewed it for correctness. 

JPL-Devin was used indirectly, borrowing from changes in https://github.com/nasa/fprime/pull/5823/changes.


Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -104,11 +104,12 @@ void CommandDispatcherImpl::compCmdStat_handler(FwIndexType portNum,
 void CommandDispatcherImpl::seqCmdBuff_handler(FwIndexType portNum, Fw::ComBuffer& data, U32 context) {
     Fw::CmdPacket cmdPkt;
     Fw::SerializeStatus stat = cmdPkt.deserializeFrom(data);
+    bool portIsConnected = this->isConnected_seqCmdStatus_OutputPort(portNum);
```
michaelkiper: Evaluate the expression once.

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -117,24 +118,24 @@ void CommandDispatcherImpl::seqCmdBuff_handler(FwIndexType portNum, Fw::ComBuffe
     // look up opcode in dispatch map
     FwIndexType entryPort;
     Fw::Success findStatus = this->m_entryTable.find(cmdPkt.getOpCode(), entryPort);
-    if (findStatus == Fw::Success::SUCCESS and this->isConnected_compCmdSend_OutputPort(entryPort)) {
+    if (findStatus == Fw::Success::SUCCESS && this->isConnected_compCmdSend_OutputPort(entryPort)) {
+        Fw::Success pendingInsertStatus = Fw::Success::SUCCESS;
```
michaelkiper, line 122: Initialize to `Fw::Success::SUCCESS` so that the `if (CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull &&
            pendingInsertStatus != Fw::Success::SUCCESS)` conditional is only taken if `portIsConnected` is true and we call `this->m_sequenceTracker.insert`.

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -117,24 +118,24 @@ void CommandDispatcherImpl::seqCmdBuff_handler(FwIndexType portNum, Fw::ComBuffe
...
             pendingCmd.opCode = cmdPkt.getOpCode();
             pendingCmd.context = context;
             pendingCmd.callerPort = portNum;
 
-            const Fw::Success pendingInsertStatus = this->m_sequenceTracker.insert(sequenceNumber, pendingCmd);
+            pendingInsertStatus = this->m_sequenceTracker.insert(sequenceNumber, pendingCmd);
 
             // if we couldn't find a slot to track the command, quit
-            if (pendingInsertStatus != Fw::Success::SUCCESS) {
+            if (not CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull &&
```
michaelkiper: This still is elided even on `-O0` for clang 22.1 on arm64 - likely for other targets as well.

```
(base) makiper@MT-400290 fprime % clang++ --version
Homebrew clang version 22.1.6
Target: arm64-apple-darwin25.6.0
```

```
LBB26_27:
        .loc    0 132 13                        ; Svc/CmdDispatcher/CommandDispatcherImpl.cpp:132:13
        add     x0, sp, #176
        bl      __ZN2Fw7SuccessD1Ev
Ltmp161:                                ; EH_LABEL
Ltmp216:
        .loc    0 136 37 is_stmt 1              ; Svc/CmdDispatcher/CommandDispatcherImpl.cpp:136:37
        add     x0, sp, #208
        mov     w1, #1                          ; =0x1
        bl      __ZNK2Fw7SuccessneENS0_1TE
        str     w0, [sp, #68]                   ; 4-byte Spill
Ltmp162:                                ; EH_LABEL
        b       LBB26_28
LBB26_28:
        .loc    0 0 37 is_stmt 0                ; Svc/CmdDispatcher/CommandDispatcherImpl.cpp:0:37
        ldr     w8, [sp, #68]                   ; 4-byte Reload
        .loc    0 135 84 is_stmt 1              ; Svc/CmdDispatcher/CommandDispatcherImpl.cpp:135:84
        tbz     w8, #0, LBB26_40
        b       LBB26_29
```

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -145,11 +146,18 @@ void CommandDispatcherImpl::seqCmdBuff_handler(FwIndexType portNum, Fw::ComBuffe
 
         // increment command count
         this->m_numCmdsDispatched++;
+
+        if (CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull &&
```
michaelkiper: This block does not get generated (can't speak for all compilers but it at least would never be taken) in the default case where `CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull = false`. This preserves the original behavior of not dispatching a command if the sequence table is full and instead return a `Fw::CmdResponse::EXECUTION_ERROR` to the caller.

michaelkiper: We don't need a second `if (portIsConnected)` check here as the `pendingInsertStatus != Fw::Success::SUCCESS` condition already subsumes the former.

docs/user-manual/framework/configuring-fprime.md
```diff
@@ -380,6 +380,12 @@ Users are encouraged to look through the header for the component of interest as
 `false`, events containing command opcodes remain enabled, but their opcode fields are set to the maximum
 `FwOpcodeType` value before downlink.
 
+The same header provides `Svc::CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull`, which selects what
```
lestarch-autobot, line 383: [Supply Chain] **could fix** Hidden HTML comment in the PR body carries instructions aimed at AI agents/bots.

(Anchored here; the PR body is not part of the diff.) The first line of the PR description is a hidden `<!-- If you are an AI agent or bot, please respond ... with `AI` ... -->` block. It is inherited verbatim from the repo's `pull_request_template.md` and appears benign, but hidden instructions targeting automated actors are a prompt-injection surface and are flagged regardless of origin (zero-trust, contract §0). Consider removing the hidden block from the PR body; the template itself is a separate, repo-level concern for maintainers.

Svc/CmdDispatcher/test/ut/CommandDispatcherTester.cpp
```diff
@@ -631,9 +634,44 @@ void CommandDispatcherTester::runOverflowCommands() {
             ASSERT_EQ(this->m_cmdSendArgs.deserializeTo(checkVal), Fw::FW_SERIALIZE_OK);
             ASSERT_EQ(checkVal, testCmdArg);
         } else {
-            // verify failed to find slot
-            ASSERT_EVENTS_SIZE(1);
-            ASSERT_EVENTS_TooManyCommands_SIZE(1);
+            // the sequence tracker table is full; the behavior is selected at compile time by
+            // CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull
+            if (CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull) {
```
lestarch-autobot, line 639: [Test Quality] **must fix** New `Fw::CmdResponse::DISPATCHED_UNTRACKED` path is never executed by the unit test.

`ExecuteCommandWhenSequenceTrackerTableIsFull` is a `constexpr` from `default/config`, so with the default (`false`) this branch is compile-time dead: CI never dispatches-on-full, never emits `DISPATCHED_UNTRACKED`, and never checks the `OpCodeDispatched` → `TooManyCommands` ordering. The new enum constant and the new behavior in `seqCmdBuff_handler` therefore ship with zero executed coverage, and the TestMain `COMMENT` ("for both settings") overstates what one build verifies.
Consider making the policy injectable for UT (e.g. a UT-only config override / `-D` for the `CmdDispatcher` UT target, or a second UT build of this test with the flag flipped) so both branches run in CI.

cc @LeStarch @thomas-bc — low-confidence finding, please confirm (same pattern already exists for `IncludeCommandOpcodesInEvents`).

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
lestarch-autobot: [C++ Design] **suggestion** `portIsConnected` is never reassigned; declare it `const` (CPP-11).

Matches the neighbouring `const U32 sequenceNumber` idiom and makes clear the connection state is sampled once per dispatch.

```suggestion
    const bool portIsConnected = this->isConnected_seqCmdStatus_OutputPort(portNum);
```

docs/user-manual/framework/configuring-fprime.md
lestarch-autobot, line 383: [Documentation] **could fix** `docs/reference/system-functional/command-dispatch.md` "Off Nominal" still describes the tracker-full case as an unconditional rejection.

(Anchored here; the stale line, `command-dispatch.md:74`, is not in the PR diff.) The reference page's "Too Many Pending Commands" bullet now only describes the `false` default of the new config; readers of the subsystem reference will not learn that the dispatch-and-report-`DISPATCHED_UNTRACKED` path exists. Proposed replacement for that bullet:

```markdown
- **Too Many Pending Commands** — If the pending command tracker is full, a `TooManyCommands` warning event is logged. By default (`Svc::CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull = false`) the command is rejected with `EXECUTION_ERROR`; when the setting is `true`, the command is dispatched anyway and the source receives `DISPATCHED_UNTRACKED`, meaning its completion status will never be reported.
```

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -145,11 +146,18 @@ void CommandDispatcherImpl::seqCmdBuff_handler(FwIndexType portNum, Fw::ComBuffe
...
+            pendingInsertStatus != Fw::Success::SUCCESS) {
+            this->log_WARNING_HI_TooManyCommands(CmdDispatcherCfg::getEventOpcode(cmdPkt.getOpCode()));
+            this->seqCmdStatus_out(portNum, cmdPkt.getOpCode(), context, Fw::CmdResponse::DISPATCHED_UNTRACKED);
+            return;
+        }
```
lestarch-autobot: [Maintainability] **suggestion** `maint-code-duplication`: the table-full outcome is split across two mirrored blocks whose correctness rests on an unwritten invariant.

This block and the one at L135 carry inverted copies of the same predicate and the same `TooManyCommands` log; a future third mode (or a changed event) must be edited in both places. It also only stays correct because `pendingInsertStatus` is pre-set to `SUCCESS` on the disconnected-port path — that coupling is explained in PR comments, not in the code. The trailing `return;` is a no-op today but will silently skip any statement later appended after this block inside the branch.

```suggestion
        // pendingInsertStatus is only non-SUCCESS for a connected caller whose insert failed (see check above)
        if (CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull &&
            pendingInsertStatus != Fw::Success::SUCCESS) {
            this->log_WARNING_HI_TooManyCommands(CmdDispatcherCfg::getEventOpcode(cmdPkt.getOpCode()));
            this->seqCmdStatus_out(portNum, cmdPkt.getOpCode(), context, Fw::CmdResponse::DISPATCHED_UNTRACKED);
        }
```

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -117,24 +118,24 @@ void CommandDispatcherImpl::seqCmdBuff_handler(FwIndexType portNum, Fw::ComBuffe
...
             pendingCmd.context = context;
             pendingCmd.callerPort = portNum;
 
-            const Fw::Success pendingInsertStatus = this->m_sequenceTracker.insert(sequenceNumber, pendingCmd);
+            pendingInsertStatus = this->m_sequenceTracker.insert(sequenceNumber, pendingCmd);
 
             // if we couldn't find a slot to track the command, quit
-            if (pendingInsertStatus != Fw::Success::SUCCESS) {
+            if (not CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull &&
+                pendingInsertStatus != Fw::Success::SUCCESS) {
```
lestarch-autobot: [Maintainability] **suggestion** `maint-misleading-comment`: comment says "quit" on a full table, but the new condition only quits when the config is `false`.

With `ExecuteCommandWhenSequenceTrackerTableIsFull = true` the same insert failure now falls through to dispatch and is reported at L150, so the comment describes half the behavior. Separately, `not` mixed with `&&` in one expression diverges from the file (L121 in this very diff was normalized from `and` to `&&`); use `!` for a single operator dialect.

```suggestion
            // table full: reject here unless configured to dispatch untracked (then reported after dispatch below)
            if (!CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull &&
                pendingInsertStatus != Fw::Success::SUCCESS) {
```

Fw/Cmd/Cmd.fpp
```diff
@@ -23,6 +23,7 @@ module Fw {
     EXECUTION_ERROR = 4 @< Command had execution error
     BUSY = 5 @< Component busy
     CLEARED = 6 @< Command tracking was cleared before the command completed
+    DISPATCHED_UNTRACKED = 7 @< Command was dispatched but the response tracking failed, indicating unknown state
```
lestarch-autobot, line 26: [Design] **could fix** New `DISPATCHED_UNTRACKED = 7` is not mirrored in the WASM guest ABI enum `FprimeCmdResponse` (`Svc/WasmSequencer/spacewasm_include/fprime.h`).

`WasmSequencer::cmdResponseIn_handler` forwards `response.e` verbatim to guest code via `hostResumeI32`, so guests can now receive value 7 with no named constant. That header already lags `Fw::CmdResponse` (`FPRIME_CMD_CLEARED = 5` collides with `BUSY`; being corrected in #6001), so please add `FPRIME_CMD_DISPATCHED_UNTRACKED = 7` alongside so the two enums do not drift further.

Svc/CmdDispatcher/docs/sdd.md
```diff
@@ -57,6 +58,7 @@ Note #1: this requires that the component sending the command buffer have connec
...
+Note #5: the pending sequence command table is a fixed-size table (`CMD_DISPATCHER_SEQUENCER_TABLE_SIZE`), and it is only consulted when the `seqCmdStatus` output port for the calling port is connected. When a command arrives, that port is connected, and the table is full, the dispatcher's behavior is selected at compile time by `Svc::CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull`. When `false` (the default), the command is not dispatched: a `TooManyCommands` warning is emitted and the caller receives `Fw::CmdResponse::EXECUTION_ERROR`. When `true`, the command is dispatched, a `TooManyCommands` warning is emitted, and the caller receives `Fw::CmdResponse::DISPATCHED_UNTRACKED`, indicating that the command is executing but that its completion status cannot be matched back to the caller and will never be reported.
```
lestarch-autobot: [Design] **suggestion** The design leaves unstated how existing callers react to `DISPATCHED_UNTRACKED`: for `Svc::CmdSequencer` any non-`OK` status (`cmdResponseIn_handler`, `response != Fw::CmdResponse::OK`) calls `commandError` + `performCmd_Cancel`, so with the opt-in enabled a critical command executes but the sequence still aborts at that step, exactly as it does today with `EXECUTION_ERROR`.

That may be the intended contract (the issue's goal is that the command runs), but a project enabling the flag should learn it here rather than from a cancelled sequence. Documenting the stock-sequencer behavior keeps the SDD honest about what the new status buys.

```suggestion
Note #5: the pending sequence command table is a fixed-size table (`CMD_DISPATCHER_SEQUENCER_TABLE_SIZE`), and it is only consulted when the `seqCmdStatus` output port for the calling port is connected. When a command arrives, that port is connected, and the table is full, the dispatcher's behavior is selected at compile time by `Svc::CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull`. When `false` (the default), the command is not dispatched: a `TooManyCommands` warning is emitted and the caller receives `Fw::CmdResponse::EXECUTION_ERROR`. When `true`, the command is dispatched, a `TooManyCommands` warning is emitted, and the caller receives `Fw::CmdResponse::DISPATCHED_UNTRACKED`, indicating that the command is executing but that its completion status cannot be matched back to the caller and will never be reported. Note that the stock sequencers do not treat `DISPATCHED_UNTRACKED` as success: `Svc::CmdSequencer` fails and cancels the sequence on any non-`OK` response, `Svc::WasmSequencer` counts it as a failed command, and `Svc::FpySequencer` pushes the raw response value for the sequence to branch on.
```

Svc/CmdDispatcher/docs/sdd.md
```diff
@@ -16,6 +16,7 @@ CD-003 | The `Svc::CmdDispatcher` component shall provide an interface to regist
 CD-004 | The `Svc::CmdDispatcher` component shall process command status from components and report the results to the command buffer sender. | Unit Test
 CD-005 | The `Svc::CmdDispatcher` component shall drop incoming commands to avert a queue overflow (DOS attack). | Unit Test
 CD-006 | The `Svc::CmdDispatcher` component shall count every dropped command, including drops occurring concurrently on multiple caller threads, and report the count in the `CommandsDropped` telemetry channel. | Unit Test
+CD-007 | The `Svc::CmdDispatcher` component shall not dispatch a command that it cannot track in the pending command table, unless configured to do so, in which case it shall dispatch the command and report that the command is untracked. | Unit Test
```
lestarch-autobot, line 19: [Design] **could fix** The PR body says it resolves the CCB comment on #5812, but only recommendations 2 and 3 (drop-vs-dispatch config, new response type) are implemented; recommendation 1 (a per-port configuration to bypass tracking entirely, e.g. for uplink whose status is discarded) is not addressed here or in the SDD.

If item 1 is intentionally deferred, please say so in the PR description / Future Work (and optionally a follow-up issue) so the CCB comment is not marked resolved with a gap.

Svc/CmdDispatcher/docs/sdd.md
lestarch-autobot: [Operational] **could fix** Note #5 omits the worst-case operational consequence of `ExecuteCommandWhenSequenceTrackerTableIsFull = true` for in-repo callers.

Finding class: ops-doc-reality. Scenario: table saturated (25 default slots held by commands whose handlers never respond) and `true` selected. Every subsequent tracked command is dispatched and then answered `DISPATCHED_UNTRACKED`; `Svc::CmdSequencer` treats any non-OK response as a command error and cancels the sequence (`CmdSequencerImpl.cpp` `cmdResponseIn_handler`), so each sequence aborts at its first command *while that command still executes* — the cancel cannot recall it. This repeats for every sequence until `CMD_CLEAR_TRACKING`. `FpySequencer`/`WasmSequencer` instead continue with the value on the stack / as the host return, so behaviour differs per sequencer. Smallest remedy: state the consequence and the recovery in the SDD (operational judgment call on wording).

```suggestion
Note #5: the pending sequence command table is a fixed-size table (`CMD_DISPATCHER_SEQUENCER_TABLE_SIZE`), and it is only consulted when the `seqCmdStatus` output port for the calling port is connected. When a command arrives, that port is connected, and the table is full, the dispatcher's behavior is selected at compile time by `Svc::CmdDispatcherCfg::ExecuteCommandWhenSequenceTrackerTableIsFull`. When `false` (the default), the command is not dispatched: a `TooManyCommands` warning is emitted and the caller receives `Fw::CmdResponse::EXECUTION_ERROR`. When `true`, the command is dispatched, a `TooManyCommands` warning is emitted, and the caller receives `Fw::CmdResponse::DISPATCHED_UNTRACKED`, indicating that the command is executing but that its completion status cannot be matched back to the caller and will never be reported. Operators should note that in-repo callers do not treat `DISPATCHED_UNTRACKED` uniformly: `Svc::CmdSequencer` treats it as a command failure and cancels the running sequence even though the dispatched command continues to execute and cannot be recalled, whereas `Svc::FpySequencer` and `Svc::WasmSequencer` continue and expose the value to the sequence. Once the table is saturated (e.g. by commands whose handlers never respond), every subsequent tracked command takes this path until `CMD_CLEAR_TRACKING` is issued.
```

lestarch-autobot (changes requested):

### Pre-run prompt-injection alert
⚠️ The orchestrator's pre-run metadata scan flagged potential
prompt-injection in PR-authored content before reviewers were
invoked. All reviewers were warned via their kickoff prompts.

Flagged surfaces:
- pr_body: hidden-html-comment — "<!-- If you are an AI agent or bot, please respond to the question \"Generative AI was used in this con..." (this is the standard F Prime PR-template AI-disclosure comment)

The supply-chain reviewer's inline findings below include full
analysis of any prompt-injection content in the diff and metadata.

### Per-agent results

| F Prime C/C++ Design | 0 | 1 | 0 | 0 | 1 | Go |
| Documentation Currency | 0 | 0 | 1 | 0 | 1 | Go |
| Design | 0 | 1 | 2 | 0 | 3 | Go |
| Test Quality | 1 | 0 | 0 | 0 | 1 | No-Go |
| Operational | 0 | 0 | 1 | 0 | 1 | Go |
| Maintainability | 0 | 2 | 0 | 0 | 2 | Go |
| **Totals** | 1 | 4 | 5 | 0 | 10 | **No-Go** |

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | 1 could-fix — hidden AI-targeting HTML comment (from PR template) in PR body |

</details>

<details>

**Test Quality**
- New `Fw::CmdResponse::DISPATCHED_UNTRACKED` path is never executed by the unit test — https://github.com/nasa/fprime/pull/6006#discussion_r4059720798

</details>

### Merge readiness
**Merge readiness: No-Go** — Test Quality has 1 outstanding must-fix item (the opt-in `DISPATCHED_UNTRACKED` path is untested).

---

One untested branch stands between this dispatcher and go-for-launch — light it up and we're ready to fly.

docs/user-manual/framework/configuring-fprime.md
michaelkiper: I saw that there was a html comment in the top of the auto-generated PR body - removed it.

Svc/CmdDispatcher/docs/sdd.md
michaelkiper: I'm marking this as resolved as I now clarified in the PR body

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
michaelkiper: > future third mode (or a changed event) must be edited in both places

lambdas 👀 ....

default/config/CommandDispatcherImplCfg.hpp
```diff
@@ -28,6 +28,13 @@ constexpr FwOpcodeType getEventOpcode(const FwOpcodeType opcode) {
     return IncludeCommandOpcodesInEvents ? opcode : std::numeric_limits<FwOpcodeType>::max();
 }
 
+//! When true, execute sequence command even if the sequence tracker table is full.
+//! When false, do not execute the sequence command if the sequence tracker table is full.
+#ifndef CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL
+#define CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL false
```
michaelkiper, line 35: Macro injection for easy unit testing of both true and false conditions

Svc/CmdDispatcher/docs/sdd.md
michaelkiper: Fixed in latest commit

Fw/Cmd/Cmd.fpp
michaelkiper: Fixed in latest commit

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
michaelkiper: Fixed in latest commit

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
michaelkiper: Fixed in latest commit

docs/user-manual/framework/configuring-fprime.md
michaelkiper: Fixed in latest commit

Svc/CmdDispatcher/test/ut/CommandDispatcherTester.cpp
michaelkiper: Fixed to have both branches taken

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -117,24 +118,29 @@ void CommandDispatcherImpl::seqCmdBuff_handler(FwIndexType portNum, Fw::ComBuffe
     // look up opcode in dispatch map
     FwIndexType entryPort;
     Fw::Success findStatus = this->m_entryTable.find(cmdPkt.getOpCode(), entryPort);
-    if (findStatus == Fw::Success::SUCCESS and this->isConnected_compCmdSend_OutputPort(entryPort)) {
+    if (findStatus == Fw::Success::SUCCESS && this->isConnected_compCmdSend_OutputPort(entryPort)) {
```
lestarch-autobot: [Maintainability] **suggestion** Follow-up to https://github.com/nasa/fprime/pull/6006#discussion_r4059724575: the fix settled on `not`/`and` (L141, L156), but this line was flipped the other way, from the file's original `and` to `&&`.

`seqCmdBuff_handler` now reads in two operator dialects within 35 lines; a reader has to stop and check whether the difference is meaningful. Restore this line so the function uses one dialect.

```suggestion
    if (findStatus == Fw::Success::SUCCESS and this->isConnected_compCmdSend_OutputPort(entryPort)) {
```

default/config/CommandDispatcherImplCfg.hpp
```diff
@@ -28,6 +28,13 @@ constexpr FwOpcodeType getEventOpcode(const FwOpcodeType opcode) {
     return IncludeCommandOpcodesInEvents ? opcode : std::numeric_limits<FwOpcodeType>::max();
 }
 
+//! When true, execute sequence command even if the sequence tracker table is full.
+//! When false, do not execute the sequence command if the sequence tracker table is full.
```
lestarch-autobot, line 32: [Maintainability] **could fix** `maint-misleading-comment`: the doc comment does not mention that the value is macro-overridable, so the `#ifndef` block below reads as an unexplained indirection.

The reason (the UT build compiles both settings via `-DCMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL`) lives only in a PR comment; the next engineer editing this header cannot tell whether to change the `#define` default or the `constexpr`.

```suggestion
//! When true, execute sequence command even if the sequence tracker table is full.
//! When false, do not execute the sequence command if the sequence tracker table is full.
//! Overridable with -DCMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL=<bool> (the UT build compiles both settings).
```

lestarch-autobot, line 32: [Documentation] **suggestion** New `CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL` override macro is undocumented.

The doc comment above still describes only the `constexpr`; a project reading this header (or `docs/user-manual/framework/configuring-fprime.md`, which names only the `constexpr`) will not learn that the value can be overridden with a `-D` compile definition (as the UTs now do in `Svc/CmdDispatcher/CMakeLists.txt`). One line in the header is enough.

```suggestion
//! When false, do not execute the sequence command if the sequence tracker table is full.
//! Defaults to false; may be overridden with -DCMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL=true (used by the UTs).
```

lestarch-autobot, line 32: [Design] **suggestion** — This adds a second configuration surface for the same knob. F Prime's convention (see `default/config/` and AGENTS.md) is that projects tune compile-time config by overriding the config header, and no other `*Cfg.hpp` exposes a `#ifndef` preprocessor override. A deployment that sets `-DCMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL` on some targets but not others would compile `CommandDispatcherImpl.cpp` with two different values of a `constexpr` (the UT setup itself relies on this: the UT's own `CommandDispatcherImpl.o` silently shadows the one in the linked `Svc_CmdDispatcher` archive). If the macro is intended only as the unit-test hook, say so here so it is not mistaken for a supported project-configuration path.

```suggestion
//! When true, execute sequence command even if the sequence tracker table is full.
//! When false, do not execute the sequence command if the sequence tracker table is full.
//! Projects configure this by overriding this header. The CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL
//! macro exists only so the unit tests can build both variants; it is not a supported deployment knob.
```

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -117,24 +118,29 @@ void CommandDispatcherImpl::seqCmdBuff_handler(FwIndexType portNum, Fw::ComBuffe
...
+        Fw::Success pendingInsertStatus = Fw::Success::SUCCESS;
         const U32 sequenceNumber = this->allocateSequenceNumber();
 
+        // report that the sequence tracker table is full, with the given response to the caller
+        auto reportTrackerFull = [&](Fw::CmdResponse response) {
+            this->log_WARNING_HI_TooManyCommands(CmdDispatcherCfg::getEventOpcode(cmdPkt.getOpCode()));
+            this->seqCmdStatus_out(portNum, cmdPkt.getOpCode(), context, response);
+        };
```
lestarch-autobot, line 129: [C++ Design] **could fix** CPP-7 (no lambdas): `reportTrackerFull` is a capturing lambda in a flight handler.

Lambdas are outside the F Prime C++ subset (they synthesize an unnamed closure type and hide ownership of captured state). Move this to a private helper, e.g. `void reportTrackerFull(FwIndexType portNum, FwOpcodeType opCode, U32 context, Fw::CmdResponse response);` in `CommandDispatcherImpl.hpp`, and call it from both sites.

default/config/CommandDispatcherImplCfg.hpp
```diff
@@ -28,6 +28,13 @@ constexpr FwOpcodeType getEventOpcode(const FwOpcodeType opcode) {
...
+#ifndef CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL
+#define CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL false
+#endif
+constexpr bool ExecuteCommandWhenSequenceTrackerTableIsFull = CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL;
```
lestarch-autobot, line 37: [C++ Design] **could fix** CPP-8 (typed constants over `#define`): untyped macro override knob added to a config header.

`CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL` is an untyped preprocessor symbol whose value is only checked when it reaches the `constexpr bool`; no other `default/config` header exposes a compile-definition override, since the sanctioned way to change a config value is a project config override of this header. If the macro is kept for the UT matrix, at least apply the same pattern consistently and document that it is test-only.

Svc/WasmSequencer/spacewasm_include/fprime.h
```diff
@@ -109,6 +109,8 @@ enum FprimeCmdResponse {
     FPRIME_CMD_BUSY = 5,
     /// Command tracking was cleared before the command completed
     FPRIME_CMD_CLEARED = 5,
```
lestarch-autobot, line 111: [Operational] **future work** Guest ABI `FPRIME_CMD_CLEARED = 5` collides with `FPRIME_CMD_BUSY`; `Fw::CmdResponse::CLEARED` is 6.

Finding class: ops-contract-trace (preexisting; this PR extends the same enum). Scenario: a WASM guest branches on the `fprime_wasm_dispatch_command` return value. `WasmSequencer::cmdResponseIn_handler` forwards `response.e` verbatim, so a real `CLEARED` (6, emitted by `CMD_CLEAR_TRACKING`) matches no guest constant, while a `BUSY` (5) is misread as `CLEARED`: 2 of the 8 response codes are misrouted by any guest that tests this constant. Smallest remedy: set the value to 6 (already done in #6001; rebasing once that lands is sufficient).

```suggestion
    FPRIME_CMD_CLEARED = 6,
```

lestarch-autobot (approved):

### Per-agent results

| Supply Chain / Runner Safety | 0 | 0 | 1 | 0 | 0 | Go |
| F Prime C/C++ Design | 0 | 1 | 2 | 0 | 2 | Go |
| Documentation Currency | 0 | 1 | 1 | 0 | 1 | Go |
| Design | 0 | 2 | 2 | 0 | 1 | Go |
| Test Quality | 1 | 0 | 0 | 0 | 0 | Go |
| Operational | 0 | 0 | 1 | 1 | 1 | Go |
| Maintainability | 0 | 3 | 1 | 0 | 2 | Go |
| **Totals** | 1 | 7 | 8 | 1 | 7 | **Go** |

<details>

| Supply Chain / Runner Safety | 1 | 0 | 0 | 0 | 0 | 0 |
| F Prime C/C++ Design | 1 | 0 | 2 | 0 | 0 | 0 |
| Documentation Currency | 1 | 0 | 1 | 0 | 0 | 0 |
| Design | 3 | 0 | 1 | 0 | 0 | 0 |
| Test Quality | 1 | 0 | 0 | 0 | 0 | 0 |
| Operational | 1 | 0 | 1 | 0 | 0 | 0 |
| Maintainability | 2 | 0 | 2 | 1 | 0 | 0 |

</details>

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

### Merge readiness
**Merge readiness: Go** — all ten reviewers completed with zero outstanding must-fix items; the run-1 Test Quality must-fix (untested `DISPATCHED_UNTRACKED` path) is now covered by the dual-variant UT build. 7 non-blocking threads remain open (2 suggestions + 4 could-fix on the `CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL` macro/lambda, 1 preexisting future-work in `fprime.h`).

---

Both dispatcher variants now light up green on the test board — handing off to mission control @LeStarch @thomas-bc for the final go.

Svc/WasmSequencer/spacewasm_include/fprime.h
michaelkiper: I won't be applying this fix here as I don't want to conflict #6001 and it's out of scope for this PR

default/config/CommandDispatcherImplCfg.hpp
michaelkiper: This is overly redundant imo - it's obvious in the code.

michaelkiper: See https://github.com/nasa/fprime/pull/6006/changes#r4068766226

michaelkiper: This is obvious so that I don't think that adding this comment would add any information

default/config/CommandDispatcherImplCfg.hpp
michaelkiper: This macro is only kept for unit testing

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
michaelkiper: Is this a strict requirement? I don't think that this is lambda abuse - it's an appropriate use case. 

For a human reviewer, let me know if you want me to take out the lambda.

lestarch-autobot: [C++ Design] **Disagreement — escalating.** I still flag this on 257dcb9697f1a5956ca376e7e53f07e05cbb14d6; the contributor's response above indicates we disagree.

Re-checked `seqCmdBuff_handler` at the new head: `reportTrackerFull` is still a capturing lambda. The contributor's point that this is a tidy, local use is fair; CPP-7 in `.github/skills/fprime-cpp-design/SKILL.md` nonetheless excludes lambdas from the F Prime subset, and a private member helper expresses the same thing, so I leave the (non-blocking) could-fix open.

cc @LeStarch @thomas-bc — needs human adjudication.

Svc/WasmSequencer/spacewasm_include/fprime.h
lestarch-autobot: [Operational] **Improperly resolved.** This finding still applies on 257dcb9697f1a5956ca376e7e53f07e05cbb14d6.

`Svc/WasmSequencer/spacewasm_include/fprime.h:111` still defines `FPRIME_CMD_CLEARED = 5` (colliding with `FPRIME_CMD_BUSY`) while this PR extends the same enum with `DISPATCHED_UNTRACKED = 7`. The contributor's rationale (deferring to #6001 to avoid a conflict) is reasonable for a preexisting future-work item, but a non-maintainer resolution cannot close the thread under the review contract; a maintainer's resolution can.

cc @LeStarch @thomas-bc — contributor resolved without addressing; please adjudicate (accepting the #6001 deferral is fine).

default/config/CommandDispatcherImplCfg.hpp
```diff
@@ -28,6 +28,14 @@ constexpr FwOpcodeType getEventOpcode(const FwOpcodeType opcode) {
     return IncludeCommandOpcodesInEvents ? opcode : std::numeric_limits<FwOpcodeType>::max();
 }
 
+//! When true, execute sequence command even if the sequence tracker table is full.
+//! When false, do not execute the sequence command if the sequence tracker table is full.
+//! The 'CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL' macro exists only so the unit tests can build both conditions. It is not a supported deployment knob.
```
lestarch-autobot, line 33: [C++ Design] **suggestion** Follow-up to https://github.com/nasa/fprime/pull/6006#discussion_r4068211213: the new comment line is 152 columns, over the 120-column limit in `.clang-format` (CPP-26).

`clang-format` reflows it to two lines; the pre-commit `fprime-util format` hook would have done the same.

```suggestion
//! The 'CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL' macro exists only so the unit tests can build both conditions. It is
//! not a supported deployment knob.
```

lestarch-autobot (approved):

### Per-agent results

| Supply Chain / Runner Safety | 0 | 0 | 1 | 0 | 0 | Go |
| F Prime C/C++ Design | 0 | 2 | 2 | 0 | 2 | Go |
| Documentation Currency | 0 | 1 | 1 | 0 | 0 | Go |
| Design | 0 | 2 | 2 | 0 | 0 | Go |
| Test Quality | 1 | 0 | 0 | 0 | 0 | Go |
| Operational | 0 | 0 | 1 | 1 | 1 | Go |
| Maintainability | 0 | 3 | 1 | 0 | 0 | Go |
| **Totals** | 1 | 8 | 8 | 1 | 3 | **Go** |

<details>

| F Prime C/C++ Design | 1 | 0 | 1 | 1 | 0 | 1 |
| Documentation Currency | 1 | 0 | 0 | 0 | 0 | 0 |
| Design | 1 | 0 | 0 | 0 | 0 | 0 |
| Test Quality | 0 | 0 | 0 | 0 | 0 | 0 |
| Operational | 0 | 0 | 0 | 0 | 1 | 0 |
| Maintainability | 2 | 0 | 0 | 0 | 0 | 0 |

</details>

<details>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Generator output | clean |
| Prompt-injection | clean |

</details>

### Merge readiness
**Merge readiness: Go** — all ten reviewers completed with zero outstanding must-fix items; the head delta since run 2 is `&&`→`and` plus one config comment line. Two non-blocking items await human adjudication: the C++ Design CPP-7 lambda thread carries a contributor disagreement escalated to maintainers (https://github.com/nasa/fprime/pull/6006#discussion_r4068211212), and the Operational future-work thread on the preexisting `fprime.h` `CLEARED` collision was resolved by a non-maintainer (https://github.com/nasa/fprime/pull/6006#discussion_r4070164344). 1 new C++ Design suggestion (line-length follow-up) is open.

---

Third pass, all systems nominal — the dispatcher is cleared for docking pending maintainer sign-off on the two flagged threads.

Svc/CmdDispatcher/CommandDispatcherImpl.cpp
```diff
@@ -118,23 +119,28 @@ void CommandDispatcherImpl::seqCmdBuff_handler(FwIndexType portNum, Fw::ComBuffe
     FwIndexType entryPort;
     Fw::Success findStatus = this->m_entryTable.find(cmdPkt.getOpCode(), entryPort);
     if (findStatus == Fw::Success::SUCCESS and this->isConnected_compCmdSend_OutputPort(entryPort)) {
+        Fw::Success pendingInsertStatus = Fw::Success::SUCCESS;
         const U32 sequenceNumber = this->allocateSequenceNumber();
 
+        // report that the sequence tracker table is full, with the given response to the caller
+        auto reportTrackerFull = [&](Fw::CmdResponse response) {
```
LeStarch, line 126: We should remove this lambda function, it is 2 lines of code and very hard to follow.   Just inline the lines in the two places they are needed.

default/config/CommandDispatcherImplCfg.hpp
```diff
@@ -28,6 +28,14 @@ constexpr FwOpcodeType getEventOpcode(const FwOpcodeType opcode) {
...
+#ifndef CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL
```
LeStarch, line 34: This effects all command dispatchers in the system.  It also makes the command dispatcher UT harder to write as now we replicate tests over 2 UT executables.

I'd recommend that we establish this configuration as a default:

```
constexpr bool EXECUTE_WHEN_SEQUENCE_TABLE_FULL_DEFAULT = CMD_DISPATCHER_EXECUTE_WHEN_TRACKER_FULL;
```

Then pull this into a member variable through a config function with the defaulted parameter.   Then in a single UT executable you can have separate tests that call different config.

LeStarch (changes requested):

default/config/CommandDispatcherImplCfg.hpp
michaelkiper: Doing it as a member variable wouldn't allow the compiler to elide the untaken branch instruction set, which is why I made it the macro w/ 2 UT executions. It's a small runtime cost: load, cmp, Icache miss?, but wasn't 0 so that's why I valued the implementation I had chosen.

Issue #5234, opened 2026-06-03
ComStub can deadlock telemetry after send/retry failures because queue never receives SUCCESS


etiennecollin:
| | |
|:---|:---|
|**_F´ Version_**| `4.0.0a3`, but also applies to latest in `devel` |
|**_Affected Component_**| `Svc/ComStub` and all drivers under `Drv/`  |
---

## Problem Description

We have a communication chain like:

`comQueue <-> comStub <-> udpDriver`

Under certain failure conditions, a driver send failure can leave the queue permanently blocked waiting for a successful `comStatus`, which prevents further telemetry from being pushed.

> [!NOTE]
> Although this issue was observed with udpDriver, the underlying problem is not UDP-specific. Any driver implementing the ByteStreamDriver pattern (e.g. UART, TCP, etc.) can encounter the same deadlock scenario when send failures or retry exhaustion result in Fw::Success::FAILURE being propagated back to comQueue. The current interaction between comStub and drivers relies on implicit reinitialization behaviour that is not expressed in the interface, making the issue applicable across all drivers using this pattern.

#### Observed behaviour

When `udpDriver` encounters a send issue, such as `SocketIpStatus::SOCK_SEND_ERROR` after a network drop, its send handler returns `ByteStreamStatus::OTHER_ERROR` to `comStub`. `comStub` then:

- sets `m_reinitialize = true`
- returns `Fw::Success::FAILURE` to `comQueue`

This creates a stall:

- `comQueue` waits forever for a success status
- no more telemetry packets are pushed
- the driver never receives anything further
- the system remains alive and responsive to commands, but telemetry stops entirely
- even if the queue was made to push a packet, there is an assert on m_reinitialize in `comStub` that would fail

A similar failure mode exists when the retry counter limit is reached. In this case,, `m_reinitialize` is not involved, but the queue still deadlocks on the failure return.

#### Root cause

This looks like a leaky abstraction in the `comStub` design:

- `comStub` is trying to manage driver state, but it cannot actually know whether the driver really needs reinitialization
- the driver also does not know that `comStub` expects it to reinitialize and emit a ready signal

As a result, `comStub` can request state transitions that the driver cannot satisfy, leaving both sides stuck.

#### Impact

[...]

Issue #3129, opened 2025-01-13
TCP driver reconnect time


zimri-leisher:
| | |
|:---|:---|
|**_F´ Version_**| 3.5.1 |
|**_Affected Component_**| TcpClient |
---
## Problem Description

We are now having the following issue:
1. Given a TcpClient set up to send to an IP which is not currently running a TcpServer
2. When we invoke the `send` port of the TcpClient
3. We see it make a call to `connect` every time
4. This delays the calling thread by 10-20 ms, which at the rate we are scheduling it, quickly causes the rate group queue to overfill and assert.

For some reason, we were not seeing this issue before we updated. Perhaps it was only occasionally trying to reconnect?

## Context / Environment

Execute `fprime-util version-check` and share the output.

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
Python version: 3.10.12
CMake version: 3.22.1
Pip version: 22.0.2
Pip packages:
    fprime-tools==3.5.1
    fprime-gds==3.5.1
    fprime-fpp-*==2.2.1
Project submodules:```

LeStarch:
Part of the TCP refactor was to add `connect/reconnect` capability on the send side of the TCP library as well as on the recv side.  The reason is that we had issues where send would fail if the read reconnect had not happened yet.

For use cases such as this, it is expected that you connect before attempting to send and thus ensure that the connect/reconnect does not happen yet.

You can also set reconnect = false in the configure call to disable all reconnect behavior.

zimri-leisher:
Ah, I see. That makes sense. Maybe we could consider adding a timeout to the connection attempt? It makes a lot of sense to attempt to connect, but it violates the assumption we made that setting a timeout in the `::configure` method would put a bound on the execution time of the `send` port. I can make a ticket and look into implementing that if interested.

LeStarch:
We could add a connection timeout, and I wouldn't be opposed to it.  However, I think the preferred method would be to turn off auto-open, and handle the connection off the critical thread. 

I don't know if a timeout will properly bound that call anyway because you have two timeouts: SEND, and RECV.  The connect call would be composed of n number of send an receives to establish the connection.

The best way to get a pseudo-bound on the send is to not connect on send.  Then it will be limited to 1 iteration of the send timeout.

> F Prime does not currently support send timeouts.  You have to address the socket directly for that.  If this feature is needed we can consider adding it.

[...]

PR #6013, opened 2026-09-22
Retry IpSocket send on send timeout (EAGAIN)

### LINKED ISSUE #5960: Drv/Ip: IpSocket::send() treats a send timeout (EAGAIN) as fatal SOCK_SEND_ERROR
| | |
|:---|:---|
|**_F´ Version_**| `devel` @ 4dee010c30 |
|**_Affected Component_**| `Drv/Ip` (`IpSocket::send`), reached by `Drv::TcpClient`, `Drv::TcpServer`, `Drv::Udp` |
---
## Problem Description

`IpSocket::send()` reports a send timeout as a fatal error. `setupTimeouts()` (`IpSocket.cpp:75-90`) puts `SO_SNDTIMEO` on every socket the three IP drivers open (default 1 s, `SOCKET_SEND_TIMEOUT_SECONDS`). When it expires with nothing written, `::send` returns `-1`/`EAGAIN`, which falls through to the catch-all at `:170`:

```cpp
else if (sent == -1) {
    return SOCK_SEND_ERROR;
}
```

The documented contract is the opposite:

- `Drv/Ip/docs/sdd.md:47`: *"It will retry to transmit data a configured number of times on correctable errors … Interrupts and timeouts are the only recoverable errors."*
- `IpSocket.hpp:42`: `SOCK_SEND_ERROR //!< Failed to send after configured retries`
- `recv()` already handles `EAGAIN`/`EWOULDBLOCK` explicitly (`:209-212`); `send()` has no such branch, and its comment at `:161` ("Error is EINTR or timeout just try again") describes a retry the branch cannot perform.

The SDD sentence and the catch-all were written in the same commit (8b4e870082, #460), so the doc reflects the intent. Same class of defect as #5337 (recv-side EINTR), fixed in #5344. The sockets are blocking (`SOCKET_IP_SEND_FLAGS = 0`, no `O_NONBLOCK` anywhere in `Drv/Ip`), so the timeout is the only source of `EAGAIN` on this path.

**Impact.** `SOCK_SEND_ERROR` skips the retry budget: the drivers map only `SOCK_INTERRUPTED_TRY_AGAIN → SEND_RETRY` and everything else to `OTHER_ERROR` (`TcpClientComponentImpl.cpp:73-88`); ComStub then sets `m_reinitialize` and emits `FAILURE` (`ComStub.cpp:87-111`), parking ComQueue. `m_reinitialize` clears only on a driver reconnect, but the socket is not closed on `SOCK_SEND_ERROR` (`SocketComponentHelper.cpp:128-152` closes only on `SOCK_DISCONNECTED`), so on a live link no reconnect comes. One timeout on a healthy connection stops downlink until the peer closes the connection or the process restarts, and nothing is logged when it happens. Ref uses the 1 s default (`comDriver: Drv.TcpClient`, `RefTopology.cpp:80`).

## Context / Environment

```
Operating System: Darwin
CPU Architecture: arm64
Platform: macOS-26.6.2-arm64-arm-64bit
Python version: 3.12.3
CMake version: 3.26.0
Pip version: 26.2
Pip packages:
    fprime-tools==4.3.0a5
    fprime-gds==4.2.2a4
    fprime-fpp==3.3.0
```

## How to Reproduce

1. Run `Ref -a 127.0.0.1 -p 50000` against a plain TCP peer that reads for 10 s, then stops reading while keeping the socket open.
2. Watch `netstat -an -p tcp`: at Ref's ~750 B/s the peer Recv-Q fills (~401 KB, ~9 min), then the FSW Send-Q (~147 KB, ~3 min more) and goes flat — `send()` is now blocking and timing out.
3. Have the peer resume reading ~8 s after the Send-Q goes flat.

| build | received in the 40 s after resume | last byte at |
|:--|:--|:--|
| devel | 547,840 B — exactly the buffered residue (401,152 + 146,688) | +0.0 s, then nothing |
| with the fix below | 582,656 B | +39.2 s (continuous) |

(The resume must come within ~10 s of the plateau; longer stalls exhaust ComStub's `RETRY_LIMIT` on both builds, which is #5234.)

Unit-level: a mock in the style of `InterruptOnceSocket` (`TestTcp.cpp:16`) whose `sendProtocol` sets `errno = EAGAIN` and returns `-1` gets `-13` (`SOCK_SEND_ERROR`) from `send()`; expected `-6` (`SOCK_INTERRUPTED_TRY_AGAIN`).

## Expected Behavior

A send timeout is returned as `SOCK_INTERRUPTED_TRY_AGAIN`, so the drivers report `SEND_RETRY` and ComStub's retry budget applies, as the SDD describes:

```cpp
else if ((sent == -1) && ((errno == EAGAIN) || (errno == EWOULDBLOCK))) {
    return SOCK_INTERRUPTED_TRY_AGAIN;
}
```

before the catch-all, the same in `UdpSocket::send`'s zero-length path (`UdpSocket.cpp:270`), and the `:161` comment corrected. Reusing the existing status keeps every driver's `default: OTHER_ERROR` mapping correct, including downstream `ByteStreamDriver` implementations. No ComStub change. A PR with the fix and both tests is ready; existing Drv/Ip, TcpClient, TcpServer and Udp suites pass unchanged.

Out of scope: recovery after `RETRY_LIMIT` on a sustained stall is #5234 (worth noting there that the socket is not closed on send errors, so the reconnect ComStub waits for after retry exhaustion does not occur on a live link).

*AI usage: an AI assistant (Claude Code) was used to trace the code paths, build the reproduction harness and draft this text; the history, test results and the Ref reproduction were reviewed by me.*

rvaccone:
| | |
|:---|:---|
|**_Related Issue(s)_**| Fixes #5960 |

---
## Change Description

`IpSocket::send()` returns `SOCK_SEND_ERROR` when `SO_SNDTIMEO` expires with nothing written. The drivers report `OTHER_ERROR`, ComStub parks ComQueue until the driver reconnects, and a live but stalled link never reconnects. One timeout on a healthy connection stops downlink until the peer closes the connection.

- `EAGAIN`/`EWOULDBLOCK` now return `SOCK_INTERRUPTED_TRY_AGAIN`, which every driver already maps to `SEND_RETRY`, so ComStub's retry budget applies. Same branch in `UdpSocket::send`'s zero-length path.
- Existing status reused rather than a new one, so `default: OTHER_ERROR` in ByteStreamDriver implementations stays correct, as #5344 did for the recv side. This is the behaviour the SDD and the #460 review thread describe.
- Stale "or timeout" comment and the `send()` doc comment corrected.

Four new mock-socket UTs: timeout, timeout after a partial write, EPIPE still fatal, and the UDP zero-length path. They fail on devel and pass with the fix on macOS and Linux (GCC 15); Drv/Ip, TcpClient, TcpServer, Udp and ComStub suites are green.

Two calls worth a look. Returning rather than looping on the timeout keeps a driver send call bounded by one timeout (#3129), so a stall longer than `RETRY_LIMIT` × timeout still reaches the retry-exhaustion latch, which is #5234. A timeout after a partial write resends the buffer, so the peer sees the prefix twice; the retry-exhausted path already behaves this way and the frame detector resyncs.

Written with Claude Code and reviewed by me.

IAMAI


LeStarch:
One thing I don't understand here, if we get EINTR the response is to try again locally in the for loop, but if we timeout or get EAGAIN we break with an error message.

Why have divergent behavior in these two cases nominally recoverable by a retry somewhere?

